PR #13, opened 2019-09-05
Fix contributing guideline incomplete yarn command

cyan33:
The given yarn command are incomplete, which yield the error below:

image

According to [Yarn run | Yarn](https://yarnpkg.com/en/docs/cli/run), it should be:

```
yarn run build
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,24 +21,24 @@ Before doing anything else, run these commands:
 ```
 git clone git@github.com:chakra-ui/chakra-ui.git
 cd chakra-ui
-yarn install
-yarn bootstrap
-yarn build
+yarn run install
+yarn run bootstrap
+yarn run build
 ```
 
 ## Root Repo Scripts:
 
 ```sh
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -21,24 +21,24 @@ Before doing anything else, run these commands:
 ```
 git clone git@github.com:chakra-ui/chakra-ui.git
 cd chakra-ui
-yarn install
-yarn bootstrap
-yarn build
+yarn run install
```
Vynlar: I don't believe that `yarn run install` is valid. Believe this should be `yarn install`

segunadebayo:
Hi @cyan33 , I looked at the link you posted, there's a line in Yarn's doc that says:

> "It’s also possible to leave out the run in this command, each script can be executed with its name:"


I use the `yarn [cmd]` a lot during development and I don't think it's an issue. In your screenshot I see you typed `yarn boostrap` instead of `yarn bootstrap`

PR #18, opened 2019-09-05
Fixes issue #8: Slider not working on touch devices

### LINKED ISSUE #8: Slider doesn't work on iOS
I can't change the value when using iOS

devanshj:
Fixes #8 

It's tested on chrome android and also on chrome desktop's touch emulator, the slider works fine in both. Note that this is a quick fix, it's not heavily tested but it should mostly work everywhere.

You can test slider with this PR ~[here](http://chakra-ui-pr-18.surge.sh/?path=/story/slider--default)~ [here](https://deploy-preview-18--chakra-ui.netlify.com/?path=/story/slider--default) (I was a little faster than netlify bot :P)

I've also updated the docs for slider props to be `onMouseDown: React.MouseEventHander | React.TouchEventHandler` instead of `onMouseDown: React.MouseEventHander`

### DIFF
--- a/packages/chakra-ui/src/Slider/index.js
+++ b/packages/chakra-ui/src/Slider/index.js
@@ -176,7 +176,7 @@ const Slider = forwardRef(
     const getNewValue = event => {
       if (trackRef.current) {
         const { left, width } = trackRef.current.getBoundingClientRect();
-        const { clientX } = event;
+        const { clientX } = event.touches ? event.touches[0] : event;
         let diffX = clientX - left;
         let percent = diffX / width;
         let newValue = percentToValue(percent, min, max);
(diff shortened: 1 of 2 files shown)


packages/chakra-ui/src/Slider/index.js
```diff
@@ -306,14 +310,17 @@ const Slider = forwardRef(
...
           onMouseLeave={handleMouseUp}
+          onTouchEnd={handleMouseUp}
           onBlur={event => {
             handleMouseUp();
             onBlur && onBlur(event);
           }}
           py={3}
           aria-disabled={isDisabled}
           ref={ref}
+          css={{ touchAction: "none" }}
```
devanshj: `css={{ touchAction: "none" }}` was added...
1. To prevent scrolling while moving the slider's thumb
2. To not cause the `"react-dom.development.js:1506 [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080"` bug on chrome.

segunadebayo: I see what you mean. Thanks for looking into this.

I'll test it on my mobile devices and merge it.

devanshj: You're welcome :) Btw great work on chakra!

segunadebayo:
@all-contributors please add @devanshj for code

PR #30, opened 2019-09-06
[Docs-Modal] Update modal docs

lee-reinhardt:
The `Modal` component is missing from the docs sidebar. I only found it because it's linked to from the `CloseButton` docs page.

There's also a few incorrect references to `AlertDialog` in the props table.


segunadebayo:
@all-contributors please add @lee-reinhardt for docs

PR #32, opened 2019-09-07
[Button] Fixes for `xs` size

lee-reinhardt:
The `xs` size was missing from the type enum.

The `fontSize` for it was also set to `sm`, rather than `xs`.


Vynlar:
Good stuff @lee-reinhardt ⚡️

PR #33, opened 2019-09-07
[Theme] Add missing exports from theme

FadeZhanger:
The `ITheme` interface was missing from exports. This PR adds `export *` to allow all named exports from the theme folder to be available externally.

### DIFF
--- a/packages/chakra-ui/src/index.d.ts
+++ b/packages/chakra-ui/src/index.d.ts
@@ -89,6 +89,7 @@ export { default as Textarea } from "./Textarea";
 export { default as useToast } from "./Toast";
 export { default as Tooltip } from "./Tooltip";
 export { default as theme } from "./theme";
+export * from "./theme";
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this @FadeZhanger . The theme interface is being used directly by the theme type def. Not sure if it's worth re-exporting it but it might be useful in some cases.

Thanks again.

FadeZhanger:
Hey @segunadebayo, my use case is to get access to the type `ITheme` without having to extract it with `typeof theme`.

Example:
```typescript
import {theme} from '@chakra-ui/core'
const customTheme: typeof theme = {};
```
becomes:
```typescript
import {ITheme} from '@chakra-ui/core'
const customTheme: ITheme = {};
```

Pretty small change but appreciated nonetheless. Thanks for the merge.

segunadebayo:
I see what you mean. Thanks for this.

PR #66, opened 2019-09-11
Fix TypeScript definitions

### LINKED ISSUE #63: How to use toast?
I have been enjoying use this library so far but I am unfamiliar with how to use toast. I want to be able to create a toast that has a different color and description based on the result of a function, but everytime I try I get a notification telling me I'm breaking the rules of hooks. How should I set this up? I'm sorry for the second noob question today.

aputinski:
Just a few TypeScript definition updates. Thanks for the amazing work on Chakra UI!

Fixes #63

### DIFF
--- a/packages/chakra-ui/src/IconButton/index.d.ts
+++ b/packages/chakra-ui/src/IconButton/index.d.ts
@@ -3,7 +3,7 @@ import { ButtonProps } from "../Button";
 
 type PropsFromButton = Pick<
   ButtonProps,
-  "size" | "variant" | "color" | "isLoading" | "type"
+  "size" | "variant" | "color" | "isDisabled" | "isLoading" | "type"
 >;
 
 interface IIconButton extends PropsFromButton {
--- a/packages/chakra-ui/src/Toast/index.d.ts
+++ b/packages/chakra-ui/src/Toast/index.d.ts
@@ -19,10 +19,22 @@ export interface IToast extends AlertProps {
    * The description of the toast
    */
   description?: string;
+  /**
+   * Position
+   */
+  position?:
+    | "top-left"
+    | "top"
+    | "top-right"
+    | "bottom-left"
+    | "bottom"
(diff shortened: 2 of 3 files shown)


lee-reinhardt:
I toyed around with fixing the `useToast()` types as well. With this solution, there is a type error on `IToast` because `position` already exists on `AlertProps` as a styled-system prop.

Screen Shot 2019-09-11 at 12 46 39 AM

I'm far from a TS expert, but my understanding here is that since we're extending `AlertProps`, the new `position` has to be the same, or a sub-set of the original type ([source](https://stackoverflow.com/questions/44078808/how-can-i-extend-an-interface-and-override-a-propertys-type)), which would not be the case here, hence the error.

The only way I saw to override the original type was to use an `Overwrite` helper ([source](https://stackoverflow.com/a/54975267)). This is where I landed:

```typescript
import { IAlert, AlertProps as BaseAlertProps } from "../Alert";
import { BoxProps } from "../Box";
import { Position } from "toasted-notes";
import * as React from "react";

type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U;

type AlertProps = Overwrite<
  BaseAlertProps,
  {
    /**
     * One of toasted-notes positions.
     */
    position?: keyof typeof Position;
  }
>;

export interface IToast extends AlertProps {
  /**
   * The title of the toast.
   */
  title?: string;
  /**
   * If `true` adds a close button to the toast.
   */
  isClosable?: boolean;
  /**
   * Callback function to close the toast.
   */
  onClose?: () => void;
  /**
   * The description of the toast
   */
  description?: string;
  /**
   * Duration of toast in ms.
   */
  duration?: number | null;
}
```

(This also adds `duration` which is also currently missing).

aputinski:
@lee-reinhardt Good catch, in my local patch I just cheated and used `// @ts-ignore`, so I missed the error. The solution you landed on seems great.

PR #74, opened 2019-09-12
Add unit test framework + example tests

Vynlar:
Addresses #7 

Added some basic unit tests. There are still some issues to work on but I would like to get this tooling into the repo so interested folks can start working on writing tests for other components.

- [x] Add Jest + @testing-library/react
- [x] Write example tests
- [ ] Run the tests as part of the pull request checkers
- [ ] Run tests via husky on commit/push (do we want this?)

Another thing to think about:
Components that use `react-spring` either require the use of `jest.useFakeTimers()` or we need to mock out `react-spring` to disable animations in tests. Opinions?


package.json
```diff
@@ -10,8 +10,8 @@
     "lerna": "^3.15.0",
     "prettier": "1.18.2",
     "prop-types": "^15.7.2",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6",
+    "react": "^16.9.0",
```
segunadebayo: Any reason for this upgrade? Just curious

segunadebayo:
Thank for the setting up an initial version for the tests, @Vynlar .

I've looked through it and I think it makes sense. Well done! 

I'll have one contributor review this and then I'll merge it so we can start writing more tests.

Any reason for the upgrade to ^16.9 for `react` and `react-dom`?

Vynlar:
Thanks for taking the time to review this!

Yes, react testing library uses the new async version of `act` that was added in 16.9

Vynlar:
@segunadebayo I notice that there are both yarn.lock and package-lock.json files. Which dependency tool is official for chakra?

jestrux:
@Vynlar What components are you currently writing tests for so I can pick something else.

segunadebayo:
@Vynlar , I think `package-lock.json` is the correct one.

Sometimes I end up using `yarn` that's why there's a `yarn.lock` but adding dependencies using Lerna produces a `package-lock.json`, so I'd say `npm` is the official tool.

aaronmcadam:
@segunadebayo Setting `npmClient` in `lerna.json` should make it use yarn for everything:

```diff
{
  "packages": ["packages/*"],
  "version": "independent",
+  "npmClient": "yarn",
  "registry": "https://registry.npmjs.org/",
  "publishConfig": { "access": "public" }
}
```

segunadebayo:
@aaronmcadam ,

Thanks for this. I'll set the npmClient to `yarn`

aaronmcadam:
@segunadebayo If you're going to set the `npmClient`, it's recommended to configure `useWorkspaces` too, it should be a little bit more efficient. 

Here's a reference: https://medium.com/hy-vee-engineering/creating-a-monorepo-with-lerna-yarn-workspaces-cf163908965d

packages/chakra-ui/package.json
```diff
@@ -24,11 +24,13 @@
...
     "build-storybook": "npm install && build-storybook",
-    "build:watch": "nodemon -w src -x 'rm .babelrc &> /dev/null; npm run build'"
+    "build:watch": "nodemon -w src -x 'rm .babelrc &> /dev/null; npm run build'",
+    "test": "jest --watch"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "npm test && lint-staged",
+      "pre-push": "npm test && lint-staged"
```
Ronva: As the number of test cases grows, having to wait for tests to finish before commit *and* before push will become very cumbersome. I think we should decide on one or the other.

packages/chakra-ui/src/AlertDialog/test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  prettyDOM,
+} from "../utils/testing";
+import AlertDialog from "./index";
+import {
+  AlertDialogHeader,
+  AlertDialogBody,
+  AlertDialogFooter,
+} from "./components";
+
+const MockDialog: React.FunctionComponent = () => {
```
Ronva: I really like the idea of mocking the components in the test but it might make some of the files very long. Do you think there's enough value in moving mocked components to a mock directory?

segunadebayo:
@Vynlar, I'm a bit concerned that merging this PR might cause some conflicts since I've updated the components and structure.

Can you confirm and possibly resolve any conflicts so I can merge them?

packages/chakra-ui/package.json
oeduardoal:
```suggestion
      "pre-commit": "lint-staged",
      "pre-push": "npm run test"
```

I think it should be like this

Vynlar: The only reason I had it running both times is because it's annoying to make a commit, and then push, only to realize that there are failing tests, requiring either another commit or to reset the previous one.

I'm willing to do the `pre-push` route and treat it as more of a reminder mechanism than anything else.

packages/chakra-ui/src/AlertDialog/test.tsx
Vynlar: I find that co-locating the tests and mock components is helpful and keeps the tests pretty simple. If the files are getting long, I think it is best to split up isolates tests into their own files rather than separate the mocks/tests.

Vynlar:
@segunadebayo I'll take a look at the changes and see what the best course of action is. I think I'll attempt to setup `jest` at the root of the project such that we can test any of the packages in the monorepo rather than only `packages/chakra-ui`.

segunadebayo:
@Vynlar, Yes, I agree that we should setup jest at the root. I moved storybook and babel config to the root so it makes sense for jest setup and config to be at the root.

Vynlar:
The project has changed so rapidly that this PR is now too out of date to be useful. I am going to take a step back and look into adding unit tests on the `dev-ts` branch from scratch.

Issue #91, opened 2019-09-16
[Input] Flushed Variant won't accept focusBorderColor


mattrothenberg:
Hello!

I've noticed that when using the `flushed` variant of the `Input` component, the `focusBorderColor` doesn't seem to work.

```jsx
<Input
  variant="flushed"
  focusBorderColor="green"
  {...props.field}
  id="name"
  placeholder="John Doe"
/>
```
<img width="185" alt="Screen Shot 2019-09-16 at 3 16 12 PM" src="https://user-images.githubusercontent.com/5148596/64986415-f5ac9b00-d894-11e9-932d-11f6e25cb7f0.png">

Looking at the `flushedStyle` function here, it doesn't look like the prop is even being taken into account.

https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Input/styles.js#L80-L98

1. Is this intentional?
2. If so, I'm happy to update the docs accordingly.
3. If not, I'm happy to update the implementation to take `focusBorderColor` into account.

Love the library, thanks again for your hard work here.

Cheers 🎊

segunadebayo:
Hi @mattrothenberg,

Please feel free to update the implementation and submit a PR. I guess we skipped it.

Thanks for catching this.

mattrothenberg:
Resolved by https://github.com/chakra-ui/chakra-ui/pull/92

🥂

PR #92, opened 2019-09-17
Tweak Input focusBorderColor API

mattrothenberg:
As per https://github.com/chakra-ui/chakra-ui/issues/91, this PR proposes a few adjustments to the `Input` component's API.

One discrepancy that I've noted throughout the codebase is the ambiguity of the `color` prop. In some cases, users are expected to pass a "partial" color string, e.g., `teal`. In other cases, users are expected to pass a fully-formed color string that corresponds with a key in the `colors` object that powers the UI, e.g., `teal.500`.

I'm partial to the latter approach, but ultimately I care more about consistency. To that end, I've tweaked the `Input` API accordingly.

1. Change the default value of `focusBorderColor` from `blue` to `blue.500`, so as to avoid unnecessary string interpolation in the input style generator.
2. Applied the value of the `focusBorderColor` prop to the `flushed` variant, which was otherwise ignored before.

I do have a few open questions, though –

1. Should we expose a prop for `errorColor`? It feels odd to allow users to customize the focus state, but not the error state.
2. I noticed that we were treating `focusBorderColor` differently when the dark mode was applied (dropping the color from `500` to `300`). Is there any reason for this? Should we defer this opinion to users, allowing them to pass whatever `focusBorderColor` they want as a function of whether the app is in dark mode or not?

Hope this is helpful, thanks for letting me share my thoughts!

### DIFF
--- a/packages/chakra-ui/src/Input/examples.js
+++ b/packages/chakra-ui/src/Input/examples.js
@@ -1,9 +1,13 @@
 import { storiesOf } from "@storybook/react";
+import { withKnobs, select } from "@storybook/addon-knobs";
 import React from "react";
 import Box from "../Box";
 import Input from "../Input";
 
+const colors = ["blue.500", "red.200", "orange.300", "teal.600"];
+
 const stories = storiesOf("Input", module);
+
 stories.addDecorator(story => {
   return (
     <Box maxWidth="lg" mx="auto" mt={6} p={6}>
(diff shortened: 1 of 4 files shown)


segunadebayo:
Hi @mattrothenberg ,

I really like your feedback and I agree when you say that the `color` prop is quite inconsistent and I'm looking to find ways to make it consistent. Just allowing the user to pass any color they want is better.

I appreciate the changes to the Input component.

To answer your questions:
1. It's a good idea to provide the `errorBorderColor` prop to make it possible to style the border when the input is invalid. We can default it to `red.500`

2. I changed the `focusBorderColor` in dark mode following the Material Design Spec for dark theme. Dark mode is typically a low-light, dark background scenario, and to provide enough color contrast, it's advised to use a lighter color value, hence the switch from `500` to `300`.

To make it more declarative, I think we'll have to allow the user to determine whichever color they want to use in light and dark mode. Thanks for the suggestion.

I'll look into this more thoroughly and ensure more uniform usage of any `color` prop. For now, kindly add the `errorBorderColor` prop to the input, update the type definition and I'll merge the PR.

mattrothenberg:
@segunadebayo I've addressed your feedback in https://github.com/chakra-ui/chakra-ui/pull/92/commits/92525f4f8b2f24db49cdb4c3d0b9575b544235a9.

Thanks again for your work on this awesome library!

PierreAndreis:
Bump on this. It doesn't make sense to have a powerful theme when we can't change the focus outline color

with-heart:
I'm not sure why this popped up in my notifications. But while I'm here:

@allcontributors please add @mattrothenberg for code and examples

PR #104, opened 2019-09-20
Adds semantic html fixes for docs index and READMEs

electronspin:
I suggest using `<strong>` html tag for adding semantic prominence to text. Here are basic fixes (reference: https://twitter.com/thesegunadebayo/status/1174449005172051968)

- Changed `<b>` to `<strong>` for semantic html in sample component on docs index
- Changed `<b>` to `<strong>` for semantic html in READMEs


Ronva:
From MDN:

> The HTML Strong Importance Element (\<strong\>) indicates that its contents have strong importance, seriousness, or urgency.

I don't really think the usage fits that criteria in this case. It's like the difference between an alt description and an empty alt, one is for accessibility and the other is for purely presentational purposes.

I could obviously be wrong and I'm open to changing my mind, just my 2 cents 😄

electronspin:
I would suggest that if the element in question is not semantically important, but instead simply visually bolded, then css `font-weight: bold` should be used. My primary goal here is to point out that the `<b>` tag is deprecated html and might send the wrong message to people learning html and frontend design.

Ronva:
@electronspin \<b\> is no longer deprecated as of HTML 5 ([spec reference](https://www.w3.org/TR/html50/text-level-semantics.html#the-b-element)). It actually has semantic meaning in the most up to date specification, explained really well in [this stackexchange answer](https://softwareengineering.stackexchange.com/a/255588).

Quote:
>[T]he HTML5 specification says that strong should be used to mark up an important part of the text, like a warning or some very important-to-catch word in a sentence. On the other hand, b should be used to mark up things that need to be easy to find in the text, like keywords.

segunadebayo:
Looking at @Ronva's reply. I agree that `<b>` tag is more appropriate since it's purpose to draw attention to a specific keyword NOT to show importance, seriousness, or urgency.

We could also have achieved that with the CSS `font-weight:bold` just like @electronspin suggested. Either way, I guess we can close out this PR.

Thanks a lot for your suggestion and comments @Ronva  and @electronspin.

PR #106, opened 2019-09-20
[Text] modify type to allow isTruncated

### LINKED ISSUE #105: [Text] Type Definition 'isTruncated'
Hi. You've done really nice work on the library. I'm giving it a spin on a project and it looks like the types for the Text component might need some tweaking. 

The component renders/functions properly but if you're using TypeScript the `isTruncated` prop causes a compiler error (with create react app typescript). It looks like the other styled-system props work though. My TypeScript skills are super basic but I'll see if I can find a fix and submit a PR.

Thanks!

Here is a [CodeSandbox](https://codesandbox.io/s/naughty-dream-jmm1d) showing the issue, although the CodeSandbox setup still compiles.

```Type '{ children: string; isTruncated: true; }' is not assignable to type 'IntrinsicAttributes & RefAttributes<HTMLElement> & HTMLAttributes<HTMLElement> & LayoutProps & ColorProps & SpaceProps<string | number> & BordersProps & BackgroundProps<string | number> & ... 12 )```

xtalx:
This should fix #105

### DIFF
--- a/packages/chakra-ui/src/Text/index.d.ts
+++ b/packages/chakra-ui/src/Text/index.d.ts
@@ -1,6 +1,15 @@
 import { BoxProps } from "../Box";
 import * as React from "react";
 
-declare const Text: React.FC<BoxProps>;
+interface IText {
+  /**
+   * Boolean prop for text truncation.
+   */
+  isTruncated?: boolean;
+}
+
+export type TextProps = IText & BoxProps;
+
+declare const Text: React.FC<TextProps>;
 
 export default Text;


segunadebayo:
Hi @xtalx ,

Thank you for taking a shot at this. I've resolved this issue and will release a patch soon.

I'll close this PR.

xtalx:
Alright cool, just curious... Is there even a strong reason to keep the Text component since all it does at this point is render a paragraph tag by default? I was thinking the idea would be you could still add the css properties to the Box component that make up `isTruncated` but only use the prop on `Text` or other non primitives where appropriate.  Just like how `Flex` and `Grid` work by providing some good defaults and shorthands.

Thanks for your time in reviewing this and getting a fix in, even if it wasn't what I had in mind!

segunadebayo:
I agree. At the moment, there's no strong reason to keep the Text component. It only renders a `p` and inherits all the functionality of the `Box` component. I'd say it's for semantics.

In a second thought, I like your suggestion on restricting the `isTruncated` prop to Text and Heading component. However, there might be come scenarios where the prop is needed outside the context of Text and Heading (for example, in a `Tag` or `MenuItem`)

I'll give this some more thought. Thanks again.

xtalx:
Yeah, it can get tricky. I have a little utility package that I use that has Box, Text, Flex, and Grid components where they're all variations of Box (Text,Flex,Grid, don't return a Box) with different subsets of props. This has drawbacks too, of course. All of them accept the css prop for edge cases where you need something else but it looks like one of the goals with Chakra is to avoid having to mix in the css prop. Anyway, good luck!

segunadebayo:
Correct! As much as possible we want to limit the use of the `css` prop.

Thanks again.

Issue #52, opened 2019-09-08
[Checkbox] [Radio] Add 'isReadOnly' attribute


kripod:
While checkboxes and radio buttons may be disabled, they cannot be made read-only through Chakra UI. According to the HTML specification, every input type [has a `readonly` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Attributes_common_to_all_input_types).

I would like to propose adding an `isReadOnly` prop to checkable input fields, as it is suitable for making forms which cannot be modified after submission.

kripod:
@segunadebayo This doesn’t seem to be stale for me. Has it been solved?

tioluwani94:
Hi @kripod, this has been fixed in the dev-ts branch.

kripod:
Awesome, thank you!

PR #145, opened 2019-10-04
[Checkbox] Add 'isReadOnly' attribute

Rahmon:
I've added the isReadOnly prop to Checkbox component.

#52

### DIFF
--- a/packages/chakra-ui/src/Checkbox/examples.js
+++ b/packages/chakra-ui/src/Checkbox/examples.js
@@ -51,6 +51,16 @@ stories.add("disabled checkbox", () => (
   </React.Fragment>
 ));
 
+stories.add("readonly checkbox", () => (
+  <React.Fragment>
+    <Checkbox isReadOnly>Readonly </Checkbox>
+    <Checkbox isChecked isReadOnly>
+      Readonly
+    </Checkbox>
+    <br />
+  </React.Fragment>
+));
+
 function IndeterminateExample() {
   const [checkedItems, setCheckedItems] = React.useState([true, false]);
 
--- a/packages/chakra-ui/src/Checkbox/examples.js
+++ b/packages/chakra-ui/src/Checkbox/examples.js
@@ -17,11 +17,11 @@ stories.addDecorator(story => {
 
 stories.add("Default", () => (
   <React.Fragment>
-    <Checkbox isIndeterminate defaultChecked>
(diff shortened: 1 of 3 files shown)


segunadebayo:
Thanks for attempting to fix this @Rahmon.

I'm trying to fix the Storybook deployment issue but I wanted to ask, based on your changes, does it prevent the user from changing the checked state of the Checkbox?

It seems likes you only applied a visual change. Ideally, we want to prevent the "onChange" from firing if the `isReadOnly` prop is present. Or did you account for this already?

Rahmon:
@segunadebayo 

Yes, I've applied only a visual change because the specification I've found in MDN Web Docs about [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#readonly) attribute say that:

> Note that since the value of the checkbox is not exposed by the browser to the user this attribute **only changes the appearance** of the checkbox, typically by displaying with a gray rather than black outine and  gray rather than white background when not checked.  However this attribute does not apply to the checked state of the attribute, which can **still be altered by the user**.  If you also want to prevent the user from changing the checked state of the element you must also specify the disabled attribute.

Anyway, I can update to prevent the "onChange" firing if this is the expected behavior.

segunadebayo:
Hi @Rahmon, I see what you mean and this is a controversial issue.

My thought will be that when users pass `isReadOnly`, they expect that the checkbox should be able allowed to change state. I understand native HTML doesn't do it this way but we could account for this.

Let's prevent the `onChange` from firing. Thank you for working on this.

Rahmon:
Ok @segunadebayo. I'll do that.

segunadebayo:
Hi @Rahmon, I'd like to merge your work. 

Let me know when you can finish this. Thanks

Rahmon:
I'm thinking about prevent from firing, can I just set disabled as true? So when isReadOnly is true the input will be readonly and disabled.

segunadebayo:
> @segunadebayo
> 
> I'm thinking about prevent from firing, can I just set disabled as true? So when isReadOnly is true the input will be readonly and disabled.

I'm not sure but if we mimic the behavior of  `<input type="text">`, when it has `readonly` set to `true`, the input doesn't look disabled, it looks normal but you just can't change the text.

In a similar way, the checkbox will look normal but you can't toggle it.

Rahmon:
Hi @segunadebayo. I've finished. Could you review it?

PR #190, opened 2019-10-17
[Stack] Add `isReversed` prop

### LINKED ISSUE #189: [Stack] Add `isReversed` prop
In order to avoid using the "low-level" flex-direction API and follow the design of `isInline`, I propose adding an `isReversed` prop to `Stack`, defaulting to `false`.

In my recent pull request (#188), I already [added the variable](https://github.com/chakra-ui/chakra-ui/blob/c4c8b42246b93c0fe1774979ee8a013abfef281a/packages/chakra-ui/src/Stack/index.js#L19), but its value is bound by `direction`. I think it should be overridable through a prop of the same name, leading to the modification below:

```js
isReversed |= direction.endsWith("reverse");
```

kripod:
Closes #189.

### DIFF
--- a/packages/chakra-ui/src/Stack/index.js
+++ b/packages/chakra-ui/src/Stack/index.js
@@ -30,7 +30,7 @@ const Stack = ({
         if (!isValidElement(child)) return;
         let isLastChild = children.length === index + 1;
         let spacingProps =
-          isInline || direction === "horizontal"
+          isInline || direction === "row"
             ? { mr: isLastChild ? null : spacing }
             : { mb: isLastChild ? null : spacing };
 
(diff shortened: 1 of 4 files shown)


packages/chakra-ui/src/Stack/index.js
```diff
@@ -7,32 +7,26 @@ import Box from "../Box";
...
-  if (isInline != null) {
-    flexDirection = isInline ? "row" : "column";
-  }
-
-  if (direction != null) {
-    flexDirection = direction;
-  }
+  direction = direction || (isInline ? "row" : "column");
+  isInline |= direction.startsWith("row");
+  isReversed |= direction.endsWith("reverse");
```
segunadebayo: The bitwise operator looks more concise I'd like to keep it a bit verbose so a developer doesn't have to google what `|=` means in JavaScript. You get what I mean.

This works though.

PR #235, opened 2019-11-08
fix(stack): convert React children to array before processing

### LINKED ISSUE #234: Broken Stack due to use of children.filter
In https://github.com/chakra-ui/chakra-ui/commit/04f9e7e4e64f43fdb12dc54e24c74041666b6149#diff-1553e77db4a16e801255073a8e943a2bR40, `children` is React is not an array, it's a wrapper object for none, one, or more elements. It does not have `.filter` property.

image

We can convert it to array by using `React.Children.toArray` first before performing the operation.

donysukardi:
fix #234

### DIFF
--- a/packages/chakra-ui/src/Stack/index.js
+++ b/packages/chakra-ui/src/Stack/index.js
@@ -37,12 +37,12 @@ const Stack = ({
     _direction = "column";
   }
 
-  const validChildren = children.filter(isValidElement);
+  const validChildrenArray = Children.toArray(children).filter(isValidElement);
 
   return (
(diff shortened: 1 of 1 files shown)


cschroeter:
That was my bad. This should do the trick

``` 
const validChildren = Children.map(children, (child) => {
    if (isValidElement(child)) {
      return child
    }
  })```

donysukardi:
That will still leave `undefined` in the `validChildren`.

The idea is to skip invalid elements, i.e. falsy values, isn’t it? The initial code seemed ok to already be doing that.

segunadebayo:
That was my bad too for not catching that. Thanks for this fix @donysukardi

PR #236, opened 2019-11-08
fix(button): unset hover and active styles when Button is disabled

donysukardi:
Currently disabled button still gives feedback (e.g. change in background color) when hovered as shown below,

image

This change unsets the hover and active styles for disabled button so that there will no be additional feedback given for user when user hovers on the button

### DIFF
--- a/packages/chakra-ui/src/Button/index.js
+++ b/packages/chakra-ui/src/Button/index.js
@@ -46,13 +46,14 @@ const Button = forwardRef(
     },
     ref,
   ) => {
+    const _isDisabled = isDisabled || isLoading;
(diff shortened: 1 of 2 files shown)


donysukardi:
Updated the code to only apply disabled style with `disabled` prop

segunadebayo:
Hi @donysukardi,

Sorry for the late response. We thought it might be best to move control of styles to the users and not have to deal with issues like these.

In the next release, all component styles will be available in the theme and the user will be able to change the styles to suit their needs.

I sincerely appreciate your work on this.

Thanks again.

PR #243, opened 2019-11-12
Allow intentionally null NumberInput value

kripod:
E.g. for specifying placeholders without providing a default value.

Currently, when a nullish value is specified for `NumberInput`, it thinks that the component is uncontrolled. By using `null`, a developer may denote the intentional absence of a value, leaving space for placeholders.


kripod:
@segunadebayo Sorry for disturbing, but is there a chance this will be published in a new release anytime soon?

segunadebayo:
Yes, I'm looking to publish a new version (not TS) over the weekend.

kripod:
Fantastic, thank you!

PR #265, opened 2019-11-23
Fixes gatsby-plugin-chakra-ui transpilation errors

alexlbr:
This PR fixes this issue [gatsbyjs/19042](https://github.com/gatsbyjs/gatsby/issues/19042)

The problem is the plugin is not transpiled before being published to npm.
image

With this PR the code will be transpiled before publishing as follows:
image


alexluong:
Hey @alexlbr, thank you so much for the PR.

So just to clarify, for plugin authors, we should always transpile `gatsby-browser.js` and `gatsby-ssr.js` to make sure it works in all situation?

alexlbr:
@alexluong official Gatsby plugins are always transpiled e.g.  https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-link/.babelrc. I would personally transpile any package of any project that goes to npm because we don't know where and how it will be run

segunadebayo:
Thank you for this PR @alexlbr. @alexluong can you confirm this is ready to merge?

PR #268, opened 2019-11-28
fix: prevent bugs with multiple modals opened at the same time

DouglasGabr:
### Motivation
When more than one modal is opened at the same time, once one of them is closed, the other one will "close" as well, because the portal div will be removed:

https://github.com/chakra-ui/chakra-ui/blob/db84d090c5f94da8c5979219d139d46a770ef4f2/packages/chakra-ui/src/Modal/index.js#L53-L57

The other modal disapears, but the `isOpen` state remains `true`, thus not being possible to open it again.

### Reproduction
https://codesandbox.io/s/keen-swirles-uj9yq
1. Open parent modal
2. Open nested modal
3. Close nested modal
4. Parent is gone, and can't be opened anymore :(

Alternative: always open a new portal div (instead of using the id). It's the same aproach react-modal uses

### DIFF
--- a/packages/chakra-ui/src/Modal/index.js
+++ b/packages/chakra-ui/src/Modal/index.js
@@ -76,6 +76,7 @@ const Modal = ({
   addAriaLabels = true,
   preserveScrollBarGap,
   formatIds = id => ({
+    portal: `chakra-portal-${id}`,
     content: `modal-${id}`,
     header: `modal-${id}-header`,
     body: `modal-${id}-body`,
@@ -93,6 +94,7 @@ const Modal = ({
   const contentId = formatIds(_id)["content"];
   const headerId = formatIds(_id)["header"];
   const bodyId = formatIds(_id)["body"];
+  const portalId = formatIds(_id)["portal"];
 
   let addAriaLabelledby = false;
   let addAriaDescribedby = false;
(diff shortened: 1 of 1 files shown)


caiorios:
Hello, I am facing a similar problem. When I close the Drawer and there is a modal opened, the action closes both Drawer and Modal. In my case I have a Modal it is fired when the user tries to navigate with changes unsaved (react-router getUserConfirmation).

I saw that Drawer uses the same internal component of Modal, so I think it is correlated.

Maybe it's the case of adding a custom id for Drawer too?

https://github.com/chakra-ui/chakra-ui/blob/db84d090c5f94da8c5979219d139d46a770ef4f2/packages/chakra-ui/src/Modal/index.js#L78-L81

DouglasGabr:
@caiorios I'm not sure what's the best solution. I think it's better to remove the portal custom id from the `formatIds` Modal prop and just use the `chakra-portal-${id}` string directly inside the Modal component. What do you think?

caiorios:
I think it too. I don't see any usage in making this id parametrize.. Besides Drawer and Modal, AlertDialog uses the Modal component too. Keeping the custom id portal inside Modal component will avoid changing all other components.

PR #277, opened 2019-12-03
refactor(stack): manages spacing with css

### LINKED ISSUE #207: [Stack] Spacing overrides margin values given directly to children
When using a `Stack`, it sets the `mt`, `mr`, `mb` or `ml` props of its children to value determined by `spacing`. Unfortunately, they cannot be overrided manually:

```tsx
function Component() {
  return (
    <Stack spacing={2}>
      <Box>foo</Box>
      <Box mr={100 /* Gets overrided */}>bar</Box>
      <Box>baz</Box>
    </Stack>
  );
}
```

jescowuester:
fixes #207

### DIFF
--- a/packages/chakra-ui/src/Stack/examples.js
+++ b/packages/chakra-ui/src/Stack/examples.js
@@ -6,7 +6,15 @@ import Box from "../Box";
 const stories = storiesOf("Stack", module);
 
 stories.add("vertical stack", () => (
-  <Stack spacing={4}>
+  <Stack spacing={5}>
+    <span>ooooooo</span>
+    <span>ahhhhh</span>
+    <span>Woah!</span>
+  </Stack>
+));
+
+stories.add("shouldWrapChildren", () => (
+  <Stack shouldWrapChildren spacing={5}>
     <span>ooooooo</span>
     <span>ahhhhh</span>
     <span>Woah!</span>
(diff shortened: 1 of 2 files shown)


jescowuester:
I also think the `isInline` prop is quite confusing as it sounds more like `display: inline` than `flex-direction` and we also have the `direction` prop that handles that one.

segunadebayo:
Thanks for this PR @jescowuester.

I agree that the `isInline` prop is a bit confusing. We intend to deprecate it in favor of the `direction` prop.

I'm a bit weary to merge this PR at the time because it might cause some unintended breaking changes. I'll take out time to test this solution a bit more and merge it soon.

Thanks once again.

ianstormtaylor:
Just wanted to say +1 to this PR.

It took me a while to understand why `<Stack>` wasn't doing anything on custom component children... because they didn't pass through the margin properties. Doing it via CSS instead would be a lot more natural.

segunadebayo:
Hi @jescowuester,

I've resolved this in the dev-ts branch. We'll be using css to manage spacing in the Stack component. 

Thanks for this PR.

jescowuester:
Nice! I noticed 1 potential issue, it computes the styles based on `finalDirection === "row"` so responsive values like `direction={['row','column']}` won't work.

I'd fix it like this
```
function modifyResponsiveValue(value, cb) {
  switch(true){
    case Array.isArray(value):
      return value.map(cb)
    case typeof value === 'object': {
      const newObj = {}
      Object.keys(value).forEach(k => (newObj[k] = cb(value[k])))
      return newObj
    }
    default: 
      return cb(value)
  }
}
```
```
   function parseDirection(direction) {
    const isRow = direction.startsWith('row')
    const isReversed = direction.endsWith('reverse')

    return isRow
      ? {[isReversed ? 'mr' : 'ml']: space, mb: 0, mt: 0}
      : {[isReversed ? 'mb' : 'mt']: space, ml: 0, mr: 0}
  }

  const stackStyle = modifyResponsiveValue(commonDirectionProp, parseDirection)
```

can add a pr to ts-dev if desired

segunadebayo:
@jescowuester. Yes please, PR to the dev-ts branch.

Thanks

PR #282, opened 2019-12-04
Unit Testing

royGil:
Trying to continue the work started on #7 and #74.

Maybe we should start another project named "tests" where we can keep track of the progress and make it easy for other developers to pick up components and contribute.

Please note that:

- Changes were made in the Accordion, Avatar and Alert components so check them out before merging
- I've tried to avoid ```data-testid``` as much as possible although there are cases where I didn't find a different sensible approach, we should probably remove them on production with [babel-plugin-react-remove-properties](https://www.npmjs.com/package/babel-plugin-react-remove-properties).
- I've only tested how the components work and not how they look, I think that we should run visual tests for that instead of testing css properties.
- This is my personal approach for testing, feel free to suggest different approaches.


segunadebayo:
This looks great! @royGil, thanks for taking the time to work on this.

Looking through the code, I'd love if we don't include `data-testid` at the component level, but at the test level (I mean at the `renderComponent` level). Since most of the components are composable, we should be able to do this without issues.

I also have in mind to include some `data-*` attributes to all components, as a signature to make it easy to spot Chakra UI components and also make it easy to style the components. For example, accordion might have `data-chakra-accordion-item`, as a way to identify it.

Let me know your thoughts.

royGil:
That's a good point about `data-testid`, i'll keep that in mind.


> I also have in mind to include some data-* attributes to all components, as a signature to make it easy to spot Chakra UI components and also make it easy to style the components.

can you give an example of a some common use cases?

segunadebayo:
Hi @royGil. Thanks for all your work on the tests.

I think we're getting close to merging your tests, but before then, can you update your code? We've changed the repo structure significantly and moved to a per component package.

Kindly let me know when you've done that. Thanks again.

segunadebayo:
Closing this as it's pretty stale already.

I feel bad about this but will make sure you're added to the contributors' list considering that I learned a lot from reading your tests.

Thank you 💖

PR #290, opened 2019-12-10
docs: add "Divider" component to the documentation

cjthompson:
Add the `Divider` component to the documentation


packages/chakra-ui-docs/pages/divider.mdx
```diff
@@ -0,0 +1,40 @@
+# Divider
+
+Dividers are used to display a thin horizontal line.
```
segunadebayo: We could say "horizontal or "vertical line". I've just pushed a change to support vertical dividers. Thanks

Issue #319, opened 2019-12-30
[Tooltip] showDelay does not handle mouse leave


franky47:
**Describe the bug**
I have a table of components with tooltips that should show up after a given amount of time (1.5 seconds). However, if the mouse exits the component, the tooltip is still shown, and gets stuck:
![tooltip-delay](https://user-images.githubusercontent.com/1174092/71582156-cb5bb200-2b00-11ea-8c06-c504f79cf2d2.gif)

**Expected Behavior**
Tooltips should only appear when the mouse has been inside the target component for a time greater than the `showDelay` time. When the mouse exits the target component before that time has elapsed, the tooltip should not be shown.

**To Reproduce**
1. Go to the docs > Tooltip
2. Replace the last demo (placement) with this code to enable `showDelay`:
```jsx
<Stack shouldWrapChildren spacing={6}>
  <Tooltip label="Right end" placement="right-end" showDelay={1500}>
    <Button variant="outline">Right-End</Button>
  </Tooltip>
  <Tooltip label="Left" placement="left" showDelay={1500}>
    <Button variant="outline">Left</Button>
  </Tooltip>
  <Tooltip label="Auto start" placement="auto-start" showDelay={1500}>
    <Button variant="outline">Auto-Start</Button>
  </Tooltip>
  <Tooltip label="Auto end" placement="auto-end" showDelay={1500}>
    <Button variant="outline">Auto-End</Button>
  </Tooltip>
  <Tooltip label="Left end" placement="left-end" showDelay={1500}>
    <Button variant="outline">Left-End</Button>
  </Tooltip>
</Stack>
```
3. Move your mouse from the first target component to the last
4. See the tooltips appear delayed and getting stuck.

**Suggested solution(s)**
Not sure if the fix is here or in Popper.js, will investigate and report, but the gist of it would be to cancel the setTimeout that handles showDelay when detecting a mouse exit.

**Desktop**
 - OS: macOS 10.14.6
 - Browser: Firefox 71
 - Version 0.5.2

franky47:
It can be fixed by cancelling the enter timeout on exit:

https://github.com/chakra-ui/chakra-ui/blob/bebd06798739ef913a70d0ae3bf22c1d91147bba/packages/chakra-ui/src/Tooltip/index.js#L50
```diff
  const closeWithDelay = () => {
+   clearTimeout(enterTimeoutRef.current);
    exitTimeoutRef.current = setTimeout(onClose, hideDelay);
  };
```

[...]

PR #321, opened 2020-01-01
fix(tooltip): Don't show tooltip after mouse leaves the target component

franky47:
See discussion in #319.


kltdwrds:
I'm also seeing this bug - thanks for the fix @franky47 !

PR #355, opened 2020-01-23
feat(numberinput): store value as `number | null`

kripod:
BREAKING CHANGE: NumberInput now uses null to denote an empty value. In-between states (e.g. "3.")
are no longer interpreted as a value change, thankfully to [react-typed-inputs](https://github.com/kripod/react-typed-inputs).

### DIFF
--- a/packages/Counter/src/Counter.hook.tsx
+++ b/packages/Counter/src/Counter.hook.tsx
@@ -16,7 +16,7 @@ export interface CounterOptions {
   /**
    * The callback fired when the value changes
    */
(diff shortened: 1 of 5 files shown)


segunadebayo:
Hi @kripod,

We're not going forward this PR because we think it's important to support but string and number output considering the `precision` prop we currently have. Adding this constraint will also bring about a breaking change.

I know we've been through it several times 😅. We'll revisit this in the future. I sincerely appreciate your help on this.

kripod:
I understand your decision. Thank you for reviewing!

Issue #349, opened 2020-01-18
[ColorModeProvider] ColorMode and SSR (next.js) not working very well together


santialbo:
**Describe the bug**
When using colorMode with SSR (in this case with next.js) the pages are rendered from the server with light mode.
However on the client the pages are rendered with whatever setting is stored on localstorage by `use-dark-mode`. If the value store is `darkMode = true` this leads to hydration missmatches.
This causes some of the components to be rendered with light mode even if the colorMode is dark.

Check the issue by yourself on https://next-colormode-test.santialbo.now.sh (repo on [santialbo/next-colormode-test](https://github.com/santialbo/next-colormode-test/)).
- Set dark mode on
- Refresh the page

**Expected Behavior**
I would expect that by default the client renders twice, a first time with light mode so that there are no missmatches and the turns into dark mode.

**To Reproduce**
Check [santialbo/next-colormode-test](https://github.com/santialbo/next-colormode-test/) for a minimal reproduction of the issue.

**Desktop (please complete the following information):**
Tried with both Firefox and Chrome.

davae1an:
experiencing the same issue

Penspinner:
Also experiencing the same issue.

VladLegkowski:
Would it be better not to set the theme in the local storage? 

Funny, but if you change the theme on https://chakra-ui.com/, it is not setting the cookie to true.

kolyasapphire:
Also can't get it to work both on my project (Next.js) and https://chakra-ui.com/

In the file [NoFlashSript.js](https://github.com/chakra-ui/chakra-ui/blob/28f19337a29bc6d2196708a1c76ac0d823821cc6/packages/chakra-ui-docs/components/NoFlashSript.js) on [chakra-ui-docs ](https://github.com/chakra-ui/chakra-ui/tree/28f19337a29bc6d2196708a1c76ac0d823821cc6/packages/chakra-ui-docs) I see localStorage 'darkMode' being referenced.

Returns true for me every time:
`localStorage.getItem('darkMode')`

However, in issue #214 it's 'isDarkMode'

Returns null for me every time:
`localStorage.getItem('isDarkMode')`

mdreizin:
The same issue for Gatsby as well.

franky47:
[...]

PR #362, opened 2020-01-29
fix: Make the Gatsby plugin's theme configurable

### LINKED ISSUE #347: [gatsby-plugin-chakra-ui]: Not overwriting theme
I am trying to use gatsby-plugin-chakra-ui with a custom theme.

If I understand the documentation correctly putting theme.js inside the folder "src/gatsby-plugin-chakra-ui/theme.js" should overwrite the theme. Putting a new theme.js inside this folder does not do anything. Have also tried putting it inside "src" and inside the project directory.

What I understand from the GitHub source code theme.js looks like this:
```
import { theme } from "@chakra-ui/core";
export default theme;
```

Is this the reason it always imports the default theme?

Tagging @alexluong 

"@chakra-ui/core": "^0.5.2"
"@emotion/core": "^10.0.27",
"@emotion/styled": "^10.0.27",
"emotion-theming": "^10.0.27", 
"gatsby": "^2.18.12",
"gatsby-plugin-chakra-ui": "^0.1.2",

### LINKED ISSUE #425: [Theme] gatsby-plugin-chakra-ui
I am using the plugin with gatsby-starter-blog, following the documentation with src / gatsby-plugin-chakra-ui / theme.js
And the file is ignored.

trevorblades:
Since the `src` directory was not being published by this plugin, its `theme.js` file was not shadowable. This was happening because we were transpiling the plugin files and moving them all to the package root before publishing. More info on this subject [in this comment](https://github.com/chakra-ui/chakra-ui/issues/347#issuecomment-579620254).

Since the transpile step is not required since [Gatsby runs all dependencies through babel-loader](https://github.com/gatsbyjs/gatsby/pull/14111), this PR moves the plugin's `gatsby-*.js` files to the package root and keeps `src/theme.js` in place so it can be shadowed again.

Fixes #347
Fixes #425


trevorblades:
@alexluong @segunadebayo what are your thoughts on this change? Currently, users of the Gatsby plugin are unable to override the default theme, making it unusable in many cases.

I've put together a [reproduction of the error](https://codesandbox.io/s/vigorous-wilbur-x1yjb) and an [example of the fix in action](https://codesandbox.io/s/nervous-franklin-6jqiw) to help understand the problem/solution.

alexluong:
Hey @trevorblades, sorry for the late response. The reason we transpiled the plugin was some issues in regards to monorepo:

- gatsbyjs/gatsby#19042
- #265

We apologize for the regression. It seems like this PR is just a revert of #265. Any chance you can confirm it will work inside a monorepo?

trevorblades:
@alexluong for sure! I've put together an example of the configuration described in gatsbyjs/gatsby#19042. You can check out the full example at https://github.com/trevorblades/gatsby-plugin-chakra-ui-monorepo and run it locally to make sure it works. 👍 

```bash
git clone https://github.com/trevorblades/gatsby-plugin-chakra-ui-monorepo.git
cd gatsby-plugin-chakra-ui-monorepo
yarn
yarn workspace www start
# everything should be a-ok
```

In this example I'm using a non-transpiled version of the Chakra plugin from [another git repo](https://github.com/trevorblades/gatsby-plugin-chakra-ui) within a Gatsby theme in one of the monorepo directories. I'm then installing and using the theme within a Gatsby site in a different monorepo directory using Yarn workspaces, as the original issue does.

```bash
# install plugin + peer dependencies in theme
yarn workspace gatsby-theme-example add gatsby-plugin-chakra-ui@github:trevorblades/gatsby-plugin-chakra-ui @chakra-ui/core @emotion/core @emotion/styled emotion-theming
# install theme in gatsby site
yarn workspace www add gatsby-theme-example
```

Let me know if this works for you!

segunadebayo:
@trevorblades, I don't know much about gatsby theme that's why I haven't merged it. Sorry for the delay.

I'm relying on the @alexluong to give me the green light. Kindly review @alexluong or anyone familiar with the gatsby theme and I'll merge it.

Thanks

alexluong:
Hey @trevorblades, thank you for taking the time to create the reproduction.

However, there is some issue with it. Here is the repoduction step on top of your `gatsby-plugin-chakra-ui-monorepo`:

1. In `www`, try to override the default Chakra theme:

```js
// www/src/gatsby-plugin-chakra-ui/theme.js

import { theme } from "@chakra-ui/core";

export default {
  ...theme,
  colors: {
    ...theme.colors,
    blue: {
      ...theme.colors.blue,
      200: "mediumseagreen"
    }
  }
};

```

2. In the index page, use a Button:

```jsx
import React from "react";
import { Button } from "@chakra-ui/core";

export default function Index() {
  return <Button variantColor="blue">Hello</Button>;
}
```

3. Try `yarn workspace www start`: it will behave as expected

4. Add the `build` and `serve` commands and try `yarn workspace www build && yarn workspace www serve`: in my setup it doesn't work as expected.

Let me know if you can confirm the issue. Thanks for your hard work @trevorblades!

trevorblades:
@alexluong this works for me. What `yarn` version are you working with? I'm on `1.17.3` and I'm not seeing any issues.

Here's a video of my experience going through your repro steps: https://www.youtube.com/watch?v=Ptr3Z9R-S7A

alexluong:
Interesting. Lemme take a look into it again.

I'm using `yarn 1.21.1`.

trevorblades:
@alexluong ok I think I've tracked down the cause of this issue. I've updated [my monorepo example](https://github.com/trevorblades/gatsby-plugin-chakra-ui-monorepo) to print out the overwritten variable, and color the text with `blue.500` like so:

```js
import React, {Fragment} from 'react';
import {Button, Text, useTheme} from '@chakra-ui/core';

export default function Index() {
  const theme = useTheme();
  return (
    <Fragment>
      <Button variantColor="blue">Hello</Button>
      <Text color="blue.500">{theme.colors.blue[500]}</Text>
    </Fragment>
  );
}
```

Here's what I get in development:

Screen Shot 2020-02-23 at 9 41 16 PM

And in production:

feaea32f349a312a0ad4801be58c580a

In the first image, the button color is not overwritten because the `Button` component in dark mode uses the `200` variant of the color passed in. However in light mode, the `500` variant is used. This explains the appearance of the button in production, since `colorMode` doesn't play nicely with SSR at the moment, as described in #349 

The white flash in the production GIF is what happens when the page loads and the static-rendered page starts in light mode. Due to the bug mentioned above, once the page switches over to dark mode, the button color doesn't get updated appropriately.

One way to get consistent results across development and production in a SSR app like Gatsby would be setting `isUsingColorMode` to false, unfortunately. One could still `wrapRootElement` in a `<DarkMode>` component if they want to force dark mode across the app in server and client renders alike. 

```js
import React from 'react';
import {DarkMode} from '@chakra-ui/core';

export const wrapRootElement = ({element}) => <DarkMode>{element}</DarkMode>;
```

All that being said, I think this issue is outside of the scope of this PR, and will likely need to be addressed separately. What do you think?

alexluong:
Thanks for doing the investigation, @trevorblades. As I said, besides that, everything else seems right. You can merge and publish the new version @segunadebayo! Thanks.

Issue #204, opened 2019-10-22
Deal with custom icons in TS


cschroeter:
Following the guide to add custom icons I get an TypeScript eror that the name is not assignable.

```tsx
<Icon name="custom-icon" size="20px" color="gray.400" />
```

```
Type '"custom-icon"' is not assignable to type '"link" | "time" | ...
```

Is there a ways to override the `export type Icons ..` ?

segunadebayo:
Hi @cschroeter, 

We'll look into this in the next TS release. I hope you can find a walkaround for now.

cschroeter:
yeah somehow ^^ @segunadebayo  if you need any help with the ts release please let me know

segunadebayo:
Yes, I need some help with the ts-release, particularly around theming and supporting custom types for icons, component variants, sizes, etc.

If you have any ideas, I'll love to hear them.

cschroeter:
@segunadebayo regarding the icons:

I think it is not the best approach having the icons within your theme. From a developer point of view its pretty nice to write something like `<Icon name="close" />`. 

The downside on this approach that you have no tree shaking option here, since you always will add all the icons to your bundle. So if I only add one icon in my app, all other icons also gets bundled. 

So instead of this

```tsx
<Button leftIcon="add">Hello World</Button>
```

I would rather go with this API

``` tsx
import AddIcon from './Icons'     

<Button>
  <AddIcon mr={2} />
  Hello World
</Button>

```

tony:
Is there a workaround we have in the meantime? If you have one, feel free to mention it here or link to it :) Thank you!

tony:
@cschroeter 

Hi!  I created a PR that relaxes the typing here to allow custom icons.

This means the area isn't as strongly typed, but sense any icon name can be used, it may also be more accurate in a sense.

Technically, there may not be a benefit to use `Icons` either as a typing if any string is acceptable.

tony:
In my situation, I opted to use font awesome bundles from CDN

There isn't an easy way to do this with `Icon`, but here's what I did:

```typescript
import Box from "@chakra-ui/core/dist/Box";
import Icon from "@chakra-ui/core/dist/Icon";

const FaIcon = ({ name }: { name: string }) => (
  <Box
    as="em"
    className={`fas fa-${name}`}
    lineHeight="1.2rem"
    marginLeft="-.25rem"
    marginRight="0.5rem"
    verticalAlign="middle"
    display="block"
  />
);

[...]

PR #390, opened 2020-02-08
Typings: Support custom icons

tony:
Hi, first, thank you for the project!

Related: https://github.com/chakra-ui/chakra-ui/issues/204

This means using custom icons will cause an error:

https://chakra-ui.com/icon#adding-custom-icons

For me:

```
ERROR in /hsk/_ux/js/Layout.tsx(109,30):
TS2322: Type '"faHome"' is not assignable to type 'Icons'.
```

This this case, make the check on `icons` more lenient, allow any string in `name` of `<Icon>`

### DIFF
--- a/packages/chakra-ui/src/Icon/index.d.ts
+++ b/packages/chakra-ui/src/Icon/index.d.ts
@@ -11,7 +11,7 @@ interface IIcon {
   /**
    * The name of the icon.
    */
-  name?: Icons;
+  name?: Icons | string;
   /**
    * The color of the icon.
    */


tony:
Hi, thank you again for the great project @segunadebayo! I am very grateful. This is probably the best UX library on the internet right now. Every day I find new things that save so much time.

Icons are tricky, since they relay on third party packages sometimes (some of these don't treeshake well) and various import strategies (loading svg raw from file, etc), may need a bit more flexibility / orthogonality since the user will need to find their own strategy to bring Icons in.

This typing may be helpful to users experiencing the problem in #204 - but I ultimately worked around this without using `Icon` at all: https://github.com/chakra-ui/chakra-ui/issues/204#issuecomment-583758561

benadam11:
Any chance this is going to get merged?

benadam11:
♥️

PR #403, opened 2020-02-13
Add Skeleton component

### LINKED ISSUE #402: Request to add Skeleton component
I have created a skeleton component ([sandbox here](https://codesandbox.io/s/skeleton-chakra-ui-bj0k5)), would you like to have it in the component lib?

It can be used as container or as a normal Box

remorses:
Closes #402

### DIFF
--- a/packages/chakra-ui-docs/pages/skeleton.mdx
+++ b/packages/chakra-ui-docs/pages/skeleton.mdx
@@ -1,0 +1,45 @@
+import SEO from "../components/SEO";
+
+<SEO
+  title="Skeleton"
+  description="Skeleton is used to display the loading state of some component"
+/>
+
+# Skeleton
+
+Skeleton is used to display the loading state of some component.
+
+## Import
+
+```js
+import { Skeleton } from "@chakra-ui/core";
+```
+
+## Usage
+
+You can use it as a standalone component
+
+```jsx
+<div>
+  <Skeleton height="20px" my="10px" />
+  <Skeleton height="20px" my="10px" />
+  <Skeleton height="20px" my="10px" />
+</div>
+```
+
+Or to wrap another component to take the same height and width
+
+```jsx
+<Skeleton>
+  <div>contents wrapped</div>
+  <div>won't be visible</div>
+</Skeleton>
+```
+
+### Skeleton color
+
+You can change the animation color with `colorStart` and `colorEnd`.
+
+```jsx
+<Skeleton colorStart="pink" colorEnd="orange" height="20px" />
+```
--- a/packages/chakra-ui-docs/pages/skeleton.mdx
+++ b/packages/chakra-ui-docs/pages/skeleton.mdx
@@ -36,6 +36,24 @@ Or to wrap another component to take the same height and width
 </Skeleton>
 ```
 
+Useful when fetching remote data
+
+```js
+import { Box } from "@chakra-ui/core";
+
+function Card() {
+  const { data, loading, error } = useRemoteData();
+  if (error) return <Box children="error" />;
+  return (
+    <Box>
+      <Skeleton isLoaded={!loading}>
+        <Heading>{data.title}</Heading>
+      </Skeleton>
+    </Box>
+  );
+}
+```
+
 ### Skeleton color
 
 You can change the animation color with `colorStart` and `colorEnd`.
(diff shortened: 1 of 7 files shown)


packages/chakra-ui/src/Skeleton/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+  colorStart?: string;
+  /**
+   * The color at the animation end
+   */
+  colorEnd?: string;
+}
+
+export type SkeletonProps = ISkeleton & BoxProps;
+
+declare const Badge: React.FC<SkeletonProps>;
```
segunadebayo: Thank you for this PR. It looks great.

I noticed an error with `useTheme` when trying it in CodeSandbox. Not sure what the issue it.

In this line and below, you need to export `Skeleton` not `Badge`, I guess you missed that 😅

remorses: oops, didn't notice

remorses:
I can't reproduce the `useTheme` problem, the storybook and docs seem fine, maybe codesandbox is using previous commit?

remorses:
i have added the `isLoaded` prop to skip the skeleton from rendering more easily, let's see if the sandbox updates 🤔

packages/chakra-ui/src/Skeleton/index.js
```diff
@@ -0,0 +1,69 @@
...
+  const defaultEnd = { light: colors.gray[400], dark: colors.gray[500] };
+  const {
+    colorStart = defaultStart[colorMode],
+    colorEnd = defaultEnd[colorMode],
+    isLoaded = false,
+    ...rest
+  } = props;
+  if (isLoaded) {
+    return <Fragment {...props} />;
+  }
```
segunadebayo: This looks amazing! Thank you for working on it. Here are my thoughts:

1. In real-life scenarios, we might not need the `isLoaded` prop since rendering a Skeleton will be based on some data-fetching logic and it might look like this:

```jsx
function Card(){
    const { data, loading, error } = useFetcher("/api/chakra")
    if(isLoading) return <Skeleton />
    if (error) return <ErrorDisplay /> 
    return <Component />
}
```

Hence, I'd suggest we remove the `isLoaded` prop. I hope this makes sense?

2. Expose ability to change the animation duration,  we could call this prop `speed ` whose default value is 1,  and pass that to the CSS template string in  Line 33. Make sense?

Other than that, it's good to go for the first version. In the future, we might consider creating primitives like `Circle`, `Square`, `Picture` loaders just like `react-content-loader` so it's even easier to use.

Thanks again.

packages/chakra-ui-docs/pages/skeleton.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+### Skipping the skeleton when content is loaded
+
+You can prevent the skeleton from rendering using the `isLoaded` prop
+
+```jsx
+<Skeleton isLoaded>
+  <span>Chakra ui is cool</span>
+</Skeleton>
+```
```
segunadebayo: To keep the docs as relatable as possible, I thought it might be great to add a data-fetching example here. Just like I showed below.

Maybe even use a sample card from a popular site like Airbnb. Thank you

packages/chakra-ui/src/Skeleton/index.js
remorses: I added the `isLoaded` prop for cases like this
```tsx
function Card() {
    const { data, loading, error } = useFetcher('/api/chakra')
    if (error) return <ErrorDisplay />
    return (
        <Box>
            <Skeleton isLoaded={!loading}>
                <Heading>{data?.title}</Heading>
            </Skeleton>
        </Box>
    )
}
```
Where i want to show the skeleton in a child component,
if there was not `isLoaded` prop i would need to  do this
```tsx
function Card() {
    const { data, loading, error } = useFetcher('/api/chakra')
    if (error) return <ErrorDisplay />
    return (
        <Box>
            {isLoading ? (
                <Skeleton>
                    <Heading> </Heading>
                </Skeleton>
            ) : (
                <Heading>{data?.title}</Heading>
            )}
        </Box>
    )
}
```
I prefer the first one, it is simpler and i use the child component to give the same size to the skeleton

I will add the speed prop

segunadebayo: Hmm. I agree with you. In this case, we might not want to forward all props to the `Fragment`, we should only pass children to it, to prevent any React warnings.

```jsx
  if (isLoaded) {
    return <Fragment> {props.children} </Fragment>;
  }
```

PR #422, opened 2020-02-19
feat: Make icon as optional for FormErrorMessage

gopinav:
Add conditional rendering to FormErrorMessage icon rather than default warning icon


segunadebayo:
Hi @gopinav,

Thanks for this PR and I appreciate your effort to fix this.

Unfortunately, we're no longer merging changes to `master` to help focus on the next release but I can assure that we've fix this in preparation for the next release.

There's now a `FormErrorIcon` component that can rendered by the user if they need to show an icon.

Thanks again.

Issue #427, opened 2020-02-20
Rollup fails to bundle chakra component lib


export-mike:
I'm using create-react-library and trying to use chakra-ui as the core.

I've found the lib exenv to cause issues with rollup and how its exported. I notice you have a copy of the canUseDOM in another package chakra-ui-docs so I've copied this and place it into utils.

PR to follow.

segunadebayo:
Hi @export-mike.

Since this has been fixed, I'll close it for now.

Thanks for the PR earlier. I will release a patch soon.

PR #428, opened 2020-02-20
pull canUseDOM from exenv - issue: 427

export-mike:
Roll up is failing to bundle this sub dependency. the logic for this is simple so I've removed usage of exenv
 https://github.com/chakra-ui/chakra-ui/issues/427

### DIFF
--- a/packages/chakra-ui/package.json
+++ b/packages/chakra-ui/package.json
@@ -26,7 +26,6 @@
     "aria-hidden": "^1.1.1",
     "body-scroll-lock": "^2.6.4",
     "color": "3.1.2",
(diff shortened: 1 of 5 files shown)


export-mike:
Oh right looks like this has been addressed in a different way

```<<<<<<< master
import { canUseDOM } from "../utils";
=======
import exenv from "exenv";
>>>>>>> master

////////////////////////////////////////////////////////////////////////

const { canUseDOM } = exenv;```

segunadebayo:
@export-mike, yes it has been addressed and merged

Thanks for this PR, though. I'm going to close it now.

PR #460, opened 2020-02-28
docs(input): pointer to making <Input> accessible

Vinnl:
This includes a pointer in the docs that <Input>s can be wrapped
in <FormControl>s and combined with a <FormLabel>s.

It also removes the <Text> element from one example - it is not
relevant to the example, and fulfils the role of a <FormLabel>,
hence nudging developers to inaccessible code.


segunadebayo:
Thank you!

Issue #417, opened 2020-02-18
[Select] <option> labels are not readable in dark mode


maggo:
**Describe the bug**
`<option>` elements are inheriting the white color of the `<select>` element in dark mode and are not readable.

![image](https://user-images.githubusercontent.com/384953/74699586-5acc3a00-5201-11ea-8f6f-8f527d977c27.png)

**Suggested solution(s)**
A CSS reset for the <option> elements should do the trick (`option { color: initial; }`)

**Desktop**
Windows Chrome 80

balibebas:
Suggested solution works. It would be preferable, however, for the background color of the select element to match the theme:

![Peek 2020-06-05 20-52](https://user-images.githubusercontent.com/65664227/83879103-b4845000-a76f-11ea-82a5-87bde826b8ce.gif)

with-heart:
I've verified that `Select` is now using different styling for light and dark modes in 1.0: https://codesandbox.io/s/rough-meadow-3vyr2?file=/src/App.js

nikhilpatil-git:
> I've verified that `Select` is now using different styling for light and dark modes in 1.0: https://codesandbox.io/s/rough-meadow-3vyr2?file=/src/App.js

This sandbox is not working please check

with-heart:
@nikhilpatil-git Fixed!

leomotta121:
Still not matching dark mode

balibebas:
@leomotta121 which pre-release version are you using? perhaps there was a regression

stefan-girlich:
This phenomenon still exists as of @chakra-ui/react v1.0.1. 
Observations:
- The white background is only applied when a color is passed to `<Select />` via the `bg` or `bgColor` prop.
- Issue reproducible on Firefox 83.0 on MacOS 10.15.7
- Not reproducible on Chrome 86.0.4240.198 on MacOS 10.15.7; unstyled native dropdown overlay is used
- Issue only exists in _light_ mode (if `colorMode` returned by `useColorMode()` equals _"light"_)

with-heart:
> * The white background (behind the white text) is only applied when a color is passed to `<Select />` via the `bg` or `bgColor` prop.
> * Issue reproducible on Firefox 83.0 on MacOS 10.15.7
> * Not reproducible on Chrome 86.0.4240.198 on MacOS 10.15.7; unstyled native dropdown overlay is used
[...]

PR #464, opened 2020-03-02
Fix select option white background 

noartem:
#417


segunadebayo:
Hi @noartem, 

Sorry for the late response and I appreciate your attempt to fix this. I see you're introducing a new color hue to the `gray`, and this might not be the best approach.

We've resolved this in preparation for the next release.

Thank you once again.

PR #478, opened 2020-03-07
build: Skeleton component improvements

remorses:
This pull request adds:
- skeleton renders Box instead of Fragment if `isLoaded` is true
- fadeIn animation of children after `isLoaded` is true
- styles are cached with  `useMemo`

Working on my project using the `Skeleton` component I notices that using `Fragment` when the prop `isLoaded` is true removes all the other styling properties applied to `Skeleton`, like margins and spacing and i cannot use `Stack` component

I also added a customizable fadeIn animation that makes everything nicer

This is a small change but improves the usage a lot.


remorses:
Hey @segunadebayo can you review this?

PR #490, opened 2020-03-12
chore(storybook): expand stories path

with-heart:
The existing storybook configuration was pathed to
`packages/Alert/src/*.stories.tsx` specifically. This change expands the
path to any stories in `packages/` (`packages/**/*.stories.tsx`), so
that all stories appear in storybook.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -2,7 +2,7 @@ const path = require("path")
 const TsconfigPathsPlugin = require("tsconfig-paths-webpack-plugin")
 
 module.exports = {
-  stories: ["../packages/Alert/src/*.stories.tsx"],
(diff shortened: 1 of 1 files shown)


aaronmcadam:
This was done during development to show just the one story at a time. Should there be an environment variable to support this? If it's not set, you could show all stories.

segunadebayo:
@aaronmcadam is right. I typically do this to focus my Storybook explorer to just one component at a time. I find Storybook a bit slow and annoying when I load all stories, and I'm just trying to refine one component at a time.

@with-heart, feel free to tweak this during development. When all is done and complete, then I'll revert back to loading all stories. Sounds good?

with-heart:
Sounds like a plan! I'll go ahead and close this.

Issue #406, opened 2020-02-14
Allow to use `usePortal` on the Menu component.


santialbo:
When the menu trigger is within an overflown element it's necessary to render the menu using a portal in the same way it's already possible with the `Popover` component.

**Describe the solution you'd like**
Same interface as the Popover component, just another prop `usePortal: boolean;`

**Describe alternatives you've considered**
I tried implementing myself by passing the prop all the way down to the Popper component but I got an error that I was not sure how to fix with the `focusableItems`

segunadebayo:
Hi @santialbo,

Do you have a codesandbox replica for this issue? I'd like to take a look.

Thanks

santialbo:
This is a feature request, I just tried changing [this line](https://github.com/chakra-ui/chakra-ui/blob/06b65dd4117d8c1207d3ed80f4932d8c2b0d7dee/packages/chakra-ui/src/Menu/index.js#L317) but the stories would not work, they would throw an error because [the effect in L51](https://github.com/chakra-ui/chakra-ui/blob/06b65dd4117d8c1207d3ed80f4932d8c2b0d7dee/packages/chakra-ui/src/Menu/index.js#L51-L61) would be executed before `menuRef` has a value.

bjgrosse:
@segunadebayo here is a sandbox that reproduces this. If you don't set usePortal=true on the MenuList, then the menu gets cut off. If you do usePortal then you get an error about the focusableItems. focusableItems is null at the time of opening the menu.

Turning autoSelect off gets past this error, but then other things don't work such as the closeOnBlur. Seems that a various number of things need to be changed to allow the portal. 

https://codesandbox.io/s/beautiful-cray-dk9eq

bjgrosse:
It should be noted that the Popover component works fine in this same scenario, and it would be lovely if the Menu could as well.

Sebastiansc:
I'm facing this same issue. It seems that when `usePortal` is true this line here isn't called in the order that's expected https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Popper/index.js#L144. That line should be called before this https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Menu/index.js#L52 (as it does when `usePortal` is false).

Since `menuRef` is defined in Popper/index.js#L144, Menu's `useEffect` in L52 doesn't set the focusable items causing auto selection to error out

Sebastiansc:
[...]

PR #498, opened 2020-03-15
[WIP] fix: Allow Portal to initialize with provided container

Sebastiansc:
## Summary
* Menu expects `menuRef` to be defined at this point https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Menu/index.js#L52. This ref is attached to a `PseudoBox` inside of a `Portal` component. `Portal` returns `null` on initial render since it initializes `mountNode` as `null`, so it attaches no ref until it's `useEnhancedEffect` and `setMountNode` are called, but this are called after the ancestors `useEffect`, so any ancestors relying on that ref being attached would fail.
 I considered changing the `useEnhancedEffect` to `useEffect` but the `setMountNode` re-render was still getting run after the ancestors `useEffect`. By initializing `mountNode` to the provided `container` we can ensure that there is an attached ref by providing a valid DOM node for `container` on the initial render.
*  Since `finDOMNode` will return `null` if the container is not defined this won't cause any extra renders for component who's `container` is dynamic (`Modal`).

## Issue
* https://github.com/chakra-ui/chakra-ui/issues/406

### DIFF
--- a/packages/chakra-ui/src/Portal/index.js
+++ b/packages/chakra-ui/src/Portal/index.js
@@ -16,7 +16,7 @@ function getContainer(container) {
 
 const Portal = forwardRef(
   ({ children, container, isDisabled = false, onRendered }, ref) => {
-    const [mountNode, setMountNode] = useState(null);
+    const [mountNode, setMountNode] = useState(getContainer(container));
     const handleRef = useForkRef(children.ref, ref);
 
     useEnhancedEffect(() => {


Sebastiansc:
Found a better solution for this. Closing this pr will open another one soon

segunadebayo:
Hi @Sebastiansc,

Thanks for this PR and I appreciate your attempt to fix this.

We've resolved this in preparation for the next release. You can check here:
https://github.com/chakra-ui/chakra-ui/blob/3b31020478532fc429d53efddaf7861a3d896404/packages/portal/src/Portal.stories.tsx#L22

Thanks again.

PR #510, opened 2020-03-21
fix: erroneous event handler in Image component

blnk-space:
Image component had incorrectly typed `onError` event in camelcase, leading to the native DOM element having a property on it that it didn't recognize. Renaming it to `onerror` did the trick.


segunadebayo:
Thank you for catching and fixing this!

PR #512, opened 2020-03-22
Text: Add text cropping feature

Vynlar:
# Summary
This PR adds a feature that I call "baseline crop" (better name suggestions welcome).
By default in HTML, text nodes have extra white space above and below them. In the design world, text usually sits directly on a baseline. This adds a feature to the `Text` component that removes this extra white space.

# Usage
```jsx
<Text isBaselineCropped>
  ChakraUI rocks!
</Text>

<Heading isBaselineCropped>
  ChakraUI rocks!
</Heading>
```

# Context
Here are some resources that help clarify the _why_ surrounding this change.
- This talk demonstrates a similar concept, albeit slightly different approach: https://www.youtube.com/watch?v=TGHbkTGVqoU
- The Braid Design System implements this exact concept in their Text component: https://github.com/seek-oss/braid-design-system
  - In this podcast, Mark Dalgleish discusses the rationale behind their decision to implement this cropping in the Braid Design System

# Technical Considerations
Getting this feature to work is quite finicky and requires a small amount of manual measuring on the part of the end-user. I plan to create a small widget for the docs website (here is [a similar widget](http://text-crop.eightshapes.com/)) that allows the user to measure their fonts and copy values into their theme.
The current implementation handles the following tricky things:
- `isTruncated` still works even with the cropping
- Responsive styles, including responsive fontFamilies, still works

# Potential Problems
- If someone uses `"normal"` as a line height, I convert it to `1.2` because I need to know the line height in order to perform the cropping. This only occurs when using the `baselineCrop` prop and therefore it is opt-in. It would be good to note this in the docs.

# Remaining Work
- [x] My implementation currently does _not_ work if you use the object syntax for responsive styles.
Example: `<Text baselineCrop lineHeight={{ sm: 'tall', md: 'normal' }} />`. I have not yet figured out a way to make this possible, and therefore may be a necessary compromise. To mitigate this downfall, we could move this feature into its own component like `<Text.Cropped />` in order to isolate the issue.~
- [x] The `baselineCrop` prop still needs to be added to the `Heading` component. It should be quite simple since it is quite similar to the `Text` component.


segunadebayo:
Hi @Vynlar,

This is amazing! Thanks for this detailed explanation of the text cropping feature, I think it'll be a great addition to Chakra UI.

To do some testing, we can either move this feature to its own component "CroppedText" and "CroppedHeading", or use an experimental prop like `unstable__baselineCrop` to the Text and Heading components.

Getting a feature like this will definitely come at a cost, so I agree we might need to lock down the "normal" line-height to `1.2`. I believe it's a safe assumption, if it turns out unsafe then we can simply add a setting on the theme object to allow the user to specify it.

For the object responsive styles, I'll take a look at the code and suggest how we can make it work. If you've solved it for array responsive styles, then we can make it work for the object. I'll send some ideas shortly.

Thanks for working on this!

Vynlar:
@segunadebayo 

Thanks for taking the time to look at this!

I'll elaborate on the challenge that the object styles pose:

# The Problem

In order to calculate the amount to crop from the top/bottom of the text, I need to know two values: `font-family` and `line-height`. This is simple when they are both single values, and I have implemented a way to figure it out when they one or both are arrays. Here's an example of that:

Given the props: `lineHeight={[1, null, 2]}` and `fontFamily: ["mono", "body", null]`
I can line up the two arrays like such:
```
[1,      null,   2]
["mono", "body", null]
```
Then I have all the info that I need to create responsive crop values.

The issue comes when one or both of `lineHeight` or `fontFamily` is an object. Here is an example of such a time:
`lineHeight={{ sm: 1, lg: 2 }}` and `fontFamily={{ sm: "mono", md: "body" }}`
I can no longer line up the values, since lineHeight does not define an `md`. And since objects are not ordered, I cannot "calculate" the value as I could with arrays that contains nulls.

This whole thing may be a non-issue if we assume that users define their breakpoints as an array in their theme, _and then add aliases_. As shown in the snippet from the chakra docs:
```
// First, create an alias for breakpoints
const breakpoints = ["30em", "48em", "62em", "80em"];
// aliases
breakpoints.sm = breakpoints[0];
breakpoints.md = breakpoints[1];
breakpoints.lg = breakpoints[2];
breakpoints.xl = breakpoints[3];
```
An issue would arise if the user defined their breakpoints like this (not even sure if this is valid according to the styled-system theme spec):
```
breakpoints: {
sm: "30em",
md: "48em",
lg: "62em",
xl: "80em"
}
```

Let me know if you or anyone else (looking at you @lee-reinhardt) have any insight into this.

🤘 Keep rockin' it!

Vynlar:
🎉 Success!

By making the user required to define their breakpoints as an array (only required if they use the baselineCrop prop), I was able to make this work for _object responsive style props_.

Next step:
Implement this for the `Heading` component.

packages/layout/src/layout.utils.ts
```diff
@@ -0,0 +1,274 @@
...
+  if (isArray(value)) {
+    return value
+  } else if (value === undefined) {
+    return []
+  } else if (typeof value === "object") {
+    // In order for the text crop feature to work, we must know the _order_ of the breakpoints,
+    // meaning we require the breakpoints to be an array (rather than an object)
+    // Note: aliases as members of the array are still supported
+    if (!isArray(breakpoints)) {
+      throw new Error(
```
segunadebayo: Amazing! Great work @Vynlar.

One thing: we're actually looking to enforce object breakpoints going forward (for better intuitiveness). Most users find it difficult to understand the aliasing part.

I see you're internally converting the object to an array, so I thought we can avoid throwing.

Vynlar: Are you saying that you plan to require users define breakpoints like this?:
```
breakpoints: {
sm: "30em",
md: "48em",
lg: "62em",
xl: "80em"
}
```
Or as an array?:
```
breakpoints: ["30em", "48em", "62em", "80em"]
```

segunadebayo: Yes. Like this:
```js
breakpoints: {
sm: "30em",
md: "48em",
lg: "62em",
xl: "80em"
}
```

Vynlar: Thanks for the clarification.

In order for me to calculate these text crop values, I need to know the **order** of the breakpoints. If they are defined as an object, I no longer have a way to determine the order. Perhaps we could create a helper function for users to use in their theme files?:
```
breakpoints: makeBreakpoints([
  ["sm", "30em"],
  ["md", "48em"],
  // and so on...
])
```

Vynlar:
@segunadebayo I've updated the code to assume an object breakpoint definition by default.

To ensure this, I added in some runtime checks that will ensure that the theme is setup in a way that will work with this feature. These checks only run if you enable the `baselineCrop` mode on the `Text` or `Heading` components, and also only run in development mode.

I plan to add documentation for this feature once the documentation site build is working again.

Additionally, I plan to create a small widget that will help our users measure their fonts and put the correct values into their theme file.

🎉Cheers!

Vynlar:
One thing of note is we may want to think of a better name for the prop. Right now it is `baselineCrop`. It is a boolean, so perhaps we should follow the convention of starting boolean props with `is`. For example: `isCropped` or `isBaselineCropped`.

segunadebayo:
Great work @Vynlar, keep going ⚡️!

As for the prop name, maybe we can use `isBaselineCropped` for now.

Vynlar:
Alrighty! 

I've done a lot of code cleaning and reorganization for this feature. Additionally, I wrote documentation into the `text` and `heading` pages, though I was unable to get the docs site to run in order to check them.

At this point, I'm ready for feedback!

segunadebayo:
Amazing work @Vynlar, I will review this over the weekend and leave my feedback.

Thank you!

segunadebayo:
@Vynlar, please rebase this with `master` considering we just switch branches

kumar303:
@Vynlar this is great work! Did the work on this move to another PR? The text cropping feature is essential for consistent spacing.

kumar303:
I added https://github.com/chakra-ui/chakra-ui/issues/2096 to keep track of this awesome work.

kumar303:
Here is another issue about integrating with the Capsize tool: https://github.com/chakra-ui/chakra-ui/issues/1199

Vynlar:
@kumar303 I started this PR much earlier in the Chakra 1.0 beta process and things were changing too quickly for me to keep up. Additionally, in that time [Capsize](https://github.com/seek-oss/capsize) was released and is likely a better tool to tackle this problem. Yes, I got the cropping to work; however, the world has changed enough that I think it would be worth taking a step back on this problem and approaching it from a different angle.

Once Chakra 1.0 is officially released, I would be willing to work on integrating Capsize. I do have some questions for the community about the best way to approach this that I will add into #1199.

PR #527, opened 2020-03-26
build: added forwardRef to Stack

### LINKED ISSUE #526: [Stack] not using `forwardRef`
**Describe the bug**
The `Stack` component should forward ref using `forwardRef`
I get the following error
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `Step`.
    in Stack (at HowItWorks.tsx:75)
```

**Expected Behavior**
The `Stack` component should forward ref using `forwardRef`

remorses:
closes #526

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   ],
   "scripts": {
     "commit": "git-cz",
-    "postinstall": "yarn core:build",
+    "postinstall": "yarn build",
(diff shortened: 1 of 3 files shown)


remorses:
The build fails because the `postinstall` script references the old and now missing `core:build`, my second commit here resolves the issue

PR #528, opened 2020-03-27
build(parser): use tsc to transpile code

cschroeter:
This is a proof of concept how you could speed up build time by only using the typescript compiler.
In this example we only build a ESM variant. We could also build a CJS version for example:

```
    "build": "concurrently yarn:build:*",
    "build:esm": "tsc --project tsconfig.esm.json",
    "build:cjs": "tsc --project tsconfig.cjs.json",
```

```bash
➜  parser git:(dev-ts) ✗ yarn build:tsdx 
yarn run v1.22.0
$ tsdx build
✓ Creating entry file 2.2 secs
✓ Building modules 14.8 secs
Done in 17.46s.


➜  parser git:(dev-ts) ✗ yarn build
yarn run v1.22.0
$ rimraf dist
$ tsc --project tsconfig.build.json
Done in 2.05s. 
```


aaronmcadam:
What other benefits is `tsdx` providing us?

cschroeter:
`tsdx` is using Roll-Up under the hood.

aaronmcadam:
Have you measured Babel @cschroeter?

cschroeter:
Babel is also a good choice - but since you anyway want to run a typecheck like `tsc --noEmit` you don't gain any benefits, since the longest part is the typechecking.

aaronmcadam:
> Babel is also a good choice - but since you anyway want to run a typecheck like `tsc --noEmit` you don't gain any benefits, since the longest part is the typechecking.

I understand the trade-offs, I've been using Babel with TypeScript for a long time. I was curious to see if it's any faster.

cschroeter:
If you just want to transpile your code as fast as you want - checkout https://github.com/swc-project/swc

we managed to compile our 150k loc project at LogMeIn in < 1s

aaronmcadam:
> If you just want to transpile your code as fast as you can - checkout https://github.com/swc-project/swc
> 
> We managed to compile our 150k loc project at LogMeIn in < 1s. Unfortunately we don't run it in production since it is missing type declarations

Aren't type declarations supposed to be removed from the final build?

cschroeter:
Yeah so if you want to build an app that is totally fine. They also provide an webpack loader :)

But for libs like ours or ChakraUI we need this declaration files for people using it getting good intellisense support

aaronmcadam:
That's cool @cschroeter, thanks for the background. 

I'm thinking we need to be careful that just because a tool is faster doesn't make it better for our use cases.

What things do we care about for our build? 

For example, Babel can be good because it allows us to use other plugins and can be more efficient with module output.

segunadebayo:
I'm also a bit careful about this considering that we have a working system. Not an expert in build tools but I think we can try both `tsc` and `babel` approach. I've used babel and it works pretty well.

cschroeter:
`tsdx` provides a lot of features that does not make much sense for a library like ChakraUI. So `tsdx` is not an ideal tool since it brings a lot of feature that ChakraUI does not benefit from but also increase the build time by 10-15 times.

Babel is good choice but it also makes the tool chain more complex. Eventually if you use Babel you still need to have TypeScript as a dependency since you still want to do type checking and create declaration files. 

My recommendation is to start with TypeScript which simplifies the setup and speed up the build time

segunadebayo:
@cschroeter , I agree with your thought. Let's start with TypeScript and if need babel at any point, it'll pretty easy to add it.

TSDX definitely has some stuff we don't need and this slows down the build time.

segunadebayo:
Thank you so much for your help on this @cschroeter.

Based on your work, we've been able to improve the build time significantly and also ensured test and lint scripts work correctly.

We're grateful ❤️

Issue #534, opened 2020-03-28
[PseudoBox] Use pseudo element instead of pseudo class.


wonism:
**Describe the bug**
It is not a bug. But, We should use `::before` instead of `:before` in CSS3. It is more semantically right.
See: https://developer.mozilla.org/en-US/docs/Web/CSS/::before#Syntax

**Expected Behavior**
Have pseudo element instead of pseudo class

**To Reproduce**
n/a

**Suggested solution(s)**
Use pseudo element instead of pseudo class

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] All
 - Browser [e.g. chrome, safari] All
 - Version [e.g. 22] All

**Additional context**
n/a

PR #535, opened 2020-03-28
fix: use pseudo element instead of pseudo class

wonism:
#534


segunadebayo:
@wonism, thanks for this PR. What's the rationale for this?

wonism:
@segunadebayo, As I described in #534 

Pseudo element is a selector for specifying a part of the element.
Pseudo class is a selector for specifying a special state of selected element.

In case of before/after content inside of an element, pseudo element is more semantically correct than pseudo class.

And we should use **double colons(::)** for pseudo element. not a single colon(:). This distinguishes pseudo class from pseudo element.

---

Specification of pseudo elements. See:
https://www.w3.org/TR/selectors-3/#pseudo-elements

segunadebayo:
@wonism, I see what you mean. Thanks for the clarification.

One thing though, can you PR against the `dev-ts` branch?  We're no longer merging changes to master to help us focus on the next release.

This is the file you're looking for: https://github.com/chakra-ui/chakra-ui/blob/dev-ts/packages/parser/src/configs/pseudo.selector.ts

Thanks

wonism:
@segunadebayo, Sure. I re-created a PR. #552

PR #536, opened 2020-03-28
fix(radio-group): call querySelector from rootNode directly

with-heart:
Calling `querySelector` by itself causes a `TypeError: Illegal Invocation` error because it uses `this` and must be bound to `rootNode`. Calling `rootNode.querySelector` fixes this binding issue.


segunadebayo:
Thank you buddy!

PR #537, opened 2020-03-28
fix(radio-group): reverse isNative check

with-heart:
I mentioned this issue in Discord, but I believe that `isNative` logic is currently backwards. I assume we want components to work for web by default.

Before:

`isNative` value | prop used
---|---
`undefined` (default) | `isChecked`
`false` | `isChecked`
`true` | `checked`

After:

`isNative` value | prop used
---|---
`undefined` (default) | `checked`
`false` | `checked`
`true` | `isChecked`


with-heart:
This is intended behavior meant to make `useRadioGroup` work directly with `useRadio`. Misunderstanding on my part.

PR #538, opened 2020-03-29
fix(select): pass placeholder prop to SelectField

with-heart:
This PR fixes an issue where `placeholder` was being incorrectly passed to the root `chakra.div` by destructuring it from `props` and passing it to `SelectField`.


segunadebayo:
This is a good catch! Thank you Mark ❤️

PR #541, opened 2020-03-29
test: add tests for array utils

amadk:


packages/utils/src/tests/array.test.ts
```diff
@@ -1,4 +1,50 @@
...
+  getLastItem,
+  getPrevItem,
+  getNextItem,
+  removeIndex,
+  addItem,
+  removeItem,
+  getNextIndex,
+  getPrevIndex,
+  chunk,
+  getNextItemFromSearch,
```
with-heart:
```suggestion
```

Not using either of these imports ;)

amadk: thanks, good catch

segunadebayo:
Thanks for your work on this, Amad, I noticed you didn't get to complete the tests. I'll take it up from here.

PR #554, opened 2020-04-01
chore: create component test checklist document

with-heart:
This PR creates a component test checklist document, which outlines some basic items that should be verified for each component module to ensure that it meets our testing standards:

- [ ] Common use cases snapshotted
- [ ] Common use cases run through `axe`/`toHaveNoViolations`
- [ ] `role`/`aria`/`data` attributes tested
- [ ] Component behaviors tested (reacts to events, handles callbacks
      appropriately, updates state correctly, etc.)
- [ ] Controlled/uncontrolled use cases tested
- [ ] Associated utils/helpers/etc. tested

Since accessibility is the core of The Chakra Way™️, we have multiple related checklist items to help us guarantee that our components remain as accessible as possible.

### DIFF
--- a/.team/CODE_REVIEW_GUIDELINE.md
+++ b/.team/CODE_REVIEW_GUIDELINE.md
@@ -68,6 +68,8 @@ TypeScript:
 
 Testing:
 
+- [Chakra Component Test Checklist](COMPONENT_TEST_CHECKLIST.md)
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thank you for this @with-heart ❤️

PR #559, opened 2020-04-02
Chore: review radio component

tioluwani94:
This PR includes updates made to the radio component package in preparation for release. The following activities were carried out:

- Tests review
- Types review
- Storybook examples
- Addition of the migration
- Comments review
- Readme review
- Dependency list review
- Addition of the RadioGroup component

### DIFF
--- a/packages/radio/README.md
+++ b/packages/radio/README.md
@@ -1,4 +1,39 @@
-Base Package
+# Radio
 
-Don't run any npm script in here. It's just a basic folder to enable us create
-new packages faster.
+Radio component is used in forms when a user needs to select a single value from
+several options.
+
+## Installation
+
+```sh
+yarn add @chakra-ui/radio
+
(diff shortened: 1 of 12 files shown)


packages/radio/README.md
```diff
@@ -1,4 +1,39 @@
...
+<Radio>This is a radio</Radio>
+```
+
+### RadioGroup
+
+RadioGroup is used to bind multiple radios into a group, and it indicates
+whether which option is selected.
+
+```jsx
+<RadioGroup defaultValue={["one", "two"]}>
```
segunadebayo, line 34: RadioGroup only takes a single value not an array, since only one radio can be checked at a time. Please change this.

tioluwani94: Thanks for spotting this @segunadebayo. It has been resolved.

PR #563, opened 2020-04-03
Review spinner component

tioluwani94:
This PR includes updates made to the spinner component package in preparation for release. The following activities were carried out:

- Tests review
- Types review
- Storybook examples
- Comments review
- Readme review
- Dependency list review

### DIFF
--- a/packages/spinner/README.md
+++ b/packages/spinner/README.md
@@ -1,1 +1,66 @@
 # Spinner
+
+Spinners provide a visual cue that an action is processing awaiting a course of
+change or a result.
+
+## Installation
+
+```sh
+yarn add @chakra-ui/spinner
+
+# or
+
+npm i @chakra-ui/spinner
+```
+
+## Import component
+
+```jsx
+import { Spinner } from "@chakra-ui/spinner"
+```
+
+## Usage
+
+```jsx
+<Spinner />
+```
+
+## Spinner with different sizes
+
+Change the size of the spinner by passing the `size` prop.
+
+```jsx
+<>
+  <Spinner size="xs" marginRight={3} />
+  <Spinner size="sm" marginRight={3} />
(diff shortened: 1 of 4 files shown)


packages/spinner/README.md
```diff
@@ -1 +1,66 @@
...
+Change the background color of the moving section of the spinner by passing the
+`color` prop.
+
+```jsx
+<Spinner color="red.500" />
+```
+
+## Spinner with empty area color
+
+Change the background color of the spinner's by passing the `emptyColor` prop.
```
with-heart, line 53:
```suggestion
Change the background color of the spinner by passing the `emptyColor` prop.
```

packages/spinner/README.md
```diff
@@ -1 +1,66 @@
...
+
+```jsx
+<Spinner color="red.500" emptyColor="gray.200" />
+```
+
+## Spinner with different speed
+
+Change the spinner's animation speed area by passing the `speed` prop. The unit
+of the value is seconds
```
with-heart, line 62:
```suggestion
of the value is seconds.
```

with-heart, line 62: We could also reference `animation-duration` instead:

```suggestion
Change the spinner's animation speed area by passing the `speed` prop. The unit
of the value matches the [`animation-duration CSS property`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-duration) property.
```

packages/spinner/README.md
tioluwani94: Thanks for catching this @with-heart.

packages/spinner/README.md
tioluwani94: Thanks for catching this @with-heart.

tioluwani94: Thanks for this @with-heart.

PR #568, opened 2020-04-04
feat: allow disabling focus lock modal

Drakota:
Hello there! 👋

We're currently using chakra for an electron app and since we have a titlebar with dropdowns and buttons, but also use modals in the application we couldn't have the focusLock enabled. I thought I'd drop a quick PR, since more options the better right?

Thank you for you work on Chakra 😊


segunadebayo:
Hi @Drakota, thanks for this PR.

Unfortunately, we're no longer merging updates to master. I've gone ahead to do this in the `dev-ts` branch. You can see it here: https://github.com/chakra-ui/chakra-ui/blob/e2df94b8b445a24dbe3d9786d77b9ed2b41afd58/packages/modal/src/Modal.tsx#L60

Thanks again for this, and I'm glad you find Chakra UI useful.

PR #588, opened 2020-04-07
test: implement all remaining util tests except for html/react related functions

amadk:
Implemented tests for the following utilities:
- Array
- Assertion
- Function
- Number
- Object
- Responsive
- Dom


packages/utils/src/tests/responsive.test.ts
```diff
@@ -0,0 +1,11 @@
+import { mapResponsive } from "../responsive"
+
+test("mapResponsive - 'prop: [1, 2, 3], mapper: item => (item += 1)' should equal [ 2, 3, 4 ]", () => {
+  expect(mapResponsive([1, 2, 3], item => (item += 1))).toStrictEqual([2, 3, 4])
```
segunadebayo, line 4: Can we make this test a bit more relatable? We mostly use the `mapResponsive` function to handle responsive style props in Chakra's components.

Perhaps changing the tests to be like
```jsx
// sample usage in SimpleGrid component
const columns = mapResponsive({ sm: 2, md: 3 }, val => `grid-template-columns(${val}, 1fr )` )
```
I mean the goal is to make it as realistic as possible. You get what I mean?

amadk: thanks, i made the new changes, let me know what you think

PR #602, opened 2020-04-12
Review progress component

tioluwani94:
This PR includes updates made to the progress component package in preparation for release. The following activities were carried out:

- Tests review
- Types review
- Storybook examples
- Addition of the migration
- Comments review
- Readme review
- Dependency list review


packages/preset-base/src/components/Progress.ts
```diff
@@ -0,0 +1,73 @@
+import { ComponentTheme, mode } from "./utils"
+import { getColor } from "@chakra-ui/color"
+
+const getProgressBg = (props: any) => {
```
gnyani, line 4: Why is this `any`? I think we can define an inexact type for these props right?

PR #641, opened 2020-04-25
fix: use same image crossOrigin in preloading and img tag

matchu:
The `` HTML tag already supports the `crossOrigin` property, and Chakra's `<Image>` will already forward the `crossOrigin` prop to the tag, hooray!

But, when we _preload_ the image, we do so without `crossOrigin` set, which means that the preloaded image can't actually be reused for the final `` tag 😓

This means that the browser sends a second request, and preloading is not only unhelpful, but more expensive on bandwidth, and approximately doubles the image load time!

In this change, we update the `<Image>` component to use the `crossOrigin` prop in its image preload request, if present. Now, when an `<Image>` is provided a `crossOrigin` prop, we'll make only one request, with the `crossOrigin` behavior that the caller requested.

### DIFF
--- a/packages/chakra-ui/src/Image/index.js
+++ b/packages/chakra-ui/src/Image/index.js
@@ -4,7 +4,7 @@ import { useEffect, useState, forwardRef, useRef } from "react";
 import Box from "../Box";
 
 export function useHasImageLoaded(props) {
-  const { src, onLoad, onError, enabled = true } = props;
+  const { src, onLoad, onError, crossOrigin, enabled = true } = props;
   const isMounted = useRef(true);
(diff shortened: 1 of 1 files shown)


matchu:
Oh hey, I'm noticing that I didn't follow all of the instructions re branch name in the original repository; I built this on `master` of my fork. I can go tweak that if it's useful to y'all!

segunadebayo:
Thanks for this PR @matchu,

Can you update this PR to point to the `dev-ts` branch instead? That's where we're actively working for the next major release.

PS: In the `dev-ts` branch, this will go to `useImage` hook in `packages/image`.

Thanks

PR #648, opened 2020-04-27
fix: validate variantColors

### LINKED ISSUE #644: [Button] using a color without shades throws a warning
**Describe the bug**

The Button has a check here to validate the variant color, but is not considering a color like white or black or any plain colors that has no shades 

https://github.com/chakra-ui/chakra-ui/blob/67c0243a4becd2c2509934d59e52b4a25ff199ab/packages/chakra-ui/src/utils/index.js#L168

Also if you do send a color like blue.500 then the check also warns because the color is not part of the theme

**Expected Behavior**
It shouldn't warn in case of an existing color with or without shades

tafelito:
fixes #644

### DIFF
--- a/packages/chakra-ui/src/utils/index.js
+++ b/packages/chakra-ui/src/utils/index.js
@@ -163,18 +163,24 @@ export function useVariantColorWarning(label, variantColor) {
         );
       }
 
-      // if variant color exists and is an object
-      // TODO: Check for the 100 - 900 keys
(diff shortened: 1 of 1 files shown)


segunadebayo:
Hi @tafelito. I'm sorry for this PR is no longer relevant considering we're moving to v1 shortly.

Thank you for putting some work into this!

Issue #632, opened 2020-04-23
[Badge] TypeError: Cannot use 'in' operator to search for 'red' in undefined


bklynate:
**Describe the bug**
I've encountered a TypeError when running jest snapshot test on a component that imports the Badge component. It seems to be related to the `variantColor` prop

<img width="571" alt="Screen Shot 2020-04-22 at 10 05 08 PM" src="https://user-images.githubusercontent.com/4186033/80053606-55ce8400-84eb-11ea-8d1e-93c1ae1b2f78.png">

Here is the simple snapshot test I have put together:
```
import React from 'react';
import { Header } from '../index';

describe('Header', () => {
  it('renders correctly', () => {
    const tree = renderer.create(<Header />).toJSON();
    expect(tree).toMatchSnapshot();
  });
});
```

Here is the Header component:
```
/* eslint-disable consistent-return */
import React, { Component } from 'react';
import { Flex, Text, Badge } from '@chakra-ui/core';
import { connect } from 'react-redux';

class Header extends Component {
  getIfUserIsAuthenticated(user = {}) {
    if (user === null) return false;
    if (user === undefined) return false;
    return typeof user === 'string';
  }

  renderSignInContent(user) {
    const isUserAuthenticated = this.getIfUserIsAuthenticated(user);
    if (isUserAuthenticated) return <a href="/api/logout">Logout</a>;
    return <a href="/auth/google">Log in with Google</a>;
  }

  render() {
    // ----------------------------------------------------------------
    // TODO: Leaving this code here; Will use to wire up authentication
    // ----------------------------------------------------------------
    // const { request } = this.props;
    // const { cookies } = request;
    // const { asx_data: userToken } = cookies;

    return (
      <Flex
        display="flex"
        flexDirection="column"
        alignItems="center"
        justifyContent="center"
        textAlign="center"
        mt={4}
      >
        <Flex
          display="flex"
          flexDirection="row"
          alignItems="flex-start"
          justifyContent="flex-start"
        >
          <Text fontSize="2rem" fontWeight="bold" mb={0}>
            NYC JOB SEARCHER
          </Text>
          <Badge variant="subtle" variantColor="red" ml={1} mt="1.8rem">
            BETA
          </Badge>
        </Flex>
        <Text color="gray.500">Easily Find City Job Listings</Text>
      </Flex>
    );
  }
}

const mapStateToProps = state => {
  return state;
};

export { Header }

export default connect(mapStateToProps)(Header);
```

[...]

Issue #627, opened 2020-04-21
TypeError: Cannot use 'in' operator to search for 'gray' in undefined on React 3.4.1


fadhilx:
**Describe the bug**
i got error `TypeError: Cannot use 'in' operator to search for 'gray' in undefined`

**Expected Behavior**
it should run without error

**To Reproduce**
1. install `create-react-app` to latest version, in this case (3.4.1)
2. create new project with `create-react-app`
3. add button on page
```jsx
import React from "react";
import PropTypes from "prop-types";
import "./LoginPage.scss";
import { Button } from "@chakra-ui/core";
const LoginPage = () => (
  <div className="LoginPage" data-testid="LoginPage">
    LoginPage Component
    <Button>Test</Button>
  </div>
);

LoginPage.propTypes = {};

LoginPage.defaultProps = {};

export default LoginPage;

```
4. save and See error
![image](https://user-images.githubusercontent.com/15516786/79863140-d321c800-8409-11ea-8723-709b3cdd7704.png)

**Suggested solution(s)**
only possible solution so far is to (maybe) downgrade react version. but to fix this issue must done on affected code 

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 80

segunadebayo:
Hi @fadhilx,

This error shows up with you're not wrapping the entire application in the `ThemeProvider`. 

Kindly go to the root of your application, import ThemeProvider and wrap your application in it.

Thanks.

PR #658, opened 2020-04-30
fix: Throw error when theme is empty object

terrierscript:
I got error `TypeError: Cannot use 'in' operator to search...` like those issues.
https://github.com/chakra-ui/chakra-ui/issues/632
https://github.com/chakra-ui/chakra-ui/issues/627

This error happen when forget `<ThemeProvider>`.
`ThemeProvider` throw error when `theme` is `undefined`.
But sadly, `@emotion/core`'s `ThemeContext` returns empty object by default and this check is not passed
https://github.com/emotion-js/emotion/blob/master/packages/core/src/context.js#L17

This pull request append empty object checking.

### DIFF
--- a/packages/chakra-ui/src/ThemeProvider/index.js
+++ b/packages/chakra-ui/src/ThemeProvider/index.js
@@ -14,7 +14,7 @@ ThemeProvider.defaultProps = {
 
 const useTheme = () => {
   const theme = useContext(ThemeContext);
-  if (theme === undefined || Object.keys(theme).length === 0 ) {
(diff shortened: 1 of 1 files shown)


terrierscript:
My point in this PR is not "I want to add a check", but "the condition of `theme === undefined` is wrong".
(Removed `theme === undefined` to get the intent right.)

The reason is due to the specification of `emotion` as described above.

Also, this error, which I found unfriendly this time, is caused by not adding `<ThemeProvider>`, so it is not possible to make `<ThemeProvider>` have a check.

Issue #659, opened 2020-04-30
Request to add property to Tabs


vonwao:
**Is your feature request related to a problem? Please describe.**
Usability issue, repetitive code

There may be another way to do it, but in this case it's not documented. 

**Describe the solution you'd like**
Add _selected property to Tabs or TabList  component.
for example: _selected={{ color: "white", bg: "blue.500"}}

**Describe alternatives you've considered**
Currently can use property in Tab inside TabList, but this becomes repetitive:
<Tab _selected={{ color: "white", bg: "blue.500"}}>Price</Tab>

I could create my own custom component inheriting Tab or Tabs, but I think that would be confusing

segunadebayo:
Hi @vonwaol, we have a permanent solution for this in the next release.

For now, why not create a wrapper around the Tab component and use that instead of repeating code?

with-heart:
See the docs for the new component theming api for more information about how this will work in 1.0: https://next.chakra-ui.com/docs/theming/overview

PR #671, opened 2020-05-04
feat: impl check shiftKey in Editable #659

keslert:
Before submitting when the "Enter" key is pressed, check to ensure the shiftKey or metaKey are not pressed. This allows for adding new lines when the editable is a textarea.


segunadebayo:
Hey @keslert,

Thanks for this PR. Can you possibly PR against the dev-ts branch instead?

PR #674, opened 2020-05-04
fix(docs): change false to true in the props of the Editable component

vicradon:
If the `isPreviewFocusable` and `submitOnBlur` are set to false, changes made to the component would not be shown


with-heart:
Good catch @Vicradon!

PR #680, opened 2020-05-06
feat: added loading attribute to Image

goncy:


with-heart:
Would it make sense to also include `"auto"`? 

From https://web.dev/native-lazy-loading/:

image

PR #709, opened 2020-05-14
feat: <ModalContent /> can now have minGap specified for mobile screens

ansonkao:
## Summary
Currently in ChakraUI, when a modal is displayed in a mobile device, there is no gap on the sides. If you prefer there to be a gap (see screenshots below), you can now pass the minGap prop to specify a desired thickness.

- Adds `minGap` prop to `<ModalContent />` with example in docs
- Fixes Table overflow problem in docs when viewing from mobile screens

Thoughts on the name `minGap`?

## Screenshots
image
image
image


segunadebayo:
Thanks for this PR, @ansonkao.

We're changing the DOM structure of the modal in the next release and that makes this PR incompatible.

I sincerely appreciate your effort, and we'll consider your idea in this PR.

PR #720, opened 2020-05-15
feat: add iconColor prop to <Select />

dannytatom:
This PR adds an `iconColor` prop to the `<Select />` component which allows you to set an icon color separate from the text color. This is helpful for when the design calls for the icon matching the border color.

It's not an existing issue, but need it for my job. If I need to create a matching issue or something, let me know.


segunadebayo:
Hi @dannytatom, thanks for this!

We've fixed this in our working branch `dev-ts` but will merge and release an update considering that it's urgent for your work.

dannytatom:
@segunadebayo awesome, I appreciate it!

Issue #714, opened 2020-05-15
[MenuList] closes incorrectly on IE 11


lightDer:
**Describe the bug**
MenuList closes incorrectly on IE11.

**Expected Behavior**
The menu list should keep open.

**To Reproduce**
1. Click the menu button, and the menu list is opened.
2. Move the cursor to the menu list, and it triggers `onBlue` event.
3. At this moment, the menu list closes incorrectly.

**Suggested solution(s)**
See
https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Menu/index.js#L304L305
and find that `event.relatedTarget` is null on IE 11.

Use `const target = event.relatedTarget || document.activeElement` instead.

**Desktop (please complete the following information):**
 - Windows
 - IE
 - 11

PR #738, opened 2020-05-20
fix: #714 MenuList closes incorrectly on IE 11

lightDer:
Please see #714.


with-heart:
From what I've been able to find, this does appear to be the best solution! Thanks @lightDer!

Issue #739, opened 2020-05-20
localStorage issues


giovanni-bozzano:
I am working on a desktop webapp that relies on Chakra for the UI rendering.
I am having issues using `ColorModeProvider`. Specifically using `<DarkMode>`.
It appears that on macOS (Safari WKWebView) the page is not able to access `localStorage` and it just crashes.

It would be nice to have a switch option to use/not use the `localStorage` feature provided by the use-dark-mode library and just let my application handle the user preference.

with-heart:
I believe this should no longer be an issue once 1.0 is released, but I'll let @segunadebayo speak to that.

Aidurber:
I think the quickest way would be to pass a storage interface to the `ColorModeProvider` component. That's is how [redux-persist](https://github.com/rt2zz/redux-persist#v6-upgrade) handles it. 

However, since the `ColorModeProvider` value is `light | dark`. This requires a breaking change unless the provider is wrapped and takes `value` and `storage` as props. So you're looking at a v1 regardless 😬 

**Storage prop**
```typescript
interface IStorage{
 setItem(key:string, value:string);
 getItem(key:string):string;
 removeItem(key:string):string
// etc.
}
```

You then get into the problem of [creating adapters](https://github.com/rt2zz/redux-persist#storage-engines). 

One possible way to do this without a breaking change would be to add an optional config call.

```jsx
function App(){
  React.useEffect(() => {
    Chakra.configure({
      storage: localStorage
    })
  }, [])
 return //
}
```
[This approach is used by Atomic Layout](https://redd.gitbook.io/atomic-layout/api/layout/configure)

Then the ColorModeProvider takes that value (defaulting to localStorage for backwards compatibility). Or even as part of the theme object, but I'm not a fan of that approach since it's mixing concerns.

dedo1911:
I just got around the issue with a solution like this:

```jsx
import { ColorModeContext } from '@chakra-ui/core'

const DarkMode = props => {
  const darkMode = {
    colorMode: 'dark',
    toggleColorMode: _ => {}
  }
  return <ColorModeContext.Provider value={darkMode}>
    {props.children}
  </ColorModeContext.Provider>
}
```

What I did is basically defining my own `DarkMode` color provider that doesn't use anything but a `colorMode` property.
You should implement your custom `toggleColorMode` or set `colorMode` with your rules.

Aidurber:
@dedo1911 Good thinking! Much simpler than anything I suggested 😁

[...]

PR #740, opened 2020-05-20
fix: #739 ColorModeProvider correctly use manual mode

dedo1911:
In case of value provided to ColorModeProvider, don't rely on use-dark-mode result


dedo1911:
Should fix #739 avoiding access to localStorage in case of "manual mode"

PR #742, opened 2020-05-21
chore(website): add canonical urls

with-heart:
This PR adds canonical urls using the post slug via `gatsby-plugin-react-helmet-canonical-urls` and the `SEO` component. It also removes trailing slashes on URLs using
`gatsby-plugin-remove-trailing-slashes`.


with-heart:
Thanks to @KenzoBenzo for pointing me in the right direction!

PR #749, opened 2020-05-22
update Button loadingText type

probablyup:
Doesn't specifically need to be a string


with-heart:
Hi @probablyup, I'd love to get this merged but could you fix your commit message to start with `fix: ` so it passes our semantic PR rules? We use them for changelog generation, so we want to make sure this is correctly included and categorized.

probablyup:
@with-heart done

with-heart:
Thank you!

PR #753, opened 2020-05-23
feat: v1 codemods

jmiazga:
I am creating this PR to track what needs to be done, and hopefully others can help with this.

### General tasks
- [ ] write README
- [ ] update package.json
- [ ] determine if we can write tests against the codemods
- [ ] need to figure out how to only modify chakra components (jscodeshift findJSXElementsByModuleName method doesnt work with es modules)

### Codemods (some might not be worthwhile or feasible)
These tasks are coming from the [migration doc](https://github.com/chakra-ui/chakra-ui/blob/07bc4e864623c061070727d296d925e68a4e5d7c/docs/pages/migration.mdx)

- [ ] Update your dependencies - remove emotion from package.json (might not want to do)
- [ ] Clone Chakra's default theme (might not be worth doing, can accomplish with npx)
- [ ] Update the ThemeProvider (might not be worth doing, simple change in a single file)
- [ ] Rename variantColor to colorScheme - in progress
- [ ] Update layout size prop - in progress
- [ ] Update these style props - in progress
- [ ] ...

### DIFF
--- a/packages/codemods/src/variantColor-to-colorScheme.ts
+++ b/packages/codemods/src/variantColor-to-colorScheme.ts
@@ -1,28 +1,128 @@
-import { Transform } from "jscodeshift";
+import { Transform } from "jscodeshift"
+import { describe } from "jscodeshift-helper"
+import { types as t } from "@babel/core"
 
-const transformer: Transform = (file, api) => {
-  const j = api.jscodeshift;
-  const root = j(file.source);
+const renames = [
+  ["variantColor", "colorScheme"],
+  // ["size", "boxSize"], // need to break out into w and h props in the future
+  ["rounded", "borderRadius"],
+  ["roundedLeft", "borderLeftRadius"],
+  ["roundedRight", "borderRightRadius"],
+  ["roundedTop", "borderTopRadius"],
+  ["roundedTopLeft", "borderTopLeftRadius"],
+  ["roundedTopRight", "borderTopRightRadius"],
+  ["roundedBottomLeft", "borderBottomLeftRadius"],
+  ["roundedBottomRight", "borderBottomRightRadius"],
+  ["roundedBottom", "borderBottomRadius"],
+  ["size", "boxSize"], // do we need to change all size to boxSize or just for Box component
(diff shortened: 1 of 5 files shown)


packages/codemods/src/variantColor-to-colorScheme.ts
```diff
@@ -0,0 +1,128 @@
...
+  ["rounded", "borderRadius"],
+  ["roundedLeft", "borderLeftRadius"],
+  ["roundedRight", "borderRightRadius"],
+  ["roundedTop", "borderTopRadius"],
+  ["roundedTopLeft", "borderTopLeftRadius"],
+  ["roundedTopRight", "borderTopRightRadius"],
+  ["roundedBottomLeft", "borderBottomLeftRadius"],
+  ["roundedBottomRight", "borderBottomRightRadius"],
+  ["roundedBottom", "borderBottomRadius"],
+  ["size", "boxSize"], // do we need to change all size to boxSize or just for Box component
```
segunadebayo, line 17: We need to careful with this one. Some components use the `size` prop to refer to visual size and not width+height. Just like button, checkbox, etc.

For these components, we won't change the prop. but we'll update for the rest. I guess the best solution is to keep an array of the components to ignore.

segunadebayo:
Great work @jmiazga and @with-heart. Thank you for working on this.

For cloning the default theme, we're already creating a CLI for this. So users will only do `npx chakra init` and it'll set it up for them.

packages/codemods/src/variantColor-to-colorScheme.ts
with-heart: Yeah we're definitely going to use `width`+`height`. Unfortunately `jscodeshift` is kind of a pain, so we very quickly got stuck but we're going to keep working on it and see where we can get.

with-heart: Oh, my comment only addressed half of your point lol. We're planning on changing the prop name only if the value of `size` isn't one of the existing t-shirt sizes we use in `master`.

PR #772, opened 2020-05-27
feat: add PortalContainerProvider

with-heart:
This PR adds a `PortalContainerProvider` component, allowing the user to globally specify a `container` for portals to be appended to. 

```jsx
const App = () => {
  return (
    <ThemeProvider>
      {/* Modals+Drawers are attached to this div */}
      <div ref={portalContainer} />
      <PortalContainerProvider container={portalContainer}>
        <Modal>
          <ModalOverlay />
          <ModalContent>
            <ModalHeader>Header</ModalHeader>
            <ModalBody>Body</ModalBody>
          </ModalContent>
        </Modal>
      </PortalContainerProvider>
    </ThemeProvider>
  )
}
```

We ran into the need for this feature at work. 

We have a large legacy system of custom styling and bootstrap, and we're attempting to add Chakra into the mix. Chakra styles are frequently overridden the specificity of the existing styles. 

We discovered that we can use `CacheProvider` from `@emotion/core` with the `styling-plugin-extra-scope` plugin to scope the rules generated by Chakra to a container with an id (we're using `#is-chakra`), thus allowing us to give much greater specificity to Chakra's rules. See [this CodeSandbox](https://codesandbox.io/s/upbeat-morning-8rfur?file=/src/App.js) for an example.

The problem we've run into with this approach is that portals are by default attached to `document.body`, which causes `Modal`s and `Drawer`s to lose all of their Chakra styling. We can pass `container` to `Modal` or `Drawer` to fix this, but it's a large burden to our users to ask them to specify the `container` prop every time they use one of these components. This change allows us to remedy that by specifying the `container` in one place.

### DIFF
--- a/packages/chakra-ui/src/Modal/index.js
+++ b/packages/chakra-ui/src/Modal/index.js
@@ -10,7 +10,7 @@ import { disableBodyScroll, enableBodyScroll } from "body-scroll-lock";
 import FocusLock from "react-focus-lock/dist/cjs";
 import { wrapEvent, useForkRef, getFocusables } from "../utils";
 import Box from "../Box";
-import Portal from "../Portal";
+import Portal, { usePortalContainer } from "../Portal";
 import CloseButton from "../CloseButton";
 import { hideOthers } from "aria-hidden";
 import { useId } from "@reach/auto-id";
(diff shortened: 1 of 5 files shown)


with-heart:
If this is an approach we like, I'll open a PR for `dev-ts` as well!

with-heart:
This functionality exists (in a better form) in `dev-ts` via `PortalManager`, so I'm going to close this.

PR #780, opened 2020-05-30
fix: button storybook in docs header

mikqi:
Previously in docs page there is no button to storybook page, but in homepage has it. This PR add button storybook in docs pages.

_Before_
image

_After_
image


with-heart:
Tested this locally and it works well! Thanks @mikqi!

PR #782, opened 2020-05-30
fix(number-input): use value string to check converting to number

with-heart:
This PR fixes an issue where `NumberInput` throws `TypeError: value.indexOf is not a function` when `max` or `min` values are specified as numbers by first converting `value` to a string.


with-heart:
Closing in preparation for 1.0.

PR #783, opened 2020-05-30
fix(number-input): allow left and right arrow keys for movement

### LINKED ISSUE #711: [NumberInputField] ArrowLeft & ArrowRight doesn't work
**Describe the bug**

In NumberInputField component `ArrowLeft` & `ArrowRight` doesn't work.

**Expected Behavior**

I'm not sure if it's intended to work like this, but I'd really like to navigate through the entered data using keyboard arrow left & right keys

**To Reproduce**
- Use `NumberInputField` component:
1. Enter any number in the field
2. Click on `ArrowLeft` or `ArrowRight` to navigate through entered data
3. We cannot navigate between the entered data

**Suggested solution(s)**

We could allow keyboard keys `37` (ArrowLeft) & `39` (ArrowRight) for the navigation
Also I found [this](https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/useNumberInput/utils.js#L4-L10), where chakra only allows [numeric keypad](https://github.com/chakra-ui/chakra-ui/issues/254)

 - OS: macOS Catalina (10.15.3)
 - Browser: Chrome

with-heart:
fixes #711 

This PR enables left and right arrow keys to function as movement keys inside the input. Currently the user can only click in order to change their position in the input, which is non-standard input behavior.

### DIFF
--- a/packages/chakra-ui/src/useNumberInput/index.js
+++ b/packages/chakra-ui/src/useNumberInput/index.js
@@ -2,7 +2,7 @@ import { canUseDOM } from "exenv";
 import { useRef, useState, useEffect, useCallback } from "react";
 import {
   calculatePrecision,
-  preventNonNumberKey,
+  preventNonNumberNonArrowKey,
   roundToPrecision,
 } from "./utils";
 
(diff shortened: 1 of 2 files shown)


with-heart:
Closing in preparation for 1.0. Will create a new PR for `dev-ts` to resolve the same issue.

PR #791, opened 2020-06-01
docs: remove unnecessary memoization

with-heart:
I'm not sure what changed from the time this was added to now, but it appears we no longer need to memoize any of these components. Things are loading fine locally and the sidebar nav no longer scrolls to the top when changing to a new page. Memoization was also breaking hot reload, so that's working again as well.


with-heart:
Memoization already removed in `dev-ts`. Closing.

PR #797, opened 2020-06-02
feat(media-query): add additional hooks

with-heart:
This PR adds the following `media-query` hooks:

- `useBreakpointValue`
- `useAnimationPreference`
- `useColorModePreference`
- `useViewportSize`


packages/media-query/src/media-query.hook.tsx
```diff
@@ -148,3 +150,82 @@ const createMediaQueries = (breakpoints: Dict) => {
     return mediaQuery
   })
 }
+
+type BreakpointValues<T> = { [breakpoint: string]: T }
```
with-heart, line 154: Should we include the responsive array style as well?

packages/media-query/src/media-query.hook.tsx
```diff
@@ -148,3 +150,82 @@ const createMediaQueries = (breakpoints: Dict) => {
...
+  const breakpoint = useBreakpoint()
+
+  // if breakpoint is undefined, we'll return undefined
+  if (!breakpoint) return
+
+  // if we have no value for the current breakpoint, return the `base`
+  // breakpoint value
+  const baseValue = values.base
+
+  return values[breakpoint] || baseValue
```
with-heart, line 175: Do we already have tooling somewhere for finding a value from a breakpoints object that defaults to `base`? I saw `assignObject` but that wasn't quite what we needed here.

packages/media-query/src/media-query.hook.tsx
```diff
@@ -148,3 +150,82 @@ const createMediaQueries = (breakpoints: Dict) => {
...
+export function useViewportSize(
+  ref: React.RefObject<HTMLElement>,
+): Readonly<[number, number]> {
+  const [size, setSize] = useState<Readonly<[number, number]>>([0, 0])
+
+  useEffect(() => {
+    if (!ref.current) return undefined
+
+    const observer = new ResizeObserver(([entry]) => {
+      const { inlineSize: width, blockSize: height } = entry.contentBoxSize[0]
```
with-heart, line 220: `ResizeObserver` gives us `contentBoxSize` or `borderBoxSize`, and I'm not sure which would be more appropriate here. Should this include the padding/border? Or maybe we should return the width/height for both and let the user decide?

segunadebayo: Should we consider using "@reach/rect" for this part? It's much smaller than the resize observer package and works really well.

What do you think? I feel we can simply re-export `useRect` from `@reach/rect`

packages/media-query/src/media-query.hook.tsx
segunadebayo: We're enforcing breakpoints in the theme to be an object in v1. It makes the logic a lot easier.

packages/media-query/src/media-query.hook.tsx
segunadebayo: Oh no. We need to make a decision on this. I was thinking on calling it either `base` or `xs`. Then we can update the docs and code to be consistent

with-heart: Is it just trying to figure out the right now or would they work differently?

packages/media-query/src/media-query.hook.tsx
with-heart: That's a better idea than using the larger package. Though maybe the best idea is not even re-exporting and letting users decide which to use if they want this functionality?

packages/media-query/src/media-query.hook.tsx
segunadebayo: Let's leave it at `base` for now.

segunadebayo:
Again, please rebase this with `master` and send a PR

PaulPCIO:
image

This happens when resizing the window?

ljosberinn:
@PaulPCIO hi, is this related to this PR or is it a separate issue?

PaulPCIO:
Hi, no sorry it's unrelated, I was about to open up an issue but this seems like a ghost in the machine, I tried replicating this again but it does not seem to happen anymore.  ¯\_(ツ)_/¯

PR #805, opened 2020-06-05
docs(guides): add file contributors functionality

with-heart:
This PR adds `fields.contributors` to `Mdx`. This is a field intended for use with `guides` pages, so we can list users that contributed to each guide.

### DIFF
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -1,11 +1,10 @@
 const path = require("path")
 const { createFilePath } = require("gatsby-source-filesystem")
-const getFileContributors = require("file-contributors").default
-require("isomorphic-fetch")
 
 const {
   sortPostNodes,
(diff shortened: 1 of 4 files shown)


docs/gatsby-node.js
```diff
@@ -58,6 +50,36 @@ exports.onCreateNode = async ({ node, actions, getNode }) => {
   }
 }
 
+exports.createSchemaCustomization = ({ actions, schema }) => {
```
with-heart, line 53: This creates a `Contributor` type (that defaults to `[]`) and adds it to the `Mdx.fields`. 

We do this because:
- it's an optional field meant only for `guides` collection pages
- we currently have no nodes with `fields.contributors` since all the guides are on `dev-ts`, so gatsby fails when it sees a query for `contributors` as that type isn't defined

PR #814, opened 2020-06-06
chore(site): add team page

with-heart:
This PR adds a team page to the site, fetching the team's data using `gatsby-source-graphql` and the GitHub API.

image

In order for this to work, we need to make sure we have a `GITHUB_API_TOKEN` in env with `repo` and `read:org` permissions available in the build step.

### DIFF
--- a/docs/src/pages/team.js
+++ b/docs/src/pages/team.js
@@ -9,6 +9,8 @@ import {
   Avatar,
   SimpleGrid,
   Link,
+  Flex,
+  Tooltip,
 } from "@chakra-ui/core"
 import { IoLogoTwitter, IoLogoGithub, IoIosGlobe } from "react-icons/io"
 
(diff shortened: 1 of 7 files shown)


docs/src/pages/team.js
```diff
@@ -0,0 +1,104 @@
...
+          </Stack>
+          <Text>{bio}</Text>
+        </Stack>
+      </Stack>
+    </Box>
+  )
+}
+
+const sortMembers = (a, b) => {
+  // segun comes first!
```
segunadebayo, line 51: 💖

segunadebayo:
Thanks for this PR @with-heart.

Can you also include the list of all contributors below that? I think Github has a way of showing all the code contributors to a repo.

If it's possible, let's do this and possibly filter out the core members since we're already at the top :)

What do you think?

with-heart:
@segunadebayo Love it! I'll implement that shortly.

with-heart:
According to [this post on github.community](https://github.community/t/graphql-get-repository-contributors/14422/7), there is no way to get the contributors. I can just read .all-contributorsrc though, so I'll go that route :)

with-heart:
Added a list of contributors. Each contributor displays an avatar that links to the user's github profile page. 

Updated screenshot:

image

Issue #777, opened 2020-05-29
[IImage interface] Bug in onLoad definition (and possibly onError)


wweaver:
**Describe the bug**

The `<Image>` component's `onLoad` function is defined incorrectly in typescript. It's defined as `() => void`, but the first argument could be an `event`.

![image](https://user-images.githubusercontent.com/1482976/83261053-c0c54780-a188-11ea-8a45-699a0fee367d.png)

![image](https://user-images.githubusercontent.com/1482976/83261108-da668f00-a188-11ea-8f91-6603975c48f9.png)

**Expected Behavior**

If I just remove the `onLoad` definition in the `IImage` interface the type error goes away as the underlying BoxProps, really the React DomAttributes, already supports what is needed.

**To Reproduce**

Create an Image with an onLoad method that takes an event

```
<Image
    src={'/path/to/image'}
    onLoad={(event: SyntheticEvent<any, Event>): void =>
        console.log(event.currentTarget.width)
    }
/>
```

**Suggested solution(s)**
Remove the `onLoad` definition from the `IImage` interface.

**Desktop (please complete the following information):**
 - OS: OS X
 - Browser: N / A
 - Version: N / A

**Additional context**
I'm not using the `onError` callback but I'm guessing that has the same issue.

segunadebayo:
This is no longer an issue in the `v1` pre-release version.

PR #816, opened 2020-06-07
fix: remove unnecessary event types on image

Aidurber:
This PR is resolving the missing `onLoad` and `onError` callback types for #777. As @wweaver mentioned on the bug:
> If I just remove the onLoad definition in the IImage interface the type error goes away as the underlying BoxProps, really the React DomAttributes, already supports what is needed.

Unable to add tests for the types. 

Tested by creating a symlink to @chakra-ui/core in a sandbox project of mine and ensuring the default image callback types are available.

### DIFF
--- a/packages/chakra-ui/src/Image/index.d.ts
+++ b/packages/chakra-ui/src/Image/index.d.ts
@@ -16,14 +16,6 @@ interface IImage {
    */
   alt?: string;
   /**
-   * A callback for when the image `src` has been loaded
-   */
-  onLoad?: () => void;
-  /**
-   * A callback for when there was an error loading the image `src`
-   */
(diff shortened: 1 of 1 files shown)


Aidurber:
Here's a sandbox of it in action: https://codesandbox.io/s/vigorous-sinoussi-vxm9r?file=/src/index.js

Observe the `onLoad` and `onError` callbacks being typed.

santialbo:
I think this is no longer an issue on the upcoming version.

https://github.com/chakra-ui/chakra-ui/blob/a6141a7928da52a82f3a4238dda0be460c4b5ee6/packages/image/src/use-image.ts#L20

This can probably be closed or change the destination to the v0 branch.

PR #832, opened 2020-06-09
feat(divider): revise divider and add vertical support

cschroeter:
Refactor `<Divider />` 
Implemented vertical variant
Added docs


cschroeter:
Not sure what I should put as the `divider.svg`

PR #863, opened 2020-06-15
feat(utils): replace null values in mapResponsive with previous value

remorses:
This let you omit values ti use the previous one, like `styled-system` does

```tsx
<Stack direction={['column', null, 'row']} />
```


with-heart:
Hi @remorses, can you target `develop` for this? We're merging most pre-1.0 release work there first so we can make keep `master` in a clean state. 

We'll go back to `master`-based dev after the release.

remorses:
I switched to #871 because github does not let me resolve the conflicts

PR #871, opened 2020-06-16
feat(utils): replace null in mapResponsive

remorses:


segunadebayo:
Hey @remorses,

Thanks for this PR. From my knowledge of styled-system, I think it skips the `null` breakpoint and goes to the next value.

Here's what I mean. Given these breakpoints `{ sm: 480, md: 768, lg: 1280 }`

```jsx
<Box margin={[ "20px", null, "40px" ]}/>
```

will return

```css
.css-Box {
  margin: "20px",
  @media (min-width: 768px){
   margin: "40px"
  }
}
```
Notice the "sm" breakpoint is skipped. Given that styled-system uses a min-width media query, it's unnecessary to duplicate the values.

With your solution, we'll get this:

```css
.css-Box {
  margin: "20px",
  @media (min-width:480px){
   margin: "20px"
 }
  @media (min-width: 768px){
   margin: "40px"
  }
}
```

I believe the first CSS and yours do the same thing from a CSS perspective, so we might not need to generate an extra css style for null values. We simply just skip it.

I hope this makes sense. Let me know if you have questions

remorses:
I tried the Stack component with null values and it didn’t work because the mapper function didn't handle the case Where the argument is a null value

In that case the mapper function could be changed to return null if the mapped value is null but in case where the mapper logic is more complex it is no longer the case

I don’t know, letting the user explicitly handle the null case is error prone

segunadebayo:
Ah, then the issue is in the `mapResponsive` function that the Stack component uses.

I'll take a look tomorrow.

remorses:
Another solution is to wrap the mapper in another function that handle the null value semply returning null instead of calling the mapper

PR #884, opened 2020-06-17
feat: allow responsive pseudo props with object notation

### LINKED ISSUE #876: Request to add responsive values to AspectRatioBox component
**What problem does this feature solve?**
I am using AspectRatioBox as a wrapper around my hero component, to override the image initial ratio, so I'm sure the image (and the hero content) has the right ratio depending on the screen width.

**What does the proposed API look like?** 
Being able to do `<AspectRatioBox ratio={{ base: 0.75, lg: 2.3 }} />`

**Describe alternatives you've considered**
I'm currently using ThemeUI's sx prop to force paddingBottom on the before element:
```javascript
<AspectRatioBox
      w="100%"
      sx={{
        '&::before': {
          paddingBottom: [
            `${(1 / 0.75) * 100}%`,
            `${(1 / 1) * 100}%`,
            `${(1 / 1.5) * 100}%`,
            `${(1 / 1.5) * 100}%`,
            `${(1 / 2.3) * 100}%`,
          ],
        },
      }}
    >
```

santialbo:
I might have completely missed the point here. Please guide me if this can be fixed in some other way.

This would also fix #876

The isssue was that while this worked:
```
<Box _before={{paddingBottom: [1, 2, 3, 4]}}/>
```
the equivalent object notation didn't
```
<Box _before={{paddingBottom: { base: 1, sm: 2, md: 3, lg: 4 }}}/>
```

### DIFF
--- a/packages/utils/src/responsive.ts
+++ b/packages/utils/src/responsive.ts
@@ -1,7 +1,9 @@
 import { isArray, isObject } from "./assertion"
 import { Dict } from "./types"
-import { objectKeys, merge } from "./object"
+import { objectKeys } from "./object"
 import { getLastItem } from "./array"
+
+export const breakpoints = Object.freeze(["base", "sm", "md", "lg", "xl"])
 
 export function mapResponsive(prop: any, mapper: (val: any) => any) {
   if (isArray(prop)) {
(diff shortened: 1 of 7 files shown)


segunadebayo:
Hey @santialbo.

Thanks for this Draft PR. Looking through your code and I think we could attack this another way.

**The Constraint**
The `css` function could remain as is. The constraint around the `css` function is that responsive values must be arrays. I'm not sure we want to support object responsive notation considering that the `css` should also be able to process nested styles.

**The walkaround**
1. Write a function to convert responsive object notation to responsive array notation
2. Write a function to check if an object is "breakpoint-like", so the `css` can tell when it's a nested css object or responsive value object.
3. When `css` sees a breakpoint-like object, it should convert it to the array notation and process as usual.

This is just a rough thought, but could definitely be improved over time.

PS: I've just written the first function for you in the `utils/src/responsive` perhaps it can help.

santialbo:
I modified the branch with the suggested changes.
With respect to:
>  I'm not sure we want to support object responsive notation considering that the css should also be able to process nested styles.
I think it's expected to work with objects given they can be passed everywhere else (I thought it was something that was forgotten in the implementation).

packages/utils/src/responsive.ts
```diff
@@ -28,33 +30,14 @@ export function mapResponsive(prop: any, mapper: (val: any) => any) {
 }
 
 export function objectToArrayNotation(obj: Dict) {
```
santialbo, line 32: I simplified this function a bit, please tell me if you prefer the previous version or if I've missed anything. I checked the tests to understand better what it was doing.

segunadebayo: This makes sense. Thank you

santialbo:
I rebased the branch and fixed the conflicts

segunadebayo:
Thank you @santialbo, I'll review it shortly

PR #890, opened 2020-06-18
feat(toast): remove manager z-index

mrmckeb:
When testing, I found that this wasn't required - the docs, as an example, work fine with no `z-index` value.


mrmckeb:
@segunadebayo It seems like the CI failure here is unrelated to anything I did - let me know if you think otherwise though!

segunadebayo:
Thank you @mrmckeb.

I'll take a look at the CI issue and resolve it. It's definitely unrelated

PR #895, opened 2020-06-19
feat: support global tab indicator styling

sholzmayer:


segunadebayo:
Good catch @sholzmayer. Thank you 💖

PR #896, opened 2020-06-19
fix: Updated the selector for the focus-visible polyfill

ivoilic:
Changed the selector for the focus-visible polyfill to use an attribute instead of a class.
Per this update to the polyfill documentation: https://github.com/WICG/focus-visible/commit/fda80f8401807c1cbb702fb1a15b9635828bfc6d


with-heart:
@allcontributors add @ivoilic for code

PR #900, opened 2020-06-20
chore: refactor component props style computation as an interpolation

avaneeshtripathi:
Move the style calculation for component props to utils to be calculated as an interpolation.


with-heart:
@allcontributors add @avaneeshtripathi for code

akmalnawfer (commented):
Looks good. 🚀

PR #908, opened 2020-06-21
feat: ✨ Persist Dark Mode Preferences using local-storage

### LINKED ISSUE #892: Store Dark Mode preferences in Local Storage
So I can refresh the Chakra UI website & not lose my preferences. Right now, everytime I refresh it takes me back to light mode.

anarmawala:
Close #892 

✨ Uses local storage to store updated color mode preference.

Not sure if this is the best way to go about this. Would appreciate any feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "release": "lerna publish",
     "release:ci": "lerna publish -y --loglevel verbose",
     "release:next": "lerna publish --force-publish=* --no-private --preid next --dist-tag next",
+    "publish:next": "lerna publish from-package --no-private --dist-tag next",
     "docs": "yarn workspace @chakra-ui/docs",
     "docs:start": "yarn docs start",
     "core": "yarn workspace @chakra-ui/core",


remorses:
This will not work in a server environment (nextjs, gatsby) because `localStorage` is not defined

with-heart:
I could have sworn we had code in place for conditionally managing `localStorage` while checking for `Storage` existence to handle SSR.

deadcoder0904:
Yes, @remorses is right. For more info, read this awesome blog post on how to solve it: https://joshwcomeau.com/gatsby/dark-mode/

deadcoder0904:
Another problem is after refreshing there is a second where I see light mode & then it switches back to dark mode.

remorses:
This problem is not solvable if you use ssr unfortunately, to render the page with a certain mode on the first render you would need to render it in the server and get the color mode from a cookie, this is impossible to do without adding a special `getInitialProps` in nextjs or similar similar

Even if you add a `getInitialProps` in nextjs you lose the static site optimization because you need a server to check the mode preferred by the user on every request

The post Josh Comeau linked by @deadcoder0904 is a solution only if you use css variables everywhere because the page rendered in the server must be equal to the page rendered on the client and this is possible only using css variables (thing that will never happen with `chakra-ui` given that it uses `emotion` and has a lot of runtime js for styling)

The only feasible thing is to render the page with `initialColorMode` and only after the first render on the client change the color mode based on user preferences with a slow transition to not get an ugly flash

TLDR;:
This is a problem that cannot be handled in chakra-ui without having the first render always `light` or `dark`

remorses:
I think that having the transition to the preferred color mode happen after the first render is completely fine, if you want to see a cool example of a color transition check out the [MDX site](https://mdxjs.com) (try to visit the site with dark mode enabled on the OS preferences)

with-heart:
Closing this as we've reverted the API for color mode-related stuff. Also not sure this PR is still needed.

Where supported, we get/set the color mode value from `localStorage`. If no value is found, we check `theme.config.initialColorMode` to determine the default value to use. See [`useColorModeState`](https://github.com/chakra-ui/chakra-ui/blob/master/packages/color-mode/src/use-color-mode-state.ts#L54-L82) for the implementation around this. 

I believe this covers the use case(s) in your PR. Feel free to update your PR and re-open if not!

with-heart:
Ahhh, okay I had a derp moment. My explanation is not correct. We get the color mode value from `localStorage`, but not until the first execution of `useEffect` inside `useColorModeState`.

https://github.com/chakra-ui/chakra-ui/blob/ac52c0e2d42da1d048b710d60c22aea9b80896ce/packages/color-mode/src/use-color-mode-state.ts#L55-L57

I think all we need to change is the state initializer:

```tsx
React.useState(() => storage.get(options?.initialColorMode || 'light'))
```

It's also possible that changing line 62 to `useLayoutEffect` would solve this instead:

https://github.com/chakra-ui/chakra-ui/blob/ac52c0e2d42da1d048b710d60c22aea9b80896ce/packages/color-mode/src/use-color-mode-state.ts#L62-L73

@anarmawala Want to try either of these?

anarmawala:
I think `initialize-color-mode.tsx` solves the Gatsby issue already by injecting the below script into the html

```
(function() { try {
  var mode = localStorage.getItem('chakra-ui-color-mode');
  if (!mode) return
  document.body.classList.add('chakra-ui-' + mode);
} catch (e) {} })();
```

As such I think the issue is already solved. Let me know if I missed something.

Defite:
> This problem is not solvable if you use ssr unfortunately, to render the page with a certain mode on the first render you would need to render it in the server and get the color mode from a cookie, this is impossible to do without adding a special `getInitialProps` in nextjs or similar similar
> 
> Even if you add a `getInitialProps` in nextjs you lose the static site optimization because you need a server to check the mode preferred by the user on every request
> 
> The post Josh Comeau linked by @deadcoder0904 is a solution only if you use css variables everywhere because the page rendered in the server must be equal to the page rendered on the client and this is possible only using css variables (thing that will never happen with `chakra-ui` given that it uses `emotion` and has a lot of runtime js for styling)
> 
> The only feasible thing is to render the page with `initialColorMode` and only after the first render on the client change the color mode based on user preferences with a slow transition to not get an ugly flash
> 
> TLDR;:
> This is a problem that cannot be solved in chakra-ui without having the first render always `light` or `dark`, which is already implemented i think

Well, if so, then why Theme UI is perfectly deals with fouc? ;)

Defite:
> I think `initialize-color-mode.tsx` solves the Gatsby issue already by injecting the below script into the html
> 
> ```
> (function() { try {
>   var mode = localStorage.getItem('chakra-ui-color-mode');
>   if (!mode) return
>   document.body.classList.add('chakra-ui-' + mode);
> } catch (e) {} })();
> ```
> 
> As such I think the issue is already solved. Let me know if I missed something.

no, it's not.

you also need to have css-variables both client and server side.

PR #992, opened 2020-06-27
fix(preview-html): classlist null when devtools open - storybook

navin-moorthy:
When the storybook is already open with devtool open and the storybook is restarted, will result in
classlist being null error in console, because of DOM being not loaded


.storybook/preview-head.html
```diff
@@ -1,40 +1,44 @@
 <script data-id="chakra-ui-no-flash">
   ;(function () {
-    var storageKey = "chakra-ui-color-mode"
-    var classNameDark = "chakra-ui-dark"
-    var classNameLight = "chakra-ui-light"
+    window.addEventListener("DOMContentLoaded", function (e) {
```
navin-moorthy, line 3: I have only added this line of code to check if the DOMContent is loaded before running the below code.
Rest of the code being formatted by prettier.

navin-moorthy:
Thanks @with-heart for approving and @segunadebayo for merging it.

Feeling excited about my first contribution to Chakra. 🎉

with-heart:
@allcontributors please add @navin-moorthy for code

PR #988, opened 2020-06-26
docs: add icon migration notes

balibebas:
@with-heart Thanks for pointing me to the correct locations to update in #922. I made a fork this time so I can keep this branch up-to-date with `master` so the PR doesn't go stale. Please note this PR leaves off the one place I did change in #922 as it occurred to me migration notes might not be needed in that package (assuming it's new).

I also spotted some more icons feedback in the #906 today. If you'd like me to tackle some of those with this PR to make it more meaningful please don't hestitate to ask.

PR #994, opened 2020-06-27
Fix broken link

brennanfife:
Old link is broken


with-heart:
Unfortunately this PR is no longer needed due to changes made in #988. Thanks for the effort though @brennanfife!

PR #1012, opened 2020-06-28
feat(image): image lazy loading support

Andu033:


with-heart:
Thanks @Andu033! 

The prop is fine but looking at it a little more, I'm not sure this will work with `Image`, at least how it is now.

https://github.com/chakra-ui/chakra-ui/blob/ac52c0e2d42da1d048b710d60c22aea9b80896ce/packages/image/src/image.tsx#L77-L101

What does it mean to have `loading="lazy"` if we render `fallbackSrc`? To make that actually work, you'd need to pass `ignoreFallback` every time in order to get `useImage` to return `loaded` status. 

@segunadebayo, do you think we should we modify `useImage` to check for the `loaded` prop and always return `loaded` status if it exists?

segunadebayo:
Thanks for this @Andu033 

yes, it makes sense to bypass the `useImage` hook if the `loading` prop is passed, it's always better to allow the browser handle image the loading strategy.

In this scenario, we can't apply `fallbackSrc`. We'll push all work to the native `img` tag.

What do you guys think?

with-heart:
@segunadebayo That makes perfect sense. I'm going to go ahead and merge this. I created #1027 to follow up on implementing the new flow around `loading` now that it's in place.

sebqq:
Hello @Andu033 , I would like to ask you if SSR is supported?

with-heart:
@all-contributors please add @Andu033 for code

lcswillems:
@with-heart I don't understand the reasonning here

For me it makes sense to have a fallback and `loading="lazy"`. I want to have the fallback displayed whenever the image is not loaded. And I want `loading="lazy"` to trigger the loading of the image as soon as the image is visible.

PR #1023, opened 2020-06-29
fix: remove emotion core direct dependency

remorses:
If we include the `@emotion/core` package as a dependency the theme context will be duplicated, making impossible to share the `chakra-ui` theme with other components that rely on `useTheme` exported from emotion

The same happens in the inverse direction, chakra cannot use theme from other packages

Making emotion a peer dependency forces the user to only use only an instance of emotion theme context

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -25,7 +25,7 @@ one by one. All you need to do is install the `@chakra-ui/core` package:
 ```sh
 $ yarn add @chakra-ui/core
 # or
-$ npm install --save @chakra-ui/core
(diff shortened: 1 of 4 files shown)


segunadebayo:
Hi @remorses,

We really want our users to migrate to Chakra UI fully and might try to encourage them to port over their use of existing libraries.

Let's leave it as is for now, and if we ever notice any problems, we might consider adding it as a peer Dep.

Thank you.

remorses:
You won’t be able to do it later because that would be a breaking change, we do it in beta stage or probably never

PR #1033, opened 2020-06-29
docs: updated collapse props. added example for using the config prop

### LINKED ISSUE #1017: [Collapse] Duration prop is not applying
**Describe the bug** No matter what I do, changing the `duration` value on a `<Collapse />` component doesn't seem to have any effect

**Expected Behavior** ... it should have an effect

**Link to minimal reproduction** https://codesandbox.io/s/eloquent-lumiere-0glyc?fontsize=14&hidenavigation=1&theme=dark

**Steps to Reproduce**:
1. Create a `<Collapse>` and fill it with a lot of content
2. Add a `duration` prop and set to some high value, say 10 seconds (`10000`)

**Suggested solution(s)** N/A

- OS: Mac OS 10.15.5
- Browser: Chrome 83

**Additional context** N/A

jmiazga:
Added an example of the `config` prop to the Collapse docs. Also updated the props chart.

Fixes #1017

### DIFF
--- a/website/docs/components/collapse.mdx
+++ b/website/docs/components/collapse.mdx
@@ -72,16 +72,58 @@ function Example() {
 }
 ```
 
+### Customizing the transition state styles
+
+The `config` prop is used to style the Transition component's initial, entered, and 
+exiting states. This allows the height, opacity, transform, transition, or any 
+other valid CSS property to be configured for each state.
+
+```jsx
+function Example() {
+  const [show, setShow] = React.useState(false)
+
+  const handleToggle = () => setShow(!show)
+  const config = {
+    init: {
+      height: 0,
+      opacity: 0,
+    },
+    entered: {
+      height: 48,
+      opacity: 1,
+      transform: "translateY(0)",
+      transition: "height 500ms ease, opacity 500ms ease, transform 500ms ease"
+    },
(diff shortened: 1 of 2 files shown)


website/docs/components/collapse.mdx
```diff
@@ -72,16 +72,58 @@ function Example() {
...
 
+### Customizing the transition state styles
+
+The `config` prop is used to style the Transition component's initial, entered, and 
+exiting states. This allows the height, opacity, transform, transition, or any 
+other valid CSS property to be configured for each state.
+
+```jsx
+function Example() {
+  const [show, setShow] = React.useState(false)
```
with-heart, line 83: Could you replace this and the related functionality with `useDisclosure`?

segunadebayo:
I'd like to review this later tomorrow before merging.

website/docs/components/collapse.mdx
jmiazga: Yep updated every example in the Collapse docs.

segunadebayo:
Thanks @jmiazga for this PR.

I've updated the `Collapse` component as mentioned here #1017. So this fix won't be needed.

I appreciate you for taking the time to fix it.

PR #1047, opened 2020-06-30
fix(stack): default align-items css value

### LINKED ISSUE #1046: [Stack] Default CSS property `align-items` value
**Describe the bug**

`Stack` default flex layout uses `align-items: flex-start`, constraining inherent display of children.
In my case, I have a vertical `Stack` and all elements appears crushed to the left.

Capture d’écran 2020-06-30 à 15 06 33

**Expected Behavior**

HTML _block_ element inside a `Stack`  component should take 100% width.

https://codesandbox.io/s/chakra-v1-stack-vertical-fffsf

**Suggested solution(s)**
Don't specify `align-items` property at all and let developer handle it.

**Desktop**

- OS: macOS
- Browser: Chrome
- Version: 83.0.4103.116

**Additional context**
Currently migrating from `@chakra-ui/core@0.8.0` to `1.0.0-next.3`.

mr-wildcard:
Fixes #1046

### DIFF
--- a/packages/layout/src/stack.tsx
+++ b/packages/layout/src/stack.tsx
@@ -77,7 +77,7 @@ export const Stack = React.forwardRef(function Stack(
 ) {
   const {
     direction = "column",
-    align = "normal",
+    align,
     justify,
     spacing = "0.5rem",
     wrap,
--- a/packages/layout/src/stack.tsx
+++ b/packages/layout/src/stack.tsx
@@ -77,7 +77,7 @@ export const Stack = React.forwardRef(function Stack(
 ) {
   const {
     direction = "column",
-    align = "flex-start",
+    align = "normal",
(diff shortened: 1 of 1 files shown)


packages/layout/src/stack.tsx
```diff
@@ -77,7 +77,7 @@ export const Stack = React.forwardRef(function Stack(
 ) {
   const {
     direction = "column",
-    align = "flex-start",
+    align = "normal",
```
segunadebayo, line 80: I see what you mean. We definitely want to reduce breaking changes as much as we can.

I think the best thing to do in this case, is to completely remove the align prop's default value.

mr-wildcard: No problemo 👌

PR #1064, opened 2020-06-30
fix: button ignores `type` prop when `as` prop is used

### LINKED ISSUE #770: [Button] Attribute type="button" remains when as prop is used
**Describe the bug**
When using the `as` prop on the Button component (e.g. to make it an anchor) the attribute `type="button"` persists.

**Expected Behavior**
When using the Button component with the `as` prop to render it as not a button, it should render the component without `type="button"` by default.

**To Reproduce**
- Use this template: https://codesandbox.io/s/kind-mountain-wcmne?file=/src/index.tsx

**Suggested solution(s)**
Theres two options to solve this issue:

1. Remove `"button"` as default value for the type property in the button component.
2. When the  `as` prop is defined on the Button component, don't use `type="button"` by default.

**Additional context**
I've checked the `dev-ts` branch and the issue also occurs in there. I'm glad to help by creating a PR! Let me know what you think would be the best solution.

jmiazga:
The `type` prop has a default of `button` which is passed down even if the `as` prop is used. Not sure if there are any edge cases I need to account for, like is there ever a time we would expect `as` and `type` to both be used at the same time?

Fixes #770

### DIFF
--- a/packages/button/src/button.tsx
+++ b/packages/button/src/button.tsx
@@ -92,6 +92,7 @@ export const Button = forwardRef<ButtonProps>(function Button(props, ref) {
     colorScheme = group?.colorScheme,
     size = group?.size,
     className,
+    as,
     ...rest
   } = props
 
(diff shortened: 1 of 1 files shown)


segunadebayo:
This make sense!

PR #869, opened 2020-06-16
Update z-index docs and remove most defaults

mrmckeb:
I've added to the docs around z-index and removed most of the default values, in the hope that this will encourage people to research and understand the alternatives.

Please feel free to make whatever changes you feel necessary to this.

PR #1084, opened 2020-07-01
fix(modal): add missing z-index to overlay and content

mr-wildcard:
AFAIK there isn't any `z-index` applied to `Modal` elements in v1 so I used the v0 way of adding them.

Note that I put `zIndex` declaration after user's props (`...props`) so that `z-index` will always output values from styled system. 👍 / 👎 ? :)


with-heart:
We removed most of the default z-index values as part of #869. There's more information about managing z-index here: https://next.chakra-ui.com/docs/z-index

mr-wildcard:
oops, didn't aware of this shift ;) I think we can close my PR.

PR #1126, opened 2020-07-04
docs(tooltip): update tooltip docs with bug fixes

navin-moorthy:
In this round, I updated the docs for `tooltip` and fixed some bugs that I see on the way.


packages/tooltip/src/tooltip.tsx
```diff
@@ -92,12 +92,12 @@ export const Tooltip = React.forwardRef(function Tooltip(
...
 
   /**
-   * If the `label` or `aria-label` is empty, there's no
+   * If the `label` is empty, there's no
    * point showing the tooltip. Let's simply return back the children
    *
    * @see https://github.com/chakra-ui/chakra-ui/issues/601
    */
-  if (!(label || ariaLabel)) {
+  if (!label) {
```
navin-moorthy, line 100: When `label` is not given but only the `aria-label`, I get the below tooltip
image

Fixed it by only checking the label. Since tooltip is rendered when aria-label is given but not label

packages/tooltip/src/use-tooltip.ts
```diff
@@ -26,11 +26,11 @@ export interface UseTooltipProps {
   /**
    * Callback to run when the tooltip shows
    */
-  onShow?(): void
+  onOpen?(): void
```
navin-moorthy, line 29: Our previous version has `onOpen` and `onClose` instead of `onShow` and `onHide`.

Throughout the package I can see both show & hide and open and close, I think it's better we have a unified naming here to clear the confusion

packages/tooltip/src/use-tooltip.ts
```diff
@@ -60,12 +60,12 @@ export interface UseTooltipProps {
 
 export function useTooltip(props: UseTooltipProps = {}) {
   const {
-    showDelay = 0,
-    hideDelay = 0,
+    showDelay = 200,
+    hideDelay = 200,
```
navin-moorthy, line 64: Based on the Typescript Interface defaults,
image

I updated the defaults

packages/tooltip/src/use-tooltip.ts
```diff
@@ -97,31 +97,31 @@ export function useTooltip(props: UseTooltipProps = {}) {
...
       clearTimeout(enterTimeoutRef.current)
     }
-    exitTimeoutRef.current = setTimeout(onClose, hideDelay)
+    exitTimeoutRef.current = setTimeout(_onClose, hideDelay)
   }
 
   const onClick = () => {
     if (closeOnClick) {
-      onClose()
+      closeWithDelay()
```
navin-moorthy, line 112: I ran into a bug caused by delay open and close immediately,
2020-07-04 at 11 53 16 - Lavender Alpaca

Fixed it by changing the `onClose` to `closeWithDelay` on all handlers, since that clears out the `open timer`.
This is caused because of the even propagation from mousedown to focus on click.

packages/tooltip/src/use-tooltip.ts
with-heart: I like this change because it also unifies `useTooltip` with an api similar to `useDisclosure`. @segunadebayo what do you think?

packages/tooltip/src/use-tooltip.ts
with-heart: If we make the above change to `onOpen`/`onClose`, we should probably change these to `openDelay` and `closeDelay` as well.

packages/tooltip/src/use-tooltip.ts
```diff
@@ -74,11 +74,11 @@ export function useTooltip(props: UseTooltipProps = {}) {
...
   } = props
 
-  const { isOpen, onOpen, onClose } = useDisclosure({
+  const { isOpen, onOpen: _onOpen, onClose: _onClose } = useDisclosure({
     isOpen: isOpenProp,
     defaultIsOpen,
-    onOpen: onShow,
-    onClose: onHide,
+    onOpen: onOpen,
+    onClose: onClose,
```
with-heart, line 81:
```suggestion
    onOpen,
    onClose,
```

navin-moorthy: Sure, I will wait until we get the feedback from @segunadebayo on changing the prop names.

packages/tooltip/src/use-tooltip.ts
segunadebayo: This change is good. I definitely want us to keep same API as the previous version.

Thank you.

packages/tooltip/src/use-tooltip.ts
segunadebayo: Please update the js doc and default value to `0`, that's what we hard in the previous version. 

I guess I skipped updating it. Thank you

packages/tooltip/src/use-tooltip.ts
```diff
@@ -74,11 +74,11 @@ export function useTooltip(props: UseTooltipProps = {}) {
     modifiers,
   } = props
 
-  const { isOpen, onOpen, onClose } = useDisclosure({
+  const { isOpen, onOpen: _onOpen, onClose: _onClose } = useDisclosure({
```
segunadebayo, line 77: Instead of renaming the functions to `_onOpen` and `_onClose`. I'll prefer it if you renamed the prop names.

`onOpen: onOpenProp`
`onClose: onCloseProp`

navin-moorthy: I will make these changes @segunadebayo, Thanks.

packages/tooltip/src/use-tooltip.ts
navin-moorthy: Done. I have made these changes.

packages/tooltip/src/use-tooltip.ts
navin-moorthy: Done, I made these changes.

navin-moorthy:
@with-heart @segunadebayo  I think I messed up with the rebase and merge to resolve the conflict.
If you guys wanted a new pull request I am happy to do so.

with-heart:
@navin-moorthy Oooooh that's a rough spot. You could try undoing the rebase!

Run `git reflog` from this branch and look for the `HEAD@{#}` pointer for the last commit before you made before the rebase. Then run `git reset --hard HEAD@{#}` and it'll take the branch back to how it was before you rebased.

navin-moorthy:
> @navin-moorthy Oooooh that's a rough spot. You could try undoing the rebase!
> 
> Run `git reflog` from this branch and look for the `HEAD@{#}` pointer for the last commit before you made before the rebase. Then run `git reset --hard HEAD@{#}` and it'll take the branch back to how it was before you rebased.

Thanks you so much for the detailed step on how to revert back.
This time I merged it 😆, no more rebase.

navin-moorthy:
@with-heart @segunadebayo  I only made the changes to the docs, did not see the docs since I am having trouble starting the docs as informed in the discord.

Make sure to just check if everything is looking great in the docs before merging or we can wait until we fix the bug with starting the docs, so I can check and confirm.

navin-moorthy:
@segunadebayo  I was able to run the docs and verify my changes, everything look great from my end.

PR #1140, opened 2020-07-07
fix: getMediaQuery sort was returning NaN, which behaved differently …

### LINKED ISSUE #1137: [General] Cross-browser inconsistencies in media-queries
Opening on behalf of pablo on Discord.

Breakpoints are differently computed, apparently based on browser?

**Expected Behavior** Breakpoints are consistent across browsers.

https://codesandbox.io/s/chakra-ui-media-issue-l1kex?file=/src/index.js

**Steps to Reproduce** 

Open in both Chrome and Firefox.
Chrome works as expected. XL > LG > MD > SM.
Firefox renders:
- MD above 1280px
- SM and LG below 1280px
- SM and XL below 992px
- SM below 768px

jmiazga:
Fixes #1137 

Ok so this was a fun one. The compare function in `sort` was returning `NaN`. That is because `'30em' - '48em'` results in `NaN`.

In Chrome nothing happened to the order of the array after `sort` was executed, but in Firefox the array was reversed.

### DIFF
--- a/packages/parser/src/utils/get-media-query.ts
+++ b/packages/parser/src/utils/get-media-query.ts
@@ -21,12 +21,7 @@ export function getMediaQuery(breakpoints?: Dict, mapper = toMediaQuery) {
 
   const asArray = objectKeys(_breakpoints)
     .map((key) => _breakpoints[key])
-    .sort((a, b) => {
-      const regex = /^\d+/
-      const aMatch = a.match(regex)
-      const bMatch = b.match(regex)
-      return aMatch[0] - bMatch[0]
-    })
+    .sort((a, b) => parseInt(a, 10) - parseInt(b, 10))
     .map(mapper)
 
   const asObject = objectKeys(_breakpoints).reduce((result, point) => {
--- a/packages/parser/src/utils/get-media-query.ts
+++ b/packages/parser/src/utils/get-media-query.ts
@@ -21,7 +21,12 @@ export function getMediaQuery(breakpoints?: Dict, mapper = toMediaQuery) {
 
   const asArray = objectKeys(_breakpoints)
     .map((key) => _breakpoints[key])
-    .sort((a, b) => a - b)
+    .sort((a, b) => {
+      const regex = /^\d+/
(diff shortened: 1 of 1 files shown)


packages/parser/src/utils/get-media-query.ts
```diff
@@ -21,7 +21,12 @@ export function getMediaQuery(breakpoints?: Dict, mapper = toMediaQuery) {
 
   const asArray = objectKeys(_breakpoints)
     .map((key) => _breakpoints[key])
-    .sort((a, b) => a - b)
+    .sort((a, b) => {
+      const regex = /^\d+/
```
ljosberinn, line 25: I'm unaware under which circumstances this is called so please forgive me if this is wrong, but but I'd assume this is performance sensitive so we might be able to gain a small win here by having the regex outside of the compare function, or even extract the whole sort-function to its own definition?

with-heart, line 25: Agree with @ljosberinn, these changes are perfect if you at least define `regex` outside of the function call so it's only created once. I'm not entirely sure of the performance implications, but we might as well avoid it anyway!

with-heart:
Interestingly enough, I know that @segunadebayo is reworking the `css` and `parser` packages into a `styled-system` package that reuses a lot of `styled-system` internals for performance reasons, and this functionality no longer exists in this form there.

You can see the new `css` function (which is where `getMediaQuery` was used previously) in the `benchmarks` branch: https://github.com/chakra-ui/chakra-ui/blob/benchmarks/packages/styled-system/src/css.ts

jmiazga:
@with-heart Should I make the changes or will this PR bc closed bc of the work on the `benchmarks` branch?

jmiazga:
Ok I pushed a fix that I like better than the original. Not sure the timing of the `benchmarks` branch, but this will at least fix array responsive prop issues that people are having.

segunadebayo:
As mentioned in #1137. I'll close this in favor of using styled-system's core functions (not styled system itself)

PR #1141, opened 2020-07-07
run `storybook` without having to run `build`

remorses:
This pull request lets you run `storybook` and `test` without having to run `build`, this makes contributing much easier given that the build step takes about 10 minutes

Now all the packages import from the `src` directory and the compilation is handled by storybook so you don't have to run `yarn build`.

Obviously you cannot import from `src` in all other cases, so i added a babel plugin that removes `/src` from all chakra imports in the distributed packages

I know this pull request may look scary because of the many lines changed but the 99% of these changes have been made just replacing `@chakra-ui/{package}` with `@chakra-ui/{package}/src', also i already have tested that build works and typescript types are fine

I made this pull request because build takes 10 minutes and i wasn't able to contribute anymore because of this, this changes makes the development cycle 20x faster and now i can return to contribute to this awesome package! ❤️

PS: to generate the documentation you still have to run `yarn build` before, i can apply the same changes there if you want

PS: given that there are a lot of changed lines it's better to merge this as soon as possible or i won't be able to resolve conflicts


with-heart:
Love this idea! I'll review this asap

with-heart:
The first thing that popped into my mind was worrying that intellisense would suggest the wrong thing. I deleted `node_modules` and ran `yarn` again, and verified that the first suggestion in VSCode is always `/src` for packages. The suggestion for the base package still exists, but it is always the second suggestion.

image

I wonder if there's a way to prevent that duplication and only allow it to resolve `/src`. Even if we can't, I think that's probably worth the hassle.

remorses:
@with-heart the auto complete double suggestions happen only on the internal code, right? After this pull request you should only import from the `/src` directories in internal code but the package consumer will consume packages as always, importing from the `/dist` directory

remorses:
There are too many merge conflicts, I need to reopen this pull request and merge it before you make any changes to `develop` because i need to change every import path to the `/src` directories

@segunadebayo can you give me a date and time to create another pull request you can merge right after?

with-heart:
> @with-heart the auto complete double suggestions happen only on the internal code, right? After this pull request you should only import from the `/src` directories in internal code but the package consumer will consume packages as always, importing from the `/dist` directory

Yeah, it's only happening on internal code, but it's still a concern because we have a lot of contributors, many of whom might not know about our unique setup. I imagine it won't be much of an issue but it was something worth thinking about.

jmiazga:
I feel like there has to be an easier solution somehow. Yarn workspaces + typescript + storybook doesn't seem that unique. I wonder how other people have solved this?

segunadebayo:
Taking a look at the updates, I think it looks great. Thank you @remorses

However, I'm a bit hesitant to merge this considering the significant changes it makes to the workflow, and its kind of against the typical workflow I'm familiar with.

I really want you to contribute to the code because I value your contributions. I'm just not sure if this is the best approach to solve this build problem. I'd rather wait for your Storybook alternative and use that instead (btw, I'm excited about it)

I can't a specific date right now as there still some things to clean up, but we can consider talking about this again once V1 is live.

Don't get me wrong, I want the build and dev experience to be 10X faster (I mean, that's the Chakra way). Permit me to hold off on this one for now till we find a better solution or approach.

dvh91:
I just stumbled here from twitter and it looks like an interesting idea!
It can be easier to review and merge if you'll use `babel-plugin-transform-imports` and change the imports in the babel config of the running storybook.

Take a look at Material UI docs for different use case but same practice:
https://material-ui.com/guides/minimizing-bundle-size/#option-2

remorses:
@segunadebayo this approach is not very popular but is perfectly functional:

What this pull request really does is:
- replace all the `from '@chakra-ui/core'` with `from '@chakra-ui/core/src'`
- add a babel plugin that transform again `from '@chakra-ui/core/src'` to `from '@chakra-ui/core'` when building the ditributed version

The only changes to the workflow will be that you now import from `/src` instead of `/dist`

This little change lets you skip 10 minutes of building every time you have to run `yarn storyboook`, `yarn test` and `yarn docs`

This also prevents a lot of bugs where you have stale code inside the `dist` directories

If you want a more smooth road to V1 please let add these changes, i will open a draft pull request and wait to add all the changes to the import paths until you are ready to merge


@dvh91 if you don't change the import paths you still need to run build to have the typings inside the `/dist` directory, so there is no benefit

remorses:
Another problem that i had are circular dependencies:

If a package A depends on B and B depends on A you cannot run `yarn build` because `tsc` needs to have the other packages  types available to be able to compile

This would be fixed with this pull request

remorses:
Moved to  #1173

PR #1167, opened 2020-07-11
docs(switch): wrap usage code with FormControl

with-heart:
This PR fixes a docs build failure due to `FormLabel` requiring a surrounding `StylesProvider`.

### DIFF
--- a/website/docs/form/switch.mdx
+++ b/website/docs/form/switch.mdx
@@ -31,10 +31,10 @@ import { Switch } from "@chakra-ui/core"
 ## Usage
 
 ```jsx
-<Flex justify="center" align="center">
+<FormControl as={Flex} justifyContent="center" alignItems="center">
   <FormLabel htmlFor="email-alerts">Enable email alerts?</FormLabel>
   <Switch id="email-alerts" />
-</Flex>
+</FormControl>
 ```
 
 ## Sizes


website/docs/form/switch.mdx
```diff
@@ -32,8 +32,10 @@ import { Switch } from "@chakra-ui/core"
 
 ```jsx
 <Flex justify="center" align="center">
```
segunadebayo, line 34: Can we move the flex styles to `FormControl` here

with-heart: I'm noticing something interesting here. I've switched it to `FormControl as={Flex}` in order to enable flex props on `FormControl`, but if I use `justify` or `align` it's ignoring those props. I have to use `justifyContent` and `alignItems` to make it work correctly. Is that expected?

with-heart:
I'm noticing two things as I remove `Flex` for the "Usage" example:

1. I've switched it to `FormControl as={Flex}` in order to enable flex props on `FormControl`, but if I use `justify` or `align` it's ignoring those props. I have to use `justifyContent` and `alignItems` to make it work correctly. Is that expected?

2. I believe the `Switch` isn't quite centered with the label text.

image

segunadebayo:
@with-heart, it's not expected but we can look into this post v1. It's most likely related to the `shouldForwardProp` function in `@emotion/styled`

with-heart:
@segunadebayo I'll make an issue without a milestone so we can revisit later

PR #1171, opened 2020-07-11
chore(color): make color hex to uppercase for consistent

neighborhood999:


with-heart:
@allcontributors please add @neighborhood999 for code

PR #1172, opened 2020-07-11
refactor(system): storing color mode in the system package

### LINKED ISSUE #1147: Store color mode value in theme
Currently we store the color mode value in a react context

This approach has a few problems that limit the extensibility with other packages
- when other packages use another color mode context the state is duplicated in multiple places
- if other UI packages want to know the current color mode they must have a peer dependency on `@chakra-ui/color-mode`
- if there are multiple installs of `chakra-ui` the color mode context is duplicated and color mode will stop working

We could instead use the emotion `ThemeContext` to store the color mode value, this way `chakra-ui` color mode can easily be used by other UI packages without having to depend on `chakra-ui`

Obviously there would be no change on the current API as we can just change the `useColorMode` and `useColorModeValue` functions to get the value from theme instead of context

remorses:
This is a work in progress

After this change the `colorMode` and `toggleColorMode` would be stored in the theme object so other packages can use them without having to depend on `@chakra-ui`

Closes #1147


packages/system/src/providers.tsx
```diff
@@ -78,3 +96,69 @@ export function GlobalStyle() {
     />
   )
 }
+
+/**
+ * React hook that reads from `ColorModeProvider` context
```
with-heart, line 101:
```suggestion
 * React hook that reads from `ThemeProvider` context
```

with-heart:
Looks good so far! I think this is a sensible change.

PR #1173, opened 2020-07-11
refactor: run `storybook` without having to run `build`

remorses:
**What does this pull request does?**:
- replace all the `from '@chakra-ui/core'` with `from '@chakra-ui/core/src'`
- add a babel plugin that transform again `from '@chakra-ui/core/src'` to `from '@chakra-ui/core'` when building the ditributed version

**What are the benefits?**
- skip 10 minutes of building every time you have to run `yarn storyboook`, `yarn test` and `yarn docs`
- prevents bugs where you have stale code inside the `dist` directories, for example when you compile a package but not its dependencies
- you can now have circular dependencies between `chakra-ui` packages
- vscode typescript plugin is faster because it has to parse less files and these files don't change

The last commit to do is replace all import paths from '@chakra-ui/pkg' to '@chakra-ui/pkg/src' but i will wait until @segunadebayo is ready to merge to prevent merge commits


segunadebayo:
I discussed this with the team and we're convinced to go with this approach given the effort, and it also feels like a deviation from the norm workflow.

In the meantime, we've set up scripts in the root package.json to help others get started. Build might take a while as you said, but it's something we can bear with for now.

We can reconsider this in the future. I really appreciate the effort put into this.

Issue #1180, opened 2020-07-11
[Slider] Incorrect step movement


Geeman201:
**Describe the bug**
Affects both pre V1 & V1 
When setting a `step` and a `min` value the slider doesn't behave the same way as the "browser" `number` input.

With the following slider code:
```
 <Slider min={1} max={7} step={2}>
```

The slider increments: 
`1, 2, 4, 6, 7`

**Expected Behavior** 
The slider should increment 
`1, 3, 5, 7`

**Link to minimal reproduction**

[![Edit serene-gates-8mkmu](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/serene-gates-8mkmu?fontsize=14&hidenavigation=1&theme=dark)

**Suggested solution(s)**
Happy to raise a PR but wanted to check this behaviour wasn't deliberate 

**Desktop (please complete the following information):**

- OS: OSX
- Browser All browser

with-heart:
This is weird in that I'm seeing different (thought still incorrect) behavior from what you reported. On macOS with Chrome 83.

When incrementing, it goes 1, 4, 6, 7. When decrementing, it goes 7, 6, 4, 2, 1.

Geeman201:
Interesting, I am on Chrome 83 - but also verified it on safari & firefox. It's quite easy to miss a number if you drag to far.

Do you want me to raise a PR ? It looks as though `roundValueToStep` needs to also be aware of the `min` value.

with-heart:
@Geeman201 Sure! If you'd like to submit a PR, we'd love that :)

PR #1185, opened 2020-07-11
fix: #1180 slider moving incorrectly when min is odd and step is even

Geeman201:
Fixes issue #1180


with-heart:
@allcontributors please add @Geeman201 for code and test

Issue #1191, opened 2020-07-12
Equalize `nextjs` js and ts examples


with-heart:
The `nextjs-typescript` example is a more complex example than its js counterpart but doesn't use any Chakra components, and the two also have much different content. The examples should be equalized so that you get the same basic example regardless of whether you're using js or ts.

Geeman201:
I'm happy to work on this one 👍 What kind of content do you want on both of them?

with-heart:
I have this PR open for creating basic `create-react-app` templates, and they have basic content inspired by the default `create-react-app` content, along with a color mode switcher: https://github.com/chakra-ui/chakra-ui/pull/936

imo it should be something similar to that, especially with the color mode switcher. Simple enough for users to understand how the package is set up and configured and how to use color mode in their components, without becoming too complex.

I'm open to ideas though!

Also feel free to join us on Discord if you'd like :) https://discord.gg/hUGpfa8

Geeman201:
Cool - I think keeping them aligned is a "good enough" thing for v1 of this 👍 ?

with-heart:
Agreed!

PR #1206, opened 2020-07-13
Feat: #1191 nextjs typescript

Geeman201:
Towards #1191


Geeman201:
Part 1 of 2 - want to get this right before replicating in normal JS.. 

Couple of things i'm unsure of: 
1. README looks like the pre-generated one from nextjs - do we want something more chakra specific there?
1. I copied a lot of the stuff from your CRA template. Wasn't sure if we wanted something slightly more "next" based. But I think its actually a good simple demo.. 
1. Do we want to add any tests?
1. We should probably sync these with the nextjs examples repo.

with-heart:
@Geeman201 Excellent work! This is looking great!

1. I'm not sure honestly. I feel like it's probably better off with the default `README.md` if people want to copy this directly. Possibly add a link to Chakra itself?
2. The only thing that jumps out to me that's especially important is showing how to use next's `Link` with Chakra's `Link`.
3. I think we're okay without tests. Probably overkill for an example that isn't doing anything.
4. There's a nextjs examples repo? 🤣

with-heart:
Sorry I haven't had a chance to review this yet @Geeman201. It's on my list for today!

examples/nextjs-typescript/components/Layout.tsx
```diff
@@ -1,41 +1,51 @@
...
-    </footer>
+    <Container maxWidth="1200px">
+      <header>
+        <Flex py={4} justifyContent="space-between" alignItems="center" mb={8}>
+          <Flex justifyContent="space-between" alignItems="center">
+            <nav>
+              <HStack spacing={12}>
+                <NextChakraLink
+                  href="/"
+                  passHref={true}
```
with-heart, line 31: Shouldn't need to set this since `passHref` is hardcoded to `true` for `NextLink` in `NextChakraLink.tsx`.

examples/nextjs-typescript/components/NextChakraLink.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { PropsWithChildren } from "react"
+import NextLink from "next/link"
+import { LinkProps as NextLinkProps } from "next/dist/client/link"
+import {
+  Link as ChakraLink,
+  LinkProps as ChakraLinkProps,
+} from "@chakra-ui/core"
+
+export type NextChakraLinkProps = PropsWithChildren<
```
with-heart, line 9: This type should include `& Omit<NextLinkProps, "passHref">` because it's hard-coded to `true` and it's not a valid prop to pass to `ChakraLink`.

examples/nextjs-typescript/pages/properties/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { properties } from "../../utils/sample-data"
+import { PropertySummary } from "../../components/PropertySummary"
+import { NextChakraLink } from "../../components/NextChakraLink"
+
+const PropertiesPage = () => (
+  <Layout title="Next.js + TypeScript example | View properties">
+    <Heading mb={4}>Available this weekend</Heading>
+    <Wrap>
+      {properties.map((property) => (
+        <NextChakraLink
```
with-heart, line 12: I think we should iterate on this (not in this PR, just sometime later) to make the summaries function as [semantic breakout buttons](https://hankchizljaw.com/wrote/create-a-semantic-breakout-button-to-make-an-entire-element-clickable/). This is totally fine for now though!

with-heart:
Along with rebasing, can you re-run `yarn` and check in `yarn.lock`? I think it's currently missing that you installed `faker`/`@types/faker`.

examples/nextjs-typescript/pages/properties/index.tsx
Geeman201: Like that idea! 

TIL the word for those kind of buttons - Thanks

Geeman201:
Rebased up from `master` instead of `develop` 

Closing in favour of #1321

PR #1226, opened 2020-07-15
chore: use theme provide radius for outline button

neighborhood999:
We can use defined borders for the outline button.

### DIFF
--- a/packages/theme/src/components/button.ts
+++ b/packages/theme/src/components/button.ts
@@ -73,12 +73,12 @@ function ghost(props: Props) {
 }
 
 function outline(props: Props) {
-  const { colorScheme: c } = props
+  const { colorScheme: c, theme } = props
   const borderColor = mode(`gray.200`, `whiteAlpha.300`)(props)
 
   return {
     container: {
-      border: "1px solid",
+      border: theme.borders['1px'],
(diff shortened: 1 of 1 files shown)


packages/theme/src/components/button.ts
```diff
@@ -73,12 +73,12 @@ function ghost(props: Props) {
...
 
 function outline(props: Props) {
-  const { colorScheme: c } = props
+  const { colorScheme: c, theme } = props
   const borderColor = mode(`gray.200`, `whiteAlpha.300`)(props)
 
   return {
     container: {
-      border: "1px solid",
+      border: theme.borders['1px'],
```
segunadebayo, line 81: This is no longer necessary, I think the best solution would be to use `border : 1px` since this will automatically resolve to the theme's border value.

neighborhood999: That's so cool, I can't imagine this way can work haha. 🤯

So, should I need to update related to the above mentioned `border: 1px`?

segunadebayo: That's exactly why I left it that way. Haha

I'll close this PR for now considering you now understand what's happening.

Thank you, man.

PR #1243, opened 2020-07-15
fix: added ColorModeOptions type to theme config

### LINKED ISSUE #1223: [Theme] Sync theme.config types with ColorMode
**Describe the bug**
Right now, the theme.config property [is not typed](https://github.com/chakra-ui/chakra-ui/blob/master/packages/theme/src/index.ts#L8) which leads to `theme.config.initialColorMode` being `string`, which is incompatible when used against the `ColorMode` type.

image

**Suggested solution(s)**

Simply adding and exporting a ThemeConfig definition should solve this:

```ts
import { ColorMode } from '@chakra-ui/core';

export interface ChakraConfig {
  initialColorMode: ColorMode;
  useSystemColorMode: boolean;
}

const config: ChakraConfig = {
  initialColorMode: 'dark',
  useSystemColorMode: false,
};
```

jmiazga:
Fixes #1223 

This PR adds the `ColorModeOptions` type to the theme config.

### DIFF
--- a/.changeset/angry-buses-camp.md
+++ b/.changeset/angry-buses-camp.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/color-mode": patch
+"@chakra-ui/theme": patch
+---
+
+Added `ColorModeOptions` type to `theme.config` and `useColorModeState`.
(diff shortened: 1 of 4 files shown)


jmiazga:
@with-heart Should I have done something to generate that changeset?

with-heart:
> @with-heart Should I have done something to generate that changeset?

Gonna document that in #1228 soon!

PR #1269, opened 2020-07-17
[WIP]: Improve `orientation` prop support

### LINKED ISSUE #1187: [Tabs] Improve `orientation` prop support
The `Tabs` component technically supports the `orientation` prop (values of `vertical` or `horizontal`), but the effect it has is to switch between `keyDown` handlers for `ArrowUp`+`ArrowDown` or `ArrowLeft`+`ArrowRight` and to alter the `aria-orientation` prop. The default `theme` provides no visual support for the `vertical` orientation, and the `orientation` prop is entirely undocumented.

- [ ] Add `theme` support for `vertical` tabs
- [ ] Document the `orientation` prop

rockchalkwushock:
Closes #1187

@segunadebayo @with-heart 

This is still very much a work in progress.

1. I'm confused about how `orientation` gets addressed within the `theme` as we merge `Tabs` into the default theme.
2. Still needs to have some tests written to verify toggling between the `orientations` gives the desired results.

### DIFF
--- a/packages/theme/src/components/tabs.ts
+++ b/packages/theme/src/components/tabs.ts
@@ -5,9 +5,11 @@ import {
   mode,
   Sizes,
   Variants,
+  Props,
 } from "@chakra-ui/theme-tools"
 
 const register = {
+  orientation: ["horizontal", "vertical"],
   parts: ["tablist", "tab", "tabpanel", "indicator"],
   sizes: ["sm", "md", "lg"],
   variants: [
@@ -40,6 +42,18 @@ const baseStyle: BaseStyle<typeof register> = (props) => {
     tablist: { justifyContent: alignments[align] },
     tabpanel: {
       padding: 4,
+    },
+  }
+}
+
+const orientation = (props: Props) => {
+  const { orientation } = props
+  return {
+    tablist: {
+      [orientation === "vertical"
+        ? "borderRight"
+        : "borderBottom"]: "2px solid",
+      borderColor: "inherit",
     },
   }
 }


segunadebayo:
@rockchalkwushock, kindly change the base branch to develop.

with-heart:
I wonder if there's a way to reset the `changeset-bot` when a PR is updated, like how `lgtm-com` was able to see that the base branch changed and updated its message. I'll look into that, or open an issue in the `changesets` repo if that functionality doesn't exist.

with-heart:
> 1. I'm confused about how `orientation` gets addressed within the `theme` as we merge `Tabs` into the default theme.

Hmm, could expand on this? I'm not sure what you're confused by, so I'm not sure how to help you understand that better.

rockchalkwushock:
@with-heart I apologize for the delay in getting back to you. I had no time to be at the keyboard this weekend. Perhaps I am just not seeing it but in looking at the data structure for the `theme` their is a `components` object that we are spreading the default styles for each component in the library. What I wasn't finding was how `orientation` fits into the component-level data structure.

```ts
// @chakra-ui/theme

const theme = {
  ...foundations,
  components,
  styles,
  config,
}

// @chakra-ui/tabs

// All of the below seem to be the case for all components but `orientation` is not used by all components.
// Is adding it to this object okay or does it need to be spread in somewhere else?
// I don't see a component level theme typing, that's where I am getting a little confused.

const tabs = {
  register,
  defaultProps,
  baseStyles,
  sizes,
  variants,
}
```

segunadebayo:
Thank you for this PR @rockchalkwushock. I just pushed an update taking ideas for this PR, and added some other style updates as well.

Issue #1146, opened 2020-07-08
[Select] Not enough padding for icon


dodas:
I expected the `Select` component to be inline by default, like native select. That is, occupy only width needed to display the longest option.

When I try to achieve that by wrapping the `Select` in `<Box display="inline-block">`, I end up with the chevron icon squished to the select text.

I tried adding the padding myself: `<Select pr={4}>`, but it adds it to wrapper element, so the icon ends up out of border.

Chakra UI version: both both pre-1.0 and 1.0 next.4.

Thanks.

<img width="509" alt="Screenshot 2020-07-09 at 01 53 39" src="https://user-images.githubusercontent.com/14360171/86981491-095c2180-c187-11ea-84ed-8ea7d22fcea7.png">

jmiazga:
I did some digging into this. The [select field theme](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/theme/src/components/select.ts) has `paddingRight: 2rem` but this value does not seem to get applied to the `SelectField` component. The `SelectField` paddingRight prop is 4. I think this prop is coming from the paddingRight defined in sizes.

trevorblades:
Good catch @jmiazga. I think you're on to something regarding the styles defined in `sizes` overriding other custom styles. I noticed similar behaviour with the `Heading` component (#1242) and I wonder if they're related in some way. Could be the result of how styles are being composed from `sizes` and other sources.

I'll poke around in the code some more and see if I can find any useful info.

segunadebayo:
I just resolve the issue with the Heading. Styles passed via props have more priority now.

Not sure if that would fix this Select issue but I'll look into it

segunadebayo:
Thanks to @neighborhood999. We've just resolved it.

PR #1273, opened 2020-07-17
fix(select): add padding on select with arrow icon

neighborhood999:
fixes: #1146

I encountered the same problem as issue #1146 and I found we don't overwrite padding from the input component.

### Screenshot
螢幕快照 2020-07-17 下午2 37 46

Sorry for not familiar TypeScript...😣


segunadebayo:
Thanks for catching this @neighborhood999. I've just used the solution from your PR and added types as well.

Thank you

PR #1314, opened 2020-07-19
feat: added storageManager prop to ColorModeProvider

jmiazga:
@segunadebayo I wanted to get the conversation started around injecting the color mode storage mechanism into `ColorModeProvider`.

### DIFF
--- a/packages/color-mode/src/index.ts
+++ b/packages/color-mode/src/index.ts
@@ -1,3 +1,4 @@
 export * from "./color-mode"
 export * from "./initialize-color-mode"
 export type { ColorModeOptions } from "./use-color-mode-state"
+export { localStorageManager } from "./color-mode.utils"
--- a/packages/color-mode/src/index.ts
+++ b/packages/color-mode/src/index.ts
@@ -1,4 +1,5 @@
 export * from "./color-mode"
 export * from "./initialize-color-mode"
 export type { ColorModeOptions } from "./use-color-mode-state"
-export { localStorageManager } from "./color-mode.utils"
+export { cookieStorageManager, localStorageManager } from "./color-mode.utils"
+export type { StorageManager } from "./color-mode.utils"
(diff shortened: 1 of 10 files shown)


packages/color-mode/src/index.ts
```diff
@@ -1,3 +1,5 @@
 export * from "./color-mode"
 export * from "./initialize-color-mode"
 export type { ColorModeOptions } from "./use-color-mode-state"
+export { cookieStorageManager, localStorageManager } from "./color-mode.utils"
+export type { StorageManager } from "./color-mode.utils"
```
with-heart, line 5: You could put these in `storage-manager.ts` instead so then you can have a barrel export for that file instead. Would make it easier to manage the exports than juggling which pieces from `utils` are exported.

with-heart:
Where are we with this? Is this something we want to try to include for 1.0 release? @segunadebayo @jmiazga

jmiazga:
@with-heart I just pushed some updates to make cookies work with SSR on next-js. Check out this commit and let me know your thoughts: e7df198065f041433c9f46f36b1990fc7e90a5de

Once the code gets finalized I need to add something to the docs.

with-heart:
@jmiazga told me earlier that he confirmed that this prevents the flash 😄

with-heart:
One thing that does suck for Next is that this breaks [Automatic Static Optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization) due to the use of `getInitialProps`. I don't think there's any way around that though.

There's more discussion around this in #349. Luckily I think this will at least effectively close that issue.

jmiazga:
Yeah it does break automatic static optimization, but at least SSR still works. I feel like solutions like this https://github.com/chakra-ui/chakra-ui/issues/349#issuecomment-665114552 would give the user a worse experience than SSR would.

with-heart:
@jmiazga Agreed

Issue #1317, opened 2020-07-20
[Link] disabled link but still can navigate to another page


neighborhood999:
![Kapture 2020-07-20 at 15 27 16](https://user-images.githubusercontent.com/10325111/87911054-919ec880-ca9d-11ea-9071-f26feea6b9bc.gif)

I think the correct behavior will not navigate to another page, I will create a PR try to fix.

PR #1318, opened 2020-07-20
fix(link): prevent navigate when link is disabled

neighborhood999:
fixes: #1317 

When link is disabled and user try to click (focus), it will not navigate to another page.

### DIFF
--- a/packages/theme/src/components/link.ts
+++ b/packages/theme/src/components/link.ts
@@ -4,28 +4,30 @@ const register = {
   parts: ["link"],
 } as const
 
-const baseStyle: BaseStyle<typeof register> = {
-  link: {
-    transition: `all 0.15s ease-out`,
-    cursor: "pointer",
-    textDecoration: "none",
-    outline: "none",
-    color: "inherit",
-    _hover: {
-      textDecoration: "underline",
-    },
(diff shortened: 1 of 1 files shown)


with-heart:
Unfortunately this is only half the battle because it doesn't apply equally to all devices: https://css-tricks.com/how-to-disable-links/

I feel like it's probably more semantically correct and accessible for users to switch to `Text` if they don't want `Link` to function as a link. I'm not sure what's best for us though! @segunadebayo thoughts?

segunadebayo:
I agree @with-heart. @neighborhood999 I've been thinking about this as well. I feel there's no point rendering a link if it's going to be disabled. It feels more like a normal text in this scenario.

From an accessibility perspective as well, we might not want. This means we'll have to remove the `isDisabled` prop so we don't mislead other users as well.

I've just done this.

Thanks for looking into this @neighborhood999. It helped us revisit our decision with the link component

neighborhood999:
@with-heart @segunadebayo appreciate helping review PR. I agree with @with-heart above mentioned. I re-think this problem and agree with you, `isDisabled` maybe can be removed from `Link` component, it's not necessary, if want, I think this is user's responsibility.

PR #1343, opened 2020-07-22
docs(breadcrumb): removes obviously from documentation

antonjb:
Ran into the term obviously when viewing the Popover docs today. That was removed earlier but found another instance. Suggest removing the term to make the documentation a little friendlier.


antonjb:
Unsure what has happened with the build. That feels unrelated to this change 🤔

with-heart:
@antonjb You're right! It's not related. Thanks for removing this :)

with-heart:
@all-contributors please add @antonjb for doc

PR #1358, opened 2020-07-23
Fix `lineHeights.none` in the theme docs

trevorblades:
This branch updates a typo in the docs. The key `base` was accidentally used twice in place of the `none` key in the `lineHeights` part of the theme docs.


with-heart:
Good catch!

with-heart:
@all-contributors please add @trevorblades for doc

Issue #1315, opened 2020-07-19
[Select] White on white option text


hissalht:
When using dark mode, the options in a Select element appear white on white which is illegible.

![image](https://user-images.githubusercontent.com/7683740/87886916-fb2cc180-ca20-11ea-8ba0-7565aab109a3.png)

**Expected Behavior** 
The text is legible.

**Link to minimal reproduction** 
https://codesandbox.io/s/pensive-rhodes-pntec?file=/src/Component.js

**Steps to Reproduce**
Render a `Select` element when dark mode is on.

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: Firefow Developer Edition, New Microsoft Edge

with-heart:
Still occurring as of `next.6`. Reproduction: https://codesandbox.io/s/rough-meadow-3vyr2?file=/src/App.js

segunadebayo:
This has been fixed in `v1.0.0-rc.0`

PR #1388, opened 2020-07-24
fix(select): select option bg was white in dark mode

leoalves:
This PR will fix the issue with Select component in dark mode where the background  is white.

The select component is getting the styles from the input styles.  In dark mode the applied style for the background is `whiteAlpha.200`.  That works for the select when closed, but when opened the browser adds white to the background of the option.  To solved I just added `gray.700` to the option background style.

image

Tested on Firefox, Chrome and the new Edge.  I don't have a Safari handy to test it there.  But I believe it will work.  background property for the option is supported in all browsers as far as I know.  The issue might be with the selected value.  But at least in the 3 browsers I tested they all seemed to work.

Link to the opened issue:
https://github.com/chakra-ui/chakra-ui/issues/1315

### DIFF
--- a/examples/create-react-app/package.json
+++ b/examples/create-react-app/package.json
@@ -15,7 +15,7 @@
     "@types/react-dom": "16.9.8",
     "eslint-config-react-app": "5.2.1",
     "react-scripts": "3.4.1",
-    "typescript": "3.9.5"
+    "typescript": "3.9.6"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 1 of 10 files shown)


with-heart:
Hi @leoalves, we're currently working from `develop` instead of `master` until 1.0 is fully released. I switched your PR's base branch to `develop`, but it's including all of the commits from `master`. Could you cherry-pick this commit onto a new branch with the same name from `develop` and force-push it? Lemme know if I can help with that!

segunadebayo:
Thank you this PR @leoalves.

Due to conflicts, I've gone ahead to add your solution to the code.

Thanks again.

leoalves:
Thanks for both of you.
I will work on the develop branch from now on. 
Cheers

PR #1392, opened 2020-07-25
fix(icons): Fix viewBox on DragHandleIcon

santialbo:
`DragHandleIcon` uses `0 0 10 10` instead of the default `0 0 24 24`


with-heart:
How it looks with this fix: image

PR #1393, opened 2020-07-25
feat(menu): add trigger, openDelay and closeDelay props

### LINKED ISSUE #1380: Need Menu open on hover
I am using menu item and it is expandable using click but I want it to expand on hover .I am not able found any way to it . 
I have tried updating isOpen render prop to true onMouseEnter event but still nothing happens

dodas:
Seems to work nice, but I am quite sure there will be some nitpicks.
Feel free to edit.
Fixes #1380


with-heart:
Also I could have sworn I had added tests for `Menu` at some point. Like I would have bet money on it. 🤷

dodas:
Considering the discussion in the issue, should I close this PR?

with-heart:
@dodas I think so. We can always re-open later if it becomes a concern. 

Solid work though! Thanks for contributing.

Issue #1375, opened 2020-07-24
Question: Why don't use 'required' attribute for Input component?


scon-io:
`aria-required` is still useful for user agents that do not yet support HTML5, but `required` is handled natively on most of browsers.

with-heart:
Hi @scon-io! Which version of Chakra are you using?

scon-io:
@with-heart I'm using 1.0.0-next.4 :)

trevorblades:
I was also wondering this. In v0, when you pass `isRequired` to an `Input`, you get the native browser `required` attribute as well: https://github.com/chakra-ui/chakra-ui/blob/v0/packages/chakra-ui/src/Input/index.js#L37

I would expect the same to be true in v1. The [`isDisabled` prop directly translates to the `disabled` attribute](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/form-control/src/use-form-control.ts#L43), so I'm surprised the same convention isn't used for `required`. Right now, we have to write required inputs like this:

```jsx
<Input
  isRequired
  required
  {...otherProps}
/>
```

The solution for this seems pretty simple, and I've put together a PR that addresses this here: https://github.com/chakra-ui/chakra-ui/pull/1400

segunadebayo:
Thank you for catching this @scon-io.

I've pushed a fix for this.

PR #1400, opened 2020-07-26
feat: add support for native required attribute in FormControl

trevorblades:
This branch adds support for the native `required` attribute when `isRequired` is passed to an `Input` or `FormControl` component. If the omission of the native `required` attribute was intentional, please disregard this 🙏 

Addresses #1375

### DIFF
--- a/packages/form-control/src/use-form-control.ts
+++ b/packages/form-control/src/use-form-control.ts
@@ -41,6 +41,7 @@ export function useFormControl<T extends HTMLElement>(
     ...props,
     id: props.id ?? field?.id,
     disabled: props.disabled || props.isDisabled || field?.isDisabled,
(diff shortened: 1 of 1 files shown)


with-heart:
From what I'm reading, it seems like `required` is the default way to go, with `aria-required` existing to support pre-HTML5 devices.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute

This seems like a sensible addition.

segunadebayo:
My concern is how this affects form libraries, like Formik. I remember running into issues with this in a project. Here's some background: https://github.com/formium/formik/issues/544

segunadebayo:
I'm not sure what caused the build issue, but I've just pushed a fix for this.

Thank you @trevorblades

PR #1425, opened 2020-07-27
fix(styled-system): Fix error in SystemStyleObject type

santialbo:
I believe this type was meant to be an intersection type rather than a union. Close this if I'm wrong.

### DIFF
--- a/packages/styled-system/src/css.types.ts
+++ b/packages/styled-system/src/css.types.ts
@@ -121,12 +121,11 @@ type PseudoStyles = {
     : SystemCSSProperties
 }
 
-export type SystemStyleObject =
-  | SystemCSSProperties
-  | CSSPseudoStyles
-  | CSSSelectorStyles
-  | ApplyPropStyles
-  | PseudoStyles
+export type SystemStyleObject = SystemCSSProperties &
+  CSSPseudoStyles &
+  CSSSelectorStyles &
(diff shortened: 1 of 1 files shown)


with-heart:
That build failure might have our answer 🐙

santialbo:
I think it's because of the type intersection with `CSSSelectorStyles`. might need to express it in some other way but I think the intention is still for it to be an intersection type.

with-heart:
Gonna let @segunadebayo take a look. I'm not entirely sure what the intent is here.

santialbo:
Might need some TS guru here.
The root of the problem is that apparently you can't do this in TypeScript:
```
type A = {
  opacity: number
  [key: string]: string
}
const foo: A = { opacity: 1, display: "" }
```
No matter how you try
```
type A = {
  opacity: number
} & {
  [key: string]: string
}
const foo: A = { opacity: 1, display: "" }
```
```

type A = {
  opacity: number
} & Omit<
  {
    [key: string]: string
  },
  "opacity"
>

const foo: A = { opacity: 1, display: "" }
```
They all fail :/

segunadebayo:
@santialbo, yeah it won't work if you used an intersection type. I've experimented with this as well and it proved to be challenging to figure out.

Let's leave this as-is for now, and improve it later if we run into any weird TS issue.

PR #1430, opened 2020-07-28
fix/stricter types

ljosberinn:
WIP!

style:
- use `import React, {}` instead of `import * as React`
- group imports at the top
- always only export from /src/index.ts

types: 
- enforce `children` props where necessary (e.g. Button, Drawer) and omit optional children props where required
- use correct `Ref<Some Element>` types where posssible

test:
- wrap tests in `describe` blocks
- add `testA11Y` test-util function as wrapper around jest-axe
- add several low hanging fruit tests
- update snapshots accordingly
- separate tests per file to streamline tests folder


segunadebayo:
I love where this is going @ljosberinn. Thank you

ljosberinn:
Due to the size of this PR which imo got far out of scope (sorry for that) and because of #1445, I'll most likely scrap this PR and make individual PRs per package which will be a lot easier to review instead of this mess. 

This will be a chain of PRs because of that.

First series: temporarily introduce `React.FC` (through a custom version of it) for all components:
```ts
import { ReactElement, ReactNode } from 'react';

/**
 * Wrapper for ModernFunctionComponent<T>
 *
 * React.FC, without the issues.
 *
 * - no implicit, always present, optional children
 * - no propTypes: use TS interfaces/types
 * - no legacy contextTypes or childContextTypes
 * - no default props: use function default parameters
 *
 * @example
 * ```ts
 * // without children
 * const MyComponent: MFC<MyComponentProps> = (props) => {}
 *
 * // with children
 * const MyComponent: MFC<MyComponentProps & WithChildren> = (props) => {}
 *
 * // or
 *
 * interface MyComponentProps extends WithChildren {
 *  value: string;
 * }
 *
 * // or as type: type MyComponentProps = { value: string } & WithChildren
 *
 * const MyComponent: MFC<MyComponentProps> = (props) => {}
 * ```
 *
 * @see https://github.com/facebook/create-react-app/pull/8177
 */
export type MFC<T = {}> = ModernFunctionComponent<T>;

interface ModernFunctionComponent<T = {}> {
  (props: T, context?: any): ReactElement<any, any> | null;
  displayName?: string;
}

export interface WithChildren {
  children: ReactNode;
}
```
This way we can then dynamically use MFC & WithChildren where we need it. I'd add this to `utils/react-helper`.

2nd series of PRs: migrate `import * as React from 'react';` to `import React, { something } from 'react';`
3rd series: introduce a11y tests to all components through `jest-axe`

Please let me know if this is feels right & better. @segunadebayo @with-heart

PR #1431, opened 2020-07-28
chore(system): remove shouldForwardProp isValid check

### LINKED ISSUE #1347: [Layout] Properties are Not Passed when using "as" (1.0.0-next.6)
~When working with a component that used an object as a prop I noticed that it would not come through when using an "as" property.~

When using non HTML property names properties are not passed when using "as". For example `name` and `href` props will pass but anything that doesn't have an HTML prop on an element does not (that is my working theory anyway).

Quite simply 

         <SimpleComponent name="Susan" nested={{ lastName: "Smith" }}>Children Work</SimpleComponent>

Will pass nested to the component but

          <Box as={SimpleComponent} name="Susan" nested={{ lastName: "Smith" }}> Children Still Work</Box>

Will not pass the nested property to simple component (name property will pass). I tried this with a few other components as well (Button, Flex) with same behavior.

**Expected Behavior** I expect as to pass props to the base component of any type

**Link to minimal reproduction** https://codesandbox.io/s/relaxed-vaughan-xjglk?file=/src/App.tsx

**Steps to Reproduce**  See sandbox

**Suggested solution(s)**

I have not looked closely but I think this would be a candidate for a unit test. If you agree it is a bug I'm happy to help create that.

- OS: MacOS
- Browser: Chrome, Safari 
- Version 84, 13.1.1

**Additional context** I love Chakra-UI, nice work! :)

with-heart:
Resolves #1347

The `isValid` check was added recently, but I don't think that logic works quite the way we were expecting it to. The only way I could set the correct "should forward" status and get the tests to pass was by removing `isValid`.

### DIFF
--- a/packages/system/src/should-forward-prop.ts
+++ b/packages/system/src/should-forward-prop.ts
@@ -1,6 +1,5 @@
 import { propNames } from "@chakra-ui/styled-system"
 import { memoizeOne } from "@chakra-ui/utils"
-import isValid from "@emotion/is-prop-valid"
 
 const allPropNames = [
   ...propNames,
@@ -13,7 +12,7 @@ const allPropNames = [
 
 function createShouldForwardProp(props: any) {
   const regex = new RegExp(`^(${props.join("|")})$`)
-  return memoizeOne((prop: string) => isValid(prop) && !regex.test(prop))
+  return memoizeOne((prop: string) => !regex.test(prop))
 }
 
 export const shouldForwardProp = createShouldForwardProp(allPropNames)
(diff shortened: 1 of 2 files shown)


segunadebayo:
@with-heart. Thanks for this, I've also tried this and it seemed to work.

However, I noticed more ReactDOM console warnings and didn't get the time to check all components to fix all console warnings. Let's look into this together.

We'll add this change

with-heart:
@segunadebayo Sounds good to me!

PR #1447, opened 2020-07-29
Adding named exports to @chakra-ui/theme

cereallarceny:
We wanted to be able to easily override theme values without needing to have a full copy/paste of the existing theme file which quickly goes out of date as new versions of Chakra are released. Our solution to this was to separate all the theming objects and functions into their own exportable space without being mangled by `multiStyleConfig` or `styleConfig` (which makes it impossible to easily copy and paste styles after the fact). This way, you're given an easily `*Styles` named export, which contains all the theming objects/functions as keys on an object. Pull down whatever you want, modify it, and then combine it with the original object and apply to `multiStyleConfig` and `styleConfig`. This way, you'll always have "up-to-date" styles when you upgrade Chakra versions, and you'll only be overriding the styles you need.

**Reviewers note:** There are many `@ts-ignore` statements in this PR. Please don't be intimidated by this. In reality there's about 2 or 3 real Typescript issues we couldn't figure out in doing this refactor. As long as we figure out what they are, it should mostly be a copy/paste solution to fix all the duplicates. We'd love some help on typing some of these theme object/functions that we couldn't figure out ourselves.


segunadebayo:
This looks good to me @cereallarceny. Thank you for taking the time and effort to refactor the theme in a way that allows for better composition of component styles.

I will review the code in-depth later today.

cereallarceny:
Fantastic, thanks @segunadebayo. We're waiting on this PR to get merged for us to continue with our own development. Feel free to correct us on the right way to fix any `@ts-ignore` issues you find and we'll amend the PR appropriately. Alternatively, you can correct them yourself if you'd like. Let me know how we can be of assistance.

packages/theme/src/components/alert.ts
```diff
@@ -5,87 +5,117 @@ import {
...
 
-  defaultProps: {
-    variant: "subtle",
-  },
+const alert = multiStyleConfig({
+  parts,
+  baseStyle,
+  variants,
+  // @ts-ignore
+  defaultProps,
```
with-heart, line 107: I think this can be resolved by defining `defaultProps` as `const`:

```ts
const defaultProps = {
  variant: 'subtle',
} as const
```

with-heart: Any consumer defining `defaultProps` as a separate object rather than inline would need to do the same.

packages/theme/src/components/checkbox.ts
```diff
@@ -1,90 +1,113 @@
...
-        },
-      },
+const parts = {
+  control: "the main checkbox container",
+  label: "the checkbox label",
+  description: "the checkbox assistive text",
+  icon: "the checkmark icon",
+}
+
+// @ts-ignore
```
with-heart, line 10: I think if you type this as `Record<string, any>` like a lot of the other components, it should work fine.

with-heart:
@cereallarceny provided me with an example of a custom component theme using these changes:

```ts
import { accordionStyles } from '@chakra-ui/theme';

const { parts, baseStyleContainer, baseStyleButton, baseStylePanel } = accordionStyles;

// Here's the modification
baseStyleButton._focus = { outline: none; }

const baseStyle = {
  container: baseStyleContainer,
  button: baseStyleButton,
  panel: baseStylePanel
};

const accordion = multiStyleConfig({
  parts,
  baseStyle
});

export default accordion;
```

An interesting side effect here is that since `baseStyle` already contains references to each of its parts, you could just do:

```ts
import { accordionStyles } from '@chakra-ui/theme';

const { parts, baseStyle, baseStyleContainer, baseStyleButton, baseStylePanel } = accordionStyles;

// Here's the modification
baseStyleButton._focus = { outline: none; }

const accordion = multiStyleConfig({
  parts,
  baseStyle
});

export default accordion;
```

Or even:

```ts
import { accordionStyles } from '@chakra-ui/theme';

const { parts, baseStyle } = accordionStyles;

// Here's the modification
baseStyle.button._focus = { outline: none; }

const accordion = multiStyleConfig({
  parts,
  baseStyle
});

export default accordion;
```

packages/theme/src/components/checkbox.ts
with-heart: This should apply to all the functions that have `@ts-ignore` right now.

with-heart:
I think these changes demonstrate that we're missing something in how the component theming system works. I can't 	quite put my finger on what's wrong or what would make it better, but I wanted to add a few thoughts/notes:

- `styleConfig` and `multiStyleConfig` are awkward and I'm not sure they provide us with much value 
  - when objects are defined on their own rather than inline, they become awkward to type
    - need to use `as const` for `defaultProps` objects
    - `baseStyle` value functions are generically typed using `Record<string, any>`
  - we could automatically call `styleConfig`/`multiStyleConfig` when processing the theme rather than calling it in each individual component theme file
- `baseStyle` would be more flexible if it was always an object whose values could be either `SystemStyleObject` or functions that take `props` and return `SystemStyleObject`

cereallarceny:
@with-heart I believe we have resolved the issues you pointed out... sans refactoring `styleConfig/mutliStyleConfig`.

segunadebayo:
@with-heart, I agree that we're not quite there yet with the theming. This is a good welcome improvement but we'll make it better over time.

I have some ideas I want to play with.

Thank you for your work on this @cereallarceny.

Issue #1377, opened 2020-07-24
[Stack] Input state is lost when new items are added to the Stack


trevorblades:
Since the v1 `Stack` component [uses indexes for keys](https://github.com/chakra-ui/chakra-ui/blob/master/packages/layout/src/stack.tsx#L162) on its cloned children, it means that components get remounted if a new element gets added to the start of the Stack.

Here's a codesandbox demonstrating this effect: https://codesandbox.io/s/reverent-sound-ytcrq?file=/src/App.js

Add values to the text fields, and then set or clear the `error` state using the buttons to see the text field values getting reset when the error message is shown or hidden.

The pre-v1 `Stack` didn't have this effect, and its children weren't re-rendered when new children were added. I was expecting the same behaviour to take place here.

This seems to be due to the addition of the `shouldWrapChildren` and `divider` props, which require us to iterate over children and change them in some way. One potential fix would be to avoid mapping over children if `!shouldWrapChildren && !divider` [at this part of the code](https://github.com/chakra-ui/chakra-ui/blob/master/packages/layout/src/stack.tsx#L157):

```tsx
const clones = !shouldWrapChildren && !divider ? validChildren : validChildren.map(...);
```

This would fix this issue for some, but not _all_ use cases of this component.

Update: I made a PR for the above proposed change [here](https://github.com/chakra-ui/chakra-ui/pull/1455) 🎉

segunadebayo:
I've pushed a fix for this issue.

Thank you @trevorblades

PR #1455, opened 2020-07-29
fix: only map over Stack children when necessary

trevorblades:
This branch makes an update to the `Stack` component to avoid cloning and assigning new `key`s to its children when it isn't necessary. This addresses an issue brought up in #1377.


segunadebayo:
Oops! @trevorblades I already pushed a fix to the develop branch based on your recommendation.

Thank you for catching the issue and taking the time to send in a PR.

Going to close this for now

PR #1458, opened 2020-07-30
Support Storybook Composition

jesstelford:
Storybook's new [Composition feature](https://github.com/storybookjs/storybook/blob/next/docs/src/pages/configurations/composition/index.md) supports auto-detecting direct dependencies which have a published storybook and embedding it in the main application's storybook instance.

From [the docs](https://github.com/storybookjs/storybook/blob/next/docs/src/pages/configurations/composition/index.md):

> **Automatic loading**
> The process above is actually often not even required! Storybook will auto-compose direct dependencies' storybooks if it's able to find them.
> 
> The ref will be auto-loaded if the direct dependency has a field called storybook in it's package.json which looks like this:
> 
> ```json
> {
>   "storybook": {
>     "title": "My Package",
>     "url": "https://my.storybook-url.com"
>   }
> }
> ```
> This is really useful if you depend on a component library that is also using storybook.
> 
> Not only do you get to view all of their stories, but their documentation too, if they use addon-docs!

NOTE: May also require setting up the netlify deployment to accept CORS.

Interestingly when running Storybook locally, I can only see a single story (for `Drawer`) - how do I run the full storybook correctly?


segunadebayo:
Thank you for this @jesstelford. I've updated the stories set in `.storybook/main.js` to load all stories.

jesstelford:
Fantastic!

@segunadebayo I've fixed the merge conflicts too 👍

with-heart:
Thanks so much @jesstelford!

with-heart:
@allcontributors please add @jesstelford for code

bladey:
@with-heart is there a version of the Storybook that is running rc code, e.g `^1.0.0-rc.7`?

https://chakra-ui.netlify.app is currently using the public release as far as I can tell.

When we link into your Storybook from ours, it attempts to load https://chakra-ui.netlify.app/stories.json which doesn't exist, my best guess is that this is because your Storybook is running version 5.2.4 when 6.0.27 is available.

Any chance we could get a version running rc and the latest Storybook?

ljosberinn:
This will eventually resolved with the release of v1

frysztak:
@ljosberinn seems like v1 is out now, but https://chakra-ui.netlify.app is still using Storybook 5. is this something you guys are working on (or, perhaps, need help with)?

with-heart:
@frysztak If you're able to assist, that would be great. I'd love to get us upgraded and even move our storybook to storybook.chakra-ui.com, but it's not work we currently have prioritized.

PR #1464, opened 2020-07-30
build(pacakge.json): adding exports field for all packages

JayaKrishnaNamburu:
## Issue
Currently when we try to load `@chakra-ui/button` using a cdn like jspm. Will throw a error, since the `main` file points to the `cjs`.

Hi, I am adding `exports` field to all the packages, so `esm` based cdn's can resolve them and make them available on request. The packages do already have `module` attribute specified in them. But my understanding says, the field is mostyl used by the bundlers and not the cdn services.

## Current Behaviour

Currently if we try to resolve any package using using a cdn service like `jspm` it starts to throw error. Since the `main` which is used by them actually refers to a cjs build. Here is the issue and the jsfiddle to reproduce the same.

A more detailed explanation of the behaviour is explained here --> https://github.com/jspm/project/issues/72#issuecomment-665837262

Here are some of the details on how these export maps help --> https://docs.skypack.dev/lookup-urls/lookup-urls#lookup-a-package-export

JSFiddle --> https://jsfiddle.net/ng3ufhe9/


guybedford:
Looks good to me, the only questions might be:

1. Should `"exports"` branch between Node.js usage and the browser? That is, is it expected for the CommonJS to load in Node.js?
2. Are there other files in the package that might need to be exported? Exports by default is encapsulating so will cause other modules not to be exposed.

Happy to answer any questions further here as well.

segunadebayo:
@with-heart can you help investigate the status of this? It's been open for too long

segunadebayo:
Hi @JayaKrishnaNamburu,

The benefits of this aren't clear to me just yet. Can you explain a bit more?

PS: Feel free to help resolve conflicts as well.

Thanks

JayaKrishnaNamburu:
## Package exporting cjs under main !

The most common approach that package authors take is, exporting `cjs` using `main` field in package.json. And `esm` using `module` field.
This is a pattern that is encouraged and followed mainly on the bundlers userland.

But in node environment you need to export using only `main` field. Since, `module` attribute is not supported or a standard in node.
Instead we have a another field called, `type : module` which is a node [standard](https://nodejs.org/api/esm.html#esm_enabling) to define if the code we are exporting is a es mdoule or cjs.

And for `esm` modules we the file extension is `.mjs` and not `.js`. But the bundlers userland will accept even if we use the `.js` extension for modules.

To overcome this issue, `cdn`'s like `jspm` does a ahead of time static analysis. This makes then understand the package structure and the files and exports.
But during this step, they mainly rely on `exports` field which is a `node` standard. And which is again not used by most of the pacakge authors, since bundlers are not using them at the moment.

In such instances, the static analysis fails and `cdn's` faile to serve the packages when request a specific module.

Here are little more details which i wrote on this topic https://jkrishna.dev/post/dependencies

Here are more details about the issue that users face if they load from cdn right now --> https://github.com/jspm/project/issues/72

Thanks, please let me know if any further information is needed. Happy to help 😄

ljosberinn:
@JayaKrishnaNamburu I'd love to see this happen but I'm afraid you rebased at a time where we accidentally had a broken develop branch.. could you kindly rebase again? 👀 if build then passes I'll release it right afterwards as rc6

JayaKrishnaNamburu:
Hi @ljosberinn i rebased with the latest develop branch. Here is some more information about export-map --> https://docs.skypack.dev/package-authors/package-checks#export-map

JayaKrishnaNamburu:
You can cross-check with the concerns that were posted above here. If you explicitly want to expose any files that are out the current export map scope.

Should "exports" branch between Node.js usage and the browser? That is, is it expected for the CommonJS to load in Node.js?
Are there other files in the package that might need to be exported? Exports by default is encapsulating so will cause other modules not to be exposed.


https://github.com/chakra-ui/chakra-ui/pull/1464#issuecomment-666448710

ljosberinn:
@allcontributors please add @JayaKrishnaNamburu for code

ljosberinn:
a small follow up:

this works in prod as evident here in this pr - however in dev, nextjs requires cjs exports, so I've adjusted the exports of all packages accordingly

PR #1467, opened 2020-07-30
fix(media-query): create media queries using breakpoint name keys only

### LINKED ISSUE #1444: [useBreakpoint/useBreakpointValue] API partially broken
As reported on [Discord](https://canary.discordapp.com/channels/660863154703695893/660865137137156146/737974879168757830), the return value of `useBreakpoint` and `useBreakpointValue` is `undefined` if no matching breakpoint is currently active; it should fallback to `base` if given.

Also, the passed argument must currently match array indices:
`{base: 'column', 0: 'column', 1: 'row', 2: 'row', 3: 'row'}`
 instead of, following the docs, 
`{base: 'column', sm: 'column', md: 'row', lg: 'row', lg: 'row'}`.

It's also broken in the [docs](https://next.chakra-ui.com/docs/hooks/use-breakpoint-value); `variant` is always `undefined` according to the devtools:

image

https://codesandbox.io/s/distracted-bird-3ifkq?file=/src/App.tsx

**Expected Behavior** API works as expected.

**Suggested solution(s)**

Investigate what goes wrong [here](https://github.com/chakra-ui/chakra-ui/blob/c38892760257b9bbf1b63c05f7f9ccf1684a90b0/packages/media-query/src/use-breakpoint-value.ts#L15).

with-heart:
Fixes #1444 

This PR resolves an issue where `createMediaQueries` was returning a broken breakpoints array, causing `useBreakpointValue` objects to only match when using array numbers as keys rather than the breakpoint names.

Also added tests for both `createMediaQueries` and `useBreakpointValue` to help us avoid this situation in the future.

Thanks to @nassimbenkirane for investigating and sharing what he found!

### DIFF
--- a/packages/media-query/src/create-media-query.ts
+++ b/packages/media-query/src/create-media-query.ts
@@ -1,4 +1,4 @@
-import { objectKeys, Dict, isNumber } from "@chakra-ui/utils"
+import { Dict, isNumber, breakpoints as bps } from "@chakra-ui/utils"
 import calculateMeasurement from "calculate-measurement"
 
 /**
@@ -6,7 +6,12 @@ import calculateMeasurement from "calculate-measurement"
  * using a combination of `min-width` and `max-width`.
  */
 function createMediaQueries(breakpoints: Dict) {
-  const sorted = getSortedKeys(breakpoints)
+  // get the non-number breakpoint keys from the provided breakpoints
+  const keys = Object.keys(breakpoints)
+
+  // use only the keys matching the official breakpoints names, and sort them in
+  // largest to smallest order
+  const sorted = bps.filter((bp) => keys.includes(bp)).reverse()
 
   // create a min-max media query string
   return sorted.map((breakpoint, index) => {
@@ -54,13 +59,6 @@ function subtract(value: any) {
  */
 function toMediaString(value: any) {
   return isNumber(value) ? `${value}px` : value
-}
-
-/**
- * Sort the breakpoints in descending order
- */
-function getSortedKeys(bps: Dict) {
-  return objectKeys(bps).sort((a, b) => parseInt(bps[b]) - parseInt(bps[a]))
 }
 
 export default createMediaQueries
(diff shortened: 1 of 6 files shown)


packages/media-query/src/create-media-query.ts
```diff
@@ -1,12 +1,17 @@
...
  * using a combination of `min-width` and `max-width`.
  */
 function createMediaQueries(breakpoints: Dict) {
-  const sorted = getSortedKeys(breakpoints)
+  // get the non-number breakpoint keys from the provided breakpoints
+  const keys = Object.keys(breakpoints)
+
+  // use only the keys matching the official breakpoints names, and sort them in
+  // largest to smallest order
+  const sorted = bps.filter((bp) => keys.includes(bp)).reverse()
```
with-heart, line 14: Instead of using all of the keys that are on `breakpoints` args (which includes stringified array indexes since it'll be an array with keys attached to its prototype), we instead use the `breakpoints` array from `@chakra-ui/utils` and only include keys that are present in the object passed to the function. 

This guarantees that we're only using valid breakpoints, and that we're using the actual breakpoint names.

We also reverse them to make sure we're returning an array of largest to smallest breakpoints.

packages/media-query/src/create-media-query.ts
```diff
@@ -1,12 +1,17 @@
...
 import calculateMeasurement from "calculate-measurement"
 
 /**
  * Create a media query string from the breakpoints,
  * using a combination of `min-width` and `max-width`.
  */
 function createMediaQueries(breakpoints: Dict) {
-  const sorted = getSortedKeys(breakpoints)
+  // get the non-number breakpoint keys from the provided breakpoints
+  const keys = Object.keys(breakpoints)
```
nassimbenkirane, line 10: I see that there was an import of objectKeys function, wonder what is different from Object.keys (maybe some polyfills ?) But it might be worth using it here ?

with-heart: It was solely for wrapping `Object.keys` with more accurate types which we're not using here, so I think we're good!

PR #1474, opened 2020-07-31
fix/transition: Change children type

### LINKED ISSUE #1473: [SlideIn] Type definition for children is not correct
**Describe the bug**

This is a 0.8 bug.

The type definition for the `SlideIn` component's `children` prop is not correct. It's being overridden by `TransitionProps`. `SlideInProps` need to come last.

Here's a reproduction: https://codesandbox.io/s/relaxed-architecture-slmh0?file=/src/App.tsx

selbekk:
This commit makes sure the `children` prop is set to whatever we specify in `IScale` and `ISlideIn`, instead of React Spring's TransitionProps's implementation.

Fixes #1473

### DIFF
--- a/packages/chakra-ui/src/Transition/index.d.ts
+++ b/packages/chakra-ui/src/Transition/index.d.ts
@@ -9,7 +9,7 @@ interface ISlideIn {
   children: (styles: Object) => React.ReactNode;
 }
 
-type SlideInProps = ISlideIn & TransitionProps<boolean>;
+type SlideInProps = TransitionProps<boolean> & ISlideIn;
 export const SlideIn: React.FC<SlideInProps>;
 
 interface IScale {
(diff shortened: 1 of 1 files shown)


selbekk:
Not sure if I've created the PR to merge towards the correct branch 🤔  Let me know if I need to do something differently.

with-heart:
I think we're gonna pass on this. We're not planning any further releases for v0 at this time.

selbekk:
Oh, not even bugfixes? That's a bit scary to me, given the large amount of breaking changes and somewhat complex migration path to v1. A lot of not-so-maintained projects might not have the resources to migrate, but will be stuck with an unmaintained version.

It's your right, of course, it's a larger commitment to support legacy versions.

If you're dropping support for v0, perhaps a note on the website should be considered at least.

PR #1527, opened 2020-08-05
Test/add a11y tests

ljosberinn:
- adds a11y tests to all components, disabling jest-axe rules only where necessary (focus-lock, looking at you)
- aims to streamline how tests are written across all components
- use `screen` instead of `utils.` where possible
- avoid using conjunctive in test names
- ~~use `describe(<Component />` wrapper~~
- snapshot updates
- adds tests for `Menu` and `AlertDialog`

edit: got up to `Menu` so far and would like to merge it earlier as its already hard to keep it up to date and I lack the time right now to do the rest :(


packages/alert-dialog/tests/alert-dialog.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+              <button type="button">Yes, delete</button>
+            </AlertDialogFooter>
+          </AlertDialogContent>
+        </AlertDialogOverlay>
+      </AlertDialog>
+    </>
+  )
+}
+
+describe("<AlertDialog />", () => {
```
with-heart, line 63: imo we should [avoid nesting when we're testing](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing). It doesn't really add anything to wrap every test in a file in the same block

with-heart: Or at the very least, add it in a separate PR. Since it indents the entire file, it's impossible to tell what's actually changing in some of these files 🙃

ljosberinn: Generally I agree with everything in Kents article there, obviously, his example is far over the top though and there to make a point.

> The describe function is intended to group related tests together and can provide for a nice way to visually separate different tests, especially when the test file gets big. But I don't like it when the test file gets big. So instead of grouping tests by describe blocks, I group them by file. So if there's a logical grouping of different tests for the same "unit" of code, I'll separate them by putting them in completely different files.

So his point is to avoid `describe` blocks if your test is so large that it should be a separate test file on itself. I don't think that applies here _yet_.

Also I'm somewhat sure if a test fails, we don't have to search for the test name but can use the describe tag right above because it adds indentation on CI too.

Since we can reintroduce it at any point anyways, I'll remove it again later today.

ljosberinn: on a sidenote, having a `top level describe` [is considered `best practice` by jest itself](https://github.com/jest-community/eslint-plugin-jest/blob/master/src/rules/require-top-level-describe.ts#L8)

packages/focus-lock/tests/focus-lock.test.tsx
```diff
@@ -12,11 +12,32 @@ test("focuses an element on render", () => {
...
   }
-  const utils = render(<Component />)
-  const input = utils.getByTestId("input")
+  render(<Component />)
+  const input = screen.getByTestId("input")
   expect(input).toHaveFocus()
 })
 
+// placeholder test until we figure out how to deal with react-focus-lock
+// being
```
with-heart, line 21: Being what?! The suspense is killing me! 😁

packages/control-box/src/index.tsx
```diff
@@ -1,86 +1 @@
...
-    </chakra.div>
-  )
-}
-
-if (__DEV__) {
-  ControlBox.displayName = "ControlBox"
-}
-
-export default ControlBox
+export * from "./control-box"
```
with-heart, line 1: Nice. Solid attention to detail!

packages/image/tests/image.test.tsx
```diff
@@ -23,40 +23,37 @@ function trackImageOnload() {
...
-  )
-}
-
 test("creates an instance of Image when mounted", () => {
-  const tools = renderComponent()
-  const image = tools.getByTestId("img")
-  expect(image).toBeInstanceOf(HTMLImageElement)
+  render(<Image src={src} fallbackSrc={fallbackSrc} />)
+
+  expect(screen.getByRole("img")).toBeInstanceOf(HTMLImageElement)
```
with-heart, line 29: 👍

packages/focus-lock/tests/focus-lock.test.tsx
ljosberinn: uh, ehrm, good question, next question 👀 I'll remove it

PR #1533, opened 2020-08-06
Combobox Component

leoalves:
Hi,

I am building a Combobox component, but I have a few questions.  
I guess the easiest would be to create a PR with what I have so far and ask the questions commenting the PR code.


packages/Combobox/src/combobox.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+if (__DEV__) {
+  Combobox.displayName = "Combobox"
+}
+
+export type ComboboxInputProps = PropsOf<typeof chakra.input> &
+  UseComboboxInputProps &
+  InputProps
+
+export const ComboboxInput = React.forwardRef(function StyledComboboxInput(
```
leoalves, line 63: The `Combobox` component has a multi part style object. I couldn't find a way to create a nested multi part style object.

For example the `Combobox` will have an `combobox-input`, `combobox-list` and `combobox-list-item` style parts.  
But the `combobox-input` will also have a `field` and an `addon` style part.  

Is there a way to create a nested multi part theme style?  
Should I create an additional multi part style object called `ComboboxInput` that inherit the styles from the input and additional combobox input specific styles?  
Or is there any other way to do this that I am not seeing?

packages/Combobox/src/combobox.hook.tsx
```diff
@@ -0,0 +1,455 @@
...
+  UseComboboxReturn
+>({
+  strict: false,
+  name: "ComboboxContext",
+})
+
+export { ComboboxContextProvider, useComboboxContext }
+
+export interface UseComboboxProps extends UsePopperProps {
+  /**
```
leoalves, line 27: What should be the API for a Combobox component.  The reason why I started building one is because I need a Google Places Autocomplete Component.  In my case the list of items will be loaded on demand.  I will need an `onChange` event and  a `isLoading` prop to render the loading icon while it waits for the server response.  I am also thinking about a `multi` prop, where multiple items can be selected.  
But what else?

leoalves:
I added 2 comments where I have some doubts on how to proceed.  

I didn't want to spend too much time before getting some feedback from the core team. I am not sure if I am going in the right direction.  Or even if the project is open for this kind of PR.

I tried to follow, as much as possible, the way the other components were build.  But there could be a few places where I misunderstood how it actually works. I am new to React Hooks and to the code base.

I am really enjoying working with Chakra-UI  I started working with it a few weeks ago and I can already see a boost in productivity.

Congrats to everyone involved on it.

package.json
```diff
@@ -70,6 +70,7 @@
     "tooltip": "yarn workspace @chakra-ui/tooltip",
     "descendant": "yarn workspace @chakra-ui/descendant",
     "clickable": "yarn workspace @chakra-ui/clickable",
+    "comboboz": "yarn workspace @chakra-ui/combobox",
```
segunadebayo, line 73:
```suggestion
    "combobox": "yarn workspace @chakra-ui/combobox",
```

packages/Combobox/src/combobox.tsx
segunadebayo: Yes, there's a way to do this @leoalves.

I'd recommend you check the theme package to see examples of how to define multipart theme styles.

Thank you for working on this 💖

segunadebayo: Based on how to set up the theming, here are some steps:

- Add `combobox.ts` to `packages/theme/src/components`
- Create a theme structure that looks like this:

```ts
const parts = {
  list: "the group of combobox items",
  input: "the combobox input",
  item: "the combobox item or option"
}

const baseStyle = {
  input: {},
  list: {},
  item: {}
}

const combobox = multiStyleConfig({
 parts,
  baseStyle,
 // you can add sizes, and variants as well
})

export default combobox
```

- Add combobox to the theme's index file
- Build the theme package `yarn theme build`

That's it! Let me know if you have any questions

timkindberg:
Have you guys considered using reakit for its amazing low-level accessible components? They have a new combo box component in beta right now. Some amazing work happening over there that honestly is in the same spiritual vein as chakra and the chakra would be an amazing layer on top.

segunadebayo:
Thanks for the initial efforts to add the Combobox component @leoalves. I'd recommend you take your time on this and open a PR once you have it ready for review.

Good point @timkindberg, I thought about this as well and I think it might be worth a try. I only wish Reakit had a way to install specific packages so we don't have to take on the entire bundle size.

Cheers.

PR #1576, opened 2020-08-09
docs(installation): add CRA templates guide note

with-heart:
This PR adds a note to the Installation page pointing CRA users to the official CRA templates guide.

image

### DIFF
--- a/website/pages/docs/getting-started.mdx
+++ b/website/pages/docs/getting-started.mdx
@@ -5,6 +5,9 @@ description: How to install and set up Chakra UI in your local project
 
 **First, you need to install Chakra UI**
 
+> For `create-react-app` installation instructions, see the
+> [official CRA templates guide](/guides/integrations/with-cra) instead.
+
 ```bash
 npm i @chakra-ui/core@next
 


website/pages/docs/getting-started.mdx
```diff
@@ -5,6 +5,9 @@ description: How to install and set up Chakra UI in your local project
 
 **First, you need to install Chakra UI**
 
+> If you're a `create-react-app` user, see the
```
ljosberinn, line 8: > If you're using `create-react-app`, see the

Maybe just me, but `a ... user` feels odd

with-heart: @ljosberinn How about "For `create-react-app` installation instructions," instead?

ljosberinn: Fine by me, feels a bit formal, but I don't mind.

with-heart: I agree but not sure how to state it better while keeping it short. I'll change it to that and if we come up with something better later, we can change it

PR #1601, opened 2020-08-10
fix(color-mode): catch when localStorage is disabled

hy2k:
This PR catches error when localStorage is disabled by browser setting. Currently, when localStorageManager tries to access disabled localStorage, it throws an error.
 
On Google Chrome,
```
Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
```

On FireFox,
```
window.localStorage is null
```

Instead, it will output warning to console.

Reference: This is how [theme-ui](https://github.com/system-ui/theme-ui) handle this type of error. https://github.com/system-ui/theme-ui/blob/a96e723bad9c0d8b03c858698045acf01587cf9d/packages/color-modes/src/index.js#L9-L32

### DIFF
--- a/packages/color-mode/src/storage-manager.ts
+++ b/packages/color-mode/src/storage-manager.ts
@@ -13,16 +13,32 @@ export interface StorageManager {
  */
 export const localStorageManager: StorageManager = {
   get(init?) {
-    const exist =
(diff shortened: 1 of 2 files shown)


jmiazga:
Thanks @hy2k for working on this!

with-heart:
@allcontributors please add @hy2k for code

PR #1607, opened 2020-08-10
Refactor: Remove as infer to speed up intellisense

### LINKED ISSUE #1582: Remove the As type infer to make intellisense faster
With latest release typescript intellisense is still very slow

I did some investigation and the problem is the `ComponentWithAs` type, i tried tweaking that types removing the `As` generic argument and everything becomes super smooth, making intellisense instantaneous

The `as` generic argument delays intellisense even if you son't use it, the delay is about `1000ms` average for me

I think we should remove the as generic and instead encourage people to use the new `chakra.tag` approach

The `as` type infer approach could be added maybe in the future, we shouldn't add it in the `1.0` release or it will be impossible to remove (because of breaking changes) and it will make everyone experience worse

remorses:
Now it takes much less time to populate auto completion suggestions, 

We can still do much better typing the component directly instead of relying on `forwardRef` and `ChakraComponent`, after that intellisense is instantaneous

Closes #1582

### DIFF
--- a/packages/pin-input/src/pin-input.tsx
+++ b/packages/pin-input/src/pin-input.tsx
@@ -60,7 +60,7 @@ if (__DEV__) {
 export interface PinInputFieldProps extends PropsOf<typeof chakra.input> {}
 
 export const PinInputField = forwardRef<PinInputFieldProps, "input">(
-  function PinInputField(props, ref) {
+  function PinInputField(props, ref: any) {
     const inputProps = usePinInputField({ ref, ...props })
     return (
       <chakra.input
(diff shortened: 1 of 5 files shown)


packages/pin-input/src/pin-input.tsx
```diff
@@ -60,7 +60,7 @@ if (__DEV__) {
 export interface PinInputFieldProps extends PropsOf<typeof chakra.input> {}
 
 export const PinInputField = forwardRef<PinInputFieldProps, "input">(
-  function PinInputField(props, ref) {
+  function PinInputField(props, ref: any) {
```
ljosberinn, line 63: that coincidentally crept in I assume? :)

remorses: I have no idea, the ref type is the same as before but now it gives a weird error, I tried using RefObject, Reg, MutableRef, AssignableRef but nothing works

Issue #1608, opened 2020-08-10
[Core] Global styles overridden by CSSReset


dustinlarimer:
In `ChakraProvider`, the `CSSReset` component is currently loaded after `GlobalStyle` ([code](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/core/src/chakra-provider.tsx#L55-L56)). Until recently, global styles were scoped under the `body` tag, which gave them priority over the reset definitions.

Since https://github.com/chakra-ui/chakra-ui/commit/e0732decc24c849e723988299a6fb2cbf87b90be, this is no longer the case – the reset definitions now have priority over global styles.

**Example:**

<img width="234" alt="Screen Shot 2020-08-10 at 12 30 14 PM" src="https://user-images.githubusercontent.com/180438/89822863-52e0b900-db05-11ea-834b-088bd82fac4a.png">

**Output with v1.0.0-next.7:**

<img width="256" alt="Screen Shot 2020-08-10 at 12 14 46 PM" src="https://user-images.githubusercontent.com/180438/89822638-22991a80-db05-11ea-9093-d5b0e5450aff.png">

**Output with v1.0.0-rc.2:**

<img width="255" alt="Screen Shot 2020-08-10 at 12 28 19 PM" src="https://user-images.githubusercontent.com/180438/89822686-25940b00-db05-11ea-9245-78e41137019f.png">

**Suggested solution(s)**

Flip load order of `CSSReset` and `GlobalStyles` in `ChakraProvider` (#1609)

**Workaround:**

Manually load `<CSSReset />` component outside of `<ChakraProvider>`:

```jsx
<CSSReset />
<ChakraProvider theme={theme}>
  <App />
</ChakraProvider>
```

<img width="253" alt="Screen Shot 2020-08-10 at 12 37 49 PM" src="https://user-images.githubusercontent.com/180438/89823431-53c61a80-db06-11ea-8247-8602b57e7b5c.png">

**System:**

- OS: macOS 10.15.5 (Catalina)
- Chrome
- 84.0.4147.105

PR #1609, opened 2020-08-10
fix(core): flip load order of CSSReset and GlobalStyles

dustinlarimer:
This PR offers a possible fix for #1608, flipping the load order of `CSSReset` and `GlobalStyles` in `ChakraProvider`. This issue surfaced as a side-effect of a recent improvement. More context is available on the [issue](https://github.com/chakra-ui/chakra-ui/issues/1608).

One observation from debugging this, is that it seems like the correct way to apply the CSS reset going forward is to use the `resetCSS` prop on `ChakraProvider`, rather than including the `<CSSReset />` component directly – is that correct?

### DIFF
--- a/packages/core/src/chakra-provider.tsx
+++ b/packages/core/src/chakra-provider.tsx
@@ -52,8 +52,8 @@ export const ChakraProvider: React.FC<ChakraProviderProps> = (props) => {
         useSystemColorMode={theme?.config?.useSystemColorMode}
         storageManager={storageManager}
       >
-        <GlobalStyle />
         {resetCSS && <CSSReset />}
+        <GlobalStyle />
         {portalConfig ? (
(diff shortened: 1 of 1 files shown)


ljosberinn:
Thank you for spotting this! We'll have to document this; although its perfectly logical and default behaviour of CSS, I feel like it might hurt DX if someone uses `GlobalStyles` and `CSSReset` and runs into this issue.

Concerning the prop - yes, `ChakraProvider` itself was added for convenience, so you should be able to use the entirety of Chakras providers through its API surface.

dustinlarimer:
@ljosberinn no problem, happy to help! Need anything else from me on this one?

dustinlarimer:
Thanks @segunadebayo!

PR #1627, opened 2020-08-12
perf: widen As prop typing

### LINKED ISSUE #1582: Remove the As type infer to make intellisense faster

with-heart:
Resolves #1582

@segunadebayo and I worked on this over the weekend. I need to run a few perf tests on it but it seemed to give a solid speed boost because we're no longer relying on `JSX.IntrinsicElement`.

### DIFF
--- a/packages/system/src/forward-ref.tsx
+++ b/packages/system/src/forward-ref.tsx
@@ -3,32 +3,21 @@
  */
 import * as React from "react"
 
-export type AssignableRef<T> =
-  | {
-      bivarianceHack(instance: T | null): void
-    }["bivarianceHack"]
-  | React.MutableRefObject<T | null>
-  | null
-
-type As<P = any> = React.ElementType<P>
+type As = string | React.ComponentType<any>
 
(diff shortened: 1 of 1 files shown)


remorses:
Just tried the new changes and it is much faster! Thank you for your great work!

with-heart:
@remorses So glad to hear it. We put a bunch of effort into it and I wasn't entirely sure it would have the result we expected 😅

I wish it was easier to test perf on tsserver other than manually playing with it

Issue #1523, opened 2020-08-05
Is there an easy way to customize component?


leadream:
**What problem does this feature solve?**

In our project we want to customize button component, such as changing its height. I just saw the [V1 theming API](https://next.chakra-ui.com/docs/theming/overview) but it needs much code. 

**What does the proposed API look like?** 

Can we have the way to customize a component like below so we can customize a component with less code:

```js
import theme from '@chakra-ui/theme';

const customTheme = theme.customizeComponent('Button', {
  sizes: {
    md: {
      height: '38px',
    }
  }
})

export default customTheme;
```

Thanks in advance.

ljosberinn:
Wouldn't be opposed to that idea. Chakra should provide a way to avoid having to spread objects yourself.

This may be related to https://github.com/chakra-ui/chakra-ui/pull/1447.

my take (which is limited to 
```ts
import defaultTheme, { customize, ComponentModification, Breakpoint } from '@chakra-ui/theme';

const breakpoints: Breakpoint[] = ['30rem', '48rem', '62rem', '80rem'];

const components: ComponentModification[] = [
  {
    component: 'Button',
    modification: {
      sizes: {
        md: {
          height: '38px'
        }
      }
    }
  },
];

const theme = customize({ breakpoints, components });

```

brennj:
Hey, this could be naive and apologies if so, what is the benefit of doing it in the proposed ways above vs something (i would consider simpler but probably debatable) like:

```
import { Button as B } from '@chakra-ui/core';

export function Button(props) {
  return <B h="38px" {...props} />;
}
```

ljosberinn:
Your solution would work, but you'd pollute your autocomplete with a second `Button` and couldn't just import from chakra directly. And if you rename it, you'd need everyone working on the project to know what your `CustomButton` or `ThemedButton` is called and I bet someone still imports from Chakra directly.

brennj:
In fairness, I've ran into that problem with autocomplete so point taken there as this is what I've done with my Chakra wrapper in our work project which uses v0. :) Interesting proposition then!

leadream:
> Your solution would work, but you'd pollute your autocomplete with a second `Button` and couldn't just import from chakra directly. And if you rename it, you'd need everyone working on the project to know what your `CustomButton` or `ThemedButton` is called and I bet someone still imports from Chakra directly.
> 
> Besides, this would only work for components itself I'm afraid. You couldn't customize e.g. breakpoints there.

[...]

PR #1631, opened 2020-08-12
refactor: styleConfigs' internalization

hericlesme:
Based on the discussion in #1447:

- `styleConfig` and `multiStyleConfig` are awkward and I'm not sure they provide us with much value 
  - when objects are defined on their own rather than inline, they become awkward to type
    - need to use `as const` for `defaultProps` objects
    - `baseStyle` value functions are generically typed using `Record<string, any>`
  - we could automatically call `styleConfig`/`multiStyleConfig` when processing the theme rather than calling it in each individual component theme file
- `baseStyle` would be more flexible if it was (as often as possible) an object whose values could be either `SystemStyleObject` or functions that take `props` and return `SystemStyleObject`

_Originally posted by @with-heart in https://github.com/chakra-ui/chakra-ui/pull/1447#issuecomment-668128127_

As proposed, this PR aims to refactor `styleConfig` and `multiStyleConfig` by automatically calling them when processing theme instead of calling each component individually.


cereallarceny:
@with-heart @segunadebayo I think you all will be very happy with this abstraction. We wanted to basically ensure that `styleConfig`/`multiStyleConfig` can continue their usage in providing types to themes... but not in a way that makes it a pain to override individual styles. Currently, you have to construct an entirely new style and override **every** property in the object, before passing it to one of these two functions. This PR would allow that step to be completely internal, and instead allow you to simply "override what you want to change, and not a damn thing more". 😉 

Because this is a significant change to the way theming works, myself and @hericlesme will be working on changing the docs to explain the way that theming **should** work given these changes.

packages/core/src/chakra-provider.tsx
```diff
@@ -45,6 +46,14 @@ export const ChakraProvider: React.FC<ChakraProviderProps> = (props) => {
     throw Error("ChakraProvider: the `theme` prop is required")
   }
 
+  Object.keys(theme.components).forEach((component) => {
+    return theme.components[component].parts
```
ljosberinn, line 50: besides `return` doing nothing in a forEach, I'd prefer this out of readability and testability reasons:

```ts
const applyStyleConfigs = (theme: Theme) => {
  Object.keys(theme.components).forEach((component) => {
    const config = theme.components[component];

    if(config.parts) {
      theme.components[component] = multiStyleConfig(config);
    } else {
      theme.components[component] = styleConfig(config)
    } 
  });

 return theme;
}
```

packages/theme/src/components/index.ts
```diff
@@ -1,38 +1,38 @@
...
-import Popover, { popoverStyles } from "./popover"
-import Form, { formStyles } from "./form"
-import Editable, { editableStyles } from "./editable"
-import Progress, { progressStyles } from "./progress"
-import Textarea, { textareaStyles } from "./textarea"
-import Stat, { statStyles } from "./stat"
-import Skeleton, { skeletonStyles } from "./skeleton"
-import Breadcrumb, { breadcrumbStyles } from "./breadcumb"
-import SkipLink, { skipLinkStyles } from "./skip-link"
+import { default as Accordion } from "./accordion"
```
ljosberinn, line 1: if we plan on aliasing anyways, we should consider using named exports in the first place

packages/core/src/chakra-provider.tsx
ljosberinn: we might keep `applyStyleConfigs` in `theme-tools` then too, importing `multiStyleConfig`/`styleConfig` would be internal then I think

cereallarceny:
@ljosberinn @with-heart @segunadebayo As we work through this PR together, @hericlesme and I have figured out what we think the ideal final syntax should be. Allow me to lay this out as best as I can:

1. All components will be exported as named exports, no default exports. We will not import and re-export the components in the `src/components/index.ts` file. There is no longer a need for this file. Instead, we will import and re-export all the components in the main `src/index.ts` file, which will also be used to construct the `components` object for the theme. This will allow us to have `import theme from '@chakra-ui/theme';` as the default export, and `import { Accordion } from '@chakra-ui/theme';` for the named exports.
2. Theming can be done in one of two ways. I have preference for the second method, but I think it's worth documenting both so that users can decide what makes the most sense for them (based on how many overrides they need to make). We're following a mentality of "import the theme, override what you need". This gives us the benefits of a) always having the latest theme file, so long as your dependencies are always up to date, b) having "smart defaults" be placed inside the core theme files and out-of-sight by the developer, and c) keep a light footprint in the developer's code, ensuring that they're only focusing on the changes they need to make, and not reconstructing entire components by themselves. Here are the two methods:

**Method 1 (_simple, but ugly_):**
```js
import theme, { Accordion } from '@chakra-ui/theme';

Accordion.baseStyle.myChange = '#f00';

theme.components.Accordion = Accordion;

export default theme;

**Comment by cereallarceny** [comment]
@ljosberinn @with-heart @segunadebayo As we work through this PR together, @hericlesme and I have figured out what we think the ideal final syntax should be. Allow me to lay this out as best as I can:

1. All components will be exported as named exports, no default exports. We will not import and re-export the components in the `src/components/index.ts` file. There is no longer a need for this file. Instead, we will import and re-export all the components in the main `src/index.ts` file, which will also be used to construct the `components` object for the theme. This will allow us to have `import theme from '@chakra-ui/theme';` as the default export, and `import { Accordion } from '@chakra-ui/theme';` for the named exports.
2. Theming can be done in one of two ways. I have preference for the second method, but I think it's worth documenting both so that users can decide what makes the most sense for them (based on how many overrides they need to make). We're following a mentality of "import the theme, override what you need". This gives us the benefits of a) always having the latest theme file, so long as your dependencies are always up to date, b) having "smart defaults" be placed inside the core theme files and out-of-sight by the developer, and c) keep a light footprint in the developer's code, ensuring that they're only focusing on the changes they need to make, and not reconstructing entire components by themselves. Here are the two methods:

**Method 1 (_simple, but ugly_):**
```js
import theme, { Accordion } from '@chakra-ui/theme';

// Make whatever change you want here
Accordion.baseStyle.someProperty = '#f00';

theme.components.Accordion = Accordion;

export default theme;
```

**Method 2 (_slightly more involved, but way more extensible_):**
```js
import theme from '@chakra-ui/theme';
import { merge } from '@chakra-ui/utils';

const Accordion = {
  baseStyle: {
    myChange: '#f00'
  }
};

const overrides = {
  components: {
    Accordion
  }
};

export default merge(theme, overrides);
```

**I know in my heart of hearts the clear winner is number 2.** You could very easily abstract any overrides you want to be in separate "component files" (`Accordion.js`, etc.) and then reimport them to your main override object, before merging into the theme.

What I like most about either of these two ways to write component styles is that there's so little to teach. Our theming documentation consists of 4 different files... why? Theming should be the easy part, the hard part should be creating new components. We intend to simplify documentation in this PR as well after we get these changes pushed up - but I think this is the API we're going to propose. Thoughts everyone?

packages/core/src/chakra-provider.tsx
hericlesme: @ljosberinn what do you think of writing this function in `@chakra-ui/theme-tools` (`theme-tools/src/style-config`)?

packages/theme/src/components/index.ts
cereallarceny: Done!

ljosberinn:
https://github.com/chakra-ui/chakra-ui/issues/1523 very much related :)

I'm also team 2 here, imo the only scalable approach because else you end up making a variable out of it and loop it yourself, or have 374698 lines of mutating the theme.

packages/core/src/chakra-provider.tsx
ljosberinn: sounds perfect

jmiazga:
I really like the idea around this change. Theming is one of Chakra's strengths, and anything we can do to improve its usability seems like a big win. 

I would vote for method 2 as well, it seems more elegant. Method 1 feels dirty to me.

cereallarceny:
@ljosberinn @jmiazga In that case, if we go with ONLY the second experience, then there’s 0 need for the named export. In which case we can reintroduce `src/components/index.ts` and remove the clutter from `src/index.ts` since we will only be default exporting the theme from there. You can still get the components actually from the theme this way. You just call `const { Accordion } = theme.components;` instead of doing so from the named export.

This would also be nice, because we're having name conflicts with the components themselves:
`import { Accordion } from '@chakra-ui/theme';` conflicts with `import Accordion from '@chakra-ui/accordion';`

packages/core/src/chakra-provider.tsx
hericlesme: @ljosberinn I made some changes and created the function. Does it seem good to you?

hericlesme: Also, what do you think about changing the theme type in the `ThemeProvider`?

From `Dict`
```Typescript
export interface ThemeProviderProps {
  /**
   * The theme to use for your application
   */
  theme: Dict
}
```
To `Theme`

```Typescript
import { Theme } from '@chakra-ui/theme'

export interface ThemeProviderProps {
  ...
  theme: Theme
}
```

packages/theme-tools/src/style-config/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+      theme.components[component] = multiStyleConfig(config)
+    } else {
+      theme.components[component] = styleConfig(config)
+    }
+  })
+
+  return theme
+}
+
+export { applyStyleConfigs }
```
ljosberinn, line 19: can just do `export const` above :)

packages/core/src/chakra-provider.tsx
ljosberinn: I'm weary of a change of the theme provider, would it still work if you omit parts of the theme, e.g. because you're not using all components and want to reduce bundle size?

hericlesme: hmmm got your point, so let's keep it `Dict` for now

cereallarceny:
@segunadebayo @ljosberinn @jmiazga @with-heart @hericlesme 

I think we've gotten all the changes in that we want here, and things seem to be working normally. I think we should move on to addressing the theming documentation for the Chakra website next in this same PR. A few things:

1. We've gotten rid of default exports for components, but we're no longer exporting the component styles as named exports on `@chakra-ui/theme`. This is intentional as we have clear preference to **method 2** described above, which doesn't necessitate the need for those styles to be exported at all. Likewise, if you do need those styles for some reason, you could always get them from `theme.components.MyComponent` via the default `theme` export. Problem solved.
2. We ran into some headache with `styleConfig`, `useStyleConfig`, `multiStyleConfig`, and `useMultiStyleConfig`... what a nightmare of bad variable naming! Apparently, `styleConfig` and `multiStyleConfig apply Typescript types to a style object. While `useStyleConfig` and `useMultiStyleConfig` apply the style object (whether it's been "typescript-ified" or not) to the global theme. The former is not useful to the developer, so we internalized it. The latter is useful in creating **new** components, but not in styling existing ones. @segunadebayo, it would be a pretty big refactor to overhaul this - so I'm not really willing to touch it personally, but you should definitely look to clean up and rename these functions in the future. I highly suspect that there are some inefficiencies in this system as well that could use some attention. Anyhow, enough complaining... moving on!
3. The big issue here is theming. After reading through the theming docs multiple times, I've concluded the following: a) the theming docs don't actually work if you copy/paste them to their entirety, b) they only talk about **creating new components**, but say nothing about modifying existing ones, c) they are confusing because the examples used (`Button`, `Badge`, etc.) reference existing Chakra components, but it's clear you're trying to show how to create a new one... these should be renamed to `MyButton` and `MyBadge` to be more clear, and d) there are simply too many pages. I'll address each of these changes below, and with your blessing - we'll get started on cleaning some of this up.

**Add page for modifying existing components**
Our entire PR here purely intends to change the way in which existing components are themed. I think we've got a good idea about how to do this from the `method 2` example above. We'll write a new page under the "Theming" key called "Modify existing components" that explains how to modify existing components and set up a project _properly_ for good theming practice. Currently, we don't provide the user any instructions on how to set their project up for theming appropriately, and this is definitely a step we need to elaborate on further.

**Combine and rename**
The "Overview" and "Basic" pages should be combined into one page called "Creating new components". We'll also correct the examples to actually work, and we'll rename the components to something more obvious that it's a _new_ component: i.e. `Button` --> `MyButton`, `Badge` --> `MyBadge`, etc. We'll also only be telling the user how to create new components and apply the `useStyleConfig`/`useMultiStyleConfig` at the **global-level**. Given the fact that modifying existing components should be done at the global level (_see "method 2" above_), we should not confuse users by telling them how to do this at the local-level. Whether you're modifying an existing component or creating a new one, this API should be the same, and it should be done in a dedicated, global folder specifically just for theming.

----------

That's roughly it. There may be some small corrections on the advanced page to update to the new API changes explained here, but that should definitely simplify things quite a bit. We welcome any feedback, but we're going to go ahead and get started on this. I feel that even if you're a bit skeptical about this direction initially, you'll like it once we're done with it.

with-heart:
@cereallarceny @hericlesme Full steam ahead! I appreciate the thoroughness in the plan, and think the changes will be quite welcome. Refactoring to clarify the difference between creating new components and modifying existing components, and between global-level and local-level are excellent.

segunadebayo:
Wow. After looking through the changes and comments, this makes a lot of sense @cereallarceny and @hericlesme. 

Thanks for taking the time to think of a better approach to theming 💖.

packages/theme-tools/src/style-config/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+export const applyStyleConfigs = (theme: Dict) => {
+  Object.keys(theme.components).forEach((component) => {
+    const config = theme.components[component]
+
+    if (config.parts) {
+      theme.components[component] = multiStyleConfig(config)
+    } else {
+      theme.components[component] = styleConfig(config)
+    }
+  })
```
segunadebayo, line 14: `multiStyleConfig` and `styleConfig` are function that literally returns their arguments. I initially made them to provide style autocompletion and intellisense only.

That said, I'm not sure how useful the function above is.

cereallarceny:
Documentation is done @segunadebayo @with-heart. We have a series of conflicts now relating to theming. We'll clear these up and then I think this PR should be good for review! 😄

cereallarceny:
@segunadebayo @with-heart Hey guys, I've fixed the conflicts. This PR is ready to go!

PR #1648, opened 2020-08-13
docs: fix package name in the migration guide

CodinCat:
Just a trivial typo. It should be `emotion-theming` instead of `@emotion/theming`.
By default the packages are ordered alphabetically, so I also updated the ordering here.


with-heart:
@allcontributors please add @CodinCat for doc

Issue #1118, opened 2020-07-03
[Menu] V1 Rendering Performance Regression


dmitru:
**Describe the bug**

Rendering performance of Menu component (and possibly some others) got a lot worse (~8-10 times) after migrating a codebase to V1.0.

Description of a JSX layout rendered:
- 20 Boxes each having a Menu with 6 MenuItems

**Expected Behavior** 

Rendering of such a layout should be reasonably fast.

**Link to minimal reproduction** 

### V1.0 (huge lag when rendering):
~~https://codesandbox.io/s/staging-tree-fywty?file=/src/App.js~~
Updated Sandbox: https://codesandbox.io/s/inspiring-nightingale-wmmjs?file=/src/App.js

### Pre-V1.0, rendering is instant:
https://codesandbox.io/s/condescending-hugle-40nim?file=/src/App.js

**Steps to Reproduce** 

Open Sandbox, click on "Toggle" button to hide and then un-hide the thumbnails. Notice a significant lag while rendering the layout.

**Desktop (please complete the following information):**

- OS: OS X most recent
- Browser: Chrome
- Version: most recent

**Additional info**

React profiler screenshots - before V1.0:

![image](https://user-images.githubusercontent.com/948486/86486971-b9f29d00-bd11-11ea-98fb-ad54a856e3ed.png)

React profiler screenshots - after V1.0 migration:

![image](https://user-images.githubusercontent.com/948486/86486979-bc54f700-bd11-11ea-8f9a-51c5dae17260.png)

carlreid:
I am not sure if it is related or not, but I also saw pretty bad frame rate drops too while I scroll. But for me, I think I traced it back to Popper which I believe is used for `Menu`.  
  
You can actually just see the frame rate drops if you use https://next.chakra-ui.com/docs/components/menu and compare it to say for example https://next.chakra-ui.com/docs/components/modal. Simply scrolling causes the frame rate drops, no other interactions at all.

Menu example:
![image](https://user-images.githubusercontent.com/33623601/86489141-aa03a900-bd63-11ea-9e6c-9af9ffdf61ce.png)  
  
Modal example:
![image](https://user-images.githubusercontent.com/33623601/86489154-bab41f00-bd63-11ea-9ba2-bf07b17fb56e.png)

In my own project, I have the following. Which, if I comment out the use of `Menu` then I get a constant FPS.
[...]

PR #1656, opened 2020-08-14
Performance: Add isLazy property to Tabs

### LINKED ISSUE #1384: [Tabs] Tabs get mounted when they are inactive
Hi there!
First of all I would like to thank you for this fantastic library. It's really impressive!

**Expected Behavior**

The Tabs component only mounts the contents of a `TabPanel` to the DOM when the tab is activated/clicked/selected.

https://codesandbox.io/s/chakra-tabs-mounting-early-kcign

**Steps to Reproduce**
1. Open the [codesandbox](https://codesandbox.io/s/chakra-tabs-mounting-early-kcign)
2. Open the console view
3. See console logs from both tabs, although we only ever activated the first one

**Suggested solution(s)**

Instead of using `display: none` to hide the DOM elements, there should also be an option to not mount the DOM elements at all until the tab is activated.

- OS: MacOS 10.15.5
- Browser: Chromium
- Version: 84.0.4132.0

**Additional context**

I'm currently using the Tabs component in an application where every tab fetches *a lot* of data. The fact that the Tabs component renders all tabs to the DOM and thereby fires lots of HTTP requests simultaneously is causing performance issues.
I have tried to use the `isManual` option, but the tabs still get rendered to the DOM.

hericlesme:
### Description
Add `isLazy` property to `Tabs` to only mount the contents of a TabPanel to the DOM when the tab is activated/clicked/selected.

### Linked Issues
Fixes #1384

### How has this been tested?
A new test has been added in `packages/tabs/tests/tabs.test.tsx`

#
Related:  #1118

### DIFF
--- a/packages/tabs/src/use-tabs.ts
+++ b/packages/tabs/src/use-tabs.ts
@@ -57,6 +57,12 @@ export interface UseTabsProps {
    * The id of the tab
    */
   id?: string
+  /**
+   * Performance 🚀:
+   * If `true`, the MenuItem rendering will be deferred
+   * until the menu is open.
+   */
+  isLazy?: boolean
 }
 
 /**
@@ -74,6 +80,7 @@ export function useTabs(props: UseTabsProps) {
     onChange,
     index,
     isManual,
+    isLazy,
     orientation = "horizontal",
     ...htmlProps
   } = props
@@ -157,6 +164,7 @@ export function useTabs(props: UseTabsProps) {
     setSelectedIndex,
     setFocusedIndex,
     isManual,
+    isLazy,
     orientation,
     enabledDomContext,
     domContext,
@@ -378,8 +386,11 @@ export function useTabPanels<P extends UseTabPanelsProps>(props: P) {
  */
 export function useTabPanel(props: Dict) {
   const { isSelected, id, ...htmlProps } = props
+  const { isLazy } = useTabsContext()
+
   return {
     ...htmlProps,
+    children: !isLazy || isSelected ? props.children : null,
     role: "tabpanel",
     hidden: !isSelected,
     id,
--- a/website/pages/docs/components/tabs.mdx
+++ b/website/pages/docs/components/tabs.mdx
@@ -310,6 +310,31 @@ tab by pressing <kbd>Space</kbd> or <kbd>Enter</kbd>.
 </Tabs>
 ```
 
+### Lazily mounting tab panels
+
+By default, the Tabs component renders all tabs to the DOM If you'd like to only mount the contents of a `TabPanel` to the DOM at once.
+If you want to defer rendering of content until the tab is activated/clicked/selected, you can use the `isLazy` property.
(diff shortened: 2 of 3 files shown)


packages/tabs/src/use-tabs.ts
```diff
@@ -57,6 +57,12 @@ export interface UseTabsProps {
    * The id of the tab
    */
   id?: string
+  /**
+   * Performance 🚀:
+   * If `true`, the MenuItem rendering will be deferred
+   * until the menu is open.
```
with-heart, line 63: Should probably refer to these as tabs :)

jmiazga:
I recently used something similar to this in react-bootstrap, but it was broken up as 2 props, `mountOnEnter` and `unmountOnExit`. I wonder if we want to do something like that?

https://react-bootstrap.github.io/components/tabs/#tabs-props

hericlesme:
> Very nice @hericlesme! Could you add a note about this to the docs as well?

Sure thing, @with-heart!

packages/tabs/src/use-tabs.ts
hericlesme: Oops... Done!

with-heart:
@jmiazga imo that's probably overkill!

jmiazga:
@with-heart That's fair. The only reason I call that out is if a component has a `useEffect` that makes an api call or something, that call will fire every time the component is mounted. That will force the user to pull that call up from the tab.

with-heart:
@jmiazga You raise a good point! Hadn't thought about that. I think it's still fine though, because it's pretty easy to lift something like that up into a callback passed to the component rendered by the tab.

I feel like we should go with this for now and if it turns out to be a common frustration to have to do something like that, we can add the other two later.

hericlesme:
Docs added, @with-heart :book: :tada:

website/pages/docs/components/tabs.mdx
```diff
@@ -310,6 +310,31 @@ tab by pressing <kbd>Space</kbd> or <kbd>Enter</kbd>.
 </Tabs>
 ```
 
+### Lazily mounting tab panels
+
+By default, the Tabs component renders all tabs to the DOM If you'd like to only mount the contents of a `TabPanel` to the DOM at once.
+If you want to defer rendering of content until the tab is activated/clicked/selected, you can use the `isLazy` property.
+In this way, the tab content will not be mounted to the DOM until it is open.
```
with-heart, line 317:
```suggestion
By default, the `Tabs` component renders all tabs content to the DOM, meaning that invisible tabs are still rendered but are hidden by styles.

If you want to defer rendering of each tab until that tab is selected, you can use the `isLazy` prop. This is useful if your tabs require heavy performance, or make network calls on mount that should only happen when the component is displayed.
```

Trying to edit this a little bit to help explain why you'd want to use it. How does this sound @ljosberinn?

ljosberinn: lgtm!

PR #1667, opened 2020-08-14
fix(transition): pass nodeRef to React transition group to avoid warning

### LINKED ISSUE #1496: [SlideFade] findDOMNode deprecation warning in StrictMode
**Describe the bug** 
findDOMNode warning when using StrictMode (v1 rc.0)

Reproduction: https://codesandbox.io/s/trusting-nightingale-kiij9?file=/src/index.tsx

To see the bug - toggle the animation and see the console.

This is because we don't use the `nodeRef` prop when using the `react-transition-group` (you can read more about the nodeRef prop in the [changelog](https://github.com/reactjs/react-transition-group/blob/1fd4a65ac45edd2aea3dec18eeb8b9c07c7eb93f/CHANGELOG.md#440-2020-05-05)) We can't really use it without changing the API either, so I can't really create a good PR to fix it without discussing the API implications with you.

One proposed API is changing the returned function value to include ref as well, changing the API from this

```tsx
<Fade in={isVisible}>
  {style => <Box style={style} />}
</Fade>
```

to this:

```tsx
<Fade in={isVisible}>
  {fadeProps => <Box {...fadeProps} />}
</Fade>
```

where `fadeProps` is an object containing the following:

```tsx
{
  style: style,
  ref: nodeRef,
}
```

selbekk:
This change passes a reference to the Transition component, which silences a warning. This means
that the API of our Transition component has to change as well!

BREAKING CHANGE: The Transition, Slide, Fade, Scale and SlideFade components now returns an object
with a ref prop and a style prop instead of just the style object. This means you now have to spread
the children argument on the element that is being transitioned.

fix #1496

### DIFF
--- a/packages/drawer/src/drawer.tsx
+++ b/packages/drawer/src/drawer.tsx
@@ -1,5 +1,10 @@
 import * as React from "react"
-import { Slide, SlideProps, Fade } from "@chakra-ui/transition"
+import {
+  Slide,
+  SlideProps,
+  Fade,
+  TransitionChildrenProps,
+} from "@chakra-ui/transition"
 import {
   Modal,
   ModalProps,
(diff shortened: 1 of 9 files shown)


selbekk:
Ah I'll fix all the build errors. Gimme a few

selbekk:
This should be ready to review now. Please let me know if there is any outstanding work you would like for me to do here.

with-heart (approved):

segunadebayo:
Thank you for your work on this PR.

We're migrating all motion/transition to `framer-motion`, making this PR stale in a couple of days.

Permit me to close it for now.

selbekk:
Oh that's a move I like. Very fine.

Issue #1666, opened 2020-08-14
[AvatarGroup] AvatarGroup shows max even when max hasn't been hit yet


PierreAndreis:
**Describe the bug** AvatarGroup shows max even when max hasn't been hit yet

**Expected Behavior** AvatarGroup should not display the over the max count if the length of the avatar isn't more than the max given

**Link to minimal reproduction** 
https://codesandbox.io/s/charming-cartwright-sn705?file=/src/index.tsx

**Steps to Reproduce** 
Use `<AvatarGroup>` with `max` set as more than the `<Avatars` given.

**Suggested solution(s)**
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/avatar/src/avatar-group.tsx#L102-L108
Only show excess avatar if `excess > 0`

**Desktop (please complete the following information):**

- OS: Mac
- Browser Chrome
- Version Chrome 84, Chakra 1.0 rc2

PR #1668, opened 2020-08-14
fix(avatar): avatar group shows excess even when excess isn't hit

PierreAndreis:
There was a bug with the Avatar Group showing excess even when excess hasn't been reached. 
This fixes https://github.com/chakra-ui/chakra-ui/issues/1666

No breaking changes.

### DIFF
--- a/packages/avatar/src/avatar-group.tsx
+++ b/packages/avatar/src/avatar-group.tsx
@@ -99,9 +99,10 @@ export const AvatarGroup = forwardRef<AvatarGroupProps, "div">(
         {...rest}
         className={cx("chakra-avatar__group", props.className)}
       >
-        {excess && (
+        {excess > 0 && (
           <chakra.span
             className="chakra-avatar__excess"
+            data-testid="avatar-excess"
             __css={excessStyles}
             children={`+${excess}`}
           />
--- a/packages/avatar/src/avatar-group.tsx
+++ b/packages/avatar/src/avatar-group.tsx
@@ -102,7 +102,6 @@ export const AvatarGroup = forwardRef<AvatarGroupProps, "div">(
         {excess > 0 && (
           <chakra.span
             className="chakra-avatar__excess"
-            data-testid="avatar-excess"
             __css={excessStyles}
             children={`+${excess}`}
           />
--- a/packages/avatar/tests/avatar-group.test.tsx
+++ b/packages/avatar/tests/avatar-group.test.tsx
@@ -24,3 +24,29 @@ test("renders a number avatar showing count of truncated avatars", () => {
   const moreLabel = tools.getByText("+3")
   expect(moreLabel).toBeInTheDocument()
 })
+
+test("does not render a number avatar showing count of truncated avatars if max is equal to avatars given", async () => {
+  const tools = render(
+    <AvatarGroup max={5}>
+      <Avatar />
+      <Avatar />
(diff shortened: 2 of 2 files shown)


packages/avatar/src/avatar-group.tsx
```diff
@@ -99,9 +99,10 @@ export const AvatarGroup = forwardRef<AvatarGroupProps, "div">(
         {...rest}
         className={cx("chakra-avatar__group", props.className)}
       >
-        {excess && (
+        {excess > 0 && (
           <chakra.span
             className="chakra-avatar__excess"
+            data-testid="avatar-excess"
```
ljosberinn, line 105: I'd like to avoid hvaing `data-testid` in production code, as its only an [escape hatch](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid). 

Could you try `screen.debug()` and paste the content to the top left of https://testing-playground.com/, then select the excess element on the top right and copy paste the selector from bottom left when the testid is gone? It might fall back to `container.querySelector('.chakra-avatar--excess')` or `screen.getByText('+1')`. I'd be fine with either, although I'd prefer the 2nd.

packages/avatar/tests/avatar-group.test.tsx
```diff
@@ -24,3 +24,29 @@ test("renders a number avatar showing count of truncated avatars", () => {
...
+  const tools = render(
+    <AvatarGroup max={5}>
+      <Avatar />
+      <Avatar />
+      <Avatar />
+      <Avatar />
+    </AvatarGroup>,
+  )
+  const moreLabel = tools.queryByTestId("avatar-excess")
+  expect(moreLabel).toBeFalsy()
```
ljosberinn, line 38: Please use `.not.toBeInTheDocument()` instead :) see https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/prefer-presence-queries.md

packages/avatar/tests/avatar-group.test.tsx
```diff
@@ -24,3 +24,29 @@ test("renders a number avatar showing count of truncated avatars", () => {
...
+  const tools = render(
+    <AvatarGroup max={6}>
+      <Avatar />
+      <Avatar />
+      <Avatar />
+      <Avatar />
+    </AvatarGroup>,
+  )
+  const moreLabel = tools.queryByTestId("avatar-excess")
+  expect(moreLabel).toBeFalsy()
```
ljosberinn, line 51: same as above

ljosberinn:
Thank you!

packages/avatar/src/avatar-group.tsx
PierreAndreis: makes total sense! I ended up going for testid because I couldn't find a reliable way of detecting that avatar excess is not on the page.
I thought about using `screen.getByText("+", {exact: false})` but wasn't sure if this was a valid assumption that there will only be a `+` on this component if the excess avatar is there.

I will switch out for the querySelector. I tried using that tool but unfortunately still couldn't find a better query

ljosberinn: imo the test should break if the `+` changes because its a user-facing UI change that, which is exactly what testing-lib concerns itself with :)

ljosberinn:
@allcontributors please add @PierreAndreis for code

PR #1692, opened 2020-08-16
fix(accordion): move item props to accordion context

### LINKED ISSUE #1691: [Accordion] Accordion doesn't work well when children are components
**Describe the bug**  
Setting `allowMulitple={false}` on Accordion still allows multiple items to be open at once when the Accordion's children is a `map` of React components.

**Expected Behavior** 
Setting `allowMulitple={false}` on Accordion should prevent multiple accordion items from being open at once.

https://codesandbox.io/s/accordion-allowmultiple-not-working-4dqcn?file=/src/App.js

**Steps to Reproduce** 
1. Click one of the accordion items in the demo window
2. Click a second accordion item and notice the first item did not close
3. Notice `allowMulitple={false}` is set on Accordion

**Suggested solution(s)**

As a workaround, this works - replace the map of components with a function that returns a map of `AccordionItem`s. The props will function normally now. This cuts out the children of Accordion being React components which is my guess somehow related to the issue. 

- OS: Mac OS
- Browser Chrome
- Version 84

with-heart:
Resolves #1691

I'll allow the commit message to explain the issue and what this PR changes:

Before this change, `Accordion` passed the `isOpen` and `onChange` props
to its `AccordionItem` children by inject is props via
`React.cloneElement`. This method of props injection caused two issues:

- `AccordionItem` children could only be rendered as direct children of
  `Accordion`
- `AccordionItem` children rendered as a mapped array could not receive
  their corresponding props, breaking the functionality of the accordion

This changes moves the `isOpen` and `onChange` props into a props getter
created in `useAccordion`. This props getter takes an accordion item's
`index` (determined in `useAccordionItem` via `useDescendant`) to
determine these values. This props getter is then attached to
`AccordionContext`.

The props getter is then accessed via `AccordionContext` inside
`useAccordionItem` and passed the item's index. This allows the item to
function anywhere inside the `Accordion` tree, whether as direct
children or not, and as part of a mapped array.

---

I opened this as a draft because I wanted to make sure this change would be acceptable to us. This would mean that users could no longer directly pass `isOpen` and `onChange` as props to an `AccordionItem`, though I see no reason that they would need to do this anyway.

I also don't intend to keep the story in place. It's just so you can see an example of it working. If you use this story without the change, the accordion does nothing. I'd probably replace it with a test instead.

### DIFF
--- a/packages/accordion/src/accordion.tsx
+++ b/packages/accordion/src/accordion.tsx
@@ -50,13 +50,13 @@ export interface AccordionProps
  * @see Docs https://chakra-ui.com/components/accordion
  */
 export const Accordion = forwardRef<AccordionProps, "div">(function Accordion(
-  props,
+  { children, ...props },
   ref,
 ) {
   const styles = useMultiStyleConfig("Accordion", props)
   const _props = omitThemingProps(props)
 
-  const { children, htmlProps, ...context } = useAccordion(_props)
+  const { htmlProps, ...context } = useAccordion(_props)
 
   const _context = useMemo(
     () => ({ ...context, reduceMotion: !!props.reduceMotion }),
--- a/packages/accordion/src/use-accordion.ts
+++ b/packages/accordion/src/use-accordion.ts
@@ -7,21 +7,13 @@ import {
   Dict,
   getNextIndex,
   getPrevIndex,
-  getValidChildren,
   isArray,
   mergeRefs,
   removeItem,
   __DEV__,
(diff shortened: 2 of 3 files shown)


segunadebayo:
Great work @with-heart. Is this ready for review?

with-heart:
Lemme switch the stories over to tests and then it will be! :)

with-heart:
@segunadebayo This is good to go!

PR #1699, opened 2020-08-16
chore(workflows): update current workflows to only `pr` + `publish`

with-heart:
~~This PR creates a separate workflow for PRs, so that we can have full control over the PR process.~~

~~It currently lints (code & types), tests, and builds all packages. This should serve as a pretty reliable indication that a PR can be merged without causing any headache.~~

I expanded the scope of this PR to cover our workflows as a whole.

# Changes

- removed the `deploy-website` workflow as we're now using the Vercel for GitHub integration to handle site deployments
- removed the `node` workflow as it's no longer needed
- updated the `publish` workflow to only apply to `push` rather than both `pull_request` and `push` because `pull_request` would mean we're publishing new versions anytime anyone opens a PR
- added the `pr` workflow, which lints, tests, and builds all packages, giving us a fairly reliable indication that a PR can be safely merged

# Workflows

1. Developer makes changes locally and then opens a PR against `develop`.
2. `pr` workflow runs and passes.
3. Maintainer merges PR into `develop`.
4. After # of changes are made to `develop`, maintainer wants to publish a new version, so they open a PR against `master` from `develop`.
5. `pr` workflow runs and passes.
6. Maintainer merges PR into `master`.
7. `publish` workflow runs and passes. A new version is published to `npm`.

Here's how this all looks:

image


jmiazga:
Any reason we would lint and test on PRs, but not everywhere?

with-heart:
@jmiazga Nope. I'll get to updating for `pull` also 😄 This is just to separate the two for clarity, and also so we can potentially attach additional actions only to one of the workflows.

with-heart:
I think this should be good to go.

with-heart:
I expanded the scope of this to reduce our workflows down to `pr` and `publish`. See the PR description for updated explanation and details.

segunadebayo:
@with-heart Can you update this branch and get the checks to pass? I'd like to merge it

PR #1717, opened 2020-08-18
Breakout Links

joe-bell:
One of the most common bad habits I've seen for linking "cards" of content is to wrap everything in a link. This PR aims to support `breakout` links to offer the same functionality but ensure semantic HTML

Further reading:

[Screenshot of Tweet](https://twitter.com/joebell_/status/1269554122128269313)

## Examples

Here are some examples of the technique in the wild (although not using Chakra UI)

- [bbc.co.uk/news](https://bbc.co.uk/news) (article links)
- [joebell.co.uk](https://joebell.co.uk) (post links)
- [smashingmagazine.com](https://www.smashingmagazine.com/) (article links)

## Caveats

One of the side-effects for this technique is that the content can't be "selectable" (i.e with a pointing device), however this seems to be pretty trivial compared to the benefits!

## Alternative Solution

I posted a [link to the PR on Twitter](https://twitter.com/joebell_/status/1295641715903270914) and got some really useful feedback from @peduarte and @aarongarciah; mostly around the concern that this link variant relies on `position: relative` to work.

@peduarte suggested the following:

```jsx
<BreakoutContainer>
  <BreakoutLink />
</BreakoutContainer>
```

Although I do agree it would be clearer for the consumer, I'm concerned that this could cause fragmentation from `Link` and an additional element (most likely a `div`) would do more harm than good. Perhaps a `useLinkBreakoutParent()` hook could support this?

Either way there's small trade-offs, but I think as long as the docs are clear initially this could always be tweaked

### DIFF
--- a/packages/layout/src/link.tsx
+++ b/packages/layout/src/link.tsx
@@ -11,6 +11,11 @@ import { __DEV__, cx } from "@chakra-ui/utils"
 
 export interface LinkProps extends PropsOf<typeof chakra.a>, ThemingProps {
   /**
+   * If `true`, the link will stretch to the full width and height of its
+   * nearest parent whose `position` is `relative`.
+   */
+  breakout?: boolean
+  /**
    *  If `true`, the link will open in new tab
    */
   isExternal?: boolean
@@ -32,7 +37,7 @@ export interface LinkProps extends PropsOf<typeof chakra.a>, ThemingProps {
  */
 export const Link = forwardRef<LinkProps, "a">((props, ref) => {
   const styles = useStyleConfig("Link", props)
-  const { className, isExternal, ...rest } = omitThemingProps(props)
+  const { breakout, className, isExternal, ...rest } = omitThemingProps(props)
 
   return (
     <chakra.a
@@ -41,7 +46,22 @@ export const Link = forwardRef<LinkProps, "a">((props, ref) => {
       ref={ref}
       className={cx("chakra-link", className)}
       {...rest}
-      __css={styles}
+      __css={{
+        ...(breakout && {
+          position: "static",
+          "&::before": {
+            content: "''",
+            cursor: "inherit",
+            display: "block",
+            position: "absolute",
+            top: 0,
+            left: 0,
+            width: "100%",
+            height: "100%",
+          },
+        }),
+        ...styles,
+      }}
     />
   )
 })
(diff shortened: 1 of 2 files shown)


with-heart:
Great idea! This is something I've built for a few different sites and it's a lovely concept. Andy Bell's article made it so simple and repeatable! 

I definitely love the idea of it being a separate composite component. Seems like a perfect candidate because it'd be so easy to add new variants to the theme specifically for your breakouts.

I'll review the PR today and play around with it though! The hook might be a better option :)

segunadebayo:
Thank you for this PR, Joe.

I might have to agree with Pedro's feedback. Given that this was created to solve a specific problem: "adding a link around a card or component", I'd suggest we make this a standalone component.

This API might work

```jsx
<BreakoutContainer>
  <BreakoutLink />
</BreakoutContainer>
```

Here are some action steps:
- Add these components to `packages/layout/breakout-link` (or whatever name you prefer)
- Create stories to illustrate this use case
- Update the docs to let the user know that they can either use `position: relative` on any container they have or use `BreakpointContainer`

Another thought I had was that this could be a "guide" instead of making it a component.

Let me know your thoughts.

joe-bell:
@segunadebayo @with-heart cheers for the feedback and action steps. I'll take this all into account and play around with some adjustments over the next few days

segunadebayo:
@joe-bell, on the other hand, I think we could rename the component to make it more intuitive.

What if we used this API:

```jsx
<LinkBox href="#" isExternal {...boxProps}>
    {/** Any child goes here */}
</LinkBox>
```

Internally, we can handle all the needed complexity, render the children alongside the breakout `<a>`. How does this sound?

joe-bell:
@segunadebayo I think I follow you. So are you suggesting `LinkBox` as the wrapper component?

How would a consumer be able to define which `a` inside the `LinkBox` "breaks out" to fill?

For example; an `a` wrapping the title could "break out" but an `a` wrapping the author name could be treated as normal (like in the use-case for Smashing Magazine)

Maybe this could be done automatically when there's a single `a`, but for multiple anchors this could get complex

segunadebayo:
@joe-bell, I've added this to the layout components based on PR.

Thank you so much for this. Feel free to improve this component in a different PR if you have better ideas.

Cheers buddy.

joe-bell:
@segunadebayo nice one! Sorry this got stale, really appreciate you getting it through

PR #1722, opened 2020-08-18
Feat/extend create icon

### LINKED ISSUE #1720: [createIcon] extend to allow more scenarios
I'm trying to create a few icons right now, and most of them are not entirely trivial, as in they have more than a single `path`, and in any case, also a `title` child.

It would also be nice to pass some `defaultProps` to a created icon, such as `{ focusable: false }` or `{ fill: 'none' }`.

This is [what we have right now](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/icon/src/create-icon.tsx):
```ts
interface CreateIconOptions {
  /**
   * The icon `svg` viewBox
   * @default "0 0 24 24"
   */
  viewBox?: string
  /**
   * The `svg` path or group element
   */
  path?: React.ReactElement
  /**
   * If the has a single path, simply copy the path's `d` attribute
   */
  d?: string
  /**
   * The display name useful in the dev tools
   */
  displayName?: string
}
```

I'd propose:
```ts
interface CreateIconOptions {
  /**
   * The icon `svg` viewBox
   * @default "0 0 24 24"
   */
  viewBox?: string
  /**
   * The `svg` path or group element(s)
   */
  path?: React.ReactElement | React.ReactElement[]
  /**
   * If the has a single path, simply copy the path's `d` attribute
   */
  d?: string
  /**
   * The display name useful in the dev tools
   */
  displayName?: string
  /**
   * Default props automatically passed to the component; overwriteable
   */
  defaultProps?: PropsOf<Icon> // or something similar, just an example
}
```

Curious though whether that'd be worth it because effectively, it's just another way of using `<Icon />` at this point. Then again, that's kinda the goal?

ljosberinn:
fixes #1720

### DIFF
--- a/packages/icon/src/create-icon.tsx
+++ b/packages/icon/src/create-icon.tsx
@@ -1,7 +1,7 @@
 import React from "react"
 import { Icon, IconProps } from "./icon"
 import { __DEV__ } from "@chakra-ui/utils"
-import { forwardRef } from "@chakra-ui/system"
+import { forwardRef, PropsOf } from "@chakra-ui/system"
 
 interface CreateIconOptions {
   /**
@@ -12,7 +12,7 @@ interface CreateIconOptions {
   /**
    * The `svg` path or group element
    */
-  path?: React.ReactElement
+  path?: React.ReactElement | React.ReactElement[]
   /**
    * If the has a single path, simply copy the path's `d` attribute
    */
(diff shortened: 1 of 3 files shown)


ljosberinn:
FWIW, it was previously already possible to use a fragment in `path`, which also conveniently avoids having to add `key`s to array elements when using the array api. In fact I prefer that approach, so I'm happy to remove `React.Element[]` again. Might consider renaming `path` to `children` then though.

segunadebayo:
We'll definitely remove or update `createIcon` once we create a script/CLI that convert svgs to React components.

For now, thanks for this update.

PR #1723, opened 2020-08-18
feature(standalone-toasts): create toasts outside of react components

### LINKED ISSUE #1693: [Toast] - How to use toast outside of component?
The chakra toasts are really beautiful, but unfortunately it seems like there is no way to use them outside of functional components!! 

I want to invoke the toast on the success or failure of Rest API, which obviously is in .ts file and outside of component. 

Any solution/workaround for this would be highly appreciated. Thanks!

Fo eg: 

In my user component:

```
export const Users = () => {
   .......

    useEffect(() => {
      dispatch(getUsers());
    }, [dispatch]);

   ........
 }

```

and then in my user action file : 

```
export const getUsers = () => {
    return (dispatch: Dispatch) => {
       getUsersService()
       .then((response: any) => {
         dispatch(setUserList(response));
         // i want to invoke success toast here
       })
      .catch((error: Error) => {
        // i want to invoke error toast here
      });
  };
};
```

TimKolberger:
As mentioned in #1693 application code outside of the react context should be able to create a new toast.
E.g. in API abstractions or Vanilla JS files it can simplify the code.

Fix #1693

## Example 

```
import { createStandaloneToast } from "@chakra/core"

const toast = createStandaloneToast()
// const customToast = createStandaloneToast({ theme: customTheme })

const fetchUsers = async () => {
  await fetch() // ...
  toast({
    title: "Yay",
    status: "success"
  })
}
```

### DIFF
--- a/packages/toast/test/standalone-toast.test.tsx
+++ b/packages/toast/test/standalone-toast.test.tsx
@@ -4,7 +4,7 @@ import { createStandaloneToast } from "../src"
 import theme from "@chakra-ui/theme"
 
 test("Standalone toast renders correctly", async () => {
-  const toast = createStandaloneToast(theme)
+  const toast = createStandaloneToast({ theme })
   const title = "Yay!"
   const description = "Something awesome happened"
 
--- a/packages/toast/test/standalone-toast.test.tsx
+++ b/packages/toast/test/standalone-toast.test.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from "react"
-import { render } from "@chakra-ui/test-utils"
+import { render, screen } from "@chakra-ui/test-utils"
 import { createStandaloneToast } from "../src"
 import theme from "@chakra-ui/theme"
 
@@ -21,7 +21,7 @@ test("Standalone toast renders correctly", async () => {
     return null
   }
 
-  const screen = render(<ToastContainer />)
+  render(<ToastContainer />)
 
   const allByTitle = await screen.findAllByText(title)
   const allByDescription = await screen.findAllByText(description)
--- a/packages/toast/test/standalone-toast.test.tsx
+++ b/packages/toast/test/standalone-toast.test.tsx
@@ -1,10 +1,9 @@
 import React from "react"
 import { screen } from "@chakra-ui/test-utils"
 import { createStandaloneToast } from "../src"
-import theme from "@chakra-ui/theme"
 
 test("Standalone toast renders correctly", async () => {
-  const toast = createStandaloneToast({ theme })
+  const toast = createStandaloneToast()
   const title = "Yay!"
   const description = "Something awesome happened"
 
--- a/packages/toast/test/standalone-toast.test.tsx
+++ b/packages/toast/test/standalone-toast.test.tsx
@@ -21,10 +21,10 @@ test("Standalone toast renders correctly", async () => {
     return null
   }
 
-  const { findAllByText } = render(<ToastContainer />)
+  const screen = render(<ToastContainer />)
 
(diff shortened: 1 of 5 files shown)


packages/toast/test/standalone-toast.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useEffect } from "react"
+import { render } from "@chakra-ui/test-utils"
+import { createStandaloneToast } from "../src"
+import theme from "@chakra-ui/theme"
```
with-heart, line 4: Make sure to add `@chakra-ui/theme` to `devDependencies` in `packages/toast/package.json`

packages/toast/test/standalone-toast.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+        isClosable: true,
+      })
+    }, [])
+
+    return null
+  }
+
+  const { findAllByText } = render(<ToastContainer />)
+
+  const allByTitle = await findAllByText(title)
```
with-heart, line 26: Can you switch this to use `screen.findAllByText` rather than destructuring from `render`?

packages/toast/test/standalone-toast.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import theme from "@chakra-ui/theme"
+
+test("Standalone toast renders correctly", async () => {
+  const toast = createStandaloneToast(theme)
+  const title = "Yay!"
+  const description = "Something awesome happened"
+
+  const ToastContainer = () => {
+    useEffect(() => {
+      toast({
```
with-heart, line 13: Very cool!

TimKolberger: ❤️

packages/toast/test/standalone-toast.test.tsx
TimKolberger: ✅  done

packages/toast/test/standalone-toast.test.tsx
TimKolberger: ✅  done

jmiazga:
Is there a benefit for keeping the useToast hook?

TimKolberger:
@jmiazga Yes, the hook `const toast = useToast()` will get the current chakra theme from the React context. 
This way it does not depend on the developer to pass the theme explicitly.

With the standalone version you need to pass your theme like: `const toast = createStandaloneToast(theme)`

trompx:
I'm trying to use Toast with apollo-link-error. I'm not sure if this will work but looks like it. Anything I can do to help?

TimKolberger:
I rebased the branch. The vercel Deployment should be working now.

TimKolberger:
Is there anything I can do to bring this PR forward?
Otherwise I will just keep rebasing it with `develop` 😅

TimKolberger:
Due some changes in the useToast hook I updated the PR accordingly.
I refactored the duplicated render function into one and added default values for the object parameter of `createStandaloneToast`.

@with-heart Your last code review is outdated due those changes.

Will this PR be part of v1?

ljosberinn:
@TimKolberger sorry for the delay here - we're packed with work at ... work atm. I'm sure Mark will find time soon! 🤗

TimKolberger:
Sure - no pressure intended! =) 
I am a bit confused about the Github UI which says "Changes requested", but I those are outdated.
Just want to make sure we are not waiting on each other :D

ljosberinn:
He might have to manually accept the changes and until then its pending; I can't update anything here myself either :/

PR #1763, opened 2020-08-21
docs: create TRIAGE overview document

with-heart:
I've been working on this for awhile and I need to stop and take a break. But I think it covers a lot of my thinking on triage and hopefully covers the topic pretty well so far.

Wanted to open this as a draft to hopefully get some feedback before I start working on it again.

### DIFF
--- a/TRIAGE.md
+++ b/TRIAGE.md
@@ -1,0 +1,137 @@
+Thanks for showing interesting in Chakra triage! This document serves as an
+overview of the triage team for our repository. It covers topics such as the
+importance of triage in an open source project, the goals of the triage team,
+and the triage process itself.
+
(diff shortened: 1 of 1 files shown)


with-heart:
I think this is finally good to go. The document captures a lot of my thoughts around triage, and outlines the basic triage process.

codebender828 (approved):
Nice work here, @with-heart . 👍

aaronmcadam:
This is awesome @with-heart!

Is there any guidance on how to decide if a component should be added to Chakra?

with-heart:
@aaronadamsCA Nothing yet, but it's something I think we should have coming up in the future. Need to have a better process around things like that.

PR #1789, opened 2020-08-23
chore: add selectors for :only-of-type, :only-child

tony:
MDN: [:only-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-of-type), [:only-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-child)

I pause when I type `_only` expecting to see autocompletion and don't see anything. I haven't seen this brought up on the tracker, so I'll give it a go:

My use case was a mixture of MDXProvider heading components being reused.

Another common example is `.map()` of components, when only one item exists in the list. 

Note: There's ways to play with `_first`/`_last` and moving spacing to child components to get the intended effects of these.

Still I think from a developer friendliness some may like it. Thoughts?

### DIFF
--- a/packages/styled-system/src/pseudo/pseudo.selector.ts
+++ b/packages/styled-system/src/pseudo/pseudo.selector.ts
@@ -154,6 +154,14 @@ export const pseudoSelectors = {
    */
   _odd: "&:nth-of-type(odd)",
   /**
+   * Styles for CSS Selector `&:only-of-type`
+   */
+  _onlyOfType: "&:only-of-type",
+  /**
+   * Styles for CSS Selector `&:only-child`
+   */
+  _onlyChild: "&:only-child",
+  /**
    * Styles for CSS Selector `&:first-of-type`
    */
   _first: "&:first-of-type",


with-heart:
Going to defer to @segunadebayo on API ideas like this :)

tony:
@with-heart Thank you

Both @segunadebayo and @with-heart : I rebased this - if this isn't clearly a good API move, I'm fine if you close it

(I found a way to fix my issue without using `:only-of-type` / `:only-child`)

segunadebayo:
Hi @tony,

I'm a bit hesitant to add more pseudo-style props to the APIs. When you have some time, please share your walkaround or solution so others can learn from it as well.

Thank you.

PS: I'll close this out.

tony:
Thanks @segunadebayo 

I was converting raw CSS that included `:only-of-type`:

```
const markdownTheme: SystemStyleObject = {
  baseStyle: {
    "& > p": {
      marginBlockStart: ".5rem",
      marginBlockEnd: ".5rem"
    },
    // Header links
    "& > .heading a.heading-link": {
      opacity: "0"
    },
    "& > .heading:hover a": {
      opacity: "1"
    },
    "& > h2:only-child, & > h2:only-of-type": {
      marginTop: 0
    }
  }
};
```

In this case, often `h2` with `_last` or `_first` (e.g. `_first: {{ paddingBottom: 3 }}`) in the react component gets the equivalent behavior to `only-child`:

> The :only-child CSS pseudo-class represents an element without any siblings. This is the same as `:first-child:last-child` or `:nth-child(1):nth-last-child(1)`, but with a lower specificity.

https://developer.mozilla.org/en-US/docs/Web/CSS/:only-child

As for `:only-of-type`, I just don't use it. If you do hit an edge case where you need this, please articulate it in a new issue and feel free to cite this. Same with `:only-child`.

effinrich:
@tony I know this is a very old issue, but `:only-of-type` worked for me like this:

```
const baseStyle = {
  'a:only-of-type': {
    background: 'orange'
  },
  textTransform: 'normal',
  '::first-letter': {
    textTransform: 'uppercase'
  }
}
```

PR #1805, opened 2020-08-25
feat(focus-lock): relax type on focusable refs to handle non-HTMLElement refs

santialbo:
I've faced this issue several times where the types don't allow me to pass an instance of a non-HTMLElement ref to a `___FocusRef` prop. For example when I use `react-select` or the `slate` editor inside a Modal. 

Up until now I've always opted out of the type check by doing the `initialFocusRef={myRef as any}`.

This PR relaxes the type constraint of the different `focusRefs` I've found to use a more generic one that matches with what the library needs.


packages/utils/src/tabbable.ts
```diff
@@ -22,7 +22,7 @@ export function hasFocusWithin(element: Element) {
   return element.contains(document.activeElement)
 }
 
-export function isHTMLElement(element: Element): element is HTMLElement {
+export function isHTMLElement(element: any): element is HTMLElement {
```
aaronmcadam, line 25: I'm curious as to why we needed to change the type of `element` to `any`?

santialbo: It's because it is called with a `Focusable` item which might not be an Element.

aaronmcadam: I'm wondering if we could use [Function overloads](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) to solve this, but that's only my own curiosity. 

These changes make a lot of sense because they focus on the needs of the focus API, not the actual type of the object. Don't mind me 😄

santialbo: > I'm guessing a union type of `Focusable | Element` is a bit too awkward to handle?

I had the same thought 😄

santialbo:
Rebased to the latest develop

with-heart (approved):

Issue #1820, opened 2020-08-26
[Menu] Using display: none to hide Menu breaks custom exit animations


Sebastiansc:
**Describe the bug** A clear and concise description of what the bug is.
* In https://github.com/chakra-ui/chakra-ui/blob/develop/packages/menu/src/use-menu.ts#L275 we are using `display: none` to hide the `MenuList` when it's closed. This doesn't allow for custom exit animations (animations can't run on a display none element) and is also not a performant option for toggling an element's visibility.

**Expected Behavior** A clear and concise description of what you expected to
happen.
* Menu uses properties for hiding that allow the use of custom exit animations

**Link to minimal reproduction** Please provide a codesandbox link or GitHub
repo with a minimal reproduction of the issue.

- JavaScript CodeSandbox template:
  https://codesandbox.io/s/github/chakra-ui/codesandbox-react-js-template/tree/master
- TypeScript CodeSandbox template:
  https://codesandbox.io/s/github/chakra-ui/codesandbox-react-ts-template/tree/master

**Steps to Reproduce** Clear and concise reproduction instructions are important
for us to be able to triage your issue in a timely manner.

**Suggested solution(s)**
* Use opacity and pointer events to hide menu. 

**Desktop (please complete the following information):**

**Additional context** Add any other context about the problem here, or some
background information of how you ran into this bug.

segunadebayo:
We fixed this when migrated most components to use `framer-motion` for their transitions.

PR #1821, opened 2020-08-26
fix(menu): hide menu using opacity and pointerEvents

Sebastiansc:
Description: https://github.com/chakra-ui/chakra-ui/issues/1820

## With custom React spring animation
https://www.loom.com/share/15e6a690213f41f2b3f4c19deb51a0ec

## Without animation
https://www.loom.com/share/c7d1ff2679cf4909a54915dc62b5515a

* As a side note, I think this a code smell that there's too much coupling between `MenuTransition` and the `Menu`. After these changes `MenuTransition` can probably be simplified

### DIFF
--- a/packages/menu/src/use-menu.ts
+++ b/packages/menu/src/use-menu.ts
@@ -272,7 +272,9 @@ export function useMenuList(props: UseMenuListProps) {
     tabIndex: -1,
     role: "menu",
     id: menuId,
-    hidden: !isOpen,
+    ...(isOpen
+      ? { opacity: 1, pointerEvents: "auto" }
+      : { opacity: 0, pointerEvents: "none" }),
     "aria-orientation": "vertical" as React.AriaAttributes["aria-orientation"],
     "data-placement": placement,
     style: { ...popper.style, ...props.style },
@@ -376,6 +378,7 @@ export function useMenuItem(props: UseMenuItemProps) {
     focusedIndex,
     closeOnSelect,
     onClose,
+    isOpen,
     menuRef,
   } = menu
 
@@ -431,7 +434,7 @@ export function useMenuItem(props: UseMenuItemProps) {
         focus(ref.current)
       }
     } else {
-      if (document.activeElement !== menuRef.current) {
+      if (document.activeElement !== menuRef.current && isOpen) {
         menuRef.current?.focus()
       }
     }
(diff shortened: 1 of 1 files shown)


packages/menu/src/use-menu.ts
```diff
@@ -431,7 +434,7 @@ export function useMenuItem(props: UseMenuItemProps) {
         focus(ref.current)
       }
     } else {
-      if (document.activeElement !== menuRef.current) {
+      if (document.activeElement !== menuRef.current && isOpen) {
```
Sebastiansc, line 437: since the menu is no longer being unmounted from the page we have to check if `isOpen` to avoid focusing on a not visible `MenuList`

santialbo:
Hiding things with opacity is not very accessible as some screenreaders will still read the contents and inner elements keep the taborder.
There has to be some other way of adding transitions, maybe using the new provided transitions.

Sebastiansc:
@santialbo we could add aria-hidden true to the menu when closed. Inner elements already have a tabIndex of -1 when closed so that shouldn't be a concern

Sebastiansc:
I've tested this with voiceover and content is not accessible or readable when the menu is closed

segunadebayo:
Hi @Sebastiansc,

I totally the point of this PR and it does make sense for some use-case. However, I'm a bit hesitant to include things like `opacity` and `pointerEvents` to the core `useMenu` hook as it makes the API quite rigid and we're not able to guarantee it's accessibility.

I'm also not sure how this would play with other animation libraries like framer-motion. I believe the ideal solution for you would be to consume the `useMenu` hook directly and customize its return value to match your needs.

I'll give this some more thought.

Sebastiansc:
I can customize the return value of `useMenuList` and that solves most issues but this line remains https://github.com/chakra-ui/chakra-ui/pull/1821/files#diff-e3d406848580454f6bcee5d0277a3948R438 and the way the hook is structured this isn't customizable. I would have to rewrite the entire hook with all of it's internals. 

Out of curiosity I'd like to know why `opacity` and `pointerEvents` makes the API rigid, if you wouldn't mind explaining. Thank you!

with-heart:
Sorry that this has been languishing. I agree with your feedback @Sebastiansc that the way `useMenuList` is currently set up, it wouldn't allow the flexibility needed to implement this yourself.

We've recently been discussing internally the possibility of moving to Framer Motion for our component animations. Segun opened this twitter poll to make the conversation public: https://twitter.com/thesegunadebayo/status/1307939057352278016

I'm not sure how (or if) that would play into this PR, or how we can move forward with resolving #1821. @segunadebayo thoughts?

segunadebayo:
Hi @Sebastiansc,

- We're looking to integrate `framer-motion` which would help us achieve better accessibility with transition/animation enabled.

- We're now using `visibility: hidden` to show and hide any popper related element. I believe this might help us achieve the balance we need.

We'll address this in a new PR that integrates `framer-motion`.

Thanks

PR #1842, opened 2020-08-28
fix(pr-template): remove duplicate content

ljosberinn:
Currently, the PR template has duplicate content which was removed here.

see https://github.com/chakra-ui/chakra-ui/blob/develop/.github/pull_request_template.md


with-heart:
Whoops 😬

PR #1849, opened 2020-08-28
fix: Update theme export in core package

trevorblades:
The published version of `@chakra-ui/theme` [only has a default export](https://unpkg.com/@chakra-ui/theme@1.0.0-rc.2/dist/esm/index.js), so if we want to export `theme` from `@chakra-ui/core` like this line seems to be doing, it'll need to be updated to export the default export of `@chakra-ui/theme`.

Strangely, the source for this module [specifies a named `theme` export as well](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/theme/src/index.ts#L5), so `export * from '@chakra-ui/theme'` _should_ work. This makes me think there might be something up with the compilation of `theme/src/index.ts` that is messing up the named export.


with-heart:
Thanks @trevorblades!

PR #1904, opened 2020-08-31
chore(dependabot): remove configuration to disable automatic updates

with-heart:
This PR removes the `dependabot` configuration in order to disable version update PRs. These are no longer needed due to #1865.

As part of this change, I think we should enable automatic `dependabot` security updates. The steps to enable this can be done here: https://docs.github.com/en/github/managing-security-vulnerabilities/configuring-github-dependabot-security-updates#managing-github-dependabot-security-updates-for-your-repositories


with-heart:
Speaking of disabling updates...

The Vercel check timed out because we had a backlog of version update PRs causing a very long deployment queue.

PR #1938, opened 2020-09-03
Set width and height for heroes

joe-bell:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

On safari there's a rather amusing bug that makes "Chakra Heroes" look very
L
O
N
G

Screenshot 2020-09-03 at 08 55 55

Issue Number: N/A

## What is the new behavior?

Image sizes are fixed to specific width and height

Screenshot 2020-09-03 at 08 55 59

- [ ] Yes
- [x] No

### DIFF
--- a/website/pages/index.tsx
+++ b/website/pages/index.tsx
@@ -355,7 +355,8 @@ const HomePage = ({ members, sponsors }) => {
                 {members.map((i) => (
                   <Img
                     key={i.login}
-                    htmlWidth="80px"
+                    width="80px"
+                    height="80px"
                     rounded="full"
                     alt={i.name}
                     src={i.avatar_url}


joe-bell:
Although this is just a quick fix, I think it's perhaps worth revaluating if `inline-flex` is the right option for displaying items inside `Wrap` (switching this to `inline-block` also solved the issue)

Perhaps it's best to avoid assuming `display` styles inside `Wrap` (keeping it focussed to solving the gap + layout) and instead leaving these choices down to the consumer for more flexibility and control. This is a decision I made for [`raam`'s `Wrap`](https://raam.joebell.co.uk/#wrap)

I understand we're probably keen to avoid adding more elements, but I'm not sure it's a good enough reason to assume consumers want `inline-flex`

What do we think?

segunadebayo:
@joe-bell, I think I agree with you here. I'll switch it to inline-flex.

joe-bell:
@segunadebayo nice one!

PR #1939, opened 2020-09-03
revert color mode to rc.2

### LINKED ISSUE #1878: [SSR] Color mode flash
It looks like the changes we made to fix color mode in `1.0.0-rc.3` worked for static generation, but also broke SSR. 

See https://personal-react-boilerplate.now.sh/ for an example.

### LINKED ISSUE #1880: [Color Mode] The colour mode is incorrect on load
**Describe the bug** A clear and concise description of what the bug is.

On load, the colour mode is incorrect, doesn't respect `prefers-color-scheme` and doesn't use the value from local storage.

image

**Expected Behavior** A clear and concise description of what you expected to
happen.

The colour mode from local storage should be used, or the media query result should be respected on load.

**Link to minimal reproduction** Please provide a codesandbox link or GitHub
repo with a minimal reproduction of the issue.

This application is experiencing the issue, it is an electron app.

https://github.com/rshackleton/twitch-control-centre/tree/19a3dbb24e48d43ba0376d79efefaf42034f8473

**Steps to Reproduce** Clear and concise reproduction instructions are important
for us to be able to triage your issue in a timely manner.

1. Run the application.
2. When the main window has loaded the colour scheme doesn't respect the OS theme or use the local storage value.

- OS: Windows
- Browser: Chrome (in Electron)
- Version: 83

**Additional context** Add any other context about the problem here, or some
background information of how you ran into this bug.

### LINKED ISSUE #1976: Unable to use system "prefers-color-scheme" for colorMode Chakra latest
# Bug report

Unable to use default system `prefers-color-scheme` when setting up colorMode with latest version of Chakra

## Describe the bug

Related to Chakra@next: "1.0.0-rc.3"

For NextJS to enable ColorModeScript, the code below which can also be found in the docs is wrong in 2 ways:
`<ColorModeScript defaultMode="light" />`

First: the prop should be "defaultColorMode" as defaultMode doesn't exists
Second: to use system's `prefers-color-scheme` it's stated to use "system" as `defaultColorMode` value but defaultColorMode allows only 2 values defined in `colorMode` which are locked to `"light" | "dark"` 

## To reproduce

refer to: https://next.chakra-ui.com/docs/features/color-mode

ljosberinn:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- color mode persistence is broken for CSR apps
- SSR & SSG apps see a white flash which rc.3 was supposed to prevent
- electron apps behave like CSR apps, always light mode

## What is the new behavior?

- `storageManager` will be used again
- `theme.config.useSystemColorMode` and `theme.config.initialColorMode` will be available again
- SSG apps will still see a white flash, this is not intended, but tolerated for now (see [comment](https://github.com/chakra-ui/chakra-ui/issues/1878#issuecomment-683708742))

- [x] Yes
- [ ] No

fixes #1880
fixes #1878
fixes #1976

## Other information

- we could add `withServerSideProps` and `getServerSideProps` as helpers for nextjs to be exported from the color mode package. I've already written the code [here](https://github.com/ljosberinn/personal-react-boilerplate/blob/master/karma/client/Karma.tsx#L108) for even more functionality, but as you can see by the typings alone it introduces some complexity (although it should be stable since its building on top of next types)
- alternatively, I'll add an example to both nextjs folderss in `/examples` which is what I've gone ahead with
- I can't build the `nextjs-typescript` example locally due to lots of typing issues. no clue why. anyone an idea? When I copy my tsconfig over into the example nothing changes, but if I copy the erroring components into my project, they work just fine.
- improved a few things here and there while I was at it:
  - remove no longer needed `_document.js` in `examples/next-js`
  - bumped versions for both nextjs examples and the docs
  - added types for `_app.tsx` in `examples/nextjs-typescript`
  - improved types for `useColorModeValue` which now correctly infers instead of returning `any`
  - refactored `localStorageManager` and `cookieStorageManager` a bit
  - refactored `useColorModeState` a bit and added comments
  - instead of extending `ThemeProviderProps` in `chakra-provider`, I inlined the props for clarity and the possibility to annotate them
  - removed async/await + unecessary spreading in `website/pages/_document/getInitialProps`
  - return noop in addListener to prevent using ?.() in useSyncSystemColorMode
  - reduce useColorModeState complexity
  - change signature of useColorModeState to object (its only used once and isnt exported in general)
  - return toggle instead of setter in useColorModeState


ljosberinn:
should be fine to review now - tested as far as I'm able to locally

ljosberinn:
updates:
- test for `cookieStorageManager`
- rename: `storageManager` to `colorModeCookie` - omit the cookie to use `localStorageManager`, pass it to use `cookieStorageManager`
- updated docs accordingly
- re-added ColorModeScript - I'm confused here, do we really need it? all it does is setting a class name that we don't actively use anywhere, except that we keep it in sync. [commit ref from history](https://github.com/chakra-ui/chakra-ui/commit/a5d638eda4ccf368fddade72362aaad592086187)
- set `resetCSS` default to `true` & removed it wherever we use it & update docs accordingly

regarding the docs:
- `/`, `/team` use cookie
- all other urls use localStorage because I couldnt get [usage of `getServerSideProps` to work](https://github.com/hashicorp/next-mdx-enhanced#reexportdatafetching)

ljosberinn:
once this portal test finally works out which it should now after rebasing this should be fine to merge

examples/next-js/src/Chakra.js
```diff
@@ -0,0 +1,27 @@
...
+  localStorageManager,
+} from "@chakra-ui/core"
+
+export const Chakra = ({ children, cookies }) => {
+  return (
+    <ChakraProvider
+      colorModeManager={
+        typeof cookies === "string"
+          ? cookieStorageManager(cookies)
+          : localStorageManager
```
ljosberinn, line 13: ensures that given cookies, will assume you're rendering server side. could also do `... ? cookieStorageManager(cookies) : undefined` and use the implicit fallback to `localStorageManager` but figured this is more explicit

examples/nextjs-typescript/tsconfig.json
```diff
@@ -1,12 +1,11 @@
 {
   "compilerOptions": {
     "allowJs": true,
-    "alwaysStrict": true,
```
ljosberinn, line 4: already covered by `strict: true`

packages/color-mode/src/color-mode-script.tsx
```diff
@@ -1,50 +1,17 @@
...
-  if (colorMode) {
-    const root = document.documentElement
-    root.style.setProperty("--chakra-ui-color-mode", colorMode)
-  }
-}
-
-interface ColorModeScriptProps {
-  defaultColorMode?: ColorMode
-}
+const noFlash = `(function() { try {
```
ljosberinn, line 3: do we still need this? In the old commit from `rc.2` I found it used in the gatsby plugin - but since we never check for any `className` and only sync ourselves, I wonder what its purpose was?

packages/portal/tests/portal.test.tsx
```diff
@@ -9,7 +9,7 @@ test("should render portal", () => {
     </PortalManager>,
   )
 
-  expect(tools.baseElement).toMatchSnapshot()
+  expect(tools.baseElement.innerHTML).toMatchSnapshot()
```
ljosberinn, line 12: idk whats up with that test. locally it doesnt work, then I fix it, push, and CI wants the initial version. then I reverse, and CI still complains.

website/src/components/chakra.tsx
```diff
@@ -0,0 +1,53 @@
...
+} from "@chakra-ui/core"
+import { GetServerSidePropsContext } from "next"
+import theme from "theme"
+
+type WithCookies = {
+  cookies?: string
+}
+
+/**
+ * higher order component for pages
```
ljosberinn, line 14: I didn't include this in the examples because it's imo a special case for the docs:
when you `useColorModeValue` in the top level of a component _inline_, it will break:

```js
function MyComponent() {
  return <ChakraProvider><Box color={useColorModeValue('teal.200', 'teal.500') /></ChakraProvider>
}
```

This is done heavily in the docs, so you have to either:
- use a hoc injecting the provider around it
- refactor the docs into several subcomponents

PR #1967, opened 2020-09-05
feat(extend-theme): merge function results

### LINKED ISSUE #1966: Use `lodash.mergeWith` to enable `extendTheme` to merge functions
# Feature request

## Is your feature request related to a problem? Please describe.

As it stands, `extendTheme` does not provide us with the full functionality needed to extend the theme. This is due to the fact that many parts of the default theme are defined as functions. In order make this work, a user needs to wrap their overrides with a function referencing the internal theme function, which makes `extendTheme` a less than ideal solution for many situations.

## Describe the solution you'd like

@TimKolberger provided this example implementation: https://codesandbox.io/s/chakra-v1-extend-theme-with-functions-30435?file=/src/index.tsx

This example extends `theme.components.Button.variants.solid` with a function. `theme.components.Button.variants.solid` is defined as a function [in the default Chakra theme](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/theme/src/components/button.ts#L78) as well.

Using `lodash.mergeWith` allows this to operate as expected, where the default theme function is evaluated with props, the extension is evaluated with props, and then the evaluation of extension is merged into the result of the default evaluation.

TimKolberger:
## Pull request checklist

      /start features)

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

> As it stands, extendTheme does not provide us with the full functionality needed to extend the theme. This is due to the fact that many parts of the default theme are defined as functions. In order make this work, a user needs to wrap their overrides with a function referencing the internal theme function, which makes extendTheme a less than ideal solution for many situations.
> - @with-heart 

Fix #1966 

## What is the new behavior?

The return value of functions will be merged as well.
Now you can extend the default styles of variants.

Disadvantage: to override an existing variant you have to override every property of it.

```ts
const override = {
  components: {
    Button: {
      variants: {
        solid: () => ({
          bg: "papayawhip",
        }),
      },
    },
  },
}

const customTheme = extendTheme(override)
```

### Change description

- add dependency to `lodash.mergewith` to `utils` package
- extract `extendTheme` in new file
- use merge customizer to merge function call results as well (e.g. component variants)

- [ ] Yes
- [x] No

`lodash.merge` is still exported from `utils`. Should we drop the dependency to `lodash.merge`?  
It cannot be removed due usage in `styled-system`.

### DIFF
--- a/packages/core/jest.config.js
+++ b/packages/core/jest.config.js
@@ -1,0 +1,5 @@
+const baseConfig = require("../../jest.config")
+
+module.exports = {
+  ...baseConfig,
+}
--- a/packages/core/src/chakra-provider.tsx
+++ b/packages/core/src/chakra-provider.tsx
@@ -6,8 +6,7 @@ import {
   ThemeProvider,
   GlobalStyle,
 } from "@chakra-ui/system"
-import defaultTheme, { Theme } from "@chakra-ui/theme"
-import { merge } from "@chakra-ui/utils"
+import defaultTheme from "@chakra-ui/theme"
 import * as React from "react"
 
 export interface ChakraProviderProps extends Partial<ThemeProviderProps> {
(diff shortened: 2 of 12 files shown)


segunadebayo:
This makes sense @TimKolberger. Thank you for working on this!

PR #1980, opened 2020-09-07
docs: document usage of close/closeAll instance methods

mattrothenberg:
## Pull request type
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Docs do not explain the usage of the `close` and `closeAll` instance methods on the `toast` object.

@ljosberinn I might be missing something here, but I don't think the `closeAll` method is working as I thought it should be in the example story. Any ideas?

#63 

- [ ] Yes
- [x] No

### DIFF
--- a/website/pages/docs/feedback/toast.mdx
+++ b/website/pages/docs/feedback/toast.mdx
@@ -79,6 +79,48 @@ function Example() {
 }
 ```
 
+## Closing Toasts
+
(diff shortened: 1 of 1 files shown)


with-heart (approved):
Nice work @mattrothenberg!

@allcontributors please add @mattrothenberg for docs

PR #2027, opened 2020-09-14
Feat/improve breakpoints dx

### LINKED ISSUE #1954: [Theme] Provide a function to create the breakpoints property via DI
# Bug report

Right now, to use custom breakpoints, you can just pass `extendTheme({ breakpoints: myBreakpointArray })` [matching the spec](https://next.chakra-ui.com/docs/features/responsive-styles#what-it-does).

However, if you manually pass, [you'd be missing the properties defined on the default theme array](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/theme/src/foundations/breakpoints.ts#L10).

We could provide a function or merge functionality for that into `extendTheme`, something like:

```ts
interface BreakpointConfig {
  [key: string]: string;
  sm: string;
  md: string;
  lg: string;
  xl: string;
}

const createBreakpoints = (config: BreakpointConfig) => {
 const breakpoints = Object.values(config);

 Object.entries(config).forEach(([key, value]) => {
  Object.defineProperty(breakpoints, key, { value });
 });

  return breakpoints;
}
```

ljosberinn:
Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

 Previously, breakpoints on a theme had to be an array with your breakpoints in ascending order which made it hard to reference which value was correlating to md etc. On top, you had to manually manipulate the array by defining properties on it, which was often overlooked and not typesafe.

Issue Number: fixes #1954

To provide easier extensibility and typesafety, breakpoints now need to be an object.

- [x] Yes
- [ ] No

As described in the new docs and above, when customizing `theme.breakpoints`, you're now supposed to use `createBreakpoints` and pass an object.

- fixed a bug in `useBreakpointValue` which wasn't working as intended given an array
- increased performance of `responsive` function used within `css` by factor 3-6x
- added support for custom breakpoints in general

### DIFF
--- a/packages/media-query/src/use-breakpoint-value.ts
+++ b/packages/media-query/src/use-breakpoint-value.ts
@@ -1,6 +1,7 @@
 import { getClosestValue } from "./media-query.utils"
 import { useBreakpoint } from "./use-breakpoint"
 import { isArray, arrayToObjectNotation } from "@chakra-ui/utils"
+import { useTheme } from "@chakra-ui/system"
 
 /**
  * React hook for getting the value for the current breakpoint from the
@@ -11,7 +12,21 @@ import { isArray, arrayToObjectNotation } from "@chakra-ui/utils"
  */
 export function useBreakpointValue<T = any>(values: Record<string, T> | T[]) {
   const breakpoint = useBreakpoint()
-  if (!breakpoint) return
-  const obj = isArray(values) ? arrayToObjectNotation(values) : values
-  return getClosestValue(obj, breakpoint)
+  const { breakpoints } = useTheme()
+
+  if (!breakpoint) {
+    return
+  }
+
+  const obj = isArray(values)
+    ? Object.fromEntries(
+        Object.entries(
+          arrayToObjectNotation(values, breakpoints),
+        ).map(([_, value]) => [value, value]),
+      )
+    : values
+
+  const closest = getClosestValue(obj, breakpoint)
+
+  return closest
 }
--- a/packages/media-query/src/use-breakpoint.ts
+++ b/packages/media-query/src/use-breakpoint.ts
@@ -69,7 +69,7 @@ export function useBreakpoint(defaultBreakpoint?: string) {
       }
 
       // add media query-listender
-      mediaQuery.addListener(handleChange)
+      mediaQuery.addEventListener("change", handleChange)
 
       // push the media query list handleChange
       // so we can use it to remove Listener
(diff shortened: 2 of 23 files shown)


website/pages/docs/migration.mdx
```diff
@@ -121,6 +121,45 @@ only one prop to manage this, you can rename it to `boxSize`.
...
+
+**You may safely skip this if you didn't customize your breakpoints.**
+
+To provide easier extensibility and typesafety, breakpoints should now be
+created through `createBreakpoints` by passing an object:
+
+```diff
++ import { createBreakpoints, extendTheme } from "@chakra-ui/core"
+
+
```
ljosberinn, line 134: extra empty lines here are necessary due to the way `diff` works

packages/media-query/src/use-breakpoint.ts
```diff
@@ -69,26 +69,26 @@ export function useBreakpoint(defaultBreakpoint?: string) {
       }
 
       // add media query-listender
-      mediaQuery.addListener(handleChange)
+      mediaQuery.addEventListener("change", handleChange)
```
ljosberinn, line 72: `.add` and `.removeListener` are deprecated APIs

packages/media-query/src/use-breakpoint-value.ts
```diff
@@ -11,7 +12,21 @@ import { isArray, arrayToObjectNotation } from "@chakra-ui/utils"
...
-  const obj = isArray(values) ? arrayToObjectNotation(values) : values
-  return getClosestValue(obj, breakpoint)
+  const { breakpoints } = useTheme()
+
+  if (!breakpoint) {
+    return
+  }
+
+  const obj = isArray(values)
+    ? Object.fromEntries(
```
ljosberinn, line 22: necessary to make the resulting object of `arrayToObjectNotation` compatible to `getClosestValue`

packages/styled-system/src/css.ts
```diff
@@ -7,40 +7,100 @@ import {
...
+}
+
+/**
+ *
+ */
+const calculateBreakpointAndMediaQueries = (
+  themeBreakpoints: string[] = [],
+) => {
+  // caching here reduces execution time by factor 4-6x
+  const isCached = cache.themeBreakpoints === themeBreakpoints
```
ljosberinn, line 48: we might be able to remove caching here alltogether IF we never have a `ThemeProvider` with a different `theme` and thus potentially different `breakpoints`. Figured I'd leave it in as the original code was checking whether `theme.breakpoints` was defined at all. Unsure if this was due to a certain test scenario that I had to fix because of that, or because of some use case. @segunadebayo

packages/styled-system/src/css.ts
```diff
@@ -98,14 +164,14 @@ export const css = (args: StyleObjectOrFn = {}) => (
     const value = config?.transform?.(val, scale) ?? get(scale, val, val)
 
     if (config?.properties) {
-      for (const property of config?.properties) {
+      for (const property of config.properties) {
```
ljosberinn, line 167: as below, the `if` statement already ensures this is truthy - another case ESLint would've caught which is why after this PR and the ColorMode PR have been merged, I'd like to implement proper linting

packages/theme-tools/tests/breakpoints.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    const result = createBreakpoints(extendedBreakpoints)
+
+    expect(result).toEqual(expect.arrayContaining(expectedResult))
+
+    Object.entries(extendedBreakpoints).forEach(([key, value]) => {
+      expect(result[key]).toBe(value)
+    })
+  })
+
+  test("works with px", () => {
```
ljosberinn, line 56: just to ensure sorting works as expected across commonly used units

PR #2062, opened 2020-09-17
improve useColorModeValue types

PabloSzx:
Very simple but useful improvement, previously it was just returning the type "any" always, now it uses generic types.


ljosberinn:
Hi @PabloSzx - thank you! I'm afraid though this is already included in the larger PR over [here](https://github.com/chakra-ui/chakra-ui/pull/1939/files).

Issue #2121, opened 2020-09-29
[Toast] Close all toasts doesn't work


la55u:
## Describe the bug

Close all toasts button doesn't work on the docs page, it only removes one (the most recent) toast from the bottom.

## To reproduce

Try it here:
https://next.chakra-ui.com/docs/feedback/toast#closing-toasts

## Expected behavior

All toasts shuld be removed immediately.

## System information

- Version of @chakra-ui/core: 1.0.0-rc.x(?)

ljosberinn:
Thanks, can confirm this is a bug and found a fix :)

ljosberinn:
https://chakra-ui-1pr7qefki.vercel.app/docs/feedback/toast#closing-toasts works after fixing it

EllaSharakanski:
Also, `toast.close(id)` doesn't work if performed one-by-one for an array of toasts all in the same position. The fix solves both issues. When is it supposed to be released? Thanks

ljosberinn:
Automatically with the next release. Migrating to framer-motion is the last missing piece for another release as far as I'm aware, so probably next week.

EllaSharakanski:
Thank you! Sorry for asking here, not sure where I can ask, is there a way for me to see this information? Couldn't find the place that specifies the missing pieces for a release

ljosberinn:
Indirectly only I'm afraid. I thought we mentioned it in an issue but I was mistaken; we mentioned it on discord and there is this twitter poll: https://twitter.com/thesegunadebayo/status/1307939057352278016

EllaSharakanski:
@ljosberinn Any update on when this is going to be released? Thank you 🙏

ljosberinn:
maybe @segunadebayo can shed some light :)

juanlatorre:
Any update on this? Toast duration and close buttons aren't working either

ljosberinn:
It's been fixed technically, but I made a silly mistake in the closing logic - which also has been fixed, we're planning to release another rc soon (today or tomorrow depending on progress). See #2328

PR #2132, opened 2020-09-30
fix(toast): closeAll method actually closes all

ljosberinn:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Fixes:  #2121

## What is the new behavior?

- `closeAll` actually closes all toasts
- better performance for a few functions as well as normalization

- [ ] Yes
- [x] No

### DIFF
--- a/packages/toast/src/toast-manager.tsx
+++ b/packages/toast/src/toast-manager.tsx
@@ -76,6 +76,8 @@ export class ToastManager extends React.Component<Props, State> {
     const { position, id } = toast
 
     this.setState((prevToasts) => {
+      const isTop = position.includes("top")
+
       /**
        * - If the toast is positioned at the top edges, the
        * recent toast stacks on top of the other toasts.
@@ -83,13 +85,13 @@ export class ToastManager extends React.Component<Props, State> {
        * - If the toast is positioned at the bottom edges, the recent
        * toast stacks below the other toasts.
        */
-      const isTop = position.includes("top")
+      const toasts = isTop
+        ? [toast, ...prevToasts[position]]
+        : [...prevToasts[position], toast]
 
       return {
         ...prevToasts,
-        [position]: isTop
-          ? [toast, ...prevToasts[position]]
-          : [...prevToasts[position], toast],
+        [position]: toasts,
       }
     })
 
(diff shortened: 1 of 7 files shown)


packages/toast/src/toast-manager.tsx
```diff
@@ -120,10 +122,26 @@ export class ToastManager extends React.Component<Props, State> {
    * Close all toasts at once
    */
   closeAll = () => {
-    objectKeys(this.state).forEach((position) => {
-      this.state[position].forEach((toast) => {
-        this.closeToast(toast.id)
```
ljosberinn, line 125: spams `this.setState` for every opened toast, new function just sets once

packages/toast/src/toast-manager.tsx
```diff
@@ -143,30 +161,33 @@ export class ToastManager extends React.Component<Props, State> {
...
+      if (!position) {
+        return prevState
+      }
 
       return {
         ...prevState,
         [position]: prevState[position].map((toast) => ({
           ...toast,
-          requestClose: toast.id == id,
+          requestClose: toast.id == id ? true : toast.requestClose,
```
ljosberinn, line 183: this was leading to the actual bug reported: per position, only one toast at max was closed at a time when using `closeAll` because it was always overwriting the previous value

packages/toast/src/toast.class.tsx
```diff
@@ -49,15 +49,19 @@ class Toaster {
   }
 
   notify = (message: ToastMessage, options: Partial<ToastOptions> = {}) => {
-    return this.createToast?.(message, options)
+    return this.createToast ? this.createToast(message, options) : undefined
```
ljosberinn, line 52: refactored optional chaining in favor of if/ternaries for a small bundlesize gain

packages/toast/src/toast.utils.ts
```diff
@@ -30,47 +29,29 @@ export function findToast(toasts: ToastState, id: ToastId) {
  * Given the toast manager state, finds the position of the toast that
  * matches the `id`
  */
-export function getToastPosition(toasts: ToastState, id: ToastId) {
-  let position: ToastPosition | undefined
-
-  objectKeys(toasts).forEach((pos) => {
-    const found = findById(toasts[pos], id)
-    if (found) position = pos
```
ljosberinn, line 38: didn't bail instantly so I replaced it with `.find`

packages/toast/src/toast.class.tsx
Zaynex: Is it worth that refactor optional chaining ?It keeps code clean. Maybe the ts compiler will imporve the bundlesize in the feature?

ljosberinn: while writing this, turns out I measured the wrong data and forget the `options = {}` assignment when comparing... its not a reduction of 28% (which I'd say is a free, although micro, optimization) but only 3.4% :/ so I'll likely revert this

thanks!

ljosberinn: gone :]

PR #2139, opened 2020-10-01
Fix/use media query

ljosberinn:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

- `useMediaQuery` always returns a tuple: `[boolean, setter]` - which makes no sense for what it supposedly does: just checking whether a query matches. You'd never want to use the setter, and, as it turns out - this functionality was unused.

- `useAnimationPreference` sounds like you receive a string or something non-booleanish, when it actually returns a boolean, being `true` if the user allows animations, `false` if the animation prefers reduced animations
  - I personally found that very confusing, and still do, even after looking at the code for a while, although its implementation is very short. Maybe just me.

## What is the new behavior?

- `useMediaQuery` now accepts both a single media query as well as an array of media queries. The API surface as such is not a breaking change, as you previously had to destructure the single boolean from the array anyways. Now you can actually use multiple! 
- renamed `useAnimationPreference` to `usePrefersReducedMotion` - this is a breaking change for `v1` only, as the hook doesn't exist in `v0`, so we wouldn't have to document it in the docs as such.
- refactored implementation of `usePrefersReducedMotion` and `useColorModePreference` to adapt to the new `useMediaQuery` possibilities
- documented both hooks in the docs since they were both undocumented before
- adjusted examples & simplified logic there (e.g. removed use of `css` prop for a more canonical way).

- [x] Yes, `next` only.
- [ ] No

Would love to write tests but since `matchMedia` doesn't exist in JSDOM, you'd just be testing your mock implementation :(

### DIFF
--- a/examples/create-react-app/src/Logo.tsx
+++ b/examples/create-react-app/src/Logo.tsx
@@ -1,5 +1,10 @@
 import React from "react"
-import { Image, ImageProps, keyframes } from "@chakra-ui/core"
+import {
+  Image,
+  ImageProps,
+  keyframes,
+  usePrefersReducedMotion,
+} from "@chakra-ui/core"
 import logo from "./logo.svg"
 
 const spin = keyframes`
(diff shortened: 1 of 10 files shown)


with-heart (approved):
Super quality PR here! Thanks for taking the time to add docs for both hooks and updating the templates too!

PR #2162, opened 2020-10-04
refactor: use react-popper

segunadebayo:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

To reduce our maintenance overhead, we'll be using react-popper instead of writing our own and re-inventing the wheel. This PR update the popper package and adds more interesting options like `matchWidth` to the base `usePopper` hook coming from `react-popper`

- [ ] Yes
- [x] No


packages/popper/src/use-popper.ts
```diff
@@ -1,170 +1,155 @@
...
+      {
+        name: "preventOverflow",
+        enabled: !!preventOverflow,
+      },
+      {
+        name: "matchWidth",
+        enabled: !!matchWidth,
+        phase: "beforeWrite",
+        requires: ["computeStyles"],
+        fn: useCallback(({ state }) => {
```
ljosberinn, line 71: this useCallback as well as the one below should most likely be declared outside of `useBasePopper`

considering the generally huge object, would it makse sense to do something like this:
```js
const popperJS = useBasePopper(referenceNode, popperNode, getConfig({
  placement,
  fixed,
  offset,
  gutter,
  arrowNode,
  preventOverflow,
  /* etc */
})
```
then we could test that function separately, too, because it oculd live outside of the hook?

segunadebayo: Yes, I thought about this as well. I was indifferent about it considering we might not be testing that function in isolation.

with-heart (approved):

Issue #2184, opened 2020-10-06
[Media-Query] useBreakpointValue hook is broken when provided with an array.


candicecz:
# Bug report

## Describe the bug

The `useBreakpointValue` hook for media queries does not work when an array is provided as an argument.

## To reproduce

1. Go to Storybook
2. Click on `breakpoints` then select `Breakpoint Value Hook` or append this path to the storybook url: `?path=/story/breakpoints--breakpoint-value-hook`
3. As you resize the page, you'll notice the width changes "...px" but the background color is not changing.

## Minimal reproduction

I was not able to reproduce the bug in CodeSandbox but the code that fails to work is available in a storybook example (as explained above). The code for that example is found here: `packages/media-query/stories/media-query.stories.tsx`.
```
// line 43
export const BreakpointValueHook = () => {
  const width = useBreakpointValue({ base: "150px", md: "250px" })
  const color = useBreakpointValue(["red.500", null, "green.500"])
  return (
    <chakra.div bg={color} mx="auto" width={width}>
      I'm {width} wide
    </chakra.div>
  )
}
```
## Expected behavior

We expect for the background to be either red or green on window resize but there is no background at all. The width property works as expected.

## Screenshots

Problem:
![Screenshot from 2020-10-06 17-55-40](https://user-images.githubusercontent.com/25373313/95266344-fe09a800-0800-11eb-94ef-95108667b8ce.png)
Expected:
![Screenshot from 2020-10-06 18-02-34](https://user-images.githubusercontent.com/25373313/95266359-05c94c80-0801-11eb-8625-3665b3fb5db7.png)

## System information

- OS: Ubuntu 20.04
- Browser (if applies): Chrome
- Version of @chakra-ui/core: [e.g. 1.0.0-rc.3]
- Version of Node.js: v.10.19.0

PR #2185, opened 2020-10-06
fix(use-breakpoint-value): fix hook when given array of values

candicecz:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The useBreakpointValue hook does not work when provided with an array.

Fixes: #2184 

## What is the new behavior?

- The hook now works again when provided with an array.
- The storybook example as described in the issue works.
- The hook also works when given a custom breakpoint.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/media-query/src/use-breakpoint-value.ts
+++ b/packages/media-query/src/use-breakpoint-value.ts
@@ -2,6 +2,7 @@ import { getClosestValue } from "./media-query.utils"
 import { useBreakpoint } from "./use-breakpoint"
 import { isArray, arrayToObjectNotation, fromEntries } from "@chakra-ui/utils"
 import { useTheme } from "@chakra-ui/system"
+import { isCustomBreakpoint } from "@chakra-ui/utils"
 
 /**
  * React hook for getting the value for the current breakpoint from the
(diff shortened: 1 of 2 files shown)


ljosberinn:
Hm that's odd. Why were the tests passing before then 🤔 https://github.com/chakra-ui/chakra-ui/blob/develop/packages/media-query/tests/use-breakpoint-value.test.tsx#L107

candicecz:
So I think I can answer that.
The test "with-array" uses the following array as the test values.
 ```
const values = ["base", "sm", "md", "lg", "xl", "customBreakpoint"]
```

The original useBreakpointValue looks like this:
```
  const obj = isArray(values)
    ? fromEntries<Record<string, T>>(
        Object.entries(
          arrayToObjectNotation(values, breakpoints),
        ).map(([_, value]) => [value, value]),
      )
    : values
```
As you can see it uses the mapped value as the key as well.
This is not a problem with the test because `base`, `sm`, etc are valid keys.
However, in the media query stories the values are:
```
const values = ["red.500", null, "green.500"]
```
and the aforementioned code yields:
```
{red.500:"red.500", green.500: "green.500"}
```
which is not valid.

candicecz:
I could change the test values if you want as well..

ljosberinn:
Would be great - yeah, was about to write that haha. Would be great if we have a test case for that to not happen again, thanks in advance!

candicecz:
I updated test, please check that I updated the values correctly. Thank you!

Issue #2170, opened 2020-10-05
[Skeleton] Typing adjustment


ljosberinn:
# Bug report

When re-adding typing support for `noOfLines` in `system`, I unknowingly broke `Skeleton` typings, which I fixed in [this commit](https://github.com/chakra-ui/chakra-ui/commit/5c62c6902e81d0fd943dcaa5ddf534a2875f0304). But I believe this is a rather bad bandaid.

The issue is as follows:

`SkeletonTextProps` extends `SkeletonProps` here
https://github.com/chakra-ui/chakra-ui/blob/5c62c6902e81d0fd943dcaa5ddf534a2875f0304/packages/skeleton/src/skeleton.tsx#L119
and now tries to overwrite `noOfLines` coming from `system` which is no longer just `number`, but `number[]` and thus cannot be overwritten as `number`.

`Skeleton` doesn't care though, but should.
https://github.com/chakra-ui/chakra-ui/blob/5c62c6902e81d0fd943dcaa5ddf534a2875f0304/packages/skeleton/src/skeleton.tsx#L129
https://github.com/chakra-ui/chakra-ui/blob/5c62c6902e81d0fd943dcaa5ddf534a2875f0304/packages/skeleton/src/skeleton.tsx#L140

The solution probably involves using `theme.breakpoints` to dynamically determine the amount of breakpoints used (default 5, but customizeable), `useBreakpointValue` and adjusting the component further down in some way.

## Expected behavior

`Skeleton` should support `noOfLines` just fine.

Jaskaran121:
i can create a pr for this....can you assign this ticket to me thankx

PR #2203, opened 2020-10-09
fix(skeleton): implemented noOfLines as responsive value

dsumer:
## Pull request checklist

- [X] Docs have been reviewed and added / updated if needed (for bug fixes
      /start features)

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Fixes: #2170 

Currently the Skeleton Component implements its own noOfLines prop instead of using the system-wide one. Therefore it's also a static value and no chakra responsive value.

## What is the new behavior?

- noOfLine for Skeleton.Text Component can now be passed as responsive values (different noOfLine value for different breakpoints)

- [ ] Yes
- [X] No

### DIFF
--- a/packages/skeleton/src/skeleton.tsx
+++ b/packages/skeleton/src/skeleton.tsx
@@ -8,6 +8,7 @@ import {
   omitThemingProps,
 } from "@chakra-ui/system"
 import { cx, __DEV__ } from "@chakra-ui/utils"
+import { useBreakpointValue } from "@chakra-ui/media-query";
 import * as React from "react"
 
 export interface SkeletonOptions {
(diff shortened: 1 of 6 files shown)


dsumer:
Sadly I couldn't find any tests for Skeleton, if desired I can also try to add a test for the new noOfLines prop as responsiveValue. Would be nice if you can point out an example test how a component can be tested with different screen sizes. :)

packages/skeleton/src/skeleton.tsx
```diff
@@ -137,10 +137,11 @@ export const SkeletonText: React.FC<SkeletonTextProps> = (props) => {
     ...rest
   } = props
 
-  const numbers = range(noOfLines)
+  const noOfLinesValue = useBreakpointValue(typeof noOfLines === 'number' ? [noOfLines] : noOfLines) || defaultNoOfLines;
```
ljosberinn, line 140: the `|| defaultNoOfLines` case can't ever happen, can it? if explicitly passed as `undefined`, it would be `defaultNoOfLines` so you're always in a situation of `is number` or `is array`?

ljosberinn:
Thank you, looks good!

Re tests: if you're up for it, tests similar to `useBreakpointValue` [here](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/media-query/tests/use-breakpoint-value.test.tsx#L188) should work. Although I'd assume the tests have to be a bit unconventional by having to check for the wrappers child length.

dsumer:
nice, I've added a test for validating the "responsiveness" of the noOfLines prop

ljosberinn:
@allcontributors please add @dsumer for code

PR #2228, opened 2020-10-13
Feat/use colors

ljosberinn:
## Pull request checklist

      /start features)

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

E.g. `linear-gradient` as css property needs inline knowledge of resolved colors, this is currently not possible with 
```js
<Box background="linear-gradient(teal.100, red.300)" />
```

Until this is resolved, `useColors` allows to resolve the colors without having to use 

```js
bg={`linear-gradient(${theme.colors.teal[100]}, ${theme.colors.red[300]})`}
```
 through a theme import.

## What is the new behavior?

- `useToken` resolves any design tokens safely from the theme
- removed `memoizeOne`
- implemented caching of anything `get` is used with via a `WeakMap`
- fixed failed develop tests

- [ ] Yes
- [x] No


ljosberinn:
Snuck in [`jest-watch-typeahead`](https://github.com/jest-community/jest-watch-typeahead#readme) which lets you filter testfiles by input and with keyup/down:
image

with-heart (approved):

ljosberinn:
Talked about it a bit and settled for a more flexible API instead:
```js
function useToken(key, ...tokens) {
  const theme = useTheme();

  return tokens.map(token => theme[key]?.[token] ?? undefined)
}
```

packages/utils/src/object.ts
```diff
@@ -54,16 +56,31 @@ export function get(
...
+    }
+
+    obj = obj[key[index]]
+  }
+
+  const value = obj === undefined ? fallback : obj
+  map.set(key, value)
+
+  return value
+}
```
segunadebayo, line 83: Just looking through this and I feel like we can split the cache part and the actual `get` function, then compose them together.

We're doing 2 things within a single function, and I'd like us to separate that.

```js
const memoGet = memoize(get)
```

Can we do this?

ljosberinn: should be resolved, readded memoizedGet where it was removed previously

PR #2298, opened 2020-10-23
RFC: Use WrapItem in favour of cloning children in Wrap

joe-bell:
## Pull request checklist

      /start features)

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently, `Wrap` clones all children and "wraps" each child with item styles. Although I may have something to do with influencing this in the initial version of [`raam`](raam.joebell.co.uk/), I no longer believe this is the right way forward.

I recently hit a snag in a custom component which involved adding/removing items to `Wrap`, with the last element being an `Input` (enhanced by Downshift). Unfortunately, every time an item was added/removed, the re-render of children caused Downshift to behave unexpectedly (https://github.com/downshift-js/downshift/issues/1187#issuecomment-713641908).

With the current implementation it's also difficult to customise the flex behaviour of individual items. For example, for my use-case to adjust the flex-grow, my workaround was as follows:

```jsx
<Wrap
  sx={{
    "> ul > li:last-of-type": {
      flexGrow: 2,
    },
  }}
>
```

Whilst this works _fine_, `Wrap's` DOM structure isn't immediately obvious (that it's wrapped in a `div`) and it's not exactly intuitive given the unreliable nature of Emotion's pseudo selectors (https://github.com/emotion-js/emotion/issues/1178)

## What is the new behavior?

Introduce `WrapItem` (like `ListItem`), which should be manually implemented around each child.

This would give more flexibility to add `flex` styles and avoid re-render issues. If a consumer wants to map through children and add the component to each item, they still have the option to do so on their own terms.

In an ideal world, we could style Wrap Items using an `> *` selector, but Emotion doesn't make that easy for us (https://github.com/emotion-js/emotion/issues/1178). This gives us two potential options:
1. Style items on the `groupStyles` with a basic class (e.g. `> .chakra-wrap__item`)
2. Use `StyleProvider` and `useStyles` to get item styles.

> I'd prefer to go down the option of a `.class` but I know that's not very Chakra

In this case it didn't make sense to use `useMultiStyleConfig()` because we don't need themed styles, but I'm open to any alternative ideas to make the dev experience better.

- [x] Yes
- [ ] No

## Other information

Alert users in the migration guide that `Wrap` will require children to be wrapped individually (which file should I do this in?)

I think it's also worth considering this approach for other "mapping" components such as `Stack` too.

-----

FYI: I'm more than happy to jump on a call to pair on any requested changes 🙂

### DIFF
--- a/packages/layout/src/wrap.tsx
+++ b/packages/layout/src/wrap.tsx
@@ -3,11 +3,13 @@ import {
   css,
   forwardRef,
   PropsOf,
+  StylesProvider,
   SystemProps,
   SystemStyleObject,
   useTheme,
+  useStyles,
 } from "@chakra-ui/system"
-import { getValidChildren, mapResponsive, __DEV__ } from "@chakra-ui/utils"
+import { mapResponsive, __DEV__ } from "@chakra-ui/utils"
 import * as React from "react"
 
 export interface WrapProps extends PropsOf<typeof chakra.div> {
@@ -61,15 +63,7 @@ export const Wrap = forwardRef<WrapProps, "div">(function Wrap(props, ref) {
     return `calc(${margin} / 2 * -1)`
   })
 
-  const validChildren = getValidChildren(children)
-
-  const clones = validChildren.map((child, index) => (
-    <chakra.li key={index} margin={itemSpacing}>
-      {child}
-    </chakra.li>
-  ))
-
-  const styles: SystemStyleObject = {
+  const groupStyles: SystemStyleObject = {
     display: "flex",
     flexWrap: "wrap",
     justifyContent: justify,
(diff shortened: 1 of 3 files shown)


joe-bell:
Here's a [screen recording](https://twitter.com/joebell_/status/1319645406083559424) of the alternative solution in action

with-heart:
I like this idea. I agree that modifying `children` is a mistake. We've moved away from it in a number of other cases, but there are still places where it persists.

ljosberinn:
Wonderful, thank you @joe-bell :)

ljosberinn:
@allcontributors please add @joe-bell for code

joe-bell:
🎉

PR #2301, opened 2020-10-23
Update color-mode.mdx Next.Js

iddify:

- [x] Docs have been reviewed and added / updated if needed (for bug fixes /start features)

- [x] Bugfix
- [x] Code style update (formatting, renaming)

Next.js code example causes a 'RangeError: Maximum call stack size exceeded' 

## What is the new behavior?

- Added Next Documents Imports.
- Fixed "RangeError: Maximum call stack size exceeded" caused by the old Document initialisation.

- [ ] Yes
- [x] No

### DIFF
--- a/website/pages/docs/features/color-mode.mdx
+++ b/website/pages/docs/features/color-mode.mdx
@@ -165,11 +165,13 @@ ReactDOM.render(
 
 ```jsx live=false
 // adjust your pages/_document.js
-import { ColorModeScript } from "@chakra-ui/core"
+import { ColorModeScript } from '@chakra-ui/core';
+import Document, { Html, Head, Main, NextScript } from 'next/document';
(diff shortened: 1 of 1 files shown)


ljosberinn:
re `Fixed "RangeError: Maximum call stack size exceeded" caused by the old Document initialisation.` - that's not the case, the docs use this initialization themselves [here](https://github.com/chakra-ui/chakra-ui/blob/develop/website/pages/_document.tsx#L13). Your version is an often copied example, but it's entirely redundant. You're awaiting it and then returning a new object of the old one. So you're unnecessarily creating an object. So youre just awaiting the response. Which makes async/await redundant if you instantly return it. See [this core eslint rule](https://eslint.org/docs/rules/no-return-await).

PR #2302, opened 2020-10-23
fix: Update cookies checks in example

msurdi:
I've found out that when there are no other cookies, checking just for a truthy value of `cookie` doesn't seem to work, but checking it is an actual string as in the example for older Next versions seems to work fine. I was wondering if this change was intentional, or just a mistake. 

Please, feel free to close this PR if the change was intentional and this is not the problem. 
Thanks!

## Pull request checklist

      /start features)

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

When there are no other cookies (such as session, etc,etc) and you're not calling `toggleColorMode` or `setColorMode` and just relying on system preferences, the cookie is never set.

Fixes: [ISSUE NUMBER]

## What is the new behavior?

When there are no other cookies (such as session, etc,etc) and you're not calling `toggleColorMode` or `setColorMode` and just relying on system preferences, now the cookie with the system preference will be set, so making it available for SSR.

- [ ] Yes
- [x] No

### DIFF
--- a/website/pages/docs/features/color-mode.mdx
+++ b/website/pages/docs/features/color-mode.mdx
@@ -227,7 +227,7 @@ import {
 
 export function Chakra({ cookies, children }) {
   // b) Pass `colorModeManager` prop
-  const colorModeManager = cookies
+  const colorModeManager = typeof cookies === "string"
     ? cookieStorageManager(cookies)
     : localStorageManager
 


ljosberinn:
Thanks, forgot to change it there, its intentional because of exactly the situation you mentioned :)

PR #2351, opened 2020-10-27
docs: Generate prop types table

TimKolberger:
## Pull request checklist

      /start features)

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently the props of each component are maintained individually and by hand. An undocumented prop or incomplete types are not an exception. 

## What is the new behavior?

I have played a bit with react docgen https://github.com/styleguidist/react-docgen-typescript
and found a way to generate the props table for every component in the docs automagically.

A post-build script creates for every *.ts file in the directory _packages/core/src_ a .json file with the [react-docgen-typescript component doc](https://github.com/styleguidist/react-docgen-typescript/blob/master/src/parser.ts#L18).

I started updating the docs for accordion, breadcrumb, close-button, icon, link and tabs.
Edit: I updated many more prop tables

- [ ] Yes
- [x] No


TimKolberger:
It works only locally for now, but here is a screenshot:
image

segunadebayo:
This is amazing! @TimKolberger. I had it mind to work on this after v1 but I love your approach.

Thank you. I'll review and test soon.

TimKolberger:
After adding the dep to `@chakra-ui/core` the package `props-docs` is build last. This fixes the "works on my machine" issue. Now we have a clean set of component info files.

I refactored the build script to make it more readable and replaced many prop tables in the docs.

All styled system props (`px`, `display`, etc.) filtered at build time.

The props "isTruncated", "layerStyle", "noOfLines" and "textStyle" are filtered on the client side, to show them just at the appropriate places, like `Text` and `Heading`.

https://github.com/chakra-ui/chakra-ui/blob/f4ba18cd809bcb94e39e658e9a607792002c7e35/website/src/components/props-table.tsx#L7-L27

TimKolberger:
@segunadebayo this is now ready for review :tada:

website/src/components/props-table.tsx
```diff
@@ -0,0 +1,98 @@
+import * as React from "react"
+// @ts-ignore
```
with-heart, line 2: Looks like this may not be needed anymore?

website/pages/docs/typography/text.mdx
```diff
@@ -125,3 +125,5 @@ Element** to see the element that gets rendered in html.
 
 Text composes the [Box](/docs/layout/box) component, so you can pass all `Box`
 style props.
+
+<PropsTable of="Text" omit={null} />
```
with-heart, line 129: Unnecessary `null` here?

website/pages/docs/typography/heading.mdx
```diff
@@ -111,6 +111,4 @@ prop. No need to write `css` or `styled()`
 The Heading composes the [Box](/docs/layout/box) component, so you can pass all
 `Box` props.
 
-| Name   | Type                                               | Default | Description              |
-| ------ | -------------------------------------------------- | ------- | ------------------------ |
-| `size` | `4xl`, `3xl`, `2xl`, `xl` , `lg`, `md`, `sm`, `xs` | `xl`    | The size of the heading. |
+<PropsTable of="Heading" omit={null} />
```
with-heart, line 114: Unnecessary `null` here?

website/src/components/props-table.tsx
```diff
@@ -0,0 +1,98 @@
...
+            return !omit.includes(propName)
+          }
+          return true
+        })
+        .sort(([a, aDef], [b, bDef]) => {
+          const aRequired = aDef.required ? 1000 : 0
+          const bRequired = bDef.required ? 1000 : 0
+          const requiredOffset = aRequired - bRequired
+          return String(a).localeCompare(b) - requiredOffset
+        }),
```
with-heart, line 50: This is a nice way to handle sort!

with-heart (approved):
This is such a valuable addition! It can be frustrating to change a component's API and have to remember to update the corresponding props table. Love this contribution!

with-heart:
@allcontributors please add @TimKolberger for code

website/pages/docs/typography/heading.mdx
TimKolberger: `null` is intended to override the default value:

https://github.com/chakra-ui/chakra-ui/blob/439d992aa5ce7d1508a63cc041beb52bff70a3d4/website/src/components/props-table.tsx#L25

An empty array would be suitable as well.

website/src/components/props-table.tsx
TimKolberger: You are right, I removed it ✅

website/pages/docs/typography/text.mdx
TimKolberger: See https://github.com/chakra-ui/chakra-ui/pull/2351#discussion_r515519184

website/pages/docs/typography/heading.mdx
with-heart: Ohhhh, yeap, that makes sense! If that were more common I'd say it makes more sense to have like a separate `hasTextProps` prop or something, but since it's only `Text` and `Heading` that should ever have that, I think this is totally fine :)

TimKolberger:
There are some empty prop tables left in the docs. These are an regression after adding the default value for `omit` on the `PropsTable` component.

What do you think of throwing an Error, if no props are left to render? This should break the docs build and the developer is notified when any props table is empty.

with-heart:
> There are some empty prop tables left in the docs. These are an regression after adding the default value for `omit` on the `PropsTable` component.

What causes this?

> What do you think of throwing an Error, if no props are left to render? This should break the docs build and the developer is notified when any props table is empty.

If it's an issue that we're rendering `PropsTable` for a component but there aren't any actual props to render, I think I'd rather it just returned `null` rather than throwing an error. But I'm not sure.

TimKolberger:
I started without filtering any prop and added more filters to reduce the number of documented props.

After adding the last filter some tables are now empty, because every prop is filtered.

Throwing the Error prevents the build to succeed if any props table is empty. This helps us to keep the documentation in a non broken state, without testing it manually.

with-heart:
> Throwing the Error prevents the build to succeed if any props table is empty. This helps us to keep the documentation in a non broken state, without testing it manually.

Yeah I think that makes sense!

website/src/components/props-table.tsx
```diff
@@ -50,6 +50,11 @@ const PropsTable = ({
     [info.props, omit, only],
   )
 
+  if (!entries.length) {
+    // this error breaks the build to notify you when there would be an empty table
+    throw new Error(`No props left to render for component ${of}`)
```
with-heart, line 55: Can you update this to explain how to resolve it? Maybe something like "Remove the use of `PropsTable` for this component".

TimKolberger: Update: 

https://github.com/chakra-ui/chakra-ui/blob/9b4ed1c804ce2bacef72a7640b4d90aa81623895/website/src/components/props-table.tsx#L53-L59

with-heart: NAILED IT!

with-heart:
imo this is good to go! Great work @TimKolberger!

What do you think @segunadebayo?

segunadebayo:
Thanks for this @TimKolberger.

- How about we move this package to the `tooling/` directory?

- Is there a way we can simply some extracted types? For example, the stack component shows this for the `align` prop:
```
"-moz-initial" | "inherit" | "initial" | "revert" | "unset" | (string & {}) | "normal" | "center" | "end" | "start" | "flex-end" | "flex-start" | "self-end" | "self-start" | "baseline" | "stretch" | ResponsiveArray<...> | ResponsiveObject<...> | undefined
```

Could we somehow override that by defining a custom type of information in the TSDoc comments? Perhaps something like this:

```
  /**
   * Shorthand for `alignItems` style prop
   * @override "SystemProps['alignItems']"
   */
  align?: SystemProps["alignItems"]
```

This isn't super crucial but I'd like us to simplify the displayed type for some component.

TimKolberger:
Thank you @segunadebayo.

I moved `packages/props-docs` to 'tooling/props-docs' 👍 

This merged PR https://github.com/styleguidist/react-docgen-typescript/pull/258 allows to override the ts value with a string
It is the perfect match for your `@override` proposal 🎉 
 
We could add the `@type` annotation to every usage of `SystemProps[x]`:

```ts
export interface GridOptions {
  /**
   * Short hand prop for `gridTemplateColumns`
   * @type SystemProps["gridTemplateColumns"]
   */
  templateColumns?: SystemProps["gridTemplateColumns"]
  /**
   * Short hand prop for `gridGap`
   * @type SystemProps["gridGap"]
   */
  gap?: SystemProps["gridGap"]
  /**
   * Short hand prop for `gridRowGap`
   * @type SystemProps["gridRowGap"]
   */
  rowGap?: SystemProps["gridRowGap"]

```

segunadebayo:
Amazing! Thank you @TimKolberger

PR #2540, opened 2020-11-17
fix: match media on useBreakpoint when no default value

stellarhoof:
## Pull request checklist

      /start features)

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

`useBreakpoint` uses `undefined` for its current breakpoint if no default value is passed, causing its effect to trigger a re-render when the current breakpoint gets updated to its correct value

## What is the new behavior?

`useBreakpoint` matches media on first render

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/.changeset/unlucky-peas-yawn.md
+++ b/.changeset/unlucky-peas-yawn.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/media-query": patch
+---
+
+fix: match media on useBreakpoint when no default value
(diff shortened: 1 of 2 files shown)


ljosberinn:
Besides CI crashing for unrelated reasons, I'm afraid this would crash server side as `window` is not available there.

stellarhoof:
Hmm you're right. However there's already a `matchMedia` [here](https://github.com/chakra-ui/chakra-ui/blob/32a2e50f05083ba3c1b00c7b988ebbcfd997b224/packages/media-query/src/use-breakpoint.ts#L63). Doesn't that break on SSR too?

ljosberinn:
useEffect doesnt run server side, hence it's safe to use there; you could incorporate  `isBrowser` in your changes

stellarhoof:
@ljosberinn Done

PR #2632, opened 2020-11-26
chore(extend-theme): improve theme override typings

TimKolberger:
## Pull request checklist

      /start features)

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The typings of the theme override is not specific enough to help while working on a theme extension.

## What is the new behavior?

- Autocomplete for existing components
  image
- existing colors
  image
- new colors
  image

- [ ] Yes
- [x] No - it does **not** introduce a narrower type, just type hints

### DIFF
--- a/packages/react/src/extend-theme.ts
+++ b/packages/react/src/extend-theme.ts
@@ -1,5 +1,6 @@
-import defaultTheme, { Theme, ColorHues } from "@chakra-ui/theme"
+import defaultTheme, { Theme } from "@chakra-ui/theme"
 import { isFunction, mergeWith } from "@chakra-ui/utils"
+import { ColorHues } from "@chakra-ui/theme/dist/types/foundations/colors"
 
 type ThemeExtensionTypeHints = {
   colors: Record<string, Partial<ColorHues> | string> // typehints for color definitions
--- a/packages/react/src/extend-theme.ts
+++ b/packages/react/src/extend-theme.ts
@@ -1,13 +1,32 @@
-import defaultTheme, { Theme } from "@chakra-ui/theme"
+import defaultTheme, { Theme, ColorHues } from "@chakra-ui/theme"
 import { isFunction, mergeWith } from "@chakra-ui/utils"
+
+type ThemeExtensionTypeHints = {
+  colors: Record<string, Partial<ColorHues> | string> // typehints for color definitions
+}
+
+/**
+ * Represents a loose but specific type for the theme override.
+ * It provides autocomplete hints for extending the theme, but leaves room
+ * for adding properties.
+ */
+type DeepThemeExtension<ThemeObject, TypeHints> = {
+  [Key in keyof ThemeObject]?:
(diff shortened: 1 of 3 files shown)


with-heart:
Are you me? I had been looking into the same type of improvement 😁 Will review when I have a sec

jvmazagao (approved):
I really don't like functions that return any, can you explain why this function `(ThemeObject[Key] extends (...args: any[]) => any` return any?

with-heart:
@jvmazagao It's important to look at the full context of that ternary to understand what's happening:

```ts
ThemeObject[Key] extends (...args: any[]) => any
        ? Partial<ReturnType<ThemeObject[Key]>>
        : ThemeObject[Key]
```

This is part of a syntax known as [conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) which was added in TypeScript 2.8. It's not actually assigning a type here, it's simply checking whether `ThemeObject[Key]` _extends from_ `(...args: any[]) => any`, which is a fancy way of saying "is `ThemeObject[Key]` a function?" 

If it's a function, we type it as `Partial<ReturnType<ThemeObject[Key]>>`, meaning it's a `Partial` of whatever that function returns. If it's not a function, the type is whatever the type of the `ThemeObject[Key]` object is.

with-heart (approved):
I think this is a good start. It's still limited by the fact that it's deriving types based on the actual content of the default theme and not all allowable values at each layer of the theme, but that's a lot more work so I like starting with this!

jvmazagao:
@with-heart Yeah, this makes sense. Can we have any unit tests for this?

packages/react/src/extend-theme.ts
```diff
@@ -1,13 +1,33 @@
...
+ * Represents a loose but specific type for the theme override.
+ * It provides autocomplete hints for extending the theme, but leaves room
+ * for adding properties.
+ */
+type DeepThemeExtension<ThemeObject, TypeHints> = {
+  [Key in keyof ThemeObject]?:
+    | Omit<DeepThemeExtension<ThemeObject[Key], TypeHints>, keyof TypeHints> // recursive type clone
+    | (ThemeObject[Key] extends (...args: any[]) => any
+        ? Partial<ReturnType<ThemeObject[Key]>>
+        : ThemeObject[Key]) // allow function or object
```
with-heart, line 19: Would we also want this to be `Partial`?

with-heart:
> @with-heart Yeah, this makes sense. Can we have any unit tests for this?

There are ways to unit test types, but we don't have any of them in the codebase yet. You basically set up tests that define various shapes of data using these types, and then type-check them to validate that they're valid. I'd like to get to that point, but we're not there yet.

packages/react/src/extend-theme.ts
TimKolberger: I think is covered by the "escape hatch" `Record<string, any>` one line below, but I will add `Partial` for completeness!👍

with-heart (approved):

PR #2636, opened 2020-11-26
chore(website): add /discord invite redirect

with-heart:
This PR adds `/discord` as a redirect to the URL for our Discord server invite so we can direct users to https://chakra-ui.com/discord.

### DIFF
--- a/website/vercel.json
+++ b/website/vercel.json
@@ -1,4 +1,10 @@
 {
   "alias": "chakra-ui.com",
-  "scope": "chakra-ui"
+  "scope": "chakra-ui",
+  "redirects": [
+    {
+      "source": "/discord",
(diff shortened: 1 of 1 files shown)


jmiazga:
Is it worth updating the link in the header to to `/discord` as well?

with-heart:
Looks like that redirect isn't working 🤷 I'll fix this later and update the URL in the header as well

with-heart:
I'm going to merge this to test it in prod. If it works, I'll follow-up with a URL change for the nav link. If it doesn't work, I'll follow-up with a fix or something.

with-heart:
Didn't work. The `vercel.json`-level redirects aren't working and we're seeing the 404 page from `next`. Going to try moving the redirect into `next` next (😀 ) instead.

PR #2666, opened 2020-11-29
Fix/framer types

jvmazagao:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Fixes: [ISSUE NUMBER]

## What is the new behavior?

- Fixing type of offsetX and offsetY
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/transition/src/slide-fade.tsx
+++ b/packages/transition/src/slide-fade.tsx
@@ -57,12 +57,12 @@ export interface SlideFadeProps extends HTMLMotionProps<"div"> {
    * The offset on the horizontal or `x` axis
    * @default 0
    */
-  offsetX?: string
+  offsetX?: string | number
   /**
    * The offset on the vertical or `y` axis
    * @default 8
    */
-  offsetY?: string
+  offsetY?: string | number
   /**
    * If `true`, the element will be transitioned back to the offset when it leaves.
    * Otherwise, it'll only fade out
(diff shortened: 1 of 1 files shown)


packages/transition/src/slide-fade.tsx
```diff
@@ -57,12 +57,12 @@ export interface SlideFadeProps extends HTMLMotionProps<"div"> {
...
    * @default 0
    */
-  offsetX?: number
+  offsetX?: string
   /**
    * The offset on the vertical or `y` axis
    * @default 8
    */
-  offsetY?: number
+  offsetY?: string
```
with-heart, line 65: I think for now we should allow them to be `string | number`. It can actually handle far more than that but I need to figure out how to access that specific type within `framer-motion`, so expanding this a bit is a good start.

with-heart (changes requested):

with-heart:
@allcontributors please add @jvmazagao for code

packages/transition/src/slide-fade.tsx
jvmazagao: Ok, in my free time, I will figure out how to expand these props, but, looking inside the `framer-motion` function that we use, they have something more inside that we need to look, but I will expand to number and string.

But based on the docs, this component only accept string int both offsets.

with-heart: @jvmazagao The types in the docs are generated from the types in the codebase, so it'll match whatever you put here 😉

jvmazagao: Fixed

Issue #2621, opened 2020-11-25
Allow JSX in toast description field


raulrpearson:
# Feature request

## Problem and solution

Wasn't sure if I should file this as bug report or feature request.

I want to be able to use JSX in the `description` option for `toast`. I had the need for this because I wanted to add an anchor element within that description. It turns out adding JSX already works, but one gets a TypeScript error because the field is [typed `string`](https://github.com/chakra-ui/chakra-ui/blob/3636952d1a29fbf307f44c66bf4d8996f30bea2e/packages/toast/src/use-toast.tsx#L48).

The solution might be as straightforward as retyping `description`?

## Describe alternatives you've considered

Before I delved into the code and realized that [JSX would work](https://github.com/chakra-ui/chakra-ui/blob/3636952d1a29fbf307f44c66bf4d8996f30bea2e/packages/toast/src/use-toast.tsx#L95), I was going to use the [render prop](https://chakra-ui.com/docs/feedback/toast#custom-component) to recreate a custom toast component that would accept markup in the description field. It was a bit annoying to have to recreate the functionality and styling of the regular toast component.

## Additional context

Maybe this approach could be extended to other parts like the `title` option, which currently is in the same situation.

My current solution is to tell TypeScript to ignore the error:

```javascript
toast({
  title: "The title",
  // @ts-expect-error
  description: (
    <Text>
      Lorem ipsum dolor{" "}
      <Link href="/">
        anchor text
      </Link>
    </Text>
  ),
  status: "success",
  duration: 9000,
  isClosable: true,
});
```

Let me know if you'd like me to PR.

raulrpearson:
👍 I'll get to it this weekend

PR #2668, opened 2020-11-30
Feat/toast/allow jsx

raulrpearson:
## Pull request checklist

- [ ] *Tests for the changes have been added (for bug fixes / features)
      /start features)

***Not sure if or what tests should be added.**

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves: #2621

## What is the new behavior?

`useToast` fields title and description get type `React.ReactNode` instead of `string`.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/toast/src/use-toast.tsx
+++ b/packages/toast/src/use-toast.tsx
@@ -41,11 +41,11 @@ export interface UseToastOptions {
   /**
    * The title of the toast
    */
-  title?: string
+  title?: React.ReactNode
   /**
    * The description of the toast
    */
-  description?: string
+  description?: React.ReactNode
   /**
    * If `true`, toast will show a close button
    */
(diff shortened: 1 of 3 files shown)


packages/toast/src/use-toast.tsx
```diff
@@ -41,11 +41,11 @@ export interface UseToastOptions {
   /**
    * The title of the toast
    */
-  title?: string
+  title?: React.ReactNode
```
jvmazagao, line 44: thinking in some usability, the title can be a string to, like I want to create a toast without jsx element, and the description either the same logic, so, in my mind, we need to accept both types strin | React.ReactNode

ljosberinn: `ReactNode` includes `ReactChild`, which includes `ReactText` which is a union of `string` and `number` so that should be already covered :)

jvmazagao: Thinking better, React.ReactNode includes ReactText and this is a union with string and number, so it's ok, Just accept the review. And close this PR

jvmazagao (approved):

TimKolberger (approved):

Issue #2700, opened 2020-12-04
Style prop configured as boolean in styled-system is not working


terryyu:
# Bug report

## Describe the bug

Style prop configured as boolean in styled-system is not working.

## To reproduce

Let's take `textTransform` as an example, it's configured as true in styled-system.
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/styled-system/src/config/typography.ts#L31

If we extend the theme, and provide a textTransform token.

```javascript
const theme = extendTheme({
  textTransform: {
    header: "uppercase"
  }
}
```

Then use that token in the code

```html
<Box textTransform="header">header</Box>
```

The token is not honored

## Minimal reproduction

https://codesandbox.io/s/jolly-lalande-1ew6j?file=/src/index.js

## Expected behavior

textTransform token should be honored and the text displays in uppercase.

## System information

- Version of @chakra-ui/react: 1.0.1

## Additional context

The config value retrieved should be converted to object if it's boolean
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/styled-system/src/css.ts#L150

with-heart:
I think we need input from @segunadebayo on this to make sure, but I believe this is intentional. All of the properties listed as `true` (or as an object with a `property` key) aren't meant to be themed, they're just extensions of underlying css properties. We provide `theme.textStyles` and `theme.layerStyles` for customizing something like this:

```jsx
const theme = extendTheme({ 
  textStyles: { 
    header: { 
      textTransform: 'uppercase'
    } 
  }
})

// ...

<Box textStyle="header">Hello World</Box>
```

terryyu:
@with-heart `textTransform` might be a bad example. I totally agree it can be substituted by textStyles, but there are quite a few other properties defined as `true`
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/styled-system/src/config/color.ts#L17
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/styled-system/src/config/others.ts#L5

I assume they are following the purpose of styled-system shortcut definition.
https://styled-system.com/custom-props/#shortcut-definition
https://github.com/styled-system/styled-system/blob/master/packages/core/src/index.js#L165

PR #2702, opened 2020-12-04
fix: style prop configured as boolean in styled-system is not working

terryyu:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Fixes: #2700

## What is the new behavior?

expand the shortcut definition by setting `property` and `scale` to the `key` name

- [ ] Yes
- [x] No

### DIFF
--- a/packages/styled-system/src/css.ts
+++ b/packages/styled-system/src/css.ts
@@ -147,7 +147,15 @@ export const css = (args: StyleObjectOrFn = {}) => (
     const val = runIfFn(x, theme)
 
     const key = k in pseudoSelectors ? pseudoSelectors[k] : k
-    const config = (parser.config as Dict)[key]
+    let config = (parser.config as Dict)[key]
+    
+    if (config === true) {
+      // shortcut definition
+      config = {
+        property: key,
+        scale: key,
+      }
+    }
 
     if (key === "apply") {
       const apply = css(get(theme, val))(theme)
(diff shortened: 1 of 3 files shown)


packages/styled-system/src/css.ts
```diff
@@ -147,7 +147,15 @@ export const css = (args: StyleObjectOrFn = {}) => (
     const val = runIfFn(x, theme)
 
     const key = k in pseudoSelectors ? pseudoSelectors[k] : k
-    const config = (parser.config as Dict)[key]
+    let config = (parser.config as Dict)[key]
```
jvmazagao, line 150: What's happening if the config module is false?

rr-terry-yu: by design, the config module only takes a true value if it's meant to be translated to an config object. https://styled-system.com/custom-props#shortcut-definition
the intention for a false value needs to be defined. the current behavior for a false value is not parsing the token against the theme and output the value directly.

terryyu: by design, the config module only takes a true value if it's meant to be translated to an config object. styled-system.com/custom-props#shortcut-definition
the intention for a false value needs to be defined. the current behavior for a false value is not parsing the token against the theme and output the value directly.

jvmazagao: Approved.

jvmazagao (approved):

packages/styled-system/src/css.ts
terryyu: thanks!

PR #2709, opened 2020-12-05
fix(system): set default useStyleConfig arguments

### LINKED ISSUE #2705: useStyleConfig without props throws
# Bug report

## Describe the bug

`useStyleConfig` allows passing undefined as `props` but the code relies on the presence of this parameter.

## To reproduce

`useStyleConfig('MyComponent')`

## Minimal reproduction

https://codesandbox.io/s/restless-sound-jiq50?file=/src/App.tsx

## Expected behavior

`props` should either have a sane fallback or be typed correctly

## Additional context

typedef: https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/use-style-config.ts#L22
destructuring: https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/use-style-config.ts#L27

with-heart:
Fixes #2705

`useStyleConfig` is typed for the `props` argument to be optional, but
we then destructure `props` within the function. This change resolves
the issue by setting a default empty object value for the `props` and
`opts` arguments.

### DIFF
--- a/packages/system/src/use-style-config.ts
+++ b/packages/system/src/use-style-config.ts
@@ -23,7 +23,7 @@ export function useStyleConfig(
   opts?: { isMultiPart?: boolean },
 ): SystemStyleObject
 
-export function useStyleConfig(themeKey: any, props: any, opts: any) {
+export function useStyleConfig(themeKey: any, props: any = {}, opts: any = {}) {
   const { styleConfig: styleConfigProp, ...rest } = props
 
   const { theme, colorMode } = useChakra()
--- a/.changeset/nice-tools-jump.md
+++ b/.changeset/nice-tools-jump.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/system": patch
+---
+
+Added default empty object argument values for the `props` and `opts` arguments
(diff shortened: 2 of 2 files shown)


packages/system/src/use-style-config.ts
```diff
@@ -23,7 +23,7 @@ export function useStyleConfig(
   opts?: { isMultiPart?: boolean },
 ): SystemStyleObject
 
-export function useStyleConfig(themeKey: any, props: any, opts: any) {
+export function useStyleConfig(themeKey: any, props: any = {}, opts: any = {}) {
```
TimKolberger, line 26: Would the type `Record<string, any>` for `props` and `opts` help with this?

with-heart: I'm not sure it makes a difference. The `any` types are devoured by the overrides, and it seems like `Record<string, any>` and `any` both work fine with the defaults specified.

TimKolberger: Oh, I missed the overrides! 👍

TimKolberger (approved):

PR #2716, opened 2020-12-05
chore: plop template

segunadebayo:
**Description**

To make it easier to contribute to Chakra UI, we need to make the process of bootstrapping a new component package smoother.

This PR uses plop as a generator to set up a pre-made package structure defined in `plop-templates/` directory. To contribute, you need to run `yarn gen:pkg` and that will ask 2 questions:

- What component package would you like to create?
- Description of this component (used in package.json)

This will then bootstrap the component package in the `packages/` directory with the package.json, README, and tests all good to go!


with-heart:
I think what would make this _extremely_ valuable would be adding comments to the various files in a component template detailed some basic expectations for what kind of content should exist in them. Hooks connect props and callbacks with a11y concerns and component behavior, tests should address things in the test checklist, etc. Make it so that it comes with implicit information rather than being totally empty!

segunadebayo:
Amazing ideas @with-heart. I'll add some comments in there and link to the guides we wrote for creating components.

Issue #2670, opened 2020-11-30
Can't override `shouldForwardProp`


with-heart:
# Bug report

## Describe the bug

When using the `chakra` factory function, users should be able to pass a custom `shouldForwardProp` function to replace that functionality for their custom component.

## Minimal reproduction

https://codesandbox.io/s/distracted-cloud-ncioj?file=/src/App.tsx

## Expected behavior

Base `div` element should not have `hello="true"` attribute attached.

## Additional context

Users can specify the `shouldForwardProp` option, but it is always overridden inside `styled`.

https://github.com/chakra-ui/chakra-ui/blob/b5f1211d2093727e9733d2c6f2728adce3338676/packages/system/src/system.ts#L128-L139

jvmazagao:
I will try to fix it, assign to me

with-heart:
Should be an easy one! 😄

jvmazagao:
@with-heart Based on your issue when the shouldForwardProp are passed to the CustomComponent if the props.hello are passed, they will show an error in the component right?

with-heart:
@jvmazagao Nope! `shouldForwardProp` just determines whether a prop is also forwarded to the underlying html element or not. So in this case, `<div hello="true" class="css-0">` is rendered, but it should just be `<div class="css-0">` since `shouldForwardProp` returns `false` for `hello`.

segunadebayo:
I just pushed a fix for this. We'll release an update shortly

PR #2717, opened 2020-12-05
fix: fixing shouldFowardProps

jvmazagao:
## Pull request checklist

      /start features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Fixes: [2670]

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/system/src/system.ts
+++ b/packages/system/src/system.ts
@@ -138,14 +138,12 @@ export function styled<T extends As, P = {}>(
   const { baseStyle, ...styledOptions } = options ?? {}
 
   const mergedShouldForwardPropFn = (prop: string) => {
-    const origShouldForward = shouldForwardPropFn(prop) // our original fn
+    const origShouldForward = shouldForwardPropFn(prop)
     if (!options?.shouldForwardProp) {
-      // no custom fn provided
       return origShouldForward
     }
 
     const optionShouldForward = options.shouldForwardProp(prop)
-    // merge results
     return optionShouldForward || origShouldForward
   }
 
--- a/packages/system/src/system.ts
+++ b/packages/system/src/system.ts
@@ -135,7 +135,7 @@ export function styled<T extends As, P = {}>(
   component: T,
   options?: StyledOptions,
 ) {
-  const { baseStyle, ...styledOptions } = options ?? {}
+  const { baseStyle, shouldForwardProp, ...styledOptions } = options ?? {}
   const opts = { ...styledOptions, shouldForwardProp }
 
   const styledFn = emotionStyled(component as React.ComponentType<any>, opts)
--- a/packages/system/src/system.ts
+++ b/packages/system/src/system.ts
@@ -18,6 +18,7 @@ import emotionStyled, {
   FunctionInterpolation,
   Interpolation,
 } from "@emotion/styled"
+import { shouldForwardProp as shouldForwardPropFn } from "./should-forward-prop"
 import { As, ChakraComponent, ChakraProps, PropsOf } from "./system.types"
 import { domElements, DOMElements } from "./system.utils"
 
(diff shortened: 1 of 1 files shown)


packages/system/src/system.ts
```diff
@@ -18,7 +18,6 @@ import emotionStyled, {
   FunctionInterpolation,
   Interpolation,
 } from "@emotion/styled"
-import { shouldForwardProp } from "./should-forward-prop"
```
segunadebayo, line 21: I think the goal here is not to remove the `shouldForwardProp` function but to apply other custom `shouldForwardProp` passed in the `options` **after** the default `shouldForwardProp` has been called.

The logic is more like:

```jsx
const shouldForwardPropFn = (prop) => {
  // if prop is listed in our styled system, use chakra's shouldForwardProp
  // else if options.shouldForwardProp is defined, use that
 // else return `true`
}
```

jvmazagao: OK, I understand this right now. I will fix today.

with-heart: imo I think what we ultimately want is to merge the two together into a single function.

If `options.shouldForwardProp` does not exist: use the existing `shouldForwardProp` function

If `options.shouldForwardProp` does exist: create a new function that takes `prop`, calls `shouldForwardProp` and `options.shouldForwardProp`, and then merges the results together. 

That way `options.shouldForwardProp` becomes an _extension_ of the base `shouldForwardProp` function rather than a replacement. 

Not sure that change makes sense for this PR or if it should be separate work.

jvmazagao: @with-heart and @segunadebayo pushed some changes.

jvmazagao: not finished, this part not makes senses to me

If options.shouldForwardProp does exist: create a new function that takes prop, calls shouldForwardProp and options.shouldForwardProp, and then merges the results together.

I don't have the props in this scope, how can I merge the two in one? I have no idea.

TimKolberger: We create a new function. And this new function will be called with the prop name, and we call the existing functions to compute the result.

Define a new function which receives any arguments and pass them to the two different `shouldForwardProps` functions. If one of those functions return `true`, the props will be passed through.

An implementation could look like this:

```ts
import { shouldForwardProp as shouldForwardPropFn } from './should-forward-prop';

// ...

const mergedShouldForwardPropFn = (...args: any[]) => {
  const origShouldForward = shouldForwardPropFn(...args); // our original fn
  if (!options?.shouldForwardProp) {
    // no custom fn provided
    return origShouldForward;
  }

  const optionShouldForward = options.shouldForwardProp(...args);
  // merge results
  return optionShouldForward || origShouldForward;
};

const opts = { ...styledOptions, shouldForwardProp: mergedShouldForwardPropFn };
// ...
```

(!) This implementation would prevent the `options.shouldForwardProp` from overriding a truthy return type of our `shouldForwardProp` function. I don't know if this is the intended behaviour for this use case. I am not very familiar with this area of chakra yet.

with-heart: > I don't have the props in this scope

It's sort of like how `shouldForwardProps` is defined currently. It's a function that takes `prop`, but we're not actually passing props to it, we're handing the function to `emotionStyled` for it to call it later.

@TimKolberger I think this is close, but it's missing something. I'm going to make a new issue for this so we can discuss a little more. We might need a separate option to handle the use cases I'm thinking of.

@jvmazagao Sorry for starting this conversation in the middle of your PR lol. I think what would resolve #2670 would be to go back to how things worked originally, but `opts` should be `{ shouldForwardProp, ...styledOptions }` instead of `{ ...styledOptions, shouldForwardProp }`, so that if a user includes `styledOptions.shouldForwardProps`, it overrides the built-in one.

segunadebayo: I see where you're going @with-heart. I think a better solution will be to do this:
`{ shouldForwardProp, ...styledOptions }` and then export our `shouldForwardProp` function so that users can leverage that when they need to build their own custom function.

Similar to how emotion has `@emotion/is-prop-valid` pkg but for us, we're simply exporting it out. This is a simpler approach, nice thought @with-heart!

jvmazagao: Sorry guys, but I'm a little confused with this discussion, and I love to work in another pr to fix another issue if I'm capable of that.

segunadebayo:
I just pushed a fix and docs for this.

PR #2729, opened 2020-12-07
feat: update and refactor pin input

### LINKED ISSUE #2679: [PinInput] Support alphanumeric pins
I'd like to propose supporting alphanumeric pins in the PinInput component. Limiting to digits seems to be an arbitrary restriction pulled from the referenced `pin-code` from Uber[[1]](https://github.com/uber/baseweb/blob/master/src/pin-code/pin-code.js#L115). I think this could either come as a prop configuration or simply changing the existing regex[[2]](https://github.com/chakra-ui/chakra-ui/blob/453db5e2bf788f51fc88387d2c9d185b911d4e44/packages/pin-input/src/use-pin-input.ts#L249) to alphanumeric.

[1] https://github.com/uber/baseweb/blob/master/src/pin-code/pin-code.js#L115
[2] https://github.com/chakra-ui/chakra-ui/blob/453db5e2bf788f51fc88387d2c9d185b911d4e44/packages/pin-input/src/use-pin-input.ts#L249

segunadebayo:
Closes #2679

## 📝 Description

- Move input props logic to parent hook and expose prop-getter `getInputProps`.
  This helps to co-locate the state logic for easier debugging
- Added support for alpha-numeric and secret values.
- Added `type` prop which can be either `alphanumeric` or `number`
- Added `mask` prop to hide input value similar to `<input type='password' />`

```jsx
<PinInput type="alphanumeric" mask>
  <PinInputField />
  <PinInputField />
  <PinInputField />
</PinInput>
```

### DIFF
--- a/.changeset/curly-squids-chew.md
+++ b/.changeset/curly-squids-chew.md
@@ -1,0 +1,17 @@
+---
+"@chakra-ui/pin-input": minor
+---
+
+- Move input props logic to parent hook and expose prop-getter `getInputProps`.
+  This helps to co-locate the state logic for easier debugging
+- Added support for alpha-numeric and secret values.
+- Added `type` prop which can be either `alphanumeric` or `number`
+- Added `mask` prop to hide input value similar to `<input type='password' />`
+
+```jsx
+<PinInput type="alphanumeric" mask>
+  <PinInputField />
+  <PinInputField />
+  <PinInputField />
+</PinInput>
+```
--- a/packages/pin-input/src/use-pin-input.ts
+++ b/packages/pin-input/src/use-pin-input.ts
@@ -79,17 +79,23 @@ export interface UsePinInputProps {
   /**
    * The type of values the pin-input should allow
    */
-  type?: "string" | "number"
+  type?: "alphanumeric" | "number"
+  /**
+   * If `true`, the input's value will be masked just like `type=password`
+   */
+  mask?: boolean
 }
 
 function toArray(value?: string) {
   return typeof value === "string" ? value.split("") : undefined
 }
 
-function checkValueRegex(value: string, type: "string" | "number") {
+function validate(value: string, type: UsePinInputProps["type"]) {
   const NUMERIC_REGEX = /^[0-9]+$/
-  const ALPHA_NUMERIC_REGEX = /^[a-z0-9]+$/i
-  return value.match(type === "string" ? ALPHA_NUMERIC_REGEX : NUMERIC_REGEX)
+  const ALPHA_NUMERIC_REGEX = /^[a-zA-Z0-9]+$/i
+  return value.match(
+    type === "alphanumeric" ? ALPHA_NUMERIC_REGEX : NUMERIC_REGEX,
+  )
 }
 
 export function usePinInput(props: UsePinInputProps = {}) {
@@ -105,6 +111,7 @@ export function usePinInput(props: UsePinInputProps = {}) {
     isDisabled,
     isInvalid,
     type = "number",
+    mask,
   } = props
 
   const uuid = useId()
@@ -146,8 +153,12 @@ export function usePinInput(props: UsePinInputProps = {}) {
       nextValues[index] = value
       setValues(nextValues)
 
-      // if we're at the last input, call onComplete (no need to move focus)
-      if (index === descendants.length - 1) {
+      const isComplete =
+        value !== "" &&
+        index === descendants.length - 1 &&
+        values.every((inputValue) => inputValue !== "")
+
+      if (isComplete) {
         onComplete?.(nextValues.join(""))
       } else {
         focusNext(index)
@@ -167,11 +178,10 @@ export function usePinInput(props: UsePinInputProps = {}) {
     (value: string, eventValue: string) => {
       let nextValue = eventValue
       if (value?.length > 0) {
-        const [firstValue, secondValue] = eventValue
-        if (value[0] === firstValue) {
-          nextValue = secondValue
-        } else if (value[0] === secondValue) {
-          nextValue = firstValue
+        if (value[0] === eventValue.charAt(0)) {
+          nextValue = eventValue.charAt(1)
+        } else if (value[0] === eventValue.charAt(1)) {
+          nextValue = eventValue.charAt(0)
         }
       }
       return nextValue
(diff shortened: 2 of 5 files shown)


.changeset/curly-squids-chew.md
```diff
@@ -0,0 +1,17 @@
...
+- Added `type` prop which can be either `alphanumeric` or `number`
+- Added `mask` prop to hide input value similar to `<input type='password' />`
+
+```jsx
+<PinInput type="alphanumeric" mask>
+  <PinInputField />
+  <PinInputField />
+  <PinInputField />
+</PinInput>
+```
```
with-heart, line 17: Nice! I love putting an example of usage directly in the changelog! Makes it much easier to digest the change than trying to parse through all of the above text and understand what it means.

.changeset/curly-squids-chew.md
```diff
@@ -0,0 +1,17 @@
+---
+"@chakra-ui/pin-input": minor
+---
+
+- Move input props logic to parent hook and expose prop-getter `getInputProps`.
+  This helps to co-locate the state logic for easier debugging
+- Added support for alpha-numeric and secret values.
+- Added `type` prop which can be either `alphanumeric` or `number`
+- Added `mask` prop to hide input value similar to `<input type='password' />`
```
with-heart, line 9: Random thought on how to make something like this potentially more clear in the future:

- Added support for alpha-numeric values via new `type` prop, which can be either `alphanumeric` or `number` (default: `number`)
- Added support for secret/hidden values via the `mask` prop (default: `false`), similar to `<input type="password" />`

Just makes it a little more clear that this is 3 changes, which prop corresponds to which change, and what the defaults of these new props are. Not a huge deal but just a thought that came to mind that I wanted to share 😁

packages/pin-input/src/use-pin-input.ts
```diff
@@ -86,13 +79,26 @@ export interface UsePinInputProps {
...
+   */
+  mask?: boolean
 }
 
 function toArray(value?: string) {
-  if (typeof value === "string") {
-    return value.split("")
-  }
-  return undefined
+  return typeof value === "string" ? value.split("") : undefined
```
with-heart, line 93: Would we actually need to check type here anymore if `value` is typed to `string`?

packages/pin-input/src/use-pin-input.ts
```diff
@@ -115,22 +123,24 @@ export function usePinInput(props: UsePinInputProps = {}) {
...
 
-  useEffect(() => {
+  React.useEffect(() => {
     if (autoFocus) {
       const firstInput = descendants[0]
       firstInput?.element?.focus()
     }
-  }, [descendants, autoFocus])
+    // We don't want to listen for updates to `autoFocus` since it only runs initially
+    // eslint-disable-next-line
```
with-heart, line 140: It'd probably be safer to refactor this a little than to disable `exhaustive-deps`. If this were to be changed in the future it might mean we miss adding a dep, or removing a dep we no longer rely on. Not the biggest deal though, just a thought for future-proofing.

packages/pin-input/src/use-pin-input.ts
```diff
@@ -156,166 +170,173 @@ export function usePinInput(props: UsePinInputProps = {}) {
...
+  const getNextValue = React.useCallback(
+    (value: string, eventValue: string) => {
+      let nextValue = eventValue
+      if (value?.length > 0) {
+        if (value[0] === eventValue.charAt(0)) {
+          nextValue = eventValue.charAt(1)
+        } else if (value[0] === eventValue.charAt(1)) {
+          nextValue = eventValue.charAt(0)
+        }
+      }
```
with-heart, line 189: Having trouble parsing this. What is this block doing? Handling a backspace input?

packages/pin-input/src/use-pin-input.ts
```diff
@@ -156,166 +170,173 @@ export function usePinInput(props: UsePinInputProps = {}) {
...
+              .filter((_, index) => index < descendants.length)
+
+            setValues(nextValue)
+
+            // if pasting fills the entire input fields, trigger `onComplete`
+            if (nextValue.length === descendants.length) {
+              onComplete?.(nextValue.join(""))
+            }
+          } else {
+            return
```
with-heart, line 231: I'm not sure what the UX here would look like, but should we potentially do something in this case or provide a callback to react to this case? Could be confusing to users to have their paste just not work.

packages/pin-input/src/use-pin-input.ts
```diff
@@ -156,166 +170,173 @@ export function usePinInput(props: UsePinInputProps = {}) {
...
+  const [focusedIndex, setFocusedIndex] = React.useState(-1)
 
-export interface UsePinInputFieldProps {
-  ref?: Ref<HTMLInputElement>
-  onChange?: ChangeEventHandler
-  onKeyDown?: KeyboardEventHandler
-  onFocus?: FocusEventHandler
-  onBlur?: FocusEventHandler
-}
+  const getInputProps = React.useCallback(
```
with-heart, line 197: This is such a nice improvement. Moves `PinInput` closer to more recent component implementations.

with-heart (commented):
Solid work here buddy! Just a few thoughts as comments, but very welcome changes for sure

.changeset/curly-squids-chew.md
segunadebayo: Amazing! We can reflect this in the docs for PinInput.

In the future, feel free to update the changeset as well. I'd love that but I learned from your suggestions. Thanks for sharing!

packages/pin-input/src/use-pin-input.ts
segunadebayo: It's because the value could be `undefined` as well. I could have just done `value?.split("")` haha. Will refactor

packages/pin-input/src/use-pin-input.ts
segunadebayo: Not sure about this one but it's something we want but `eslint` gets in the way.

packages/pin-input/src/use-pin-input.ts
segunadebayo: We need to be a bit strict here given the user can enter both "alphanumeric" or "number" values. If developers want users to enter any value, then it's appropriate to use "alphanumeric".

We also want to be able to sanitize/validate the values user paste into the box (out of security concerns) with what the business logic allows.

You get what I mean?

packages/pin-input/src/use-pin-input.ts
segunadebayo: Exactly!

.changeset/curly-squids-chew.md
with-heart: @segunadebayo Good idea! Didn't even think about updating them directly lol

PR #2737, opened 2020-12-08
fix: cannot customize the radio container theme

gifaeriyanto:
## 📝 Description

Now we cannot customize the radio container like the checkbox. I want to add a part named container like the checkbox, so we can customize it.

We cannot customize the radio container

## 🚀 New behavior

Now we can customize the radio container

### DIFF
--- a/.changeset/clever-eggs-kick.md
+++ b/.changeset/clever-eggs-kick.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/radio": patch
+---
+
+Add support for styling the radio container in the theme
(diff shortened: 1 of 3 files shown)


dodas (changes requested):
You should add `container` to the `parts` array in the theme: https://github.com/chakra-ui/chakra-ui/blob/f41a231aa7241419aa8bb4a31cca508e19a3a464/packages/theme/src/components/radio.ts#L3

gifaeriyanto:
> You should add `container` to the `parts` array in the theme:
> 
> https://github.com/chakra-ui/chakra-ui/blob/f41a231aa7241419aa8bb4a31cca508e19a3a464/packages/theme/src/components/radio.ts#L3

oh nice, I'll add it. Thanks

dodas (approved):

ljosberinn:
@allcontributors please add @gifaeriyanto for bug

PR #2753, opened 2020-12-09
fix: typescript autocomplete

### LINKED ISSUE #2265: Code completion broken in WebStorm
# Bug report

## Describe the bug

Code completion is broken in WebStorm.

## To reproduce

1. Create a minimal component:
```ts
import { Link as ChakraLink } from '@chakra-ui/core';
import React from 'react';

export interface LinkProps {
  isActive: boolean;
  to: string;
  children: string;
}

export function Link(props: LinkProps) {
  const { isActive, to, children } = props;

  return <ChakraLink>{children}</ChakraLink>;
}
```
2. Position cursor after `ChakraLink` but before `>`
3. Hit ctrl-space to trigger autocomplete
4. Result:
image

## Minimal reproduction

Minimal repro: https://codesandbox.io/s/peaceful-euler-s3p8n?file=/src/Link.tsx

## Expected behavior

IDE offers completion, like VSCode does:
image

## System information

- OS: Win 10
- Browser (if applies): Firefox
- Version of @chakra-ui/core: 1.0.0-rc.5
- Version of Node.js: 12.18.1
- WebStorm: 2020.2.3
- TypeScript: 3.9.7

TimKolberger:
Closes #2265

## 📝 Description

Autocomplete was broken in Jetbrains IDEs with the complex `as` prop type.

Multiple intersection types including many usages of `Omit` took too long to resolve for Jetbrains IDEs.

## 🚀 New behavior

This PR uses less intersection types and resolves fast in PhpStorm 2020.2.

**It should not,** but this is a pretty big change in our type codebase, so I would like to be sure that it is an improvement for every editor.

## 📝 Additional Information

One explicit type annotation for a `callAllHandlers` call was needed and the `ModalTransitionProps` needed a minor change.

### DIFF
--- a/packages/system/tests/as-prop.test.tsx
+++ b/packages/system/tests/as-prop.test.tsx
@@ -1,0 +1,41 @@
+import * as React from "react"
+import { chakra } from ".."
+import { ChakraComponent } from "../src"
+
+/**
+ * These tests should fail
+ * if the typings for `ChakraComponent` change
+ * and create a type regression
+ */
+describe("`as` prop typings", () => {
+  const Comp = (props: {}) => null
+
+  type CompWithRequiredProps = { thisIsARequiredProp: boolean }
+  const CompWithRequired = (props: CompWithRequiredProps) => null
+
+  it("should has correct types for the chakra factory", () => {
+    const div = <chakra.div />
+    const divWithAsTag = <chakra.div as="img" src="/this-is-the-way.webp" />
+    const divWithAsComp = <chakra.div as={Comp} />
+    const divWithAsCompRequired = (
+      <chakra.div as={CompWithRequired} thisIsARequiredProp />
+    )
+
+    // make jest happy
+    expect(true).toBe(true)
+  })
+
+  it("should has correct types for the ChakraComponent", () => {
+    const Div: ChakraComponent<"div"> = (props) => <chakra.div {...props} />
+    const CustomComp: ChakraComponent<typeof Comp> = (props) => (
+      <chakra.div as={Comp} {...props} />
+    )
+    const CustomCompWithRequired = chakra(CompWithRequired)
+
+    // TODO this should fail due missing required prop
+    const renderedCustomCompWithRequired = <CustomCompWithRequired />
+
+    // make jest happy
+    expect(true).toBe(true)
+  })
+})
--- a/packages/system/tests/as-prop.test.tsx
+++ b/packages/system/tests/as-prop.test.tsx
@@ -31,8 +31,9 @@ describe("`as` prop typings", () => {
     )
     const CustomCompWithRequired = chakra(CompWithRequired)
 
-    // TODO this should fail due missing required prop
-    const renderedCustomCompWithRequired = <CustomCompWithRequired />
+    const renderedCustomCompWithRequired = (
+      <CustomCompWithRequired thisIsARequiredProp />
+    )
 
     // make jest happy
     expect(true).toBe(true)
--- a/packages/system/tests/as-prop.test.tsx
+++ b/packages/system/tests/as-prop.test.tsx
@@ -2,7 +2,7 @@ import * as React from "react"
 import { chakra, ChakraComponent } from ".."
 
 /**
- * These tests should fail
+ * These tests should fail while type checking
  * if the typings for `ChakraComponent` change
  * and create a type regression
  */
@@ -12,7 +12,7 @@ describe("`as` prop typings", () => {
   type CompWithRequiredProps = { thisIsARequiredProp: boolean }
   const CompWithRequired = (props: CompWithRequiredProps) => null
 
-  it("should has correct types for the chakra factory", () => {
+  it("should have correct types for the chakra factory", () => {
     const div = <chakra.div />
     const divWithAsTag = <chakra.div as="img" src="/this-is-the-way.webp" />
     const divWithAsComp = <chakra.div as={Comp} />
@@ -24,7 +24,7 @@ describe("`as` prop typings", () => {
     expect(true).toBe(true)
   })
 
-  it("should has correct types for the ChakraComponent", () => {
+  it("should have correct types for the ChakraComponent", () => {
     const Div: ChakraComponent<"div"> = (props) => <chakra.div {...props} />
     const CustomComp: ChakraComponent<typeof Comp> = (props) => (
       <chakra.div as={Comp} {...props} />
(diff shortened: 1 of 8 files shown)


with-heart:
I feel like we may have reached the point where we need type tests lol. I have basically no way of parsing this out in my head to understand what exactly we're gaining or losing here.

ljosberinn:
Works nicely in VSCode, no perf degradation at all. if it works in IntelliJ apps now, that's great!

packages/system/tests/as-prop.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  it("should has correct types for the ChakraComponent", () => {
+    const Div: ChakraComponent<"div"> = (props) => <chakra.div {...props} />
+    const CustomComp: ChakraComponent<typeof Comp> = (props) => (
+      <chakra.div as={Comp} {...props} />
+    )
+    const CustomCompWithRequired = chakra(CompWithRequired)
+
+    // TODO this should fail due missing required prop
+    const renderedCustomCompWithRequired = <CustomCompWithRequired />
```
TimKolberger, line 36: Well, there is one thing missing for now: It does not pass along required props. You could apply every prop and it won't break. 

This needs to be improved 😅  

```ts
type CompWithRequiredProps = { thisIsARequiredProp: boolean }
const CompWithRequired = (props: CompWithRequiredProps) => null
```

I converted this to a draft until this is fixed.

TimKolberger: This is the complete failing test:

```ts
type CompWithRequiredProps = { thisIsARequiredProp: boolean }
const CompWithRequired = (props: CompWithRequiredProps) => null

const CustomCompWithRequired = chakra(CompWithRequired)

 // TODO this should fail due missing required prop
 const renderedCustomCompWithRequired = <CustomCompWithRequired />
```

TimKolberger: This is fixed 🎉

TimKolberger:
## Status

Autocompletion in Jetbrains IDEs is now fast for own component props. e.g.:

image
image


It does **not** provide autocomplete for the `as` prop (I am looking for the prop `src` here :eyes:): 

image

But TypeScript is happy - the IDE knows that it is an `img src` when jumping to the prop definition, just the autocomplete lacks. I could not get this to work. But basic autocomplete is better than none 😅 

image

## Tests

I created a bunch few TypeScript type tests in a jest test. jest will never fail - but TypeScript will, if it doesn't like something at compile time of the tests.

## Does thin introduce breaking changes?

I had to alter the `ModalTransitionProps` in our codebase and needed to explicitly type an `callAllHandlers` callback. This could have implications on usages of third party or userland components like `motion.div` which are wrapped with the chakra factory. 

@chakra-ui/react-team It would be awesome if you would test this again in VS Code (it worked on my machine) and maybe in one of your existing projects if this introduces a breaking type change 🤞

TimKolberger:
If someone is eager to get a preview or test this PR:

You can test this in your own project by cloning the branch (`fix/typescript-autocomplete`) of this PR and run:

```bash
yarn boot
cd packages
cp -r . <your-chakra-project>/node-modules/@chakra-ui

# build your own project and check for typescript errors
```

Please provide feedback if any type issues arise! 💖

vacekj:
Confirmed working on my setup using Tim's branch and instructions and using WebStorm, however there still seem to be bugs.
1. image
This appears on Input and Select Chakra Components.

2. Auto-importing Chakra Components results in this import 
image
instead of just being added to the existing Chakra import.

3. The types are still slow to appear, even on a 2020 13" MBP with 10th gen i5 and 16GB RAM, and 4GB allocated for WebStorm.

Thank you for your work nevertheless, I am sure these bugs can be fixed.

TimKolberger:
Thank you for your review @vacekj!

1. The missing required props is an issue how Jetbrains interprets the type `Omit`. We could fix that by applying `Partial<Omit<TargetType>>` for component props. But I think this is out of scope of this PR.
    ```tsx
    export interface TabsProps
      extends UseTabsProps,
        ThemingProps,
        Omit<HTMLChakraProps<'div'>, 'onChange'>,
        TabsOptions {
      children: React.ReactNode;
    }
    ```
    
      ```tsx
    export interface TabsProps
      extends UseTabsProps,
        ThemingProps,
        Partial<Omit<HTMLChakraProps<'div'>, 'onChange'>,
        TabsOptions {
      children: React.ReactNode;
    }
    ```
2. The wrong import path origins from the presence of the `src` directory in the copied chakra-ui packages. This won't happen with the installed version from npm.
3. There are many props due the validity of all css properties on chakra components. And speed is relative 😅 
     IMO some autocomplete is better than none, but here is a video recording of my machine (Macbook Pro 2016 i7 with 4GB for PhpStorm):


    https://user-images.githubusercontent.com/16899513/103233291-0293b780-493d-11eb-8146-d8209dc281d9.mp4

vacekj:
Thank you for your response and all the work you have done Tim. I really appreciate it. 
Ad 1. I agree this is definitely out of scope. 
Ad 2. My bad, thank you for pointing it out.
AD 3. This looks about as fast as my machine, so it's probably as fast as it gets. Definitely agree that some autocomplete is better than none.

Thank you again for putting in the work, this will make using Chakra a much better experience not just for me, but for the many people that use JetBrains products.

I'm looking forward to the release version to start using this in production.

Have a nice day 🙂

with-heart:
> 1. The missing required props is an issue how Jetbrains interprets the type `Omit`. We could fix that by applying `Partial<Omit<TargetType>>` for component props. But I think this is out of scope of this PR.

Wait, what? It interprets `Omit` differently than TypeScript itself?

TimKolberger:
Yes I think so.
It is no TypeScript warning, it seems like another linter in the IDE is logging those - but I don't know anything about the internals of Jetbrains linters.

It appears when Omitting some keys of a Partial type -> It thinks all other keys are required - even if they are not.

 I created a simplified reproduction in [this gist](https://gist.github.com/TimKolberger/0900962f7f2c28b4a80bcd8754785cf2).

TimKolberger:
I rebased onto develop and improved the typings for `forwardRef`. 
imo this is now ready to be merged. @segunadebayo what do you think?

segunadebayo:
Thank you for your hard work on this @TimKolberger.

wvieed:
H

احصل على Outlook لـ iOS<https://aka.ms/o0ukef>
________________________________
من: Segun Adebayo <notifications@github.com>
‏‏تم الإرسال: Wednesday, January 6, 2021 10:05:46 AM
إلى: chakra-ui/chakra-ui <chakra-ui@noreply.github.com>
نسخة: Subscribed <subscribed@noreply.github.com>
‏‏الموضوع: Re: [chakra-ui/chakra-ui] fix: typescript autocomplete (#2753)


Merged #2753<https://github.com/chakra-ui/chakra-ui/pull/2753> into develop.

PR #2754, opened 2020-12-10
fix(form-control): don't omit FormHelperText when FormControl is invalid

### LINKED ISSUE #2365: FormHelperText disappears when isInvalid is set to true on the FormControl
# Bug report

## Describe the bug

When isInvalid is set to true on the FormControl the FormHelperText disappears

## To reproduce

```
<FormControl id="email" isInvalid={true}>
  <FormLabel>Email address</FormLabel>
  <Input type="email" />
  <FormHelperText>We'll never share your email.</FormHelperText>
</FormControl>
```

## Minimal reproduction

https://codesandbox.io/s/elegant-mendeleev-izs3w?file=/src/App.js

## Expected behavior

The helper text should still be there if the form is invalid

## System information

- OS: macOS
- Browser (if applies): Chrome
- Version of @chakra-ui/core: 1.0.0-rc.7
- Version of Node.js: v12.16.1

dodas:
Closes #2365

The `FormHelperText` component does not render when its parent `FormControl` has `isInvalid` prop set to true.

## 🚀 New behavior

The `FormHelperText` will always render. If user pleases not to, it's easy to implement in userland with a single AND expression.

No


packages/form-control/src/use-form-control.ts
```diff
@@ -25,18 +25,11 @@ export function useFormControl<T extends HTMLElement>(
...
-    } else {
-      describedBy.push(field.feedbackId)
-    }
-  }
-
+  /**
+   * Error message must be described first
+   * in all scenarios
+   */
+  if (field?.isInvalid) describedBy.push(field.feedbackId)
```
dodas, line 32: Do we want to add a check similar to `hasHelpText ` here? Currently it always adds feedbackId when isInvalid, even though there might not be one.

with-heart:
I think this would actually be a breaking change. There was discussion somewhere previously (don't have time to look for it atm 😦) where we decided this functionality was intentional. I'm okay with changing the default or making it configurable, but if it changes how it works for current users, we need to consider it breaking.

segunadebayo:
Looking back at v0, this isn't a breaking change. Over here: https://v0.chakra-ui.com/formcontrol#usage, you'll see that the helperText remains even when `isInvalid` is set to `true`.

What we can do is to add a `hideHelperTextOnError` prop to handle this. We'll set this to `false` by default (for v0 behavior) and allow users to set it to `true` themselves if needed.

dodas:
I can't really see a reason for the `hideHelperTextOnError` prop.

As far as I know, users can achieve the exact same result as simply as this:
`{!isInvalid && <ChakraFormHelper>helper</ChakraFormHelper>}`

Can we bring anything extra with the prop? If no, I dislike polluting the API for such a simple (and perhaps uncommon) case.

packages/form-control/src/use-form-control.ts
segunadebayo: Yes, that make sense. Please add it

segunadebayo:
Good point @dodas! Kindly add the `hasHelpText` check and I'll merge this

avaragado:
My app, which uses @chakra/react 1.0.1 at the moment, will need modifications after this change to maintain its current behaviour. I would consider that a breaking change, and would expect to see a note in the changelog explaining what's changed.

If Chakra follows semver, then Chakra's 0.x behaviour should not affect how you label a change made to 1.0.x, _even if_ you consider the behavioural change from 0.x to 1.0.0 to be a bug (which is not the case here).

This is perhaps a trivial case because the behavioural change is small. It does illustrate the need to have a firm, documented notion of what constitutes "breaking" for a project, so that developers of Chakra and consumers of Chakra have a common understanding. Chakra consumers also consume dozens of other packages - careful use of semver is important so consumers can upgrade safely without their apps exploding.

PR #2759, opened 2020-12-11
fix(checkbox): fix themable parts list

gifaeriyanto:
## 📝 Description

Currently we able to add `container` in the parts, but this is not provided in the checkbox parts type.

Error type when adding container at the checkbox theme parts:
image

## 🚀 New behavior

No error type

### DIFF
--- a/.changeset/stale-masks-stare.md
+++ b/.changeset/stale-masks-stare.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/theme": patch
+---
+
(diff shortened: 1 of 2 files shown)


dodas:
I can't really see the `description` part being used in the component, so I think you can remove that part and add to this PR.

gifaeriyanto:
> I can't really see the `description` part being used in the component, so I think you can remove that part and add to this PR.

Sure, I've removed it.

dodas (approved):
Thanks!

PR #2775, opened 2020-12-13
fix: #2644 skeleton fading on mounting

### LINKED ISSUE #2644: Skeleton fading on mounting
# Bug report

## Describe the bug

The Skeleton component is always fading upon mounting/remounting even if the `isLoaded` prop is statically set to `true`. In my use case, I have a list that the user can reorder using `react-beautiful-dnd`. That list shows remote data that is fetched from my backend (from a parent component, the component that wraps Skeleton receives only the data directly).

After fetching the data, Skeleton fades and shows it to me (as expected, since I check if there is data in my component to show). But, if I reorder the list it "blinks", because `react-beautiful-dnd` unmount and remounts the lines after reordering.

It seems that Skeleton **always** fadeIn upon mounting. Even if it has an `isLoaded` prop set to true upon mounting.

You can see the behavior I've described above [in this video](https://imgur.com/Vrf5R6y)

## To reproduce

1. Open the [provided Codesandbox](https://codesandbox.io/s/skeleton-fade-on-remounting-ye29d)
2. Clicks on the `Toggle Skeleton` button
3. It should fade
4. Click again
5. Click again
6. It should fade again

## Minimal reproduction

https://codesandbox.io/s/skeleton-fade-on-remounting-ye29d

## Expected behavior

I would like to have a prop that could prevent that. In my opinion, we should be able to define if the `Skeleton` component should fade upon mounting even if the `isLoaded` prop was set to true.

Samic8:
Closes #2644

## 📝 Description

As described in #2644 there are cases where during the render of a skeleton component we might want to statically set isLoaded to true OR cases where the loading is already completed and therefore we want to avoid using the skeleton for a better UX (less page flickering).

#2644 did mention that a solution could be to add a flag to control this behavior. But I think that we should be opinionated about this and always prevent the use of the skeleton if the isLoaded property is already true on the first load. This will make the skeleton component simpler (by avoiding excessive flags) and nudge the developer towards better UX (less noise).

If **isLoaded** is true on the first render the fade-in animation is included

## 🚀 New behavior

If **isLoaded** is true on the first render that fade-in animation is avoided

No, if we consider the current behavior a bug.

## 📝 Additional Information

I didn't include unit tests because they would need to be _implementation_ tests. We would need to check the CSS animation property directly to determine any difference. 

Also, I'm not sure if we have a way to check CSS properties in our tests?

If anyone has a way to write unit tests or thinks we should go down the _implementation_ testing route, let me know. Thanks!

### DIFF
--- a/packages/skeleton/src/skeleton.tsx
+++ b/packages/skeleton/src/skeleton.tsx
@@ -67,14 +67,13 @@ const fade = keyframes({
 
 const useIsFirstRender = () => {
   const isFirstRender = React.useRef(true)
-  
+
   React.useEffect(() => {
     isFirstRender.current = false
   }, [])
-  
+
   return isFirstRender.current
 }
-
 
 export const Skeleton = forwardRef<SkeletonProps, "div">((props, ref) => {
   const styles = useStyleConfig("Skeleton", props)
@@ -93,7 +92,7 @@ export const Skeleton = forwardRef<SkeletonProps, "div">((props, ref) => {
   const _className = cx("chakra-skeleton", className)
 
   if (isLoaded) {
-    const animation = isFirstRender ? null : `${fade} ${fadeDuration}s`
+    const animation = isFirstRender ? "none" : `${fade} ${fadeDuration}s`
 
     return (
       <chakra.div
--- a/packages/skeleton/src/skeleton.tsx
+++ b/packages/skeleton/src/skeleton.tsx
@@ -65,8 +65,20 @@ const fade = keyframes({
   to: { opacity: 1 },
 })
 
+const useIsFirstRender = () => {
+  const isFirstRender = React.useRef(true)
+  
+  React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/skeleton/src/skeleton.tsx
```diff
@@ -65,8 +65,19 @@ const fade = keyframes({
...
 
+const useIsFirstRender = () => {
+  const isFirstRender = React.useRef(true)
+
+  React.useEffect(() => {
+    isFirstRender.current = false
+  }, [])
+
+  return isFirstRender.current
+}
```
Samic8, line 76: Should this be moved to a "utils" file? or ok here for now until we need to reuse it?

segunadebayo: We'll leave it here for now. When we need to reuse it, I'll move it.

Thanks buddy

PR #2778, opened 2020-12-14
fix: tab indicator in relative container #2040

### LINKED ISSUE #2040: TabIndicator is not calculated correctly when used wihtin a Drawer
# Bug report

## Describe the bug
The `TabIndiciator` does not calculate its position correctly when used in a component like a Drawer but also when for example resizing the window.

## Minimal reproduction

https://codesandbox.io/s/hardcore-dawn-w88t5?file=/src/App.tsx

## Expected behavior

expected-tab-indicator

## Actual behavior
actual

Samic8:
Closes #2040

## 📝 Description

As described in #2040 the tab indicator becomes misaligned when inside a position relative or fixed container. 

The tab indicators position is calculated relative to the viewport

## 🚀 New behavior

The tab container is now calculated with the assumption that the tabs root element is position relative.

No, I don't think so. But if you can think of one let me know.

## 📝 Additional Information

This fix got awfully complex, especially when having to account for potential padding in the tabs container. I'm very open to any feedback on this one and would welcome a much simpler solution. 

I haven't include testing-library tests because testing positioning logic is hard without a lot of mocking. The added storybook stories would serve the purpose of testing this functionality.

### DIFF
--- a/packages/tabs/src/tabs.tsx
+++ b/packages/tabs/src/tabs.tsx
@@ -37,7 +37,7 @@ interface TabsOptions {
 }
 
 export interface TabsProps
-  extends Omit<UseTabsProps, "rootRef">,
+  extends Omit<UseTabsProps, "rootTabsElement">,
     ThemingProps,
     Omit<HTMLChakraProps<"div">, "onChange">,
     TabsOptions {
--- a/packages/tabs/src/tabs.tsx
+++ b/packages/tabs/src/tabs.tsx
@@ -10,7 +10,7 @@ import {
   HTMLChakraProps,
   ChakraProps,
 } from "@chakra-ui/system"
-import { cx, omit, __DEV__ } from "@chakra-ui/utils"
+import { cx, mergeRefs, omit, __DEV__ } from "@chakra-ui/utils"
 import * as React from "react"
 import {
   TabsProvider,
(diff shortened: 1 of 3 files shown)


packages/tabs/src/tabs.tsx
```diff
@@ -50,22 +51,35 @@ export interface TabsProps
...
+  const rootStyles: ChakraProps["__css"] = {
+    position: "relative",
+    ...styles.root,
+  }
+
+  const captureElement = React.useCallback((node) => {
+    if (node !== null) {
+      setRootTabsElement(node)
+    }
+  }, [])
```
Samic8, line 73: This is the ugliest part of this solution. It involves storing the element reference in context so that the tab indicator hook has access to the element to do positioning calculations:
https://github.com/chakra-ui/chakra-ui/pull/2778/files#diff-a67d97c904205ad37b1bd0c751d6d590e45b819f9246e455537c1f0bd040dfa6R515

Samic8: An alternative could be to get the tabs element by doing a [closest()](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/utils/src/dom-query.ts#L131) query to get the `.chakra-tabs`. 

This is not a very _react_ like approach, but it would simplify the code.

Samic8: I made the alternative approach in 81148db8c870f8989cb4017388906f030bbd9aa6

I think it's much simpler and a better approach

segunadebayo:
Thanks for taking the time to work on this @Samic8. I will review shortly

Samic8:
Thanks @segunadebayo I'm open to any other approaches you might have.

kkatsuba:
@segunadebayo any progress here?

Issue #2789, opened 2020-12-15
Adding '2xl' as the additional breakpoint fails to generate media query.


yuzima:
I tried to add `2xl` to extend breakpoints but Chakra UI couldn't generate the related media query. 

```javascript
const breakpoints = createBreakpoints({
  sm: '20em', // 320px
  md: '30em', // 480px
  lg: '48em', // 768px
  xl: '64em', // 1024px
  '2xl': '90em', // 1440px
});
```

However, if I use `xxl` as the breakpoint point key, it works.

I checked the code and found the `2xl` is filtered by `isCustomBreakpoint` function which consider it's not a valid breakpoint key. Looks like `isCustomBreakpoint` is designed to filter the numeric breakpoint key but is there any reason to filter the key like `2xl`?

PR #2790, opened 2020-12-15
fix: use Number instead of parseInt in isCustomBreakpoint #2789

yuzima:
Currently, `isCustomBreakpoint` filters the breakpoint key like `2xl` as it using `parseInt` which parses string starting with number into Integer.  I think replace it with `Number()` can solve this problem.


with-heart:
Nice work @yuzima! I added a test case to cover this.

@allcontributors please add @yuzima for code

with-heart (approved):

PR #2794, opened 2020-12-15
fix(list): enable ref forwarding for ListItem

### LINKED ISSUE #1518: [ListItem] ListItem is not forwarding refs
**Describe the bug**

The `<ListItem />` component is not forwardings `ref`s.

**Expected Behavior** 

The `<ListItem />` component *is* forwardings `ref`s.

**Steps to Reproduce** 

See the code: https://github.com/chakra-ui/chakra-ui/blob/develop/packages/layout/src/list.tsx#L97

The `<ListItem />` is defined as:

```tsx
export const ListItem = (props: PropsOf<typeof chakra.li>) => (
  <chakra.li {...props} />
)
```

**Suggested solution(s)**
```tsx
export const ListItem = React.forwardRef(function ListItem(
  props: PropsOf<typeof chakra.li>,
  ref: React.Ref<any>,
) {
  return (
    <chakra.li ref={ref} {...props} />
  )
})
```

I think this would be it? But I might be missing something and there is a reason for this. For a workaround I used `<Box as="li" />`.

with-heart:
Closes #1518 

## 📝 Description

This PR wraps `ListItem` in `forwardRef`.

No

### DIFF
--- a/.changeset/short-fans-chew.md
+++ b/.changeset/short-fans-chew.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/layout": patch
+---
+
+Enabled ref forwarding for `ListItem`
--- a/packages/layout/src/list.tsx
+++ b/packages/layout/src/list.tsx
@@ -110,11 +110,11 @@ export interface ListItemProps extends HTMLChakraProps<"li"> {}
  *
  * Used to render a list item
  */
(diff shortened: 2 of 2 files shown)


silviuaavram:
Awesome! Thank you for fixing this!

PR #2816, opened 2020-12-17
feat: add gradient support

segunadebayo:
## Description

Add gradient support to chakra style props. This PR adds to props:

- `bgGradient`: a shorthand, convenient style prop to apply theme-aware
  gradients.
- `bgClip`: a shorthand for `background-clip` CSS attribute. Useful when
  creating text gradients.
- `backgroundClip`: the typical `background-clip` CSS attribute. Useful when
  creating text gradients.

## The Background Gradient API

To add a gradient to a component, pass the `bgGradient` prop and set its value
following the API below:

- `linear(<direction>, <from>, <to>)`
- `radial(<from>, <to>)`

and other valid css gradient properties. For linear gradients, the direction can
be either of the following values:

```js
"to-t" // 'to top'
"to-tr" // 'to top right'
"to-r" // 'to right'
"to-br" // 'to bottom right'
"to-b" // 'to bottom'
"to-bl" // 'to bottom left'
"to-l" // 'to left'
"to-tl" // 'to top left'

```

```jsx
<Box w="500px" h="200px" bgGradient="linear(to-r, gray.300, pink.200)" />
```

You can use both theme-aware color tokens or raw CSS color values.

```jsx
<Box w="500px" h="200px" bgGradient="linear(to-l, #7928CA, #FF0080)" />
```

### Multiple Color Stops

This is a gradient with multiple stops

```jsx
<Box w="500px" h="200px" bgGradient="radial(gray.300,yellow.400,pink.200)" />
```

## The Text Gradient API

To add a text gradient, pass the `bgGradient` following the API and `bgClip`
prop to `text`.

```jsx
<Text
  bgGradient="linear(to-l,#7928CA,#FF0080)"
  bgClip="text"
  fontSize="7xl"
  fontWeight="extrabold"
>
  Welcome to Chakra UI
</Text>
```


with-heart:
Going to review this today or tomorrow. It's on my todo list now. Sorry for the wait

packages/styled-system/src/utils/parse-gradient.ts
```diff
@@ -0,0 +1,72 @@
...
+    .map((color) => get(theme, `colors.${color}`, color))
+
+  if (colorsArr.length === 1) {
+    colorsArr.push(DEFAULT_TO_COLOR)
+  }
+
+  const colorsString = colorsArr.join(", ")
+
+  return `linear-gradient(${direction}, ${colorsString})`
+}
```
TimKolberger, line 69: I really like the simplicity of this, but I think it is not enough for real world use cases.
This implementation is limited to linear gradients and to a single gradient. 

Most gradients I needed to implement consisted of multiple combined gradients.

We could simplify the parsing algorithm by just replacing the color values:

```ts
export function parseGradient(value: string, theme: Dict) {
  return value.replace(/[a-zA-Z0-9_.]+/g, (substring) => {
    const color = get(theme, `colors.${substring}`, substring)
    return typeof color === "string" ? color : substring
  })
}
```

```ts
parseGradient("linear-gradient(pink.300, blue.500)", theme)
// browser -> linear-gradient(#F687B3, #3182ce)

parseGradient("radial-gradient(circle at center, red 0, blue, green 100%)", theme)
// browser -> radial-gradient(circle at center, #f00 0, #00f, #008000 100%)
```

With this we would be able to support `linear-gradient()`, `radial-gradient()`, `repeating-linear-gradient()`, `repeating-radial-gradient()` etc. And chakra users won't have to learn a new gradient syntax.

This idea came to me when playing around with the new syntax. What do you think about this approach?

packages/styled-system/src/utils/parse-gradient.ts
```diff
@@ -0,0 +1,72 @@
...
+  "to-tl": "to top left",
+}
+
+export enum ErrorMessages {
+  INVALID_SYNTAX = "Invalid value passed to prop: `bgGradient`",
+  INVALID_DIRECTION = "Direction is not valid. Defaulting to `top right`",
+}
+
+const DEFAULT_DIR: keyof typeof directionMap = "to-r"
+export const DEFAULT_TO_COLOR = "rgba(255, 0, 0, 0)"
```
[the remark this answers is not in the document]
segunadebayo: Not really. The goal is to set the default to any transparent color since the default css "transparent" value doesn't work.

packages/styled-system/tests/parse-gradient.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  const output = parseGradient(input, theme)
+  // we'll add `transparent` as the `to` color by default
+  expect(output).toEqual(
+    `linear-gradient(to top left, reddish, ${DEFAULT_TO_COLOR})`,
+  )
+})
+
+test("should not parse if value is 'none'", () => {
+  const input = "none"
+  console.log(parseGradient(input, theme))
```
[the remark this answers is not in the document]
segunadebayo: No, it's not 😀

packages/styled-system/src/utils/parse-gradient.ts
segunadebayo: I like where this is going, Tim. Thanks for taking the time.

- The constraints to linear-gradient is rather intentional considering it's a common use-case. I didn't think about supporting other types of gradients. Can you show me a real-world use-case based on your project?
- This doesn't mean the user cannot use the typical `backgroundImage` prop to add more complex gradients. I'd rather say for simple linear gradients, you can use this prop, else use the `backgroundImage` or `background` prop
- I'd love to keep your simplification though coz it reduces the LOC.

Let me think on this a bit more

wvieed (approved):

wvieed:
ا

segunadebayo:
@TimKolberger Can you review the changes?

I've updated the API to support other css gradient types and still preserve some parts of the previous API.

packages/styled-system/src/utils/parse-gradient.ts
```diff
@@ -0,0 +1,70 @@
...
+  "revert",
+  "unset",
+]
+
+const trimSpace = (str: string) => str.trim()
+
+export function parseGradient(value: string, theme: Dict) {
+  if (!value || globals.includes(value)) return value
+
+  const regex = /(?<type>^[a-z-A-Z]+)\((?<values>(.*?))\)/g
```
TimKolberger, line 31: IE11 does not support named regex groups, but I am not sure if we should keep supporting / restore support for it.

packages/styled-system/tests/parse-gradient.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    // we clean up the extra space
+    expect(output).toEqual("linear-gradient(to top left, #800000, #006400)")
+  })
+
+  test("should not parse if value is 'none'", () => {
+    expect(parseGradient("none", theme)).toEqual("none")
+  })
+
+  test("should not parse null", () => {
+    expect(parseGradient(null, theme)).toBeNull()
```
TimKolberger, line 46: This test is currently failing.

TimKolberger (commented):
Feels better to support all gradient types :tada:
Personally I have a plain CSS background and I am not big fan of widening the API surface, but I can see the benefit of a simplified API for gradients for users coming from tailwind.

I collected some gradients from my personal projects [here](https://codesandbox.io/s/chakra-ui-gradient-k4dhm?file=/src/App.tsx)

```ts
{
  shinyGradient: "linear-gradient(167deg, #f5cde2 32%, #BCFAD2)",
  scrollOverflowGradient: "linear-gradient(90deg, #EDF2F7 4px, transparent 12px), linear-gradient(-90deg, #EDF2F7 4px, transparent 12px), linear-gradient(-90deg, rgba(0,0,0,0.2) 12px, transparent 24px), linear-gradient(90deg, rgba(0,0,0,0.2) 12px, transparent 24px)",
  ball: "radial-gradient(circle at 50%, #4341fff0, #24252cf0), linear-gradient(#000, #fff)"
}
```

The ones not supported by this API are these backgrounds with combined gradients. But as you mentioned the current API should be fine for most use cases 👍 

I would love to see `useTokenTemplate` hook which replaces all e.g. color tokens in a string, as proposed in my previous comment. This is out of scope for this PR.

LGTM when the failing test is fixed 👍

with-heart (approved):

PR #2834, opened 2020-12-19
feat(extend-theme): allow passing a base theme to customize

### LINKED ISSUE #2769: Specify a custom base theme in extendTheme to extend any theme
# 🚀 Feature request

`extendTheme` allows customizing the default theme in a way that avoids spreading existing properties, which is great. 

However, there are cases where subtle theme variations are required, and having the ability to tell `extendTheme` what base  theme to work on would allow transitive theming with low boilerplate.

My use-case is the following: my website has a main theme that uses the system setting for Color Mode and the system font stack, but I also use the same React tree to generate OpenGraph images, in Dark mode and with specific fonts.

Also, I use that theme across multiple websites, so I have moved most of the common overrides that define my design system to a separate package, and at the moment I find myself spreading those overrides to apply onto Chakra's default theme.

## ✅ Proposed solution or API

```diff
- extendTheme(overrides)
+ extendTheme(overrides, baseTheme = defaultTheme)
```

## ↩️ Alternatives

- Spreading overrides (as I was spreading the theme itself in v0.x)
- Using the `mergeWith` utility, but without access to the custom inline customizer

## 📝 Additional Information

I will update this description with a link to the source when my changes are published.

TimKolberger:
Closes #2769

## 📝 Description

```diff
- extendTheme(overrides)
+ extendTheme(overrides, baseTheme = defaultTheme)
```

Allows extending an already extended theme. This is helps maintaining inherited design system.

Not possible without a copy of the extendTheme function.

## 🚀 New behavior

```ts
const theme1 = extendTheme(overrides)
const theme2 = extendTheme(overrides, theme1)
```

No

### DIFF
--- a/.changeset/clean-balloons-pull.md
+++ b/.changeset/clean-balloons-pull.md
@@ -2,4 +2,15 @@
 "@chakra-ui/react": minor
 ---
 
-extendTheme: allow passing a base theme to customize
+extendTheme: added an optional second argument `baseTheme` to customize:
+
+```jsx
+const theme = extendTheme({
+  // theme overrides
+  colors: { red: { 500: "#ff0000" } },
+  // the base theme to customize with the above overrides
+  yourTheme,
+})
+```
+
+If no `baseTheme` is provided, defaults to the Chakra theme.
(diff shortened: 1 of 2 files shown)


with-heart:
Looks great! I expanded the changeset a bit with a commented code sample, just to make it a little more clear what this feature means.

with-heart (approved):

PR #2841, opened 2020-12-19
docs: add example of how to use custom colorScheme

HermanNygaard:
I think it would be nice to also show how you can use a custom theme color for the `colorScheme` prop 😄


HermanNygaard:
@segunadebayo could you add @allcontributors for this :D ?

HermanNygaard:
@allcontributors please add @HermanNygaard for docs

PR #2876, opened 2020-12-23
refactor: deprecate defaultIsChecked props

with-heart:
## 📝 Description

This PR deprecates the `defaultIsChecked` prop in favor of a new `defaultChecked` prop. I think `defaultChecked` makes more sense as it mirrors the default prop name that React uses for `radio` and `checkbox` elements, which will make this easier to pick up. It's also confusing because `Radio` and `Checkbox` have both `defaultChecked` and `defaultIsChecked` as part of their types due to reliance on the underlying `HTMLInputElement` type, so removing this extra prop will make that even more clear.

Uncontrolled `Radio` and `Checkbox` components use `defaultIsChecked` for setting the default checked element.

## 🚀 New behavior

`Radio` and `Checkbox` now use the `defaultChecked` prop to set the default checked element. 

No. This is not a breaking change because `defaultChecked` defaults to the value of `defaultIsChecked` if not provided, meaning that `defaultIsChecked` will still work for both components. We should remove this in a future `major` bump.

### DIFF
--- a/packages/checkbox/src/use-checkbox.ts
+++ b/packages/checkbox/src/use-checkbox.ts
@@ -8,6 +8,7 @@ import {
   dataAttr,
   mergeRefs,
   PropGetter,
+  warn,
 } from "@chakra-ui/utils"
 import { visuallyHiddenStyle } from "@chakra-ui/visually-hidden"
 import React, {
@@ -55,8 +56,14 @@ export interface UseCheckboxProps {
   isRequired?: boolean
   /**
    * If `true`, the checkbox will be initially checked.
+   * @deprecated Please use the `defaultChecked` prop, which mirrors default
+   * React checkbox behavior.
    */
   defaultIsChecked?: boolean
+  /**
+   * If `true`, the checkbox will be initially checked.
+   */
+  defaultChecked?: boolean
   /**
    * The callback invoked when the checked state of the `Checkbox` changes..
    */
@@ -86,6 +93,7 @@ export interface UseCheckboxProps {
 export function useCheckbox(props: UseCheckboxProps = {}) {
   const {
     defaultIsChecked,
+    defaultChecked = defaultIsChecked,
     isChecked: checkedProp,
     isFocusable,
     isDisabled,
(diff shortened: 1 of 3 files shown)


packages/checkbox/src/use-checkbox.ts
```diff
@@ -106,7 +113,7 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
 
   const ref = useRef<HTMLInputElement>(null)
 
-  const [checkedState, setCheckedState] = useState(!!defaultIsChecked)
+  const [checkedState, setCheckedState] = useState(!!defaultChecked)
```
Zaynex, line 116: Would it be better to remind users ?
```suggestion
 if(props.hasOwnproperty('defaultIsChecked')) {
   console.warn('defaultIsChecked is deprecated! Please use the `defaultChecked` prop, which mirrors default')
 }
  const [checkedState, setCheckedState] = useState(!!defaultChecked)
```

packages/radio/src/use-radio.ts
```diff
@@ -94,7 +102,7 @@ export function useRadio(props: UseRadioProps = {}) {
 
   const ref = useRef<HTMLInputElement>(null)
 
-  const [isCheckedState, setChecked] = useState(Boolean(defaultIsChecked))
+  const [isCheckedState, setChecked] = useState(Boolean(defaultChecked))
```
Zaynex, line 106: Same as https://github.com/chakra-ui/chakra-ui/pull/2876/files#r549067666

wvieed (approved):

packages/checkbox/src/use-checkbox.ts
with-heart: Great idea! Added a warning to both `useCheckbox` and `useRadio` using our `warn` utility: https://github.com/chakra-ui/chakra-ui/pull/2876/files#diff-445514188dc73f74a48ca922881ed7816acdcf8acca6e0bf2fb39c2e4ab340bfR124-R130

Zaynex (approved):

TimKolberger (approved):
💖

PR #2891, opened 2020-12-27
docs: fix typo on Responsive Styles page

dananw:
Closes # 

## 📝 Description

Let's say you have a Box that looks like this:
but the example is use Text Component.

This is Typo or on purpose?

> Let's say you have a Box that looks like this:

## 🚀 New behavior

> Let's say you have a Text that looks like this:

No


with-heart (approved):

with-heart:
Good catch @dananw!

@allcontributors please add @dananw for docs

PR #2895, opened 2020-12-27
fix(slider): call onChangeEnd even if value doesn't change

### LINKED ISSUE #2887: slider onChangeEnd does not get called when user sets value to minimum
# 🐛 Bug report

onChangeEnd does not get called when user tries to set the value to minimum. slider is in controlled state and default value is 0, minimum value is 10_000. 

## example

```ts
            <Slider
                min={10_000}
                max={50_000}
                step={10_000}
                mx={[4, null, 0]}
                w={["90%", null, "full"]}
                onChangeStart={() => setDisplaySliderTooltip(true)}
                onChangeEnd={val => {
                    setDisplaySliderTooltip(false);
                    if (displaySliderTooltip) {
                        handleChange(val);
                    }
                }}
                onChange={value => {
                    setSliderLiveValue(value);
                }}
                value={sliderLiveValue}
                mt="12"
                dir="ltr"
            >
                <SliderTrack>
                    <SliderFilledTrack />
                </SliderTrack>
                <Tooltip
                    isOpen={displaySliderTooltip}
                    hasArrow
                    label={`${thousandSeparator(sliderLiveValue)} tt`}
                    bg="gray.800"
                    placement="bottom"
                    color="white"
                >
                    <SliderThumb />
                </Tooltip>
            </Slider>

```
[live example ](https://codesandbox.io/s/elated-rgb-ic3lt)
## 🧐 Expected behavior

> value should be updated to minimum like every other value

with-heart:
Closes #2887 

## 📝 Description

`onChangeEnd` can't reliably be used to detect when dragging ends because it isn't called if the new `value` is the same as the old `value`.

`Slider` currently has a check that prevents calling `onChangeEnd` if `value` matches the previous `value`.

## 🚀 New behavior

`onChangeEnd` is now called whether `value` matches the previous `value` or not.

Possibly? Need to discuss. If users don't expect `onChangeEnd` to be called, it will now be called. Created as a draft so we can discuss this before merging.

### DIFF
--- a/.changeset/perfect-melons-serve.md
+++ b/.changeset/perfect-melons-serve.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/slider": patch
+---
+
+fix(slider): call onChangeEnd even if value doesn't change
--- a/packages/slider/src/use-slider.ts
+++ b/packages/slider/src/use-slider.ts
@@ -184,7 +184,6 @@ export function useSlider(props: UseSliderProps) {
    * or greater than max
    */
   const value = clampValue(computedValue, min, max)
-  const prev = useRef<number>()
 
   const reversedValue = max - value + min
   const trackValue = isReversed ? reversedValue : value
@@ -246,7 +245,6 @@ export function useSlider(props: UseSliderProps) {
     (value: number) => {
       // bail out if slider isn't interactive
       if (!isInteractive) return
-      prev.current = value
       value = parseFloat(roundValueToStep(value, min, oneStep))
       value = clampValue(value, min, max)
       setValue(value)
@@ -390,8 +388,7 @@ export function useSlider(props: UseSliderProps) {
   }, [value])
 
   useUpdateEffect(() => {
-    const shouldUpdate =
-      !isDragging && eventSource !== "keyboard" && prev.current !== value
+    const shouldUpdate = !isDragging && eventSource !== "keyboard"
 
     if (shouldUpdate) {
       onChangeEnd?.(value)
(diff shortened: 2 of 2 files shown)


with-heart:
I also removed the use of `prev` throughout `useSlider`, as it no longer serves a purpose. I'm not sure what we were originally going for with the use of that value and preventing `onChangeEnd` from being called, so I could use some input on that.

I also noticed that `onChangeEnd` is called whenever the value changes even if `eventSource` is `"keyboard"`, which means that it's called anytime the keyboard is used to change the value, but `onChangeStart` is never called for `"keyboard"` which seems inconsistent. Maybe we should discuss how to make that more consistent, and whether `onChangeEnd` should even be called for keyboard events?

TimKolberger:
IMO we should remove the check for `eventSource !== 'keyboard'`.
As user I use the `onChangeEnd` prop to determine the latest value of the slider if I don't care about live updates (while dragging === using `onChange`). It does not matter with which input device it was changed. 

Currently the user could miss some change events if she/he only uses `onChangeEnd`. With removing the check we could fix an unrecognised bug in userland applications.

Missing events in `onChangeStart` when using the keyboard is valid, if `onChangeEnd` is called every time the value was changed via keyboard. The intention of this callback (of which I can think of) is to start an animation or similar while changing the value. This is not necessary while changing the value with the keyboard.

with-heart:
@segunadebayo Thoughts on Tim's comments and this PR?

segunadebayo:
@with-heart, please add a changeset for this

with-heart:
@segunadebayo Whoops! Thanks for adding that for me 😬

PR #2900, opened 2020-12-28
docs(guide): add react-table integration guide

TimKolberger:
## 📝 Description

Add integration guide with react-table

### DIFF
--- a/website/pages/guides/integrations/with-react-table.mdx
+++ b/website/pages/guides/integrations/with-react-table.mdx
@@ -1,0 +1,110 @@
+---
+title: Chakra UI + React Table
+description: Building a Data Table with Chakra UI and React Table
+tags: ["table", "data table", "react-table"]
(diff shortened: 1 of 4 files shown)


website/pages/guides/integrations/with-react-table.mdx
```diff
@@ -0,0 +1,110 @@
...
+    [],
+  )
+
+  const {
+    getTableProps,
+    getTableBodyProps,
+    headerGroups,
+    rows,
+    prepareRow,
+  } = useTable({ columns, data }, useSortBy)
```
jmiazga, line 61: Do you want to include the import for `useTable` so its apparent where its coming from?

TimKolberger: Good idea 💡  Added an "Imports" section.

jmiazga (approved):

PR #2911, opened 2020-12-29
docs(use-token): replace gradient example with boxShadow example

dodas:
## 📝 Description

Since we now have gradients supported "natively" in Chakra, it might be better to not show how to use `useToken` to make one. 

## ⛳️ Current behavior

`useToken` docs show how to use it to make a background gradient.

## 🚀 New behavior

Docs show how to use it to make a `boxShadow` with colors from theme.

No.


TimKolberger (approved):
✨

PR #2927, opened 2021-01-02
fix document using <input type="phone">

walf443:
## 📝 Description

> Add a brief description

`<input type="phone">` is not exist on specification. and `<input type="tel">` should be corrrect.

SEE https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

Code example is not good

## 🚀 New behavior

Code example is valid.

No


TimKolberger (approved):
Thank you for this PR @walf443 ✨

segunadebayo:
Good catch @walf443!

PR #2930, opened 2021-01-02
fix(deprecation): deprecate isFullWidth on Input, NumberInput, Radio,…

### LINKED ISSUE #2865: "React does not recognise the `isFullWidth` prop on a DOM element" on Select component
# 🐛 Bug report

When passing the isFullWidth prop to the Select component, the width of the component is not changed and this warning from React is logged to the terminal. 
```
Warning: React does not recognize the `isFullWidth` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isfullwidth` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```
## 💥 Steps to reproduce
Add this component to a chakra project (I used the create-react-app template)
```react
<Select isFullWidth={false}>
    <option value="Hello">
        Hello
    </option>
    <option value="World">
        World
    </option>
</Select>
```

## 💻 Link to reproduction

[CodeSandbox](https://codesandbox.io/s/chakra-react-doesnt-recognise-c1v3m?file=/src/index.tsx)

## 🧐 Expected behavior

The Select component's width should be reduced to just be enough to contain the options.

## 🌍 System information

| Chakra UI        |        1.0.3    |
| Operating System |      MacOs      |

## 📝 Additional information

This worked about a month ago but doesn't work anymore. I believe the problem also exists for the `rightIcon` prop of the `MenuButton` component.

TimKolberger:
… Select and Textarea

Closes #2865 

## 📝 Description

`isFullWidth` is present in the type definitions of Input, NumberInput, Radio, Select and Textarea - but unused as they are all defaulting to a 100% width.

I marked this prop as deprecated and we can remove them with the next major release.

Select passes `isFullWidth` to the DOM element.

## 🚀 New behavior

Select no longer passes `isFullWidth` to the DOM - and is marked as deprecated, because it has no effect.

No, but we should remove `isFullWidth` on those mentioned components with the next major version.

### DIFF
--- a/packages/input/src/input.tsx
+++ b/packages/input/src/input.tsx
@@ -25,6 +25,10 @@ interface InputOptions {
   errorBorderColor?: string
   /**
    * If `true`, the input element will span the full width of its parent
+   *
+   * @deprecated
+   * This component defaults to 100% width,
+   *  please use the props `maxWidth` or `width` to configure
    */
   isFullWidth?: boolean
 }
--- a/packages/number-input/src/number-input.tsx
+++ b/packages/number-input/src/number-input.tsx
@@ -46,6 +46,10 @@ interface InputOptions {
   errorBorderColor?: string
   /**
    * If `true`, the input element will span the full width of its parent
+   *
+   * @deprecated
+   * This component defaults to 100% width,
+   * please use the props `maxWidth` or `width` to configure
    */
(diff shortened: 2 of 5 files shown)


ugcgt:
[@]()

PR #2964, opened 2021-01-06
docs(style props): add note regarding the order of border style props

maxbeier:
## 📝 Description

I had a bug due to the order of (border) style props and it took me some time to find it. So I thought this note might help others to not repeat the mistake.

## ⛳️ Current behaviour (updates)

Having border style props in different orders produces different visual results, which might be unintended.

## 🚀 New behaviour

Added a note to inform users about this behaviour.

No

### DIFF
--- a/website/pages/docs/features/style-props.mdx
+++ b/website/pages/docs/features/style-props.mdx
@@ -274,6 +274,9 @@ import { Box } from "@chakra-ui/react"
 </Box>
 ```
 
+> **Note**: The order of properties matter. So while the above example works,
+> `<Box borderColor="gray.200" border="1px">` does not.
+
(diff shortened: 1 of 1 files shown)


dodas:
This is a CSS feature.
Please read this article: https://css-tricks.com/accidental-css-resets/

In this example you posted, you can use
`<Box borderColor="gray.200" borderWidth="1px">`
instead of 
`<Box borderColor="gray.200" border="1px">`

segunadebayo:
I agree with @dodas that this a CSS feature. I understand that it's easy to neglect them.

Let's keep this for now, and if it becomes a recurring problem, we'll add this note to the docs.

Thanks

PR #2965, opened 2021-01-06
fix(guides): unsued box import

chr-ge:
## 📝 Description

- Used the imported `Box` instead of `chakra.div`
- added `forwardRef` import

The framer example is not using the imported `Box`

## 🚀 New behavior

Easier to understand the framer example.

No.


TimKolberger (approved):
Good catch @chr-ge

PR #2993, opened 2021-01-08
fix(system): add "main" to domElements list

dodas:
## 📝 Description

add "main" to domElements list

No


segunadebayo:
Hi @dodas, as I mentioned on Discord I think we can do without this one.

Users can always do this:

```jsx
const Main = chakra("main")
```

Thanks

Issue #2730, opened 2020-12-07
Container no longer uses sizes.container values


prayogoa:
# 🐛 Bug report

Since https://github.com/chakra-ui/chakra-ui/pull/2602 the `Container` component no longer uses size values defined in `sizes.container`, but instead uses `sizes.largeSizes`. This is not mentioned in the PR, and I couldn't find any other uses of `sizes.container` in the codebase so either `sizes.container` is deprecated without documentation or this change is unintended.

## 💥 Steps to reproduce
As can be seen from the codesandbox, specifying `maxW="lg"` results in a max width value of `512px = (sizes.largeSizes.ls)` instead of `1024px (sizes.containers.lg)`

## 💻 Link to reproduction
https://codesandbox.io/s/white-moon-7sxnh?file=/src/App.tsx

## 🧐 Expected behavior

Container should use size values defined in `sizes.container`

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Chakra UI        |      1.0.2      |
| Browser          |      N/A      |
| Operating System |     N/A       |

## 📝 Additional information

![Screen Shot 2020-12-07 at 11 34 45 PM](https://user-images.githubusercontent.com/1696217/101370860-0061ae80-38e5-11eb-9c36-40180f4481f1.png)
Offending PR https://github.com/chakra-ui/chakra-ui/pull/2602
[sizes.container](https://github.com/chakra-ui/chakra-ui/blob/3fbe55c5a55d8c4697759a06ad4d492df41003e4/packages/theme/src/foundations/sizes.ts#L23)

prayogoa:
This might be an intended change, since you can still do `<Container maxW="container.lg" />` to achieve the same effect. This does make the behaviour less surprising, so imo I don't mind this change as long as it's reflected in the docs / changelog.

with-heart:
Whoops! Looks like that PR removed some functionality that was built in to `Container` previously.

cc @chakra-ui/react-team: How do we want to handle this? I'm leaning towards reverting it and coming up with a better plan for adding theming to `Container`, as this ended up being an unintentional breaking change.

jmiazga:
Do you think we can follow up quickly with a fix? If not reverting might be the better way to go.

TimKolberger:
Reverting would introduce another breaking change due the removal of the possibility to style the container. 
IMO a bugfix is the best way to go.

To resolve this, we have to map the width props with this removed function:

[...]

PR #2997, opened 2021-01-09
fix(container): restore former width prop behaviour

trevorblades:
This branch restores the former behaviour of width props (`w`, `maxW`, `minWidth`, etc.) in the `<Container />` component. I took the approach used previously in this component from [`770466a`](https://github.com/chakra-ui/chakra-ui/blob/770466ae99443d12c214e26032a5ebc704397853/packages/layout/src/container.tsx), and combined it with the current implementation.

Fixes: #2730

### DIFF
--- a/packages/layout/src/container.tsx
+++ b/packages/layout/src/container.tsx
@@ -48,6 +48,7 @@ export const Container = forwardRef<ContainerProps, "div">((props, ref) => {
   } = omitThemingProps(props)
 
   const theme = useTheme()
+
   const widthProps = transform(theme, {
     maxW,
     maxWidth,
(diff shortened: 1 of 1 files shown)


TimKolberger:
@trevorblades Thank you for this PR! 💖 

Could you kindly rebase onto develop that we can see if the build is working?

trevorblades:
@timkindberg no problem! Just rebased on develop. Let's see if that works 😄

TimKolberger:
Hi @trevorblades,

Thank you for your time and this PR! 💖 

We thought about the Container implementation and concurred that we will keep the current behaviour. 
This will simplify the introduction of a stricter type system and improve the autocomplete values.

The width of the Container will use the default scale from the theme. The user can pass one of the container widths e.g. `w="container.lg"` to use the bigger scale.

We will update the documentation to reflect those changes.

Cheers 🍻

trevorblades:
@timkindberg ok, that works for me. Thanks for your consideration!

PR #2999, opened 2021-01-09
docs: Update props table to display valid size and variant props values

### LINKED ISSUE #2984: docs: Improve PropsTable values for `size` and `variant` props
# 🚀 Feature request

All valid values for the props `size` and `variant` should be displayed in the Props table for each component.

E.g. the `Input` component has prop definition `size: string`, but with the default theme it is `size: xs | sm | md | xl`.

To save time I do not want to look up all valid values in the source code of the default theme.

## ✅ Proposed solution or API

The component [PropsTable](https://github.com/chakra-ui/chakra-ui/blob/develop/website/src/components/props-table.tsx) is responsible to render all valid props.

It accepts a prop `of` which is the component name. We can determine all valid values by inspecting the default theme at render.

```pseudo
// pseudo-code
import theme from chakra/theme

of // name of component
themeComponent = theme.components[of]

variantValues = Object.keys(themeComponent.variants)
sizeValues = Object.keys(themeComponent.sizes)

override the rendered values for prop `variant` and `size` with those values
```

## ↩️ Alternatives

Maintain the values manually - I am not a big fan of this

## 📝 Additional Information

image

Related issue: #2983

tioluwani94:
Closes #2984 

## 📝 Description

Updated the `PropsTable` component to display valid values for the `size` and `variant` props for each component on the docs.

All valid values for the `size` and `variant` props are not being displayed in the props table for each component.

image

## 🚀 New behavior

Display all valid values for the `size` and `variant` props in the props table for each component.

Screenshot 2021-01-09 at 15 29 51

No

### DIFF
--- a/website/src/components/props-table.tsx
+++ b/website/src/components/props-table.tsx
@@ -1,6 +1,7 @@
 import * as React from "react"
 import * as ComponentProps from "@chakra-ui/props-docs"
 import MDXComponents from "./mdx-components"
+import { theme } from "@chakra-ui/react"
 
 export type PropsTableProps = {
   /**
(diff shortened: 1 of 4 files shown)


website/src/components/props-table.tsx
```diff
@@ -35,6 +36,27 @@ const PropsTable = ({
...
+  const sizeValues = themeComponent?.sizes && Object.keys(themeComponent.sizes)
+  const variantValues =
+    themeComponent?.variants && Object.keys(themeComponent.variants)
+
+  /**
+   * If component has size prop, override the rendered values
+   * for `size` prop with  the component's size values
+   */
+  if (info.props.size && sizeValues) {
+    info.props.size.type.name = sizeValues
```
TimKolberger, line 49: When overriding the value with an array, we need to introduce the conditional rendering below (L89+) with `renderPropType`. What do you think about using 

```suggestion
    info.props.size.type.name = sizeValues.join(', ')
```

to concat all array values to a string?

TimKolberger:
First of all: Thank you @tioluwani94 ✨ ! 💖

website/src/components/props-table.tsx
tioluwani94: This makes sense and is cleaner. I'll update the PR to use this. Thanks @TimKolberger

segunadebayo:
Thanks for adding this Kola @tioluwani94  💖

PR #3005, opened 2021-01-10
fix(docs): format size and variant values as TS types

dodas:
## 📝 Description

Currently some props are formatted as TS types and some as loose human-readable strings. 
I updated the `size` and `variant` values formatting to be a valid TS string literal union type (same as `orientation` for example). 

For example, before:
`ghost, outline, solid, link, unstyled`

After:
`"ghost" | "outline" | "solid" | "link" | "unstyled"`

No.

### DIFF
--- a/website/src/components/props-table.tsx
+++ b/website/src/components/props-table.tsx
@@ -42,19 +42,23 @@ const PropsTable = ({
     themeComponent?.variants && Object.keys(themeComponent.variants)
 
   /**
-   * If component has size prop, override the rendered values
-   * for `size` prop with  the component's size values
+   * If component has size prop, override the rendered value
+   * for `size` prop with the component's size values formatted as TS type.
    */
   if (info.props.size && sizeValues) {
-    info.props.size.type.name = sizeValues.join(", ")
+    info.props.size.type.name = sizeValues
+      .map((size) => `"${size}"`)
+      .join(" | ")
   }
 
   /**
    * If component has variant prop, override the rendered value
-   * for `variant` prop with the component's variant values
+   * for `variant` prop with the component's variant values formatted as TS type.
    */
   if (info.props.variant && variantValues) {
-    info.props.variant.type.name = variantValues.join(", ")
+    info.props.variant.type.name = variantValues
+      .map((variant) => `"${variant}"`)
+      .join(" | ")
   }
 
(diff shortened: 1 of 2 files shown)


website/src/components/props-table.tsx
```diff
@@ -42,19 +42,23 @@ const PropsTable = ({
...
 
   /**
-   * If component has size prop, override the rendered values
-   * for `size` prop with  the component's size values
+   * If component has size prop, override the rendered value
+   * for `size` prop with the component's size values formatted as TS type.
    */
   if (info.props.size && sizeValues) {
-    info.props.size.type.name = sizeValues.join(", ")
+    info.props.size.type.name = sizeValues
```
tioluwani94, line 49: Hi @dodas, thanks for this, any reason why you are using the `.map` before the `.join`? What do you think about using this`:

```suggestion
    info.props.size.type.name = sizeValues.join(' | ')
```

dodas: Hi @tioluwani94!

Without the `.map`, you would achieve following:
`ghost | outline | solid | link | unstyled`

instead of the following:
`"ghost" | "outline" | "solid" | "link" | "unstyled"`

I am trying to achieve a valid TS type here (same as `"horizontal" | "vertical"` in orientation).

tioluwani94: This makes sense. Thanks again @dodas

TimKolberger (approved):
Love seeing this getting more and more refined. 
The changes looks good to me!

One thing we could add to this feature is to remove the props from those components, where no variant or size is present in the theme. E.g. Link.
This does not have to be part of this PR, but I think this would be valuable as well.

dodas:
I absolutely agree. I will make a new PR with those changes.

with-heart (approved):

dodas:
Do we actually want to _omit_ `variant` and `size` where it is not implemented in chakra's theme? User can still extend theme and implement these, so it kinda makes sense to have them listed in component.

Shouldn't we instead add a note to Description of those props, that they are not implemented in chakra's theme, but you can do it yourself? Perhaps with a link where you can find how to do it.

with-heart:
I think that's preferable. We want to try to put information as close to where a user would look for it, so I think having it right in the description would be literally the best choice.

Issue #3009, opened 2021-01-11
TypeError: (0 , _utils.compose) is not a function


lednhatkhanh:
Package version:

```
"@chakra-ui/icons": "1.0.2",
"@chakra-ui/react": "1.1.0",
```

It's happening with a Next.js project when running `next dev`

lednhatkhanh:
I checked again and this only happens when using `react@16` in a next.js project

anubra266:
Nah, I use react 17, in a Create react app, same issue

poizoncc:
Same
```
[!] Error: 'compose' is not exported by ../../node_modules/@chakra-ui/utils/dist/esm/index.js, imported by ../../node_modules/@chakra-ui/styled-system/dist/esm/utils/logical-prop.js
```

objmj-itminds:
Updated from @chakra-ui/react 1.1.2 to 1.1.3 as the only dependency update also brings forth this error

I have a pipeline [here](https://github.com/IT-Minds/client_comby_oilcontrols/pull/12/checks?check_run_id=1680641017) should it help. The main takeaway is this stack trace.

```
 TypeError: (0 , _utils.compose) is not a function

    > 1 | import {
        | ^
      2 |   Button,
      3 |   Container,
      4 |   Heading,

      at logicalTransform (node_modules/@chakra-ui/styled-system/src/utils/logical-prop.ts:27:22)
      at Object.logical (node_modules/@chakra-ui/styled-system/src/utils/logical-prop.ts:43:16)
      at Object.<anonymous> (node_modules/@chakra-ui/styled-system/src/config/position.ts:19:23)
      at Object.<anonymous> (node_modules/@chakra-ui/styled-system/src/config/index.ts:8:1)
      at Object.<anonymous> (node_modules/@chakra-ui/styled-system/src/index.ts:1:1)
      at Object.<anonymous> (node_modules/@chakra-ui/system/src/index.ts:2:1)
      at Object.<anonymous> (node_modules/@chakra-ui/react/src/chakra-provider.tsx:3:1)
      at Object.<anonymous> (node_modules/@chakra-ui/react/src/index.tsx:1:1)
      at Object.<anonymous> (src/components/Demo/Demo.tsx:1:1)
      at Object.<anonymous> (src/components/Demo/Demo.test.tsx:8:1)
  ```

MichaelGrigoryan25:
Having the same issue after upgrading to 1.1.3

with-heart:
[...]

PR #3015, opened 2021-01-11
fix: un-break broken 1.1.3 release

### LINKED ISSUE #3006: 'useCallbackRef' is not exported from '@chakra-ui/hooks'
[Discord Link](https://discord.com/channels/660863154703695893/758645914868514827/797919545394331658)
 On a fresh install with CRA, the following compiler error appears 
 
 ```
 ./node_modules/@chakra-ui/portal/dist/esm/portal.js
Attempted import error: 'useCallbackRef' is not exported from '@chakra-ui/hooks'.
```

with-heart:
Fixes #3006, #3009

## 📝 Description

This PR adds changesets for changes that were intended to be part of the `1.1.3` release. Since those changesets were missing, new versions of `@chakra-ui/hooks` and `@chakra-ui/utils` weren't published which broke some things in other packages that rely on those new versions.

I also re-added `useLatestRef`, `useEventCallback`, and `useMouseDownRef` and tagged them as `@deprecated`. Since we're not publishing a new `major` version, removing part of our public API would be a `semver` violation. We should hold off on removing them until the next `major` version.


with-heart:
@cschroeter also reported additional issues relating to #3006 but I haven't had a chance to touch base with him yet to find out what those are.

PR #3022, opened 2021-01-12
fix(menu): update popper position only while open

### LINKED ISSUE #2531: Performance issue using <Menu /> when scrolling
# Bug report

## Describe the bug

When closed, `<MenuList/>` re-calculates its position on scroll. This greatly impacts performance when lots of `<Menu />` components are rendered at once. This scenario is common when rendering tables/lists where each item has associated options. The `isLazy` prop, [mentioned here](https://github.com/chakra-ui/chakra-ui/issues/2415), helps but doesn't solve the issue.

## Minimal reproduction

1. https://codesandbox.io/s/chakra-uireact100-menu-perf-dejjg?file=/src/index.tsx
1. enable "Highlight updates when components render." in React DevToolsScreen Shot 2020-11-16 at 10 02 35
1. scroll right pane

## Expected behavior

When scrolling, `<MenuList />` should only re-calculate its position if it is open.

## Screenshots

Screen Shot 2020-11-16 at 10 08 18

## System information

- OS: macOS
- Browser (if applies): Chrome, Safari
- Version of @chakra-ui/react: 1.0.0, 1.0.1

dodas:
Closes #2531

## 📝 Description

I added a `enabled` option to `usePopper`. While this option is `false`, the attrs and styles won't be updated.
`Menu` uses this option, so its position won't update while it is not open.

### DIFF
--- a/packages/popper/src/react-popper.ts
+++ b/packages/popper/src/react-popper.ts
@@ -11,6 +11,7 @@ import { dequal } from "dequal"
 import * as React from "react"
 
 type Options = Partial<PopperOptions> & {
+  enabled?: boolean
   createPopper?: typeof defaultCreatePopper
 }
 
@@ -44,6 +45,8 @@ export function usePopper(
   popperElement: HTMLElement,
   options: Options = {},
 ) {
+  const { enabled = true } = options
+
   const prevOptions = React.useRef<any>(null)
 
   const optionsWithDefaults = {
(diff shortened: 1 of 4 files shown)


packages/popper/src/react-popper.ts
```diff
@@ -70,12 +71,15 @@ export function usePopper(
       phase: "write",
       fn: ({ state }) => {
         const elements = Object.keys(state.elements)
```
ljosberinn, line 73: nit: can be declared only within the `if` then, otherwise its unused?

dodas:
Thanks for reviews! I agree with both and I force-pushed the updates. Cheers!

packages/popper/src/react-popper.ts
```diff
@@ -69,13 +72,16 @@ export function usePopper(
       enabled: true,
```
segunadebayo, line 72: How about passing the `enabled` option here? instead of adding the `if` block below

dodas: Ha! I didn't realized that's an option. Great suggestion, thanks!

I updated the PR.

PR #3025, opened 2021-01-12
chore(docs): split installation codeblocks to allow easier copying

sampoder:
## 📝 Description

When I wanted to copy the commands, I might try to use the copy button but if I do it won't work :/

This PR slightly reworks that section to enable this flow

No


with-heart (approved):

with-heart:
Thanks @sampoder!

@allcontributors please add @sampoder for docs

sampoder:
amazing! glad I could help!

PR #2918, opened 2020-12-30
fix(rtl-feature): apply fixes for several components on RTL

MohamedSayed008:
Closes # 

## ⛳️ Current behavior (updates)
 - stack spacing for right and left isn't working as expected.
 - borderStartStyle , borderEndStyle, borderStartColor, borderEndColor doesn't exist in props.
 - Close button for tag , modal , drawer doesn't flip with direction.
 - icon spacing for list isn't RTL friendly

## 🚀 New behavior

> Fix alignment for Stack on RTL
> Add Missing borderStart borderEnd   for Style  and Color
> Close button alignment for tag , modal , drawer is RTL friendly now
> Icon alignment for list
>sort Object.assign keys in border.ts for better readability.

## 💣 Is this a breaking change (Yes/No):
 No

PR #3048, opened 2021-01-13
fix(slide): iOS Safari overlays DrawerFooter

### LINKED ISSUE #2468: DrawerFooter doesn't appear in mobile browsers
# Bug report

## Describe the bug

The `DrawerFooter` component is not appearing when a  `Drawer` is open in mobile browsers.  Tested on Chrome/Brave/Firefox/Safari on both Android and iOS

## To reproduce

1. Click on "Open Drawer"
2. Drawer will open
3. Should see "Cancel" and "Upload File" buttons in lower right in `DrawerFooter` but do not appear on mobile browsers

## Minimal reproduction

https://5neo3.csb.app/

## Expected behavior

DrawerFooter component should appear when drawer is open

## System information

- OS: Android/iOS
- Browser (if applies): Chrome/Brave/Firefox on Android - Safari on iOS
- Version of @chakra-ui/core:  1.0.0-rc.8

maxbeier:
Closes #2468

## 📝 Description

The Drawer currently uses a height of `100vh` (introduced in e7cbdeb762c38767268cf5fd3fafb649e0f6691e) which is the whole viewport. Unfortunately, depending on the scrolling direction, Safari overlays the page content with a bottom bar (to share the URL or switch tabs). So, if a user scrolls up and opens a Drawer, parts of it might be hidden by the bottom bar.

The drawer is already `position: fixed`. It is therefore easy to replace the explicit hight by an implicit one by setting `top: 0; bottom: 0` as this relates to the currently available space (i.e. takes the bottom bar into account). (There' no need to re-calculate in case the bar appears / disappears, as there's a scroll-lock in place when the Drawer opens and the bar can't change.)

| without Safaris UI | with Safaris UI |
| --- | --- |
| with-1 | without-1 |

## 🚀 New behavior

| without Safaris UI | with Safaris UI |
| --- | --- |
| with-2 | without-2 |

No

## 📝 Additional Information

I like this project a lot. But it was kinda time consuming to find this bug as code is often spread (which makes it hard to trace the source of a property) and code is distributed over different packages (and they are not compiled on changes, so code might be stale – but recompiling everything takes some time.) It would be great to have a general watcher service that recompiles a package on changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "prestart": "yarn && yarn bootstrap",
     "boot": "yarn prestart && yarn build",
     "start": "yarn build && yarn storybook",
-    "start:all": "yarn build && concurrently \"yarn docs develop\"  \"yarn storybook\" ",
+    "start:all": "yarn build && concurrently \"yarn docs dev\"  \"yarn storybook\" ",
     "prebuild": "yarn pkg babel-plugin build",
     "build": "lerna run build --ignore @chakra-ui/test-utils --no-private --stream",
     "test": "lerna run test --no-private --stream",
--- a/packages/transition/src/slide.tsx
+++ b/packages/transition/src/slide.tsx
@@ -28,9 +28,9 @@ const directions = {
     motion: { x: "-100%" },
     baseStyle: {
       width: "100%",
-      height: "100vh",
       left: 0,
       top: 0,
+      bottom: 0,
     },
   },
   right: {
@@ -39,7 +39,7 @@ const directions = {
       width: "100%",
       right: 0,
       top: 0,
-      height: "100vh",
+      bottom: 0,
     },
   },
 }


acolytec3:
@segunadebayo When will this commit hit a release?

Fedeorlandau:
Hello Max! How are you?

By testing it on the link you provided on #2468  I found the following bugs:

iPhone 11 Pro MAX Chrome and Safari

Close Icon on the left side.
WhatsApp Image 2021-01-17 at 14 20 05 (1)

Looks broken on Safari
WhatsApp Image 2021-01-17 at 14 20 05

Cheers,

maxbeier:
Ok, can reproduce this on a simulated iPhone 11 Pro MAX and Desktop Safari, but from what I see, it's not related to this PR.

So, the `right` property is missing on the button:

<figure>
Screenshot 2021-01-17 at 18 52 33
<figcaption>Left: production; right: develop</figcaption>
</figure>

---

It seems that the property was removed 2 weeks ago with https://github.com/chakra-ui/chakra-ui/commit/beef7534dc0164a29045ea24234db5a1736b4961 in the PR https://github.com/chakra-ui/chakra-ui/pull/2918 and replaced with `insetEnd`. As `inset` [does not seem to be widely supported](https://caniuse.com/mdn-css_properties_inset), it needs to be converted, which was introduces in [styled-system 1.2.0](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/styled-system/CHANGELOG.md#120).

I don't yet know why it doesn't work, but maybe the pointers help and you can chase the reason with me.

Fedeorlandau:
> Ok, can reproduce this on a simulated iPhone 11 Pro MAX and Desktop Safari, but from what I see, it's not related to this PR.
> 
> So, the `right` property is missing on the button:
> 
> Screenshot 2021-01-17 at 18 52 33
> 
> Left: production; right: develop
> It seems that the property was removed 2 weeks ago with [beef753](https://github.com/chakra-ui/chakra-ui/commit/beef7534dc0164a29045ea24234db5a1736b4961) in the PR #2918 and replaced with `insetEnd`. As `inset` [does not seem to be widely supported](https://caniuse.com/mdn-css_properties_inset), it needs to be converted, which was introduces in [styled-system 1.2.0](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/styled-system/CHANGELOG.md#120).
> 
> I don't yet know why it doesn't work, but maybe the pointers help and you can chase the reason with me.

Thanks for your answer!

100% agree this PR isn't related with the DrawerCloseButton. (Which I fixed it by setting the right prop manually`<DrawerCloseButton right={3} />`

Regarding the Drawer height on Chrome iOS. Do you want me to create a PR with the [--vh](https://css-tricks.com/the-trick-to-viewport-units-on-mobile/) solution? Cheers,

| Toolbar |  w/o toolbar |
|---------|--------------|
|   WhatsApp Image 2021-01-17 at 16 20 08 |      WhatsApp Image 2021-01-17 at 16 20 09        |

maxbeier:
Regarding the CloseButton placement bug, there was an update an hour ago, so I would consider it solved: https://github.com/chakra-ui/chakra-ui/releases/tag/%40chakra-ui%2Fstyled-system%401.5.0

Regarding the hidden footer buttons: I can't reproduce the bug on iOS Chrome with an iPhone 11 Pro MAX and it works in iOS Chrome on an iPhone 7 Plus. Can you check if the bug persists with the new release? If it does, I would prefer a CSS solution over one that relies on JavaScript, but if there's no other way, sure, I would be happy about a PR (but I'm just another user here).

PR #3061, opened 2021-01-14
docs(progress): correct description about thickness prop

dominictwlee:
Closes # 

## 📝 Description
I noticed that the description for the `thickness` prop on CircularProgress is erroneous and may cause confusion. It says that it sets things by ratio, expecting a value between 0 and 1, but in reality all it's doing is setting stroke width by pixel.

n/a

## 🚀 New behavior
n/a

No

### DIFF
--- a/packages/progress/src/circular-progress.tsx
+++ b/packages/progress/src/circular-progress.tsx
@@ -51,7 +51,7 @@ interface CircularProgressOptions {
    */
   min?: number
   /**
-   * The thickness of progress indicator as a ratio of `size`. Must be between `0` and `1`
+   * This defines the stroke width of the svg circle.
    */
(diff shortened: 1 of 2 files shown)


dodas (approved):

TimKolberger:
Good catch @dominictwlee! Would you mind updating the prop documentation as well? This would update the prop description in the table at the bottom at the page.

https://github.com/chakra-ui/chakra-ui/blob/fbecdbd6738aac3c7e72539c58856282fcc44e83/packages/progress/src/circular-progress.tsx#L54

with-heart (approved):

with-heart:
Thanks @dominictwlee!

@allcontributors please add @dominictwlee for docs

PR #3069, opened 2021-01-15
docs: update homepage stat numbers

dodas:
The numbers are going up fast!

Also, how do we count the "Core contributors"?


segunadebayo:
@dodas, we'll write a script to update these numbers in the future but this manual update works.

Thanks.

PR #3077, opened 2021-01-17
feat: allow responsive values for `variant`, `size`, and `orientation`

### LINKED ISSUE #1028: Allow responsive values for `variant`, `size`, and `orientation`
The new modifier versions of the `variant` and `size` props don't accept responsive values (specified as objects or arrays). Responsive values can currently be achieved using `useBreakpointValue`:

```jsx
const ResponsivelySizedModal = () => {
  const size = useBreakpointValue({base: 'full', lg: 'xl'})
  return (
    <Modal size={size}>
      {/* ... */}
    </Modal>
  )
}
```

```jsx
const ResponsivelySizedModal = () => {
  const size = useBreakpointValue(['full', null, 'xl'])
  return (
    <Modal size={size}>
      {/* ... */}
    </Modal>
  )
}
```

This functionality should instead become native to these props, like most other props in our system:

```jsx
const ResponsivelySizedModal = () => (
  <Modal size={{base: 'full', lg: 'xl'}}>
    {/* ... */}
  </Modal>
)
```

```jsx
const ResponsivelySizedModal = () => (
  <Modal size={['full', null, 'xl']}>
    {/* ... */}
  </Modal>
)
```

dodas:
Closes #1028

## 📝 Description

I added the support for responsive `variant`, `size`, and `orientation` to the system as well as to component implementations that need special care for `orientation` or `size`.

No


segunadebayo:
This looks interesting @dodas. I see you're using `useBreakpointValue` and that's an issue for SSR. I'd love to use a pure CSS approach to solve this responsive problem.

That will require a modification to this hook: https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/use-style-config.ts.

dodas:
Hey @segunadebayo !
Yes, I am modifying the `useStyleConfig.` now as well.
I will experiment with that tonight and see if it can be done with CSS.

Thanks for taking a look!

with-heart:
> Also: Any idea why the workflow failed with the error `Cannot find module '@chakra-ui/media-query' or its corresponding type declarations.`? It builds fine on my computer.

`lerna` hooks up dependencies for you locally but the CI pipeline uses the actual dependency graph, so you'd need to add `@chakra-ui/media-query` as a `dependency` to `system`.

with-heart:
Also just want to say that I'm really glad to see someone working on this. Thanks @dodas!

packages/system/src/use-style-config.ts
```diff
@@ -72,9 +82,62 @@ export function useStyleConfig(themeKey: any, props: any = {}, opts: any = {}) {
...
 }
+
+interface ResolveResponsivePropStylesOptions {
+  responsiveValue: ResponsiveValue<string>
+  responsiveStyles: Record<string, SystemStyleObject>
+  breakpoints: string[]
+  props: Record<string, any>
+}
+
+function resolveResponsivePropStyles({
```
dodas, line 99: I am not 100% satisfied with the `resolveResponsivePropStyles` implementation. Suggestions are welcome!

dodas:
Closing in favor of #3258 where responsive`orientation` is out of scope. It will be implemented in a separate PR.

PR #3092, opened 2021-01-18
feat: autocomplete for chakra theme tokens

TimKolberger:
## 📝 Description

Add TypeScript AutoComplete for the chakra UI theme (VSCode only for now, due some TypeScript bugs in JetBrains IDEs).
With the CLI tool `@chakra-ui/cli` you can generate theme typings for your application theme.

All theme tokens are typed by extending the CSS types with an additional `string` type to allow passing values like `fontSize="sm"`.  So you have to look up the theme tokens in the documentation or in the source code: 😢 

Component specific props like `variant` and `size` are generically types as `string`.  As user you don't know which values the component supports 😢 

## 🚀 New behavior

A codegen creates a TypeScript interface `ThemeTypings` and overrides the corresponding file of `@chakra-ui/styled-system`. This happens at build time in the chakra repository and can be implemented as npm `postinstall` script in userland projects.

### Manual

```bash
npx @chakra-ui/cli tokens <path/to/your/theme.(js|ts)>
```

### Postinstall script

```json
"scripts": {
  "gen:theme-typings": "chakra-cli tokens <path/to/your/theme.(js|ts)>",
  "postinstall": "npm run gen:theme-typings"
}
```

### 📸  Screenshots

image
image
image
image

No, the types are not narrowed down - they still include `string` as escape hatch.

## 📝 Additional Information

Need to figure out how to test this outside of the chakra codebase 🤔


tooling/theme-typings/src/extract-color-schemes.ts
```diff
@@ -0,0 +1,37 @@
...
+  "100",
+  "200",
+  "300",
+  "400",
+  "500",
+  "600",
+  "700",
+  "800",
+  "900",
+]
```
segunadebayo, line 14: What happens when a user defines custom color keys like

```js
const colors = {
   light: "",
   dark: ""
}
```

Is this a scenario we need to account for?

segunadebayo:
This is great work @TimKolberger. I'm glad we're able to get a working solution for this.

Thanks for improving it even further 💖

tooling/theme-typings/src/extract-color-schemes.ts
TimKolberger: Yes, the interface generator accounts for this scenario. Please see this test case: 

https://github.com/chakra-ui/chakra-ui/blob/3591e33f522c1515e71562d45f66bc19cfc6ae43/tooling/theme-typings/test/theme-typings.test.ts#L9-L11

https://github.com/chakra-ui/chakra-ui/blob/3591e33f522c1515e71562d45f66bc19cfc6ae43/tooling/theme-typings/test/theme-typings.test.ts#L75

dodas:
Hey Tim! Absolute game-changer for error prevention and developer experience. Excellent job!

One idea: have you heard about Prisma Client? They solve generating and the postinstall issue in their own way. 
[Read more about the `@prisma/client` npm package](https://www.prisma.io/docs/concepts/components/prisma-client/generating-prisma-client#the-prismaclient-npm-package).
Maybe you will like and take some ideas from them.

Cheers!

TimKolberger:
Hey Jozef! 

Thanks for the awesome suggestion 💖 

@segunadebayo and I discussed the "prisma way" as well. The question I have not solved currently is: _How can we determine the path to the userland theme file?_

Segun offered to pair on this issue and I am looking forward to it! I hope that we are coming up with a solution.

304s (approved):

304s (approved):

tooling/cli/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+For convenience, you can add a `postinstall` script to your `package.json`, so
+you don't have to think about this every time you re-install your dependencies.
+
+```json title="package.json"
+"scripts": {
+  "gen:theme-typings": "chakra-cli tokens <path/to/your/theme.(js|ts)>",
+  "postinstall": "npm run gen:theme-typings"
+}
+```
```
with-heart, line 36: This is totally fine for a first pass. I'm wondering if there's a way we could simplify this and make it more automated though, perhaps by adding something like this to `package.json` that, if present, would automatically produce theme typings when `@chakra-ui/theme` is installed:

```json
{
  "chakra": {
    "generatedTypes": true,
    "theme": "./src/theme.ts"
  }
}
```

tooling/cli/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "test:cov": "yarn test --coverage",
+    "lint:src": "eslint src --ext .ts,.tsx --config ../../.eslintrc",
+    "lint:types": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@chakra-ui/utils": "^1.0.2",
+    "cli-check-node": "^1.3.4",
+    "cli-handle-unhandled": "^1.1.1",
+    "cli-welcome": "^2.2.2",
+    "commander": "^6.2.1",
```
with-heart, line 60: Glad we decided to use a command manager for this pass of `chakra-cli`. Nice to just have an out-of-the-box pattern for this kind of thing.

website/pages/docs/theming/advanced.mdx
```diff
@@ -97,3 +97,44 @@ interface StyleConfig {
...
+}
+```
+
+The autocomplete for theme token **only works in _Visual Studio Code_ for now**.
+We keep an eye on the TypeScript support of Jetbrains IDEs, but I we have to
+wait until it is improved.
+
+Please note that the CLI cannot read non-JavaScript/TypeScript files. If you
+import other file types (`.css`, `.woff` or `.svg`) in your theme file, you'll
+need to move those imports out of the theme file.
```
with-heart, line 140: I feel like these both probably deserve to live in an `Alert` or `>` blockquote to call more attention to them, just to make it a little less likely that we have issues opened for them.

website/pages/docs/theming/advanced.mdx
```diff
@@ -97,3 +97,44 @@ interface StyleConfig {
...
+```json
+"scripts": {
+  "gen:theme-typings": "chakra-cli tokens <path/to/your/theme.(js|ts)>",
+  "postinstall": "npm run gen:theme-typings"
+}
+```
+
+The autocomplete for theme token **only works in _Visual Studio Code_ for now**.
+We keep an eye on the TypeScript support of Jetbrains IDEs, but I we have to
+wait until it is improved.
```
with-heart, line 136:
```suggestion
We'll keep an eye on the TypeScript support of Jetbrains IDEs, but we have to
wait until it is improved.
```

with-heart (approved):
This is such an excellent DX addition! I'm so pumped for it.

I have two questions:

1. What happens if the user's theme file exports `extendTheme`?
2. Would either of you be interested in writing about our approach here? Maybe on the blog? I feel like this is good knowledge to share in public.

with-heart:
Whoops, I missed reviewing before the merge. My bad! I'll add the suggestions as a follow-up PR.

website/pages/docs/theming/advanced.mdx
with-heart: @TimKolberger Do you know if there's a specific issue on the JetBrains bug tracker that would enable us to support this? Would love to link to that if possible.

tooling/cli/README.md
TimKolberger: Yes, this is definitely a point where we can improve the DX. @dodas mentioned this approach as well, I think this is a great idea. 
Lets get v1 of this feature out there and improve it step by step 👍

packages/styled-system/src/theming.types.ts
```diff
@@ -0,0 +1,24 @@
...
+  sizes: string
+  space: string
+  zIndices: string
+  components: {
+    [componentName: string]: {
+      sizes: string
+      variants: string
+    }
+  }
+}
```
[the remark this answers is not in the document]
TimKolberger: This file will be replaced during the build of chakra. See the postbuild script of the theme package.

website/pages/docs/theming/advanced.mdx
TimKolberger: There are many different issues with the TypeScript support of JetBrains IDEs.

I created a reproduction for required prop warnings [WEB-48161](https://youtrack.jetbrains.com/issue/WEB-48161) [here](https://gist.github.com/TimKolberger/0900962f7f2c28b4a80bcd8754785cf2).

Missing auto import of components here [WEB-48085](https://youtrack.jetbrains.com/issue/WEB-48085).

And I stopped looking after these two..

packages/system/src/system.types.tsx
```diff
@@ -2,15 +2,20 @@ import {
...
-  variant?: string
-  size?: string
-  colorScheme?: string
+export interface ThemingProps<ThemeComponent extends string = string> {
+  variant?: ThemeComponent extends keyof ThemeTypings["components"]
+    ? ThemeTypings["components"][ThemeComponent]["variants"] | (string & {})
+    : string
+  size?: ThemeComponent extends keyof ThemeTypings["components"]
+    ? ThemeTypings["components"][ThemeComponent]["sizes"] | (string & {})
+    : string
```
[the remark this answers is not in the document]
TimKolberger: Yes, I think this would be great as well! This would increase the type safety, but also mean a major bump. Lets test the current approach if it works for the majority of our users and improve it step by step.

PR #3116, opened 2021-01-21
Docs/menu auto select

tmikeschu:
## 📝 Description

Hello! 👋🏼 

I missed this property for a couple days because I was searching on the docs page for the keyword `focus`. While I should have read each prop carefully, I think clarifying that the opened menu applies focus is more accurate compared to actually making a selection. Is there a reason the prop isn't `autoFocus`? I realize that would be a breaking change, so I figured a docs update is the next best thing.

No

### DIFF
--- a/packages/menu/src/use-menu.ts
+++ b/packages/menu/src/use-menu.ts
@@ -55,7 +55,7 @@ export interface UseMenuProps extends UsePopperProps, UseDisclosureProps {
    */
   closeOnBlur?: boolean
   /**
-   * If `true`, the first enabled menu item will receive focus
+   * If `true`, the first enabled menu item will receive focus and be selected
    * when the menu opens.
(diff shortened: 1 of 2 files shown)


packages/menu/src/use-menu.ts
```diff
@@ -55,7 +55,7 @@ export interface UseMenuProps extends UsePopperProps, UseDisclosureProps {
    */
   closeOnBlur?: boolean
   /**
-   * If `true`, the first enabled menu item will be selected
+   * If `true`, the first enabled menu item will receive focus
```
segunadebayo, line 58:
```suggestion
   * If `true`, the first enabled menu item will receive focus and be selected
```

tmikeschu:
@segunadebayo thanks!

Issue #2615, opened 2020-11-25
useBreakpointValue breaks MS Edge


ssijak:
# Bug report

## Describe the bug

useBreakpointValue breaks MS Edge

## To reproduce

If I try to use useBreakpointValue hook in my nextjs project it breaks the website in ms edge, all JS stops executing after it encounters the hook.

It breaks with ```SCRIPT438: SCRIPT438: Object doesn't support property or method 'endsWith'```

I've tried to polyfill with ```import "core-js/features/string/ends-with"; ``` and in several other ways per nextjs docs, but it did not help.

## Minimal reproduction

https://codesandbox.io/s/gracious-waterfall-dt01w?file=/src/index.js

## System information

- OS: Windows 10
- Browser (if applies): MS Edge
- Version of @chakra-ui/react: 1.0.0
- Version of Node.js: 12

with-heart:
This is interesting. Edge should theoretically support `endsWith`, so it's strange that this is causing an issue. We use `endsWith` in a single place in the codebase: https://github.com/chakra-ui/chakra-ui/blob/dd5a4974db93d080a6603b3c1ca2c2b91f5136d9/packages/media-query/src/create-media-query.ts#L86-L86

I'm not sure why this would cause a failure for Edge though. @ssijak could you try the solution [in this comment on the `next.js` repo](https://github.com/vercel/next.js/issues/7993#issuecomment-537530895) and see if it resolves your issue?

gitmonster:
I have the same issue on Edge Legacy with a NextJS Project and applied the polyfill the way @with-heart 
mentions. Nothing helps.

ssijak:
@with-heart I've tried that, and several other solutions and nothing worked for MS Edge. It feels strange, it should work, but it keeps fatally breaking with that error.

Even doing (taken from [mozzila docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith)) :
```js
if (!String.prototype.endsWith) {
  String.prototype.endsWith = function(search, this_len) {
    if (this_len === undefined || this_len > this.length) {
      this_len = this.length;
    }
    return this.substring(this_len - search.length, this_len) === search;
  };
}
```

Before calling the hook does not help. Either I'm missing something obvious or there is some weird thing happening with my Edge.

gitmonster:
Has anyone a solution?

objmj-itminds:
I, too, have stumbled into this issue.

[...]

PR #3164, opened 2021-01-26
fix(extend-theme): extend theme should not include polyfill function

ivanyeoh:
Closes https://github.com/chakra-ui/chakra-ui/issues/2615

## 📝 Description

> When we have polyfill for `Array` or `Object`, calling useBreakpoint will throw error `Object doesn't support property or method 'endsWith'`. This is because extendTheme are also extending the polyfill functions of Array and Object

> When we have polyfill like `Array.prototype.each = () => {}` in our code, `extendTheme` will treat `each` as a props in theme 

## 🚀 New behavior

> When we have polyfill like `Array.prototype.each = () => {}` in our code, `extendTheme` should not include `each` as a props in theme

No

## 📝 Additional Information

Sometimes we don't have control over which polyfill to load. For my case, I am building an embeddable widget to be included in another website, but the website are loading `prototype.js` which includes bunch of polyfills

Here's what happen:

Screenshot 2021-01-26 at 4 44 13 PM

Screenshot 2021-01-26 at 4 47 33 PM

### DIFF
--- a/.changeset/flat-files-provide.md
+++ b/.changeset/flat-files-provide.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react": patch
+---
+
+fix: Extend theme should not include polyfill function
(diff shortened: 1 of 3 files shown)


TimKolberger (approved):
Good catch! 👍

PR #3165, opened 2021-01-26
fix: workaround for react concurrent mode focus and change events issue

### LINKED ISSUE #2350: Checkbox requires two clicks (Concurrent Mode)
# Bug report

## Describe the bug

The `<Checkbox />` component is missing some accessibility features when used in a React project with Concurrent Mode.

- The element must be focussed with an initial click before it can be toggled with a second click
- The spacebar has no effect on toggling the focussed component
- When the spacebar is pressed, an additional click is required to toggle the component

## To reproduce

1. Create a new React project
2. Enable the unstable Concurrent Mode
3. Render and interact with a `<Checkbox />` component

## Minimal reproduction

https://codesandbox.io/s/chakraui-checkbox-bug-concurrent-mode-d9rm9?file=/src/index.js

The index file has a constant `ENABLE_CONCURRENT` that can be easily toggled to enable/disable the bug. The issue is present when concurrent mode is enabled. Switching React versions is not necessary, and the component behaves normally when concurrent mode is disabled, even with the experimental React version.

## Expected behavior

The checkbox should have the same behaviour as when legacy React is used.

## Screenshots

Not applicable.

## System information

- OS: Windows
- Browser (if applies): Chrome
- Version of @chakra-ui/core: 1.0.0-rc.7
- Version of React: 0.0.0-experimental-4ead6b530

## Additional context

Let me be clear, this issue is present on the experimental version of React _and_ the unreleased preview of Chakra UI. I'm not surprised that it is broken, but wanted to give a heads up for the future release of React! I'm not sure whether you want to track such issues.

I will keep on using the experimental version of React and can keep this issue updated in case it gets resolved by the next experimental release.

anteprimorac:
Closes #2350

## 📝 Description

Fixes checkbox and radio components in React concurrent mode by applying workaround from https://github.com/facebook/react/issues/18591#issuecomment-697999763. This can be removed after the mentioned issue is resolved in React.

Currently in React concurrent mode checkbox and radio inputs require two clicks to update a checked state.

## 🚀 New behavior

This fix only applies the workaround for projects that use experimental React. It should not affect other projects.

No.

### DIFF
--- a/packages/checkbox/src/use-checkbox.ts
+++ b/packages/checkbox/src/use-checkbox.ts
@@ -18,6 +18,7 @@ import React, {
   useRef,
   useState,
 } from "react"
+import ReactDOM from "react-dom"
 
 export interface UseCheckboxProps {
   /**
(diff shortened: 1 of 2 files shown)


lcswillems:
Is this fix present in 1.2.1? Because I use it currently and still have to double click.

PR #3179, opened 2021-01-27
fix(number-input): always cast counter.value

### LINKED ISSUE #3178: `NumberInput` not clamping with `precision` on blur
# 🐛 Bug report

The `NumberInput` component `clampValueOnBlur` prop defaults to `true`, but the value is not clamped using `precision` when the `input` is blurred. For example, with a `precision` of `2` and a value of `1.1234`, blurring should clamp the value to `1.12`, but instead the value doesn't change on blur.

## 💥 Steps to reproduce

> Clear and concise reproduction steps are important for us to be able to triage
> your issue in a timely manner

1. Go to https://codesandbox.io/s/clever-noether-03l2d?file=/src/App.tsx
2. Type in `1.1234`
3. Blur the field and notice that the value doesn't change
4. Press the increment or decrement stepper and notice that the value is clamped using `precision` (to `1.13` or `1.11`)

## 💻 Link to reproduction

https://codesandbox.io/s/clever-noether-03l2d?file=/src/App.tsx

## 🧐 Expected behavior

When the field is blurred, the value should clamp in the same way that pressing the increment or decrement stepper clamps.

## 🧭 Possible Solution

It appears that this check of `counter.value !== next` is blocking `counter.cast(next)` from being called, which would otherwise clamp the value: https://github.com/chakra-ui/chakra-ui/blob/6e8fac47b5281298913f8fcd4f24c06d378bc8d6/packages/number-input/src/use-number-input.ts#L298-L300 

`cast` should likely always be called, as `counter.value` will always equal `next` on blur as `counter.value` was already updated by `onChange` and `onBlur` causes no change in value on its own.

## 🌍 System information

| Chakra UI        | 1.1.6           |
| Browser          | Any           |
| Operating System | Any           |

with-heart:
Closes #3178 

## 📝 Description

This PR resolves an issue where clamping to a `precision` prop value doesn't occur on blur.

Blurring the `input` when `precision` prop is present doesn't clamp the value to the `precision` value.

## 🚀 New behavior

Blurring the `input` clamps the value using the `precision` prop.

No

### DIFF
--- a/packages/number-input/src/use-number-input.ts
+++ b/packages/number-input/src/use-number-input.ts
@@ -295,9 +295,7 @@ export function useNumberInput(props: UseNumberInputProps = {}) {
      * - sanitize the value by using parseFloat and some Regex
      * - used to round value to computed precision or decimal points
      */
-    if (counter.value !== next) {
-      counter.cast(next)
-    }
+    counter.cast(next)
   }, [counter, max, min])
 
   const onBlur = React.useCallback(() => {
--- a/packages/number-input/tests/number-input.test.tsx
+++ b/packages/number-input/tests/number-input.test.tsx
@@ -128,7 +128,7 @@ test("should increase/decrease by 0.1*step on ctrl+Arrow", () => {
   expect(input).toHaveValue("0.00")
 })
 
-it("should behave properly precision value", () => {
+it("should behave properly with precision value", () => {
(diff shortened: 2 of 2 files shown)


with-heart:
Forgot to add a changeset! Will open a PR for that in a sec.

PR #3195, opened 2021-01-28
docs: remove function not utilized in doc example code

ryanahamilton:
## 📝 Description

Simplifies Color Mode doc code example by removing unnecessary function.


dodas (changes requested):
Hey @ryanahamilton! Thanks for contributing!

We are trying to follow [Semantic Commit Messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716). Your commit doesn't follow it and that's also the reason why the checks aren't passing.

Could you please force-push a new commit with semantic message?

Thanks!

dodas (approved):
✨

Issue #3223, opened 2021-01-30
Option to disable Tooltip's focus outline.


yassinebridi:
# 🚀 Feature request

An option to disable Tooltip focus indicator(outline), when moving around with the keyboard.

## 🧱 Problem Statement / Justification

I'm using the tooltip component as a parent of a button, and the button has a focus outline of it's own, so moving aroung with keyboard becomes painful, as you have to travese two outline indicators.

## ✅ Proposed solution or API

An option to disale the tooltip's focus indicator.

## 📝 Additional Information

![Peek 2021-01-31 00-45](https://user-images.githubusercontent.com/18403595/106370786-bd104300-635d-11eb-96e3-4aa12d32d88a.gif)

dodas:
Hey @yassinebridi,
can you provide a reproduction?

I can't reproduce it myself in the docs:
![tooltip](https://user-images.githubusercontent.com/14360171/106388681-ff7a6400-63df-11eb-92e9-14431b7fdac8.gif)

yassinebridi:
@dodas What browser are you using, i'm using chrome 88, and this gif is from the chakra docs:

![pee](https://user-images.githubusercontent.com/18403595/106389377-3ef67f80-63e3-11eb-8eb1-9923dbe39aa1.gif)

yassinebridi:
I just checked firefox dev edition, and it has the same issue, i'm not sure if this is a system issue, i'm using these browsers on a linux based system(arch).

with-heart:
It appears that this is caused by the `shouldWrapChildren` prop, which wraps `children` in a `span` with `tabIndex={0}`, causing the `span` to receive focus. This is great for cases where `children` isn't tabbable, but causes this issue to appear when `children` is an element like `button` which is already tabbable.

Are you using `shouldWrapChildren` @yassinebridi?

yassinebridi:
@with-heart You are right, i removed `shouldWrapChildren`, and it works as excpected.

with-heart:
We should probably note that somewhere 🙃

PR #3229, opened 2021-01-31
docs(tooltip): remove shouldWrapChildren prop from examples

with-heart:
Discovered in #3223 

## 📝 Description

This PR removes the `shouldWrapChildren` prop from `Tooltip` examples. The presence of that prop causes `children` to be wrapped in a `span` with `tabIndex={0}`. Since `children` are already buttons, this causes them to have two focusable wrappers, requiring a double-tab to move to the next element.


304s:
Yup 👍

PR #3232, opened 2021-02-01
fix(extend-theme): merge fn overrides with objects

dodas:
## 📝 Description

Fixed issue in `extendTheme` where overrides defined as function replaced all base styles defined as plain object.

## ⛳️ Current behavior

If styles (`baseStyles`, an variant, or an size) are defined in a plain object in the base theme and user provides a function as override, all base theme styles will be dropped.

## 🚀 New behavior

Styles returned from user's override function will be merged to base theme styles.

Hopefully no?

### DIFF
--- a/.changeset/funny-monkeys-build.md
+++ b/.changeset/funny-monkeys-build.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react": patch
+---
+
+Fixed issue in `extendTheme` where overrides defined as function replaced all
+base styles defined as plain object.
--- a/packages/react/src/extend-theme.ts
+++ b/packages/react/src/extend-theme.ts
@@ -41,11 +41,11 @@ export function extendTheme<
     object: any,
   ) {
     if (
-      isFunction(source) &&
+      (isFunction(source) || isFunction(override)) &&
       Object.prototype.hasOwnProperty.call(object, key)
     ) {
       return (...args: unknown[]) => {
-        const sourceValue = source(...args)
+        const sourceValue = isFunction(source) ? source(...args) : source
 
         const overrideValue = isFunction(override)
           ? override(...args)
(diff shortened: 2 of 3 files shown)


TimKolberger (approved):
Nice work @dodas

with-heart:
Is this breaking? I could def see this being breaking for users if their theming is based on the (incorrect, but around for awhile) old implementation

dodas:
I see what you mean. Depends on what we call ,,breaking".
Should we call a breaking change each bug fix, because users relied on different behaviour so far?

TimKolberger:
It does not break the API surface, but it introduces a different behaviour for some theme extensions.

I think the intention of the `extendTheme` function is to extend, not to override complete style definitions.

So this change is imo a bugfix, not a breaking change.

with-heart:
I think it's more about the backwards compatibility part. It's a weird case since we consider this a bug, but it's a bug that has always existed and fixing it changes the behavior. If instead we had intended for it to work that way to begin with and we decided to change this, we'd consider that breaking right? I don't see how it technically not working how we intended it to is any different.

with-heart:
To me, to answer the question of if something is breaking, I imagine I'm one of our users. If I upgrade to this as a patch version and my things break (and not because of a new bug that was introduced), it was quite literally a breaking change.

dodas:
I think it's really about point of view. Fix of any bug that users might have accounted into their code could be a breaking change.

What about just documentating this change better?
Better changeset and maybe a note in docs as well might do?

TimKolberger:
Defining what a breaking change is, is a gray area. I am glad we are talking about this!

To me as a user, a breaking change is something, where I have to update my code base to successfully build my project.

With the mindset, the UI looks different -> it is a breaking change, we would have to ship a new major often to inform the user that something could be different after this update. But imo this is always the case after updating dependencies.

with-heart:
> Defining what a breaking change is, is a gray area. I am glad we are talking about this!

Same! The visual aspect of things makes it far more abstract and difficult to define.

PR #3237, opened 2021-02-01
fix: Add support for halved values in Spacing

psujit:
## 📝 Description

> Halved values are mentioned in the documentation but are not actually in the code. If halved values are used, they are replaced to the same value in 'px'. So, if we use paddingBottom={2.5}, it is replaced by padding-bottom="2.5px" instead of the expected padding-bottom="0.625rem"

> Currently halved values cannot be used for spacing although the docs say so.

## 🚀 New behavior

> As per documentation, https://chakra-ui.com/docs/theming/theme#spacing , halved values are supported in spacing
0.5: "0.125rem",
1.5: "0.375rem",
2.5: "0.625rem",
3.5: "0.875rem"

## 📝 Additional Information
Merging this PR will allow using halved values as spacing


Ridvanovskyy (approved):

cschroeter:
I think that's a little inconsistent. Why do we have half values for everything less than 4 but not greater than 4?

In my opinion, the developers should be allowed to use their theme

psujit:
Just implemented the values that are mentioned in the docs..
https://chakra-ui.com/docs/theming/theme#spacing

cschroeter:
yeah thats true. I think it could make sense to have those half values below 16px (1rem)

PR #3244, opened 2021-02-01
Remove inaccessible box example from documentation

Ekwuno:
Closes # 

## 📝 Description

> Add a brief description

If the Box component defaults to a Div, adding a role of a button makes it seem like a button but this is not semantic. I propose removing that bit from the example, the existing example is already a good fit.

Remove as prop example from box component 

## 🚀 New behavior

I think the card example provided is sufficent

### DIFF
--- a/website/pages/docs/layout/box.mdx
+++ b/website/pages/docs/layout/box.mdx
@@ -114,13 +114,3 @@ function AirbnbExample() {
 }
 ```
 
-### as prop
-
-You can use the `as` prop to change the element render, just like
-styled-components.
-
-```jsx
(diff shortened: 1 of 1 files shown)


dodas (commented):
Hey @Ekwuno 

Chakra's `as` prop changes what element/component will comopnent render. That means `<Box as="button">` will render a `<button>` element. It is semantically the same as using the `Button` component. We are not overriding the `role` attribute.

Screenshot 2021-02-02 at 00 51 54

segunadebayo:
Thanks for chiming in @dodas. I discussed with @Ekwuno already about this.

PR #3245, opened 2021-02-02
feat: add support for text and layer style in component theme

segunadebayo:
## 📝 Description

This PR adds support for using `layerStyle` and `textStyle` in component theme.

- Extracted the layer and text style code to chakra's styled-system package
- Extracted the apply style to the styled-system package config

 It's not possible to use layer and text styles

No

### DIFF
--- a/.changeset/eighty-ties-melt.md
+++ b/.changeset/eighty-ties-melt.md
@@ -1,0 +1,28 @@
+---
+"@chakra-ui/styled-system": minor
+---
+
+- Add support for `textStyle` and `layerStyle` in styled-system package. This
+  makes it possible to use them in the component theme, `css` function, and `sx`
+  prop as well.
+
+  ```jsx
+  const theme = {
+    textStyles: {
+      caps: {
+        fontWeight: "bold",
+        fontSize: "24px",
+      },
+    },
+  }
+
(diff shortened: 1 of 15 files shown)


dodas (approved):

with-heart (approved):
This is awesome!

TimKolberger (approved):
I am not a big of fan the duplicated functionality of the props `apply`, `layerStyle` and `textStyle`.  They all serve the same purpose - reusing styles from the theme object. I would like to discuss these props, when we plan the next major release.

I really like the responsive behaviour of `apply` and the TS type cleanup ✨  Great job @segunadebayo

segunadebayo:
@TimKolberger agreed! I'm not a fan either but we'll keep it until the next major release.

jrolfs:
Yeah, so this came out of a bit of a @jrolfs + Chakra odyssey that started with #1028. I came out with a workable interim solution for responsive `size` in my use case, thanks to `apply`. That said, I think I'm pretty well-positioned to consider implementing responsive `size`, `variant` etc, but I'd be curious whether y'all would consider a PR before the next major. It might be a little tricky to make it non-breaking with all of the style object merging precedence but I'm willing to take a crack at it.

PR #3255, opened 2021-02-02
docs: refactor props table

dodas:
## 📝 Description

Refactored the props table while improving few details.

- strip the `(string & {})` from type in `size`, `variant` and `colorScheme`
- handle components that use foreign theme key, so we can still show default props for them (e.g. `IconButton` uses `theme.components.Button`)
- show default `colorScheme` value where possible
- note where `colorScheme` is not implemented in the default theme
- show `required` tag in front of `type` for required props
- don't mutate objects imported from `@chakra-ui/props-docs`

## ⛳️ Current behavior

example shown: `IconButton`

Screenshot 2021-02-02 at 23 08 02

## 🚀 New behavior

Screenshot 2021-02-02 at 23 07 11

No

### DIFF
--- a/website/src/components/props-table.tsx
+++ b/website/src/components/props-table.tsx
@@ -4,6 +4,15 @@ import * as ComponentProps from "@chakra-ui/props-docs"
 import { Tag, theme } from "@chakra-ui/react"
 import MDXComponents from "./mdx-components"
 
+/**
+ * A map of components that use foreign theme key.
+ * The key is name of the component and value is the theme key it uses.
+ */
+const themeComponentKeyAliases = {
+  AlertDialog: "Modal",
+  IconButton: "Button",
+}
+
 export type PropsTableProps = {
   /**
    * displayName of the target component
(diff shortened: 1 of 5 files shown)


website/src/components/props-table.tsx
```diff
@@ -141,39 +63,45 @@ Remove the use of <PropsTable of="${of}" /> for this component in the docs.`,
...
             <MDXComponents.td>
+              {prop.required && (
+                <Tag
+                  size="sm"
+                  colorScheme="red"
+                  mr={1}
+                  verticalAlign="baseline"
+                >
+                  required
+                </Tag>
```
TimKolberger, line 81: I really like that we visualise the required props.

From a design perspective this looks a bit too heavy, but I have no other proposal at hand.
Maybe we can iterate on that =)

TimKolberger (approved):
Nice work @dodas!  :sparkles: 
I could not find any false positives in the docs. Looks good to me :rocket:

website/src/components/props-table.tsx
dodas: Does shrinking it down a little like this help?
Seems like required props are quite rare in chakra, so the tag won't be seen often or in big concentration anywhere.

Screenshot 2021-02-04 at 20 41 13

PR #3267, opened 2021-02-03
docs: fix command for changeset

psujit:
## 📝 Description

> fix command that needs to be run before running `yarn changeset`.

> The current command `git fetch origin master:master` is erroneous because we do not have a master branch in the repo and the `main` branch is the source of truth.

## 🚀 New behavior

> Updated the docs with the proper command that needs to be run before running `yarn changeset`


psujit:
Created a PR into the wrong branch

PR #3283, opened 2021-02-04
feat: update focus util to handle unsupported browsers

segunadebayo:
## 📝 Description

Improves the `focus` to allow for opting out for `requestAnimationFrame` and added polyfill for Safari.

Using the `focus` method from the `@chakra-ui/utils` package or the DOM `focus({preventScroll: true})` doesn't work in Safari.


dodas (approved):
ah, the new IE in action 🤡

PR #3291, opened 2021-02-05
fix: support negative values for css variable theme tokens

### LINKED ISSUE #3111: support for css custom properties and negative numbers
https://github.com/chakra-ui/chakra-ui/blob/26ca4cc53b8cc0ca696f2130f832965f7dc0ee53/packages/styled-system/src/utils/positive-or-negative.ts#L17-L28

we use CSS custom properties for all of our theme token values, for instance:

```ts
const customSpace = {
  'px': 'var(--ps-size-px)',
  '0.5': 'var(--ps-size-0\\.5)',
  '0': 'var(--ps-size-0)',
  '1': 'var(--ps-size-1)',
  '1.5': 'var(--ps-size-1\\.5)',
  '2': 'var(--ps-size-2)',
  '2.5': 'var(--ps-size-2\\.5)',
  '3': 'var(--ps-size-3)',
  '3.5': 'var(--ps-size-3\\.5)',
  '4': 'var(--ps-size-4)',
  '5': 'var(--ps-size-5)',
  '6': 'var(--ps-size-6)',
…
```

When I want to use a "negative" value, it simply breaks on custom properties:

in:
```tsx
ml="-4"
```

out:
```css
margin-left: -var(--ps-size-4);
```

what _would_ work is to return a `calc`

out:
```css
margin-left: calc(var(--ps-size-4) / -1);
```

Would you be open to a PR which expands the `positive-or-negative` helper to check for and apply this transformation to negative theme values?

Thanks!

dominictwlee:
Closes #3111 

## 📝 Description
This handles negative scale values for css variables used in custom themes

If css variables are used for theme values, negative scales will be ignored and be treated as positive.

## 🚀 New behavior
This PR will use `calc()` to convert them to convert css variable to negative values.

no.

### DIFF
--- a/.changeset/silly-bugs-dance.md
+++ b/.changeset/silly-bugs-dance.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/utils": minor
+"@chakra-ui/styled-system": patch
+---
+
+Support negative scale values for css variables.
--- a/packages/styled-system/src/utils/positive-or-negative.ts
+++ b/packages/styled-system/src/utils/positive-or-negative.ts
@@ -1,4 +1,4 @@
-import { StringOrNumber, isString, isNumber } from "@chakra-ui/utils"
+import { StringOrNumber, isString, isNumber, isCssVar } from "@chakra-ui/utils"
 import unit from "./get-unit"
 
 const startsWith = (string: string, target: string) =>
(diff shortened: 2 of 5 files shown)


TimKolberger (approved):
Thank you for this PR @dominictwlee 💖

with-heart (approved):

with-heart:
@allcontributors please add @dominictwlee for code

PR #3296, opened 2021-02-06
Fix: pin input last entry update fails to call onComplete

mattwells19:
## 📝 Description

Previously `isComplete` was being determined based off of old state values as it was determined before the setState took effect. This caused the last pin-input entry to not call `onComplete` when updated. Updating the `isComplete` check to reference the next values (the value of the state after rerender) allows the check to use the most up to date values in the pin-input fields fixing the last digit update issue.

Previously, when updating the last entry in a `PinInput`, the `onComplete` function would not be called.

Before

## 🚀 New behavior

This PR changes the `isComplete` flag to check the next values which is the most up to date value the user entered. With this change, the `onComplete` is now called when the last entry is updated as expected.

After

No.

### DIFF
--- a/.changeset/tiny-apples-wait.md
+++ b/.changeset/tiny-apples-wait.md
@@ -1,0 +1,10 @@
+---
+"@chakra-ui/pin-input": minor
+---
+
+Previously isComplete was being determined based off of old state values as it
+was determined before the setState took effect. This caused the last pin-input
+entry to not call onComplete when updated. Updating the isComplete check to
+reference the next values (the value of the state after rerender) allows the
+check to use the most up to date values in the pin-input fields fixing the last
+digit update issue.
--- a/.changeset/tiny-apples-wait.md
+++ b/.changeset/tiny-apples-wait.md
@@ -2,9 +2,4 @@
 "@chakra-ui/pin-input": minor
 ---
 
-Previously isComplete was being determined based off of old state values as it
-was determined before the setState took effect. This caused the last pin-input
-entry to not call onComplete when updated. Updating the isComplete check to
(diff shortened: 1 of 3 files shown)


.changeset/tiny-apples-wait.md
```diff
@@ -0,0 +1,10 @@
+---
+"@chakra-ui/pin-input": minor
+---
+
+Previously isComplete was being determined based off of old state values as it
+was determined before the setState took effect. This caused the last pin-input
+entry to not call onComplete when updated. Updating the isComplete check to
+reference the next values (the value of the state after rerender) allows the
+check to use the most up to date values in the pin-input fields fixing the last
+digit update issue.
```
with-heart, line 10: Always appreciate lots of detail but this is more appropriate for the PR description than the release notes. Better to just let users know what was fixed, rather than what was wrong and how it was fixed.

```suggestion
Resolved an issue where completing character entry in `PinInput` failed to call `onComplete`.
```

with-heart (changes requested):
Great work! Will definitely merge after updating the changeset!

.changeset/tiny-apples-wait.md
mattwells19: Thanks for the suggestion! Updated.

TimKolberger (approved):

.changeset/tiny-apples-wait.md
with-heart: It's all good! Thanks for taking the time to fix this 😄

with-heart:
@allcontributors please add @mattwells19 for code

PR #3307, opened 2021-02-07
feat: Replace React.Children with ThemingPropsProvider

TimKolberger:
## 📝 Description

The goal is to make the styling less dependant on React.

Currently the Input components (and some others) rely on `React.Children` to apply conditional styles.

## 🚀 New behavior

A new introduced `ThemingPropsProvider` provides `size`, `variant` and `colorScheme` props via React Context to its children, but the conditional styles for `InputLeftAddon` etc. are applied with the general sibling combinator selector.

No, but I added deprecations for the `id` properties on the affected components, because they are not used anymore.

## 📝 Additional Information

@chakra-ui/react-team What do you think about this approach?


packages/input/src/input-group.tsx
```diff
@@ -75,10 +45,26 @@ export const InputGroup = forwardRef<InputGroupProps, "div">((props, ref) => {
         width: "100%",
         display: "flex",
         position: "relative",
+        ".chakra-input__left-addon ~ .chakra-input": {
+          borderLeftRadius: leftAddonBorderLeftRadius ?? 0,
```
ljosberinn, line 49: instead of `?? 0` we could assign the default when destructuring the props above:

```js
  const {
    children,
    className,
    leftElementWidth,
    rightElementWidth,
    leftAddonBorderLeftRadius = 0,
    righAddonBorderRightRadius,
    ...rest
  } = omitThemingProps(props)
  ```

dodas:
Hey Tim.

Have you found a way to style the input if it is followed by a right addon?
I was playing with it yesterday and really haven't found a CSS way to do it.

TimKolberger:
The CSS sibling selector `~` does not care where the sibling is, but with our explicit classes e.g. `.chakra-input__left-addon` and `.chakra-input__right-addon` we can determine on which side it is.

dodas:
I am afraid that's not the case:

https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator
> The general sibling combinator (~) separates two selectors and **matches all iterations of the second element, that are 
 following the first element** (though not necessarily immediately), and are children of the same parent element.

https://www.w3.org/TR/selectors-4/#general-sibling-combinators
> The subsequent-sibling combinator is made of the "tilde" (U+007E, ~) code point that separates two compound selectors. The elements represented by the two compound selectors share the same parent in the document tree and **the element represented by the first compound selector precedes** (not necessarily immediately) **the element represented by the second one.**

TimKolberger:
image
Thanks @dodas, you are right. I did not see these border radii there. Hm 🤔

TimKolberger:
Closing this PR, because this approach wont solve the issue.

PR #3338, opened 2021-02-10
Add Copy button to code snippets in readme

Lalit2005:
## 📝 Description

Added `Copy` button to code snippets in readme file

## 🚀 New behavior

When clicked on the `Copy` button, user will be taken to Copii's webpage and redirected back. You may test it to copy `hi` to clipboard👇

[copy](https://copii.vercel.app/?ct=hi&tm=blk&lg=https://avatars.githubusercontent.com/u/54212428?s=200&v=4)

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ link to your website.
 To use Chakra UI components, all you need to do is install the
 `@chakra-ui/react` package and its peer dependencies:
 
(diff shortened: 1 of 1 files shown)


dodas (commented):
I honestly much dislike this.

How hard is it to ctrl+c? It definitely takes less time that waiting to be redirected to some site and back. And it brings more maintanence to us, to keep the encoded code in urls in sync.

with-heart:
What's the benefit of the redirect that we don't get from a js-driven copy like most other sites use? Definitely would like to add easy copying, but the redirect feels strange.

dodas:
Well, you can't put JS inside github readme :D

with-heart:
Ohhhh, for some reason I was thinking this was for the getting started docs page. Derp

segunadebayo:
Thank you for taking the time to create this PR @Lalit2005.

I think we can do without this as it adds more maintenance overhead for us. The good'ol copy+paste works.

PR #3364, opened 2021-02-11
fix(checkbox): Chakra Checkbox HTML Validation fix

### LINKED ISSUE #3363: fix(checkbox): Chakra Checkbox HTML Validation bug
# 🐛 Bug report

When validating the Chakra checkbox element in the w3 HTML validator an error for each rendered checkbox comes up: "Element div not allowed as child of element label in this context. (Suppressing further errors from this subtree.)"

## 💥 Steps to reproduce

1. Add a Checkbox to your project with text content, for example `<Checkbox>Your Label</Checkbox>`
2. In DevTools inspect the element (`<label>`). You can also check yourself optionally https://validator.w3.org/nu/#textarea by pasting the element into the text area.

## 💻 Link to reproduction

Here is a small codesandbox with a Checkbox so you can confirm.
https://9fgq4.csb.app/

## 🧐 Expected behavior

The elements inside the `<label>` should not be `<div>`s, that is invalid HTML.

## 🧭 Possible Solution

I would like to make a quick PR, but basically replace the `<div>`s with `<span>`s.

grahhham:
Closes #3363 

## 📝 Description

When validating the Chakra checkbox element in the w3 HTML validator an error for each rendered checkbox comes up: "Element div not allowed as child of element label in this context. (Suppressing further errors from this subtree.)" because each StyledContainer `<label>` contains a StyledControl `<div>` and a "chakra-checkbox__label" `<div>`. This is causing errors in my project which requires that I have 0 errors when my HTML is validated, so I would like to propose that these `<div>` elements are converted to `<span>` elements to resolve the issue.

StyledControl is a `<div>`, so is "chakra-checkbox__label"

## 🚀 New behavior

Changing those 2 elements to `<span>`s instead.

No

### DIFF
--- a/.changeset/four-bees-tell.md
+++ b/.changeset/four-bees-tell.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/checkbox": patch
+---
+
+Improve the semantic HTML structure of checkbox. `label` is a phrasing content element and should not contain block element `div`. Replaced `div` with `span` which is an inline element.
--- a/packages/checkbox/src/checkbox.tsx
+++ b/packages/checkbox/src/checkbox.tsx
@@ -170,7 +170,7 @@ export const Checkbox = forwardRef<CheckboxProps, "input">((props, ref) => {
         {clonedIcon}
       </StyledControl>
       {children && (
-        <chakra.div
+        <chakra.span
           className="chakra-checkbox__label"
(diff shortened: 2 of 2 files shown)


segunadebayo (approved):
Good catch! 👌

PR #3371, opened 2021-02-12
docs: added an example of one way transition

MichaelGrigoryan25:
## 📝 Description

Added an example of **one way transition** in the documentation because figuring it out myself was very hard

Same behavior

## 🚀 New behavior

Same behavior

No

## 📝 Additional Information
None

https://user-images.githubusercontent.com/56165400/105380247-1920ec00-5c27-11eb-9afd-4f7720eb8a5e.mp4


with-heart:
Now that I understand what this is (thanks to your comment in the old PR), I think this is definitely useful to document, but I think we need a better way to explain what this is/means. Even with the video in the old PR, you had to explain what I was looking at for me to understand, and if a user doesn't catch the initial animation in the docs, they definitely won't understand what it's supposed to do.

MichaelGrigoryan25:
Can we add the feature to the website to like set a timeout to then display the animation? I don't know if there's the possibility for that in your website that's why I'm asking...

PR #3373, opened 2021-02-12
feat: add format options to use-clipboard

tobiasz:
## 📝 Description

Users might want to copy text in `text/html` format for example from Rich-Text Editors

Format option is not supported.

## 🚀 New behavior

We can pass option object, with `format` parameter.

No.

### DIFF
--- a/.changeset/nervous-poems-grab.md
+++ b/.changeset/nervous-poems-grab.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/hooks": patch
+---
+
+Add support to format parameter - String. Optional. Set the MIME type of what
+you want to copy as. Use text/html to copy as HTML, text/plain to avoid
+inherited styles showing when pasted into rich text editor.
--- a/.changeset/nervous-poems-grab.md
+++ b/.changeset/nervous-poems-grab.md
@@ -2,6 +2,8 @@
 "@chakra-ui/hooks": patch
 ---
 
-Add support to format parameter - String. Optional. Set the MIME type of what
-you want to copy as. Use text/html to copy as HTML, text/plain to avoid
-inherited styles showing when pasted into rich text editor.
+### useClipboard
+
+Add support to `format` - Optional string. Set the MIME type of what you want to
+copy as. Use text/html to copy as HTML, text/plain to avoid inherited styles
+showing when pasted into rich text editor.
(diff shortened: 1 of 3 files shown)


packages/hooks/src/use-clipboard.ts
```diff
@@ -7,29 +7,37 @@ import copy from "copy-to-clipboard"
  * @param text the text or value to copy
  * @param timeout delay (in ms) to switch back to initial state once copied.
  */
-export function useClipboard(text: string, timeout = 1500) {
+export function useClipboard(
+  text: string,
+  timeout:
```
dodas, line 12: What do you think about renaming `timeout` to `optionsOrTimeout, as it could be either an options object or timeout number`?

packages/hooks/src/use-clipboard.ts
```diff
@@ -7,29 +7,37 @@ import copy from "copy-to-clipboard"
...
+    | number
+    | { timeout?: number; format?: "text/plain" | "text/html" } = 1500,
```
dodas, line 14: Shouldn't we allow any mimetype (any string) for `format`, so users could supply `application/json` for example?

dodas (commented):
Other that those 2 little nitpicks, nice job 💯 

Thanks!

.changeset/nervous-poems-grab.md
```diff
@@ -0,0 +1,7 @@
+---
+"@chakra-ui/hooks": patch
+---
+
+Add support to format parameter - String. Optional. Set the MIME type of what
```
dodas, line 5: Also, from this changeset it is not clear we are talking about `useClipboard` hook.

packages/hooks/src/use-clipboard.ts
tobiasz: Agree!

packages/hooks/src/use-clipboard.ts
tobiasz: you are right, seems like we can support multiple `mime/types` https://www.w3.org/TR/clipboard-apis/#mandatory-data-types-x 

Should we pass any string? or create an ENUM?

dodas: I would allow any `string`, same what `copy-to-clipboard` does.

segunadebayo:
Thanks for this PR @tobiasz

PR #3408, opened 2021-02-17
fix(use-controllable): call `onChange` only if value changed

### LINKED ISSUE #3406: Tabs onChange handler is triggered twice
## 💻 Steps to reproduce & Link to reproduction

We can easily reproduce this on the [tabs documentation page](https://chakra-ui.com/docs/disclosure/tabs#controlled-tabs). Just add a `console.log` in the `onChange` handler. Please check the screenshot attached.

## 🧐 Expected behavior

The handler has to be called only once.

## 🌍 System information

| Chakra UI        |    1.1.6  & The version being used on Chakra docs page      |
| Browser          |        Chrome    |
| Operating System |      Mac      |

## 📝 Additional information
abc

dodas:
Closes #3406

## 📝 Description

Example:
```ts
const [selectedIndex, setSelectedIndex] = useControllableState({
  defaultValue: 0,
  onChange,
})
```

Currently, if user calls the `setSelectedIndex` function with the current value, the `onChange` callback will be still called.
This happens for example [here](https://github.com/chakra-ui/chakra-ui/blob/main/packages/tabs/src/use-tabs.ts#L315), because click on a `Tab` triggers both `onClick` and `onFocus`.

## 🚀 New behavior

If the new value is strictly equal (`===`) to the current one, the `onChange` callback won't be called.

No?

### DIFF
--- a/.changeset/dirty-radios-flash.md
+++ b/.changeset/dirty-radios-flash.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/hooks": patch
+---
+
+`useControllableState`: The `onChange` callback will be called only if the new
+value isn't equal to the current one.
--- a/packages/hooks/src/use-controllable.ts
+++ b/packages/hooks/src/use-controllable.ts
@@ -43,9 +43,15 @@ export function useControllableState<T>(props: UseControllableStateProps<T>) {
   const updateValue = React.useCallback(
     (next: React.SetStateAction<T>) => {
       const nextValue = runIfFn(next, value)
+
+      if (nextValue === value) {
+        return
+      }
+
       if (!isControlled) {
         setValue(nextValue)
       }
+
       handleChange(nextValue)
     },
     [isControlled, handleChange, value],


segunadebayo:
I see the intention here but you're making an assumption that the state is a primitive value (string, number, boolean, etc). Perhaps extracting that equality to a separate option that is a function `shouldUpdate` and allow the user to control this behavior if they want to.

```js
// shouldUpdate is an option the user can pass, and defaults to (next, prev)=>next !== prev
const doUpdate = shouldUpdate(next, prev)
if (!doUpdate) return
```

I might also test this a bit more to ensure it doesn't introduce any bugs in behavior.

dodas:
I can't see the `useControllableState` hook used with non-primitive value in Chakra UI codebase.
I assume there are VERY few cases where users use this hook in their code AND with non-primitive value AND could need such behaviour.
It mostly makes sense for primitive values. Calling `onChange` with the same value as the current isn't helpful for users in any way. If the value isn't primitive, getting notified about its reference change could be hepful in most cases.

This fix is really about not calling the `onChange` with the __same__ value.
In `Tabs` component this can happe on tab click, where both `onClick` and `onFocus` will be called on an click. Also, if you click on a Tab that is already selected, `onChange` will be still called.

But I am not strongly against adding `shouldUpdate` option, if it uses `===` comparsion by default.

with-heart:
#3570 is a component-level solution (for `Slider`) to the problem that could instead be solved at a higher level by this PR, and I think it'd be better to solve it at the `useControllableState` layer rather than in each use case throughout our codebase.

I do like the idea of adding the `shouldUpdate` option though, as without that option, I would consider this a breaking change. While there might not be many users using it currently, if they are, this would potentially break their implementation. What would avoid the breaking change would be defaulting `shouldUpdate` to `() => true`, as that would make the implementation continue to work as it does today, while allowing us (and any other users) to override the behavior where we'd like it to be more performant.

We could also change that default behavior to use a default comparison in the future with our next `major` version.

PR #3414, opened 2021-02-18
docs: improved ssr info about color mode

LuckeeDev:
## 📝 Description

I was trying to setup the Color Mode manager to prevent flash issues in a Next.js app and noticed that the approach currently described in the documentation doesn't work.
I faced two main issues:
- ChakraProvider should be set at the _app level, not in the index page;
- I couldn't force dark mode to be the default because I didn't know straight away the cookie name.

Therefore, I found the cookie name and setup the ChakraProvider at the _app level, providing the `cookies` value through the individual `pageProps`.

I also noticed that, when using the `cookieStorageManager`, `ColorModeScript` in `_document.js` has no effect. Am I missing something? Or should we maybe add a note to the docs saying that "If you use `cookieStorageManager`, you can remove `ColorModeScript` from `_document.js`"?

No

### DIFF
--- a/website/pages/docs/features/color-mode.mdx
+++ b/website/pages/docs/features/color-mode.mdx
@@ -278,20 +278,28 @@ export function getServerSideProps({ req }) {
 2. Import your wrapper component setting up Chakra:
 
 ```jsx live=false
-// e.g pages/index.js
+// setup your wrapper in the _app file (e.g: pages/_app.js)
 import { Chakra } from "../src/Chakra"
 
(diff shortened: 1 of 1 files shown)


segunadebayo:
Hi @LuckeeDev,

Yes, please add a note about the _document.js if cookieManager is used. You won't need to use `ColorModeScript` in this case.

Thank you.

LuckeeDev:
Thanks @segunadebayo, I just added it

PR #3415, opened 2021-02-18
use-controllable-state-machine

with-heart:
This PR is a WIP draft of some work I've been doing on using `xstate` for `useControllableProp` and `useControllableState`. I'll expand on this PR with more thoughts and issues later.


segunadebayo:
Thanks for this PR, @with-heart. It gave me an idea of how to implement this within our machine.

PR #3417, opened 2021-02-18
docs: Link to Themera for generating custom themes

tomhicks:
Links to a tool specifically designed to preview and export Chakra
color scales.

I made Themera so it's a shameless plug, but I believe
@with-heart suggested this to be a good idea on Discord.


with-heart (approved):

with-heart:
Thanks @tomhicks! Themera is great!

PR #3445, opened 2021-02-24
fix(checkbox): respect FormControl context

### LINKED ISSUE #3419: Checkbox not getting isInvalid prop from FormControl
# 🐛 Bug report

Checbox should be able to get the isInvalid prop from FormControl if it is set, just like the Input component.

## 💥 Steps to reproduce

See the link below.

## 💻 Link to reproduction

https://codesandbox.io/s/confident-shape-ngdej?file=/src/App.js

## 🧐 Expected behavior

The first checkbox in the example should be red.

## 🌍 System information

| Chakra UI        |   1.3.3         |
| Browser          |     Firefox       |
| Operating System |    Fedora        |

dodas:
Closes #3419

## 📝 Description

Respect `FormControl` context such as `isInvalid`, `isDisabled`, `isReadOnly`, or `isRequired`.

Currently `FormControl` is not respected at all.

## 🚀 New behavior

If `<Checkbox>` is nested inside a `<FormControl>`, its props are respected.

No

## 📝 Additional Information

Any thoughts on testing input components inside `FormControl`?

### DIFF
--- a/.changeset/mighty-mugs-scream.md
+++ b/.changeset/mighty-mugs-scream.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/checkbox": patch
+---
+
+`Checkbox` now respect `FormControl` context such as `isInvalid`, `isDisabled`,
+`isReadOnly`, or `isRequired`.
--- a/packages/checkbox/src/checkbox.tsx
+++ b/packages/checkbox/src/checkbox.tsx
@@ -9,9 +9,10 @@ import {
   useMultiStyleConfig,
   HTMLChakraProps,
 } from "@chakra-ui/system"
+import { useFormControl } from "@chakra-ui/form-control"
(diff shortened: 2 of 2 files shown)


with-heart (changes requested):
This looks great! Mind adding a test case or two to validate that it works correctly with `FormControl`? Or I can add them if you'd like?

segunadebayo:
Rather than tweaking the internal code to account for this. I'm going to suggest a solution in userland (in that issue).

The Checkbox was initially designed to work with CheckboxGroup and it'll get more complex on our side to handle these use cases.

- Single Checkbox should read from FormControlContext. In this case, it should read `isInvalid`
- CheckboxGroup should not read from FormControl's Context since it doesn't make sense for all checkboxes to be `isInvalid`

Rather than adding more implicit code, I think the user can read from `useFormControlContext` and set the `isInvalid` prop themselves. Not sexy, but it works and would prevent us from adding more code.

PR #3463, opened 2021-02-26
feat: let there be css variables 🥳

### LINKED ISSUE #3319: Components should accept an array of `css` objects
# 🐛 Bug report

Chakra's TypeScript definitions suggest that a Styled Component's `css` prop accepts an array of objects that are then composed. Instead, the `css` prop is ignored completely when an array gets passed.

## 💥 Steps to reproduce

In the example below, `SpecialList` should define additional CSS rules that are merged with the regular `css` prop.

## 💻 Link to reproduction

https://codesandbox.io/s/upbeat-worker-x6l79?file=/src/App.tsx

## 🧐 Expected behavior

[Emotion allows to compose multiple styles](https://emotion.sh/docs/object-styles#arrays) by passing an array of style objects to `css(...)`. Chakra's and Emotion's TypeScript definitions suggest that this works the same, out of the box.

https://github.com/chakra-ui/chakra-ui/blob/b77049f161256b67fbf5afec99d90636a814f053/packages/system/src/system.types.tsx#L61-L64

```ts
// https://github.com/emotion-js/emotion/blob/c9b57f36e99f8cfe86c971ddb29c7a5d70644bc1/packages/serialize/types/index.d.ts#L53-L63

export interface ArrayInterpolation<Props> // 👈
  extends Array<Interpolation<Props>> {}
// ...
export type Interpolation<Props> =
  | InterpolationPrimitive
  | ArrayInterpolation<Props> // 👈
  | FunctionInterpolation<Props>
```

## 🌍 System information

| Chakra UI        |  `"@chakra-ui/react": "^1.3.2"`           |
| Browser          |  Safari 14.0.3           |
| Operating System |  macOS 10.15.7          |

## 📝 Additional information

This could possibly be fixed if `getStyledObject` just returned an array instead of `Object.assign`, since it is consumed by Emotion anyway.

https://github.com/chakra-ui/chakra-ui/blob/b77049f161256b67fbf5afec99d90636a814f053/packages/system/src/system.ts#L102-L108

```diff
 // Merge the computed css object with styles in css prop 
- const cssObject: Interpolation<StyleResolverProps> = Object.assign( 
+ const cssObject: Interpolation<StyleResolverProps> = [
   computedCSS, 
   isFunction(cssProp) ? cssProp(theme) : cssProp, 
- ) 
+ ]
  
 return cssObject 
```

segunadebayo:
**Closes #3319**

**Description**
The CSS spec has evolved and over time I think we need to start embracing the power of CSS variables and step away from the styled-system "magic" 😂 we have under the hood + emotion's ThemeProvider.

From a technical standpoint, this PR improves the speed of style computation by 2.5x and reduces the bundle size of the styled-system package by almost half.

While there is still a small runtime footprint, in the future, I think we can drive the runtime to near zero.

**Technical References**
https://epicreact.dev/css-variables/
https://www.joshwcomeau.com/css/css-variables-for-react-devs/
https://codesandbox.io/s/css-variables-vs-themeprovider-df90h?file=/src/both/theme-provider.js

**Breaking Change?**
At the moment, I tried to do this without any breaking change. I'm hopeful this is the same experience for consumers of Chakra UI. **Short answer: No breaking change**

**Approach**
Instead of doing too much at runtime, we generate the CSS variables ahead of time, attach them to the `:root` element, and build a lookup object for theme tokens. The only thing we do now is to look up the CSS variable for any token or output the value directly.


packages/styled-system/src/css-var.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export function getTransformTemplate(type?: "gpu") {
+  const withGpu = "translate3d(var(--translate-x, 0), var(--translate-y, 0), 0)"
+  const basic = [
+    "translateX(var(--translate-x, 0))",
+    "translateY(var(--translate-y, 0))",
+  ].join(" ")
+  const transform = type === "gpu" ? withGpu : basic
+  return [transform, ...transformTemplate].join(" ")
+}
```
TimKolberger, line 113: This function does two things at once. I think we can add more readability by splitting it up in two separate functions:

```suggestion
export function getTransformTemplate() {
  return [ 
    "translateX(var(--translate-x, 0))",
    "translateY(var(--translate-y, 0))",
    ...transformTemplate
  ].join(" ")
}

export function getTransformGpuTemplate(type?: "gpu") {
 return [
   "translate3d(var(--translate-x, 0), var(--translate-y, 0), 0)",
    ...transformTemplate
  ].join(" ")
}
```

packages/styled-system/src/css-var.ts
```diff
@@ -0,0 +1,186 @@
+import { Dict, isCssVar, isObject, pick } from "@chakra-ui/utils"
+import analyzeBreakpoints from "./breakpoints"
+import type { CSSMap, WithCSSVar } from "./types"
+
+const separator = ">"
+
+const replace = (value: string, str = ".") => value.replace(/>/g, str)
+
+export const toVar = (value: string) => `var(${value})`
```
TimKolberger, line 9: What do you think about adding a prefix for all chakra CSS variables to prevent name clashes with existing ones?

TimKolberger (commented):
Outstanding work @segunadebayo! This is a big step into the future of Chakra UI - love it.

I know this is still in progress, but I had to look into it 😆

tundera:
Really excited for this one. I like the approach some have taken to get around the flashing issues for color mode on first page loads using CSS variables. Reading the post on EpicReact, it sounds like the runtime footprint from emotion won't be as noticeable since there won't be as many browser repaints.

packages/styled-system/src/sort.ts
```diff
@@ -0,0 +1,34 @@
...
+
+function byOrder([keyA]: [string, any], [keyB]: [string, any]) {
+  const aSortIndex = findLastIndex(cssOrder, (key) => keyA.startsWith(key))
+  const bSortIndex = findLastIndex(cssOrder, (key) => keyB.startsWith(key))
+  const aOrder = aSortIndex === -1 ? Number.MAX_SAFE_INTEGER : aSortIndex
+  const bOrder = bSortIndex === -1 ? Number.MAX_SAFE_INTEGER : bSortIndex
+
+  const indexDiff = aOrder - bOrder
+  if (indexDiff !== 0) return indexDiff
+  return collator.compare(keyA, keyB)
```
TimKolberger, line 27: @segunadebayo I think the alphabetical sorting leads to some unexpected side effects. Please compare the IconButton in this branch and `main`. 

I think we should return here `0` and mention in our docs that in an emotion css object the order of properties behave like CSS properties -> the order matters, if they can override one another like `padding` and `padding-left`

segunadebayo: Got it! I'll take a look at this.

segunadebayo: @TimKolberger. I agree! This leads to unexpected behavior. I'll return `0` instead.

segunadebayo:
I'm not sure why the Files Changed got bloated after merging changes from `main`. Mostly changed the styled-system package.

.changeset/good-islands-listen.md
```diff
@@ -0,0 +1,6 @@
+---
+"@chakra-ui/theme": minor
+---
+
+Add new config property `cssVarPrefix` to add a custom prefix for CSS variables.
+It defaults to `ck`.
```
TimKolberger, line 6:
```suggestion
It defaults to `chakra`.
```

TimKolberger (approved):
Looks good to me :tada: This is gonna be awesome  🍾 ✨

PR #3465, opened 2021-02-26
feat: add chakra ui icon

Daggy1234:
## 📝 Description

I realised there is no chakra-ui icon for those of us who would like to showcase an icon!


segunadebayo:
I'd recommend keeping this local to your project. I don't think it belongs to the icons package.

For those who would like to showcase the Chakra UI, we've provided a directory for all Chakra UI logo assets at the root of the repo.

Thank you for this!

PR #3467, opened 2021-02-26
Allow using different Motion components in transitions

### LINKED ISSUE #3051: Request: allow using components other than div for transitions
# 🚀 Feature request

I want to animate the addition of a row to a table. However, the Collapse animation doesn't really work properly because it (and all other animations) have a hard-coded `<motion.div/>` component: https://github.com/chakra-ui/chakra-ui/blob/9ca61eca9a8be06cef650341b33603d429a9b66b/packages/transition/src/collapse.tsx#L146

## ✅ Proposed solution or API

Ideally, there would be a prop that allows a custom component to be passed in (I want to use `<motion.tr/>`).

codetheweb:
Closes #3051

## 📝 Description

This allows consumers to pass a Motion component to transitions like Collapse to override the default `<motion.div/>`.

No.

## 📝 Additional Information

I've only implemented this for Collapse so far; I wanted to make sure everything looked good so far before I implement it for the other transitions.

### DIFF
--- a/packages/transition/src/collapse.tsx
+++ b/packages/transition/src/collapse.tsx
@@ -62,6 +62,12 @@ export interface CollapseOptions {
    * @default "auto"
    */
   endingHeight?: number | string
+
+  /**
+   * The Motion element to render as.
+   * @default motion.div
+   */
+  as?: React.ElementType
 }
 
 export type ICollapse = CollapseProps
@@ -139,11 +145,13 @@ export const Collapse = React.forwardRef<HTMLDivElement, CollapseProps>(
       custom,
     }
 
+    const MotionComponent = props.as ?? motion.div
+
     if (unmountOnExit) {
       return (
         <AnimatePresence initial={false} custom={custom}>
           {isOpen && (
-            <motion.div
+            <MotionComponent
               {...ownProps}
               initial="exit"
               animate="enter"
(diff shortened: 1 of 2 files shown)


packages/transition/src/collapse.tsx
```diff
@@ -62,6 +62,12 @@ export interface CollapseOptions {
    * @default "auto"
    */
   endingHeight?: number | string
+
+  /**
+   * The Motion element to render as.
+   * @default motion.div
+   */
+  as?: React.ElementType
```
segunadebayo, line 70: I understand this is a work in progress but I thought to add that this should be a motion "element type" since the collapse animation won't work if it just any react element type.

packages/transition/src/collapse.tsx
```diff
@@ -139,11 +145,13 @@ export const Collapse = React.forwardRef<HTMLDivElement, CollapseProps>(
       custom,
     }
 
+    const MotionComponent = props.as ?? motion.div
```
segunadebayo, line 148: Perhaps it might be better to add a check if `props.as` is a motion element as well. Maybe throw if it's not a motion element? I'll leave that to you.

codetheweb:
I'm going to close this for now as I no longer need this functionality and I'm struggling with the types needed to allow for generic components.

PR #3477, opened 2021-02-28
fix(popper): sync popper instance state

### LINKED ISSUE #3160: [Tooltip] Using the "auto" `placement` causes arrow to change directions when `isOpen` is toggled
# 🐛 Bug report

Hi! I'm trying to use the `placement="auto"` prop on the `Tooltip`. This is causing behavior that can be seen in this video:

https://user-images.githubusercontent.com/69278191/105784186-27278300-5f2d-11eb-97b5-78d2e5de4e31.mp4

What I found out is that in the `use-popper.ts` file there is a `finalPlacement` variable that gets set by the `use-popper` hook. In that `use-popper` there's a discrepancy between the `popperState` and the `popperOptions`.

## 💥 Steps to reproduce

1. Use the tooltip with `placement="auto"`
```
  <Tooltip
    aria-label="Button Clicked"
    label="Button Clicked"
    placement="auto"
    bg="bgColor"
    color="textColor"
  >
    <Button>
      Open Modal
    </Button>
  </Tooltip>
```

## 💻 Link to reproduction

[CodeSandbox](https://codesandbox.io/s/blissful-hodgkin-l4vpl?file=/src/App.js)

## 🧐 Expected behavior

When you "re-hover" or "re-click" I'd expect the arrow to be in the same location

## 🧭 Possible Solution

Not entirely sure...but I think it might have to do with the `useSafeLayoutEffect` not being called again in `react-popper.ts`

## 🌍 System information

| Chakra UI        |    ^1.0.3        |
| Browser          |  Chrome 88.0.4324.96          |
| Operating System |         MacOS Catalina 10.15.6    |

tomdohnal:
Closes #3160

## 📝 Description

This fix makes sure that we don't return stale popper state from the `usePopper` hook in `popper/react-popper.ts` which was causing wrong popper behaviour with using an auto-placed tooltip together with `AnimatedPresence` component.

When we use an auto-places `Tooltip` which has an `AnimatedPresense` component in it, the `usePopper` hook returns stale state causing incorrect arrow positioning.

## 🚀 New behavior

I changed the `usePopper` behaviour to keep track of the PopperJS state property in a `useState` hook and update it in a modifier alongside updating `style` and `attrs` state.

No

## 📝 Additional Information

I've got two additional ideas:
1) I think it would be a good idea to refactor the `react-popper` a bit so that we don't both the PopperJS `state` from it as well as `style`/`attributes` (as separate object keys). As `style`/`attributes` are in fact a subset of the `state` I think this can cause problems with not having a single source of truth and not knowing if we should trust the `style`/`attrs` in PopperJS `state` or the `style`/`attributes` which are returned from the `usePopper` hook as individual properties.

2) I tried writing a test for my fix but I failed due to the fact that the tests are run in a JSDom environment where calling things `getBoundingClientRect()` on elements returns an "empty" object (something like `{ top: 0, bottom: 0, ...}`).
I think running tests in JSDom makes it really hard to test these parts of ChakraUI which deal with positioning elements and would like to know if you'd be open to trying to run these test in a real browser (puppeteer / cypress / testcafe should all be supported by React Testing Library)

### DIFF
--- a/.changeset/odd-cooks-beam.md
+++ b/.changeset/odd-cooks-beam.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/popper": patch
+---
+
+make the react-popper state in sync with the internal popper instance state
--- a/packages/popper/src/react-popper.ts
+++ b/packages/popper/src/react-popper.ts
@@ -6,6 +6,7 @@ import {
   Modifier,
   Options as PopperOptions,
   VirtualElement,
+  State as PopperState,
 } from "@popperjs/core"
 import { dequal } from "dequal"
 import * as React from "react"
@@ -56,6 +57,7 @@ export function usePopper(
     modifiers: options.modifiers || EMPTY_MODIFIERS,
   }
 
+  const [popperState, setPopperState] = React.useState<PopperState | null>(null)
   const [styles, setStyles] = React.useState<State["styles"]>({
     popper: {
       position: optionsWithDefaults.strategy,
@@ -73,6 +75,8 @@ export function usePopper(
       phase: "write",
       fn: ({ state }) => {
         const elements = Object.keys(state.elements)
+
+        setPopperState(state)
 
         setStyles(resolve(state.styles, elements))
         setAttrs(resolve(state.attributes, elements))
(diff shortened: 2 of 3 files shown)


segunadebayo:
Great work here @tomdohnal. I agree with your additional thoughts as well.

Let's go with your first point of using `state` as the source of truth and deriving styles and attrs from it. Feel free to refactor as needed and I will review.

tomdohnal:
OK, thanks for merging @segunadebayo 
I'll try to send a PR with a refactor at the end of this week

PR #3481, opened 2021-03-01
chore: speed up start

Zaynex:
Closes # 

## 📝 Description

> Add a brief description

we must build all pkgs first if we want start the project in local environment.

## 🚀 New behavior

No need build pkgs, just start it!

No

## 📝 Additional Information

add index.ts in pkg's root directory, and export pkgs from *./src*

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "contributors:gen": "all-contributors generate",
     "prestart": "yarn && yarn bootstrap",
     "boot": "yarn prestart && yarn build",
-    "start": "lerna run --scope=@chakra-ui/babel-plugin build && yarn storybook",
+    "start": "yarn prebuild && yarn storybook",
     "start:all": "yarn build && concurrently \"yarn docs dev\"  \"yarn storybook\" ",
     "prebuild": "yarn pkg babel-plugin build",
     "build": "lerna run build --ignore @chakra-ui/test-utils --no-private --stream",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "contributors:gen": "all-contributors generate",
     "prestart": "yarn && yarn bootstrap",
     "boot": "yarn prestart && yarn build",
-    "start": "yarn build && yarn storybook",
+    "start": "lerna run --scope=@chakra-ui/babel-plugin build && yarn storybook",
     "start:all": "yarn build && concurrently \"yarn docs dev\"  \"yarn storybook\" ",
     "prebuild": "yarn pkg babel-plugin build",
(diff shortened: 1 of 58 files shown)


with-heart (approved):
I love that this works.

I'm also not entirely sure I understand _why_ it works. Why is `package.json` ignored in favor of `index.ts`?

Zaynex:
> I love that this works.
> 
> I'm also not entirely sure I understand _why_ it works. Why is `package.json` ignored in favor of `index.ts`?

Actually, It follow the principle of require.

According to the [official documentation](https://nodejs.org/api/modules.html#modules_folders_as_modules):
> If there is no package.json file present in the directory, or if the "main" entry is missing or cannot be resolved, then Node.js will attempt to load an index.js or index.node file out of that directory. For example, if there was no package.json file in the previous example, then require('./some-library') would attempt to load:
./some-library/index.js
./some-library/index.node


if node can't resolve pkg/dist/cjs/index.js, then it will try to resolve './pkg/index.ts'.

Zaynex:
It was inspired by react-spectrum.
https://github.com/adobe/react-spectrum/blob/30377b7558c3e621e8e52d6daf75bd69bc93dbb9/package.json#L18
https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/actiongroup/index.ts

with-heart:
- [x] `yarn start` works locally on fresh install of `node_modules`
- [x] Pipeline passed
- [x] CSB deployment works fine

I think we're good to go here.

package.json
```diff
@@ -33,7 +33,7 @@
     "contributors:gen": "all-contributors generate",
     "prestart": "yarn && yarn bootstrap",
     "boot": "yarn prestart && yarn build",
-    "start": "yarn build && yarn storybook",
+    "start": "lerna run --scope=@chakra-ui/babel-plugin build && yarn storybook",
```
with-heart, line 36: We don't have this documented anywhere but we've been trying to move towards using `manypkg` for handling most things. Could use the same script as `prebuild` in place of `lerna run`, but it's not a priority so I'll leave it up to you if you'd like to change it. 

The rest of this is great! Thanks @Zaynex!

with-heart (approved):

package.json
Zaynex: Done.

TimKolberger (approved):
I think this is safe to merge, because none of the edited files are getting published.
Interesting approach! ✨

segunadebayo:
Thanks for this improvement @Zaynex. I'll definitely help speed up our development process! 💖

PR #3488, opened 2021-03-01
fix(transition): use minimal `m` component

frysztak:
(Partially) closes #3369

## 📝 Description

`framer-motion>3.8.0` [supports tree-shaking in Webpack](https://github.com/framer/motion/pull/1008). by switching to `m` component (and adding obligatory `MotionConfig`) we can reduce bundle size a little. for a simple `<Collapse />` animation, gzipped bundle size dropped by 7 kB.

Chakra bundles entire `framer-motion` library

## 🚀 New behavior

Chakra bundles smaller subset of `framer-motion`

no.

### DIFF
--- a/.changeset/curly-days-bake.md
+++ b/.changeset/curly-days-bake.md
@@ -1,0 +1,11 @@
+---
+"@chakra-ui/checkbox": minor
+"@chakra-ui/menu": minor
+"@chakra-ui/modal": minor
+"@chakra-ui/popover": minor
+"@chakra-ui/toast": minor
+"@chakra-ui/tooltip": minor
+"@chakra-ui/transition": minor
+---
+
+use minimal `m` component to reduce bundle size
--- a/.changeset/fuzzy-glasses-count.md
+++ b/.changeset/fuzzy-glasses-count.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/transition": minor
+---
+
+use minimal `m` component to reduce bundle size
--- a/packages/checkbox/src/checkbox-icon.tsx
+++ b/packages/checkbox/src/checkbox-icon.tsx
@@ -1,5 +1,11 @@
 import { chakra, PropsOf } from "@chakra-ui/system"
-import { AnimatePresence, CustomDomComponent, motion } from "framer-motion"
+import {
+  AnimatePresence,
+  CustomDomComponent,
+  domAnimation,
+  LazyMotion,
+  m as motion,
+} from "framer-motion"
 import * as React from "react"
 
 // @future: only call `motion(chakra.svg)` when we drop framer-motion v3 support
@@ -62,25 +68,27 @@ const IndeterminateIcon = (props: PropsOf<typeof MotionSvg>) => (
 
 const CheckboxTransition = ({ open, children }: any) => (
   <AnimatePresence initial={false}>
-    {open && (
-      <motion.div
-        variants={{
-          unchecked: { scale: 0.5 },
-          checked: { scale: 1 },
-        }}
-        initial="unchecked"
-        animate="checked"
-        exit="unchecked"
-        style={{
-          display: "flex",
-          alignItems: "center",
-          justifyContent: "center",
-          height: "100%",
-        }}
-      >
-        {children}
-      </motion.div>
-    )}
+    <LazyMotion features={domAnimation}>
+      {open && (
+        <motion.div
+          variants={{
+            unchecked: { scale: 0.5 },
+            checked: { scale: 1 },
+          }}
+          initial="unchecked"
+          animate="checked"
+          exit="unchecked"
+          style={{
+            display: "flex",
+            alignItems: "center",
+            justifyContent: "center",
+            height: "100%",
+          }}
+        >
+          {children}
+        </motion.div>
+      )}
+    </LazyMotion>
   </AnimatePresence>
 )
 
--- a/packages/menu/src/menu.tsx
+++ b/packages/menu/src/menu.tsx
@@ -13,7 +13,13 @@ import {
   useStyles,
 } from "@chakra-ui/system"
 import { cx, runIfFn, __DEV__ } from "@chakra-ui/utils"
-import { CustomDomComponent, motion, Variants } from "framer-motion"
+import {
+  CustomDomComponent,
+  domAnimation,
+  m as motion,
+  Variants,
+  LazyMotion,
+} from "framer-motion"
 import * as React from "react"
 import {
   MenuDescendantsProvider,
@@ -167,22 +173,24 @@ export const MenuList = forwardRef<MenuListProps, "div">((props, ref) => {
       {...positionerProps}
       __css={{ zIndex: props.zIndex ?? styles.list?.zIndex }}
     >
-      <MotionDiv
-        {...menulistProps}
-        /**
-         * We could call this on either `onAnimationComplete` or `onUpdate`.
-         * It seems the re-focusing works better with the `onUpdate`
-         */
-        onUpdate={onTransitionEnd}
-        className={cx("chakra-menu__menu-list", menulistProps.className)}
-        variants={motionVariants}
-        initial={false}
-        animate={isOpen ? "enter" : "exit"}
-        __css={{
-          outline: 0,
-          ...styles.list,
-        }}
-      />
+      <LazyMotion features={domAnimation}>
+        <MotionDiv
+          {...menulistProps}
+          /**
+           * We could call this on either `onAnimationComplete` or `onUpdate`.
+           * It seems the re-focusing works better with the `onUpdate`
+           */
+          onUpdate={onTransitionEnd}
+          className={cx("chakra-menu__menu-list", menulistProps.className)}
+          variants={motionVariants}
+          initial={false}
+          animate={isOpen ? "enter" : "exit"}
+          __css={{
+            outline: 0,
+            ...styles.list,
+          }}
+        />
+      </LazyMotion>
     </chakra.div>
   )
 })
--- a/packages/modal/src/modal-transition.tsx
+++ b/packages/modal/src/modal-transition.tsx
@@ -1,6 +1,11 @@
 import { chakra, ChakraProps } from "@chakra-ui/system"
 import { scaleFadeConfig, slideFadeConfig } from "@chakra-ui/transition"
-import { HTMLMotionProps, motion } from "framer-motion"
+import {
+  domAnimation,
+  HTMLMotionProps,
+  LazyMotion,
+  m as motion,
+} from "framer-motion"
 import * as React from "react"
 
 export interface ModalTransitionProps
@@ -31,6 +36,10 @@ export const ModalTransition = React.forwardRef(
   (props: ModalTransitionProps, ref: React.Ref<any>) => {
     const { preset, ...rest } = props
     const motionProps = transitions[preset]
-    return <Motion ref={ref} {...(motionProps as ChakraProps)} {...rest} />
+    return (
+      <LazyMotion features={domAnimation}>
+        <Motion ref={ref} {...(motionProps as ChakraProps)} {...rest} />
+      </LazyMotion>
+    )
   },
 )
--- a/packages/modal/src/modal.tsx
+++ b/packages/modal/src/modal.tsx
@@ -22,8 +22,10 @@ import {
 import { createContext } from "@chakra-ui/react-utils"
 import {
   AnimatePresence,
+  domAnimation,
   HTMLMotionProps,
-  motion,
+  LazyMotion,
+  m as motion,
   usePresence,
 } from "framer-motion"
 import * as React from "react"
@@ -355,13 +357,15 @@ export const ModalOverlay = forwardRef<ModalOverlayProps, "div">(
     const motionProps: any = motionPreset === "none" ? {} : fadeConfig
 
     return (
-      <MotionDiv
-        {...motionProps}
-        __css={overlayStyle}
-        ref={ref}
-        className={_className}
-        {...rest}
-      />
+      <LazyMotion features={domAnimation}>
+        <MotionDiv
+          {...motionProps}
+          __css={overlayStyle}
+          ref={ref}
+          className={_className}
+          {...rest}
+        />
+      </LazyMotion>
     )
   },
 )
--- a/packages/popover/src/popover-transition.tsx
+++ b/packages/popover/src/popover-transition.tsx
@@ -1,5 +1,11 @@
 import { chakra, HTMLChakraProps } from "@chakra-ui/system"
-import { HTMLMotionProps, motion, Variant } from "framer-motion"
+import {
+  domAnimation,
+  HTMLMotionProps,
+  LazyMotion,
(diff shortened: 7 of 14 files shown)


vmizg:
Hi @frysztak, great addition! Do you have a timeline for the next patch release with this change included?

frysztak:
> 
> 
> Hi @frysztak, great addition! Do you have a timeline for the next patch release with this change included?

hi @vmizg, do you mean Chakra-UI or framer-motion? framer-motion is already released, Chakra is not up to me 🚀

vmizg:
> > Hi @frysztak, great addition! Do you have a timeline for the next patch release with this change included?
> 
> hi @vmizg, do you mean Chakra-UI or framer-motion? framer-motion is already released, Chakra is not up to me 🚀

I meant Chakra. Sorry, did not notice! 
I think this should be prioritized.

frysztak:
hi @with-heart @segunadebayo, not to nag you or anything, but are you interested in merging this? judging by reactions, some people are waiting for this :). this isn't an invasive PR, so it shouldn't take too long to review

with-heart:
Hi @frysztak, sorry for not getting back to this sooner! 

I would definitely love to have this because who doesn't love a smaller bundle size at almost no cost!

Unfortunately we would need to consider this a breaking change unless we could conditionally utilize `m` or the traditional `motion` component based on the user's current `framer-motion` version. Since our `peerDependencies` range for `framer-motion` is currently `3.x || 4.x`, we can't add functionality that is only supported by part of that range.

So I think our options are:

- Plan for our next `major` version release and include this in that release. I'd like to move to only supporting `framer-motion@4` so we could utilize all of the new size-reducing goodies that are on offer, and a `major` bump would allow us to do that.
- Release this as a `minor` version bump that includes conditional checks/functionality, so we can simultaneously support both `<3.8` and `>=3.8`.
- Why not both? 😁

frysztak:
hi @with-heart, thanks for getting back to me.

there's one thing I don't understand. I haven't been following framer-motion's development, but my framer-motion PR got merged and released as v3.8. `framer-motion<3.8.0` would work fine as well, they just won't tree-shake very well.

did something change with v4? I can see that `LazyMotion` got added, which is great - perhaps we can reduce bundle size even further (just as a separate PR 😉). but from what I can see, both [m component](https://github.com/framer/motion/blob/50634672504f03d9203ac793304b1248a4755371/src/render/dom/motion-minimal.ts) and [MotionConfig](https://github.com/framer/motion/blob/50634672504f03d9203ac793304b1248a4755371/src/components/MotionConfig/index.tsx) are still there. seems like all the pieces are there to support both `3.x || 4.x` without any code changes.

wildpow:
Any word on when this PR will get merged?

TheThirdRace:
**_Note: read with a normal tone, I'm not panicking or anything ;)_**

The pressing issue here is that framer-motion takes 27.98 KB (gzipped) in our bundle size **right now**.

Just to put things in perspective, the only package a normal website will use that will surpass framer-motion in gzip size is React-DOM... And nothing comes close to framer-motion gzip size in 3rd place.

Any idea what's the effect of so much **unused** Javascript in our bundles? The download time, the parsing time, the execution, etc. Web Core Vitals are tanking...

Such size is completely crazy given you could mostly replace Framer-Motion with 10 CSS classes and a few rules for like 1KB max, not even gzipped...

Given how critical this issue is for most people, the needs of the many should outweigh the needs of the select few that won't update their stuff... At this point, if you're not using framer-motion's latest version (3.x or 4.x), you're already doing yourself a disservice. There's absolutely no need for those people to hold everybody else hostage.

frysztak:
@TheThirdRace it's hard to disagree with you :). I was also under the impression that usage of framer-motion is an overkill. But then - I don't know core members' intentions. If you're planning to include complex animations in the library, it kinda makes sense to use framer-motion even for simple ones.

This PR is getting pretty old at this point, but I still see overinflated bundle size as a cause worth fighting for. So please @segunadebayo @with-heart let me ask you - if I refactored transitions package to exclude framer-motion, would you be open to reviewing and ultimately merging it? I don't want to pursue this in vain, if it's again your vision for this (may I add - awesome! 🚀 ) library.

TheThirdRace:
I think Framer-motion is a great option, the problem is really the fact Chakra is using the wrong component because it cannot be tree shaken.

Simply integrating this PR that replace `<motion>` by `<m>` and release Chakra UI as 2.x would solve the problem.

The argument made by @with-heart was that it wouldn't be backward compatible for some users that don't have the latest version of Framer-Motion (3.x). He suggests it would need to wait until Chakra 2.x to allow such a breaking change.

My previous post is more about why not do this now?  This is absolutely a required thing ASAP for most people using Chakra.

I'm not privy as to why it hasn't been done yet. My personal opinion is that Chakra is more than stable enough to seal 1.x as-is and start 2.x.

Sometimes I feel developers fear bumping  the major revision number without having a complete rewrite of something major. But that's not semantic versioning's purpose to begin with. Changing the major revision number is simply indicating there is at least 1 change that isn't backward compatible. It doesn't really matter how crucial how big or small that change is. If Chakra gets to 14.x in the next year, so be it... Eventually, it will dial down, there won't be so many breaking changes anymore, but for now, it's normal to increase the major revision number quite frequently given all the plans Chakra has.

frysztak:
> The argument made by @with-heart was that it wouldn't be backward compatible for some users that don't have the latest version of Framer-Motion (3.x). He suggests it would need to wait until Chakra 2.x to allow such a breaking change.

except that it would. `<m>` component is old news (it exists since at least 2.9). it works with both framer-motion 3.x and 4.x. it only tree-shakes with >= 3.8.

with-heart:
Yeah, what I said previously was definitely a misunderstanding about when the `m` component was introduced. For some reason I was thinking it was added in 4.0, but it has existed since 3.0, so it's fine to merge. It wasn't addressed/merged sooner because I haven't been keeping up with things as much as I need to be.

@frysztak Want to handle the conflicts? I can do it if not!

> Sometimes I feel developers fear bumping the major revision number without having a complete rewrite of something major. But that's not semantic versioning's purpose to begin with.

I agree that they do fear that. I'd prefer we be more like `jest` in regard to how often we bump `major`, and think that frequent `major` bumps mean more consumer version migrations are required to keep up but there would be far less changes to make per migration because there would be less changes per `major`. 

But that's not how it has worked out so far unfortunately. 

> So please @segunadebayo @with-heart let me ask you - if I refactored transitions package to exclude framer-motion, would you be open to reviewing and ultimately merging it? 

I'd prefer to merge this PR for now and have a separate discussion about `framer-motion` overall if need be.

frysztak:
@with-heart that's great 🎉 ! I'll handle the conflicts and ping you when I'm done

with-heart:
@frysztak Thank you ❤️

frysztak:
@with-heart looks like you were right. framer-motion v4 changed API so much it seems to be impossible to support both v3 and v4. for example, v3 code looks something like this:
```jsx
<AnimatePresence>
    <MotionConfig features={[AnimationFeature]}>
        <motion.div ... />
    </MotionConfig>
</AnimatePresence>
```

and v4 code looks something like this:
```jsx
<AnimatePresence>
    <LazyMotion features={domAnimation}>
        <motion.div ... />
    </LazyMotion >
</AnimatePresence>
```

some options I pondered:
1. importing everything (`import * as framer from 'framer-motion'`) and checking for appropriate properties would work, but also throw tree-shaking out the window
2. splitting v3 and v4-specific code into separate files (it would require using JS or littering TS files with `ts-ignore` - first red flag). then we can import `m`, check for `custom` property and then lazy-load (second red flag) appropriate component
3. using something like [ifdef-loader](https://github.com/nippur72/ifdef-loader), although I'm not convinced it would even work

if you have any ideas how to tackle this, please share. but right now it looks like we might have to wait till you guys are ready to bump peer dependency to v4

with-heart:
@frysztak As much as I love being right, I would prefer if I had been wrong in this case 😆 

Unfortunately I think you're right that it'd be best to wait until we can bump. We don't have a target date for v2, but it's definitely something we're working towards rn. I'll add this to the milestone so we can include it as part of that change.

Thanks for all your effort in putting this together! We'll get it in eventually because there's definitely a need for it.

mattcroat:
I'm so glad this is being considered. :+1:

I went down the rabbit hole the other day since I had jank animations on the first-page load using Next.js. I noticed the content was on the page but if it had animations it wouldn't show until the chunk loaded (I guess that explains why Lighthouse was mad about the first contentful paint even if it loads fast for me). When I opened it I found the exact same issue described here. I noticed the chunk is around **~90 kB** for me. :scream:

I hope this gets addressed as it's a huge issue when you're using a low-powered device, or slower internet since the loading of the chunk with Next.js blocks showing content on the page because it can't be animated.

StevenHallLtd:
First time Chakra user here – initially super-impressed and excited about using this for most of our projects going forwards, but then the bundle size generally is a bit disappointing.  I really love the direction you're heading in, but I would politely suggest you prioritise this, and as @TheThirdRace says – by all means respect SemVer, but please don't feel held back by it* 🙂

_*written on Google Chrome **v91**_

frysztak:
@with-heart I rebased this PR so that only framer-motion v4 is supported. two things that worry me:
1. more components are now using framer-motion. IIRC, previously only `transition` used it, now it's used by checkbox, menu, modal, popover, toast and tooltip components
2. for a simple `Collapse` transition, we'd get about 7 kB savings (not that much to begin with), but now it's only 5 kB

TheThirdRace:
> @with-heart I rebased this PR so that only framer-motion v4 is supported. two things that worry me:
> 
> 1. more components are now using framer-motion. IIRC, previously only `transition` used it, now it's used by checkbox, menu, modal, popover, toast and tooltip components
> 2. for a simple `Collapse` transition, we'd get about 7 kB savings (not that much to begin with), but now it's only 5 kB

Just to clarify, you're saying that even with the new tree-shakable `m` component, it will only reduce the gzip bundle by 5 to 7KB?

Because if that is true, maybe it would be wise to reconsider the use of `framer-motion` to begin with. There are alternatives far smaller that would fill the requirements just as well.

frysztak:
@TheThirdRace for a very simple example:
```tsx
export function CollapseEx() {
    const { isOpen, onToggle } = useDisclosure()
  
    return (
      <>
        <Button onClick={onToggle}>Click Me</Button>
        <Collapse in={isOpen} animateOpacity>
          <Box
            p="40px"
            color="white"
            mt="4"
            bg="teal.500"
            rounded="md"
            shadow="md"
          >
            Lorem ipsum etc
          </Box>
        </Collapse>
      </>
    )
  ```
  
bundle sizes look like this:

- Chakra UI master + Framer-Motion 4.1.17 -> **119.42 KB**
- Chakra UI this PR + Framer-Motion 4.1.17 -> **114.37 KB**

so a 5.05 KB reduction. it used to be closer to 7 KB when I opened this PR.

overall I agree with your sentiment. animations in Chakra aren't complicated, so using framer-motion does seem like an overkill. do you have recommendations for other libraries? [react-move](https://bundlephobia.com/package/react-move@6.4.0) seems to be the smallest one I could find

TheThirdRace:
@frysztak 
My knowledge on animations is basic at best, but when I'm looking for a package, I'm always looking at its stats in bundlephobia.com

Here are the stats for Framer-Motion https://bundlephobia.com/package/framer-motion@4.1.17

There's something not adding up though. If you look at the analysis, the `m` component should only be 6.6 KB compared to 30.1 KB for the `motion` component. (gzip size)

So I wonder how using the `m` component can result in only 5 KB of saving???

That would imply Chakra is using pretty much all Framer-Motion functionalities, importing almost all available modules and thus, remove practically any savings from tree-shaking...

As I said, something's fishy. Either Framer-Motion doesn't really tree-shake or you might have made a mistake somewhere. I mean no offense by it, I'm just trying to figure out why the numbers really don't add up to the expectations.

In any case, going back to the analysis on bundlephobia, you can see suggested alternatives at the bottom.

In those alternative, we can see `react-move` that you suggested, but also `react-spring` and `react-motion`.

As far as I know, the biggest competitor to `Framer-Motion` is `react-spring`. Even without tree-shaking, we should see a 10 KB reduction in bundle size. If any tree-shaking is applied, I would hope for at least an extra 5 KB. From the get go, `react-spring` looks like the best alternative.

As for the other 2 packages, I never heard of them so it's hard for me to give any opinion on them...

Personally, I don't like much animation libraries in react. I've experimented a bit in the past and all I could see is how much littering there was in the DOM just for a sliding effect... I mean, you don't need 2 or 3 wrappers if you can just add a class on the semantic tag you're already using...

I also have a bad experience with `Framer-Motion` in general. As soon as it was introduced in Chakra, in version 1.0.4 I think, my troubles started. Suddenly, I couldn't use Preact anymore because of some incompatibilities in the `Menu` component that would simply crash my page. So it added 20 KB to the bundle just for `Framer-Motion`, then another 40 KB for `react` + `react-dom`... It's a hard pill to swallow when you're forced to accept a 60 KB increase to your js bundle because of a design decision...

frysztak:
@TheThirdRace 
> 
> So I wonder how using the `m` component can result in only 5 KB of saving???
> 
> That would imply Chakra is using pretty much all Framer-Motion functionalities, importing almost all available modules and thus, remove practically any savings from tree-shaking...
> 
> As I said, something's fishy. Either Framer-Motion doesn't really tree-shake or you might have made a mistake somewhere. I mean no offense by it, I'm just trying to figure out why the numbers really don't add up to the expectations.

no offence taken ;). it's really simple: `m` component is just a part of the story, to use it you also need `LazyMotion` and `domAnimation`. check out the docs [here](https://www.framer.com/api/motion/guide-reduce-bundle-size/#available-features)

StevenHallLtd:
If that's the case then I would second the view that including any animation library, let alone such a large one, seems a bit of an overreach.  Framer is an incredibly powerful library, but I'm not convinced that a few well-crafted CSS animations wouldn't suffice here instead.  I've switched to Theme UI now, but I feel sad to be missing out on many of Chakra's better decisions.

Sorry to be such a party-pooper.  Free software and all that – I just hope you'll reconsider this one 🤞🏼

fschwalm:
Hello!
Any updates here?
i agree with @TheThirdRace!

thank you

valstu:
Also bumped into this while trying to shave of kilobytes from my bundle.

TimKolberger:
Since the savings for this major refactoring are only ~5kb we decided to drop this PR.
Addressing the bundle size one of our next focus points!

---

It's been a great time with this PR - I could always be sure to have reached the end of the PR list when I spotted it 🤡

TheThirdRace:
> Addressing the bundle size one of our next focus points!

That's an excellent news!

I created a discussion a while ago about the major pain points and possible leads for the bundle size: https://github.com/chakra-ui/chakra-ui/discussions/5277

If you're interested in having a more in-depth discussion about it, I would gladly take some time to explain and shed some light on the problems I encountered when trying to reduce my bundle size.

It's usually easier to talk about that kind of stuff, maybe over Discord, as it better conveys the nuances, missing information, etc., but this could be done through a Github discussion, a PM, etc.

Issue #3459, opened 2021-02-25
motion.custom() is deprecated. Use motion() instead


ViaxCo:
# 🐛 Bug report

> Provide a brief description of the bug here

This warning: `motion.custom() is deprecated. Use motion() instead` gets logged to the console whenever a new `chakra-ui` project is created with `framer-motion@3.7.0`.
Framer motion [changelog.](https://github.com/framer/motion/blob/main/CHANGELOG.md)

## 💥 Steps to reproduce

> Clear and concise reproduction steps are important for us to be able to triage
> your issue in a timely manner

## 💻 Link to reproduction

> Please create a minimal reproduction in CodeSandbox

https://codesandbox.io/s/delicate-glitter-hroux

## 🧐 Expected behavior

> Tell us what you expect to happen.

## 🧭 Possible Solution

> Not compulsory, but suggesting a fix would be great!

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Chakra UI        |            |
| Browser          |            |
| Operating System |            |

## 📝 Additional information

- Screenshots or code
- Notes or link to ideas

Mzaien:
i've had a look on chakra repo the motion.custom is used only in checkbox component and one storyblock /storybook (sorry if mis spilled )

smblee:
This breaks the `chakra-cli tokens` command for me.

`chakra-cli tokens src/styles/theme/theme.ts`

![image](https://user-images.githubusercontent.com/9781750/109548427-7b200d00-7a81-11eb-922d-881dc92fd5c0.png)

bartconsedine:
having the same issues and it is not loading my NextJS app as a result

Mzaien:
@segunadebayo i think it is solvable if it was only used in checkbox as mentioned here https://github.com/chakra-ui/chakra-ui/issues/3459#issuecomment-786892967 right?

dodas:
To solve this, we can either:

- bump required `framer-motion` to `^3.7.0` which introduced `motion()`, so we can use it.
Can we do it? Could it introduce some problems for users?
- do some heuristics to determine whether `motion()` is supported and conditionally use that, or `motion.custom()`.
I'm not yet sure how doable this option is.

codetective:
@dodas there's currently a PR for this implemementing your idea, hopefully the chakra team will review it soon. #3492

violabg:
[...]

PR #3492, opened 2021-03-01
chore: refactor motion custom

shunkakinoki:
hi @chakra-ui
Thanks for the awesome project!!!
The error was breaking `@chakra-ui/cli` so updated the command to the non deprecated version

### DIFF
--- a/examples/create-react-app/package.json
+++ b/examples/create-react-app/package.json
@@ -7,7 +7,7 @@
     "@chakra-ui/react": "*",
     "@emotion/react": "^11.1.4",
     "@emotion/styled": "^11.0.0",
-    "framer-motion": "^3.1.1",
+    "framer-motion": "^3.7.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1"
   },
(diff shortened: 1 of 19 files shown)


shunkakinoki:
Closes #3459

simonzhow (approved):

with-heart:
I'm not sure how I feel about this. It's potentially a breaking change for us to require that consuming projects upgrade `framer-motion` for what amounts to a `console.warn` message.

The issue in particular with `@chakra-ui/cli` is here: https://github.com/chakra-ui/chakra-ui/blob/0830ddcf0989700f9c8e23956726f22247de7ded/tooling/cli/src/command/tokens/index.ts#L28-L30

`console.warn` is reported to `stderr`, which means the deprecation warning is causing `@chakra-ui/cli` to throw this as an error.

I think we're be better off resolving the issue inside `@chakra-ui/cli` as a `console.warn` shouldn't justify forcing an upgrade of `framer-motion`.

cloud-walker:
framer-motion is not supporting the .custom version anymore

with-heart:
I'm going to close this as we can't proceed with these changes. In order to **force** users to upgrade a dependency, we'd need to publish a new `major` version (`@chakra-ui/react@2.0.0`) and I believe we can make some changes that will allow us to silence these issues without **requiring** dependency upgrades.

For now, #3623 adds support for `framer-motion` v4, so you'll be able to upgrade to the latest to silence these once that is merged.

with-heart:
Thank you for the effort though @shunkakinoki!

PR #3493, opened 2021-03-02
fix: fixed vertically center modal cutoff #2871

### LINKED ISSUE #2871: [Modal] Vertically centered modal cutoff on mobile
# 🐛 Bug report

On mobile devices vertically centered modals get cutoff at the top. This happens for modals that are bigger than the screen.

## 💥 Steps to reproduce

Open a modal on mobile that is vertically centered and higher than the screen.

## 💻 Link to reproduction

https://chakra-ui.com/docs/overlay/modal

## 🧭 Possible Solution

I believe this is not solvable with the current structure. I could fix it with making the modal container `display: block;` and adding one div element between the container and the modal content with `min-height: 100%; display: flex; align-items: center; justify-content: center;`.
This needs to be tested for other modal types.

## 🌍 System information

| Chakra UI        |    v1.0.3        |

nejcm:
Closes #2871

## 📝 Description

This PR fixes a bug in vertically centered modals that get cutoff when they are larger than the screens.

Currently vertically centered modals get cutoff at the top if they are higher than the screen.

## 🚀 New behavior

This fix fixes that issue by modifying some css styles on the modal container and adding one additional div wrapper.

## 📝 Additional Information

While this change passes all tests and was tested in browsers (Chrome, Edge, Firefox) it might still be good to test it on iOS. 
Also there might be a need to add some class names to the added div element but I will leave that up to the maintainers to decide.

While this change is not breaking when using the default modal theme settings it might need some changes for users running modified modal theme settings.

### DIFF
--- a/.changeset/great-months-check.md
+++ b/.changeset/great-months-check.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/modal": patch
+"@chakra-ui/theme": patch
+---
+
+Fixed vertically center modal cutoff by changing modal container styles and
+adding another wrapping div.
--- a/packages/modal/src/modal.tsx
+++ b/packages/modal/src/modal.tsx
@@ -234,13 +234,20 @@ export const ModalContent = forwardRef<ModalContentProps, "section">(
     }
 
     const dialogContainerStyles: SystemStyleObject = {
-      display: "flex",
+      display: "block",
       width: "100vw",
       height: "100vh",
       position: "fixed",
       left: 0,
       top: 0,
       ...styles.dialogContainer,
+    }
+
+    const dialogInnerContainerStyles: SystemStyleObject = {
+      display: "flex",
+      minHeight: "100%",
+      alignItems: "inherit",
+      justifyContent: "inherit",
     }
 
     const { motionPreset } = useModalContext()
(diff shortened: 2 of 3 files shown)


segunadebayo:
Hi @nejcm,

Thanks for looking into this. I think we add this note to the Modal component regarding this:

> If the content within the modal overflows beyond the viewport, don't use this prop. Try setting the overflow behavior instead.

_Reference: https://chakra-ui.com/docs/overlay/modal#make-modal-vertically-centered_

If you have a vertically centered modal and the content might overflow the viewport's height. I think it's safe to use the `overfllowBehavior` prop.

Adding an extra node might be a breaking change and we're very careful about that. Permit me to close this for now

nejcm:
Hi @segunadebayo ,

Thank you for the explanation. I understand the reasoning the and the concern but in my eyes this is currently a bug. 
Let me explain. The scrolling modal can not be vertically centered and even the example with the vertically centered modal on the docs page is broken on mobile landscape orientation. This tells me that the current implementation is not the best for all devices/screens. You can't predict how the content/text will break (what height it will take) in the modal on different screens so it can quite fast get cutoff. This means that vertically centered modals are only usable on desktops atm.

Best regards.
Nejc

Issue #3482, opened 2021-03-01
Update docs for Toast to include a section on variants


ChasinHues:
The current documentation for Toasts does not mention anything about its' variants. 

Toasts use the same variants that Alert does. 

Update the documentation to include which variants are available.

PR #3494, opened 2021-03-02
docs(toast): add variants example to toast docs

ChasinHues:
Closes https://github.com/chakra-ui/chakra-ui/issues/3482

## 📝 Description

Current docs for toasts do not mention that variants can be used.

## 💣 Is this a breaking change? 
No

## 📝 Additional Information
- Added an example of using variants with toasts.
- Combined existing examples of statuses into one example.
- Added all positions to the example of positions.
- Use `PropsTable` component instead of manual props table.

### DIFF
--- a/website/pages/docs/feedback/toast.mdx
+++ b/website/pages/docs/feedback/toast.mdx
@@ -55,7 +55,7 @@ function ToastExample() {
 Display a custom component instead of the default Toast UI.
 
 ```jsx
-function Example() {
+function CustomToastExample() {
   const toast = useToast()
   return (
     <Button
@@ -82,7 +82,7 @@ Toasts can be closed imperatively, individually (via the `close` instance
 method) or all together (via the `closeAll` instance method).
 
 ```jsx
-function Example() {
+function ClosingToastExample() {
   const toast = useToast()
   const toastIdRef = React.useRef()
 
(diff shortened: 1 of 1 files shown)


website/pages/docs/feedback/toast.mdx
```diff
@@ -238,13 +244,4 @@ toast({
...
-| `onClose`     | `function`                                                              |          | Callback function to close the toast.                                |
-| `description` | `React.ReactNode`                                                       |          | The description of the toast.                                        |
-| `status`      | `default`, `success`, `error`, `warning`, `info`                        |          | The status of the toast.                                             |
-| `duration`    | `number`                                                                | `5000ms` | Duration before dismiss in milliseconds, or `null` to never dismiss. |
-| `position`    | `top`, `top-left`, `top-right`, `bottom`, `bottom-left`, `bottom-right` | `bottom` | Position the toast will popup out from.                              |
-| `render`      | `(props: {onClose:() => void, id: string}) => React.ReactNode`          |          | Callback function to display a custom toast.                         |
+<PropsTable of="useToast" />
```
segunadebayo, line 247: The prop table for `position` and `variant` is not visible in the docs. Not sure what could be wrong.

ChasinHues: I can see `variant` in there but `position` is definitely missing... Not sure why either 🤔

PR #3511, opened 2021-03-06
feat: add theme extension withDefaultColorScheme

### LINKED ISSUE #2828: Ability to set default colorscheme for all components
# 🚀 Feature request

> Please describe your request in one or two sentences.

I would find it very convenient if it would be possible to set the default colorscheme for all components at once. 

> Please provide valid reason(s) why this should added to Chakra UI.
>
> If this feaure is related to a problem you've noticed. Mention it as well

Currently, if you want to change the default colorscheme from blue to red, for example, you have to do it for all of the existing components individually:

```javascript
const theme = extendTheme({
  components: {
    Button: {
      defaultProps: {
        colorScheme: "red",
      },
    },
    Tabs: {
      defaultProps: {
        colorScheme: "red",
      },
    },
    // ... etc
  },
});
```

## ✅ Proposed solution or API

> Please provide code snippets, gists, or links to the ideal design or API

It would be more convenient if we could do something like this:

```javascript
const theme = extendTheme({
  config: {
    initialColorMode: "dark",
    initialColorScheme: "red",
  },
});
```

## ↩️ Alternatives

As already mentioned, alternative is to set the default colorscheme for every existing component separately. It is not the end of the world to do it individually for all components, but implementing this feature request would be a great quality of life improvement.

## 📝 Additional Information

N/A

Thank you for this great library and all the hard work put into it! Cheers!

TimKolberger:
Closes #2828

## 📝 Description

This PR introduces theme extentions. `extendTheme` now composes all overrides and extensions onto the baseTheme.

`extendTheme` accepts one override object.

## 🚀 New behavior

You can apply a default color scheme to all components in your baseTheme.

```js
import {
  extendTheme,
  theme as baseTheme,
  withDefaultColorScheme,
} from "@chakra-ui/react"

const customTheme = extendTheme(
  withDefaultColorScheme({ colorScheme: "red" }),
  baseTheme, // optional
)
```

Or pass the component names you want to apply a default `colorScheme` to. This
lets you apply different color schemes to a group of components.

```js
import {
  extendTheme,
  withDefaultColorScheme,
} from "@chakra-ui/react"

const customTheme = extendTheme(
  withDefaultColorScheme({
    colorScheme: "red",
    components: ["Button", "Badge"],
  }),
  withDefaultColorScheme({
    colorScheme: "blue",
    components: ["Alert", "Table"],
  }),
)
```

You can even write you own theme extensions:

```ts
import { 
  extendTheme,
  withDefaultColorScheme,
  mergeThemeOverride, 
  ThemeExtension 
} from '@chakra-ui/react'

function withBrandColorExtension(colorName: string): ThemeExtension {
  return (theme) =>
    mergeThemeOverride(theme, {
      colors: {
        brand: theme.colors![colorName],
      },
    })
}

const customTheme = extendTheme(
  withBrandColorExtension("red"),
  withDefaultColorScheme({ colorScheme: "brand" }),
)
```

No, the current API for a single override and an optional third-party base theme are still supported.

### DIFF
--- a/packages/react/src/extend-theme.ts
+++ b/packages/react/src/extend-theme.ts
@@ -3,7 +3,7 @@ import defaultTheme, {
   isChakraTheme,
   Theme,
 } from "@chakra-ui/theme"
-import { Dict, isFunction, mergeWith, warn } from "@chakra-ui/utils"
+import { Dict, isFunction, mergeWith, pipe } from "@chakra-ui/utils"
 
 type CloneKey<Target, Key> = Key extends keyof Target ? Target[Key] : unknown
 
@@ -40,29 +40,18 @@ export type ThemeExtension<Override extends ThemeOverride = ThemeOverride> = (
 export type BaseThemeWithExtensions<
   BaseTheme extends ChakraTheme,
   Extensions extends readonly [...any]
-> = Extensions extends [infer L, ...infer R]
-  ? L extends (...args: any[]) => any
-    ? ReturnType<L> & BaseThemeWithExtensions<BaseTheme, R>
-    : L & BaseThemeWithExtensions<BaseTheme, R>
-  : BaseTheme & Extensions
+> = BaseTheme &
+  (Extensions extends [infer L, ...infer R]
+    ? L extends (...args: any[]) => any
+      ? ReturnType<L> & BaseThemeWithExtensions<BaseTheme, R>
+      : L & BaseThemeWithExtensions<BaseTheme, R>
+    : Extensions)
 /**
  * Function to override or customize the Chakra UI theme conveniently.
  * First extension overrides the baseTheme and following extensions override the preceding extensions.
  *
- * @example with only overrides:
- * import { extendTheme } from '@chakra-ui/react'
- *
- * const customTheme = extendTheme({
- *   colors: {
- *     brand: {
- *       500: "#b4d455",
- *     },
- *   },
- * })
- *
- *
- * @example with extension:
- * import { theme, extendTheme, withDefaultColorScheme } from '@chakra-ui/react'
+ * @example:
+ * import { theme as baseTheme, extendTheme, withDefaultColorScheme } from '@chakra-ui/react'
  *
  * const customTheme = extendTheme(
  *   {
@@ -73,7 +62,7 @@ export type BaseThemeWithExtensions<
  *     },
  *   },
  *   withDefaultColorScheme({ colorScheme: "red" }),
- *   theme,
+ *   baseTheme // optional
  * )
  */
 export function extendTheme<
@@ -83,23 +72,22 @@ export function extendTheme<
     | ThemeOverride<BaseTheme>
     | ThemeExtension<ThemeOverride<BaseTheme>>
   )[] = (ThemeOverride<BaseTheme> | ThemeExtension<ThemeOverride<BaseTheme>>)[]
->(
-  ...extensions: [...Extensions]
-): BaseThemeWithExtensions<BaseTheme, Extensions> {
-  const overrides = [...extensions]
-    .slice(0, extensions.length - (extensions.length > 1 ? 1 : 0))
-    .reverse()
+>(...extensions: [...Extensions]) {
+  let overrides = [...extensions]
+  let baseTheme = extensions[extensions.length - 1]
 
-  const baseTheme =
-    extensions.length > 1 ? extensions[extensions.length - 1] : defaultTheme
-
-  if (!isChakraTheme(baseTheme) || !overrides.length) {
-    warn(
-      `Last parameter needs to be a valid \`ChakraTheme\`. Please check your extendTheme function call from @chakra-ui/react.`,
-    )
+  if (
+    isChakraTheme(baseTheme) &&
+    // this ensures backward compatibility
+    // previously only `extendTheme(override, baseTheme?)` was allowed
+    overrides.length > 1
+  ) {
+    overrides = overrides.slice(0, overrides.length - 1)
+  } else {
+    baseTheme = (defaultTheme as unknown) as BaseTheme
   }
 
-  return compose(
+  return pipe(
     ...overrides.map(
       (extension) => (
         prevTheme: BaseThemeWithExtensions<BaseTheme, Extensions>,
@@ -115,10 +103,6 @@ export function mergeThemeOverride<BaseTheme extends ChakraTheme = ChakraTheme>(
   ...overrides: ThemeOverride<BaseTheme>[]
 ): ThemeOverride<BaseTheme> {
   return mergeWith({}, ...overrides, mergeThemeCustomizer)
-}
-
-export function compose<R>(...fns: Array<(a: R) => R>) {
-  return fns.reduce((prevFn, nextFn) => (value) => prevFn(nextFn(value)))
 }
 
 function mergeThemeCustomizer(
--- a/packages/react/src/extend-theme.ts
+++ b/packages/react/src/extend-theme.ts
@@ -3,7 +3,13 @@ import defaultTheme, {
   isChakraTheme,
   Theme,
 } from "@chakra-ui/theme"
-import { Dict, isFunction, mergeWith, pipe } from "@chakra-ui/utils"
+import {
+  AnyFunction,
+  Dict,
+  isFunction,
+  mergeWith,
+  pipe,
+} from "@chakra-ui/utils"
 
 type CloneKey<Target, Key> = Key extends keyof Target ? Target[Key] : unknown
 
@@ -29,9 +35,9 @@ type DeepThemeExtension<BaseTheme, ThemeType> = {
     : CloneKey<ThemeType, Key>
 }
 
-export declare type ThemeOverride<BaseTheme = Theme> = DeepPartial<ChakraTheme> &
-  DeepThemeExtension<BaseTheme, ChakraTheme> &
-  Dict
+export declare type ThemeOverride<
+  BaseTheme = Theme
+> = DeepPartial<ChakraTheme> & DeepThemeExtension<BaseTheme, ChakraTheme> & Dict
 
 export type ThemeExtension<Override extends ThemeOverride = ThemeOverride> = (
   themeOverride: Override,
@@ -42,7 +48,7 @@ export type BaseThemeWithExtensions<
   Extensions extends readonly [...any]
 > = BaseTheme &
   (Extensions extends [infer L, ...infer R]
-    ? L extends (...args: any[]) => any
+    ? L extends AnyFunction
       ? ReturnType<L> & BaseThemeWithExtensions<BaseTheme, R>
       : L & BaseThemeWithExtensions<BaseTheme, R>
     : Extensions)
--- a/packages/react/src/extend-theme.ts
+++ b/packages/react/src/extend-theme.ts
@@ -1,5 +1,9 @@
-import defaultTheme, { ChakraTheme, Theme } from "@chakra-ui/theme"
-import { isFunction, mergeWith } from "@chakra-ui/utils"
+import defaultTheme, {
+  ChakraTheme,
+  isChakraTheme,
+  Theme,
+} from "@chakra-ui/theme"
+import { Dict, isFunction, mergeWith, warn } from "@chakra-ui/utils"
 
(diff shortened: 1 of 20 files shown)


packages/react/src/extend-theme.ts
```diff
@@ -21,45 +25,143 @@ type DeepThemeExtension<BaseTheme, ThemeType> = {
...
+) => Override
+
+export type BaseThemeWithExtensions<
+  BaseTheme extends ChakraTheme,
+  Extensions extends readonly [...any]
+> = Extensions extends [infer L, ...infer R]
+  ? L extends (...args: any[]) => any
+    ? ReturnType<L> & BaseThemeWithExtensions<BaseTheme, R>
+    : L & BaseThemeWithExtensions<BaseTheme, R>
+  : ThemeOverride<BaseTheme>
```
TimKolberger, line 43: This type is TS 4.1 - we need to find a way to express this with TS3.8 or consider bumping the min. TS version.

TimKolberger: Nevermind - just saw that we support TS ^4.1.3

segunadebayo:
This looks very interesting, @TimKolberger. Great work buddy!

I think we can extend this idea to be more like `withDefaultProps` so it can be used to configure the default size, variant, and colorScheme for components.

The signature of the function allows for more flexibility `withDefaultProps({  colorScheme: "red", size: "sm" })`.

Just an idea I had while reviewing. What are your thoughts?

with-heart:
The interesting thing with function composition is that if the types of the functions line up correctly, we could easily create `withDefaultProps` by passing `withDefaultColorScheme`, `withDefaultSize`, and `withDefaultVariant` functions to some other function that merges them together into a single function. So a user could use the individual functions if they only care about one of the three, or use `withDefaultProps` which is just the combination of the other three.

TimKolberger:
Thank you for your reviews 💖 

I think function composition is a powerful tool and composing a theme is a good next step for the chakra theme.

As @with-heart mentioned we easily could compose the function `withDefaultProps` from more specific ones. I will add those to this PR 👍 

Currently the types are not _that_ smart to extract the exact type of the composed theme if some properties are removed while composing. It just adds properties on top and cannot remove them. 
Those typings are a real hassle 😅 , but I will try to improve them in upcoming PRs.

I think this approach could help us in the long run as well, by instead of just overriding a theme - composing a theme from _scratch_ with only the necessary components for each application. This would help reduce the application bundle size.

with-heart:
@TimKolberger Mind if I add a few changeset files to cover `withDefaultSize`, `withDefaultVariant`, and `withDefaultProps`?

segunadebayo:
@with-heart please go ahead. If you can include a changeset for the update of `isFunction` in the `utils` package that would be great!

segunadebayo:
Any updates on this @with-heart? I'd love to get this merged

with-heart:
So sorry this has taken me so long. I've been quite busy this week, but I'm working on this now!

with-heart:
And now that I'm looking into the changesets, I'm also noticing some things that I didn't notice before when I wasn't looking quite as closely. So going to leave a review for more discussion once I figure out what it is I'm trying to say 😬 

So sorry!

packages/react/src/extend-theme.ts
```diff
@@ -21,45 +25,118 @@ type DeepThemeExtension<BaseTheme, ThemeType> = {
...
+    .reverse()
+
+  const baseTheme =
+    extensions.length > 1 ? extensions[extensions.length - 1] : defaultTheme
+
+  if (!isChakraTheme(baseTheme) || !overrides.length) {
+    warn(
+      `Last parameter needs to be a valid \`ChakraTheme\`. Please check your extendTheme function call from @chakra-ui/react.`,
+    )
+  }
```
with-heart, line 97: It seems like we could improve the DX and reduce the complexity of implementation if we made the same assumption about `baseTheme` as we do with `extendTheme` currently: the last argument can be a `theme` to modify, but if it's not, we use the default Chakra theme.

```jsx
const theme = extendTheme({
  // theme stuff
})

const theme = extendTheme(
  compose(),
  some(),
  things(),
  withNo(),
  theme()
)
```

This is similar to how [`customize-cra`](https://github.com/arackaf/customize-cra) works, where you're just defining the pipeline of functions that will be executed on the base theme, at least by default.

packages/react/src/extend-theme.ts
```diff
@@ -21,45 +25,118 @@ type DeepThemeExtension<BaseTheme, ThemeType> = {
...
-  function customizer(
-    source: unknown,
-    override: unknown,
-    key: string,
-    object: any,
+  ...extensions: [...Extensions]
+): BaseThemeWithExtensions<BaseTheme, Extensions> {
+  const overrides = [...extensions]
+    .slice(0, extensions.length - (extensions.length > 1 ? 1 : 0))
+    .reverse()
```
with-heart, line 88: `reverse()` + `compose()` is hurting my brain trying to figure out what order the functions are called in. `extendTheme(a(), b(), c())`, what order are they called in?

packages/react/src/extend-theme.ts
with-heart: Seems like from looking at the related docs, we have to make the last object a theme because the last item in the list could theoretically be a theme customization object that passes as a theme, which would obviously break things. 

That feels awkward to me, and I'm wondering if we could make this significantly better by making it a totally new API rather than modifying the existing one, so we don't have to maintain any backwards compatibility.

I'm have lots of thoughts here but I'm not sure this is the right place to share them. Maybe we could hang out and chat about it @TimKolberger?

packages/react/src/theme-extensions/with-default-color-scheme.ts
```diff
@@ -0,0 +1,35 @@
...
+  components?: string[] | Dict
+}): ThemeExtension {
+  return (theme) => {
+    let names = Object.keys(theme.components || {})
+
+    if (Array.isArray(components)) {
+      names = components
+    } else if (isObject(components)) {
+      names = Object.keys(components)
+    }
```
with-heart, line 20: It seems like this logic isn't quite what we'd want.

We set `names` to `theme.components`, and then we immediately replace that with `components` or `Object.keys(components`). That would be fine except `components` defaults to `defaultTheme.components`, so there's no way to have it use `theme.components` for the `names`. Which imo would be a major use case, so if a theme with custom component themes is passed, they're automatically included.

with-heart: Same for the other functions with the same setup.

with-heart (changes requested):
I hate to hold this up but I think we might want to think some things through a bit more before we lock ourselves into an API that will be difficult to change later. 

I have some really cool ideas in mind that build on top of this concept that I think would give our users a lot more flexibility and greatly expand the usefulness of tooling around constructing/modifying a customized theme.

packages/react/src/theme-extensions/with-default-color-scheme.ts
TimKolberger: This code defaults to all components in the theme. If the argument `components` is defined and is an array or object, it uses those components. 

In this test case the Alert component is not explicitly passed, but receives the colorScheme:

https://github.com/chakra-ui/chakra-ui/blob/1a1e980ba9b84bd50e56d167c71312f7804e7562/packages/react/tests/with-default-color-scheme.test.tsx#L5-L21

with-heart: Right! But what I'm saying is that it defaults to the wrong `theme`'s components. Like the last `expect` here fails:

```ts
const themeWithCustomComponent = extendTheme({
  components: {
    MyCustomComponent: {
      defaultProps: {
        colorScheme: 'purple',
      },
    },
  },
})
const theme = extendTheme(
  {
    colors: {
      brand: {
        500: '#b4d455',
      },
    },
  },
  withDefaultColorScheme({colorScheme: 'brand'}),
  themeWithCustomComponent,
)

expect(theme.components.Alert.defaultProps.colorScheme).toBe('brand')
expect(theme.components.MyCustomComponent.defaultProps.colorScheme).toBe(
  'brand',
)
```

packages/react/src/extend-theme.ts
TimKolberger: The order of overrides is from right to left, except the base:

```js
extendTheme(
  extionsionONEoverridesBase,      // <- called first
  extensionTWOoverridesONE,        // <- called second
  extensionNEXToverridesPRECEDING, // <- called third
  base,
)
```

We need to keep this function backwards compatible, so the last param needs to stay the baseTheme. 

It thought with `reverse()`  the using side is more intuitive, but I do not have a strong opinion on this.
Without reverse  the order would be:

```js
extendTheme(
  extensionNEXToverridesPRECEDING,  // <- called third
  extensionTWOoverridesONE,         // <- called second
  extionsionONEoverridesBase,       // <- called first
  base,
)
```

packages/react/src/theme-extensions/with-default-color-scheme.ts
with-heart: `names` is initially correct in getting the names from `theme.components` rather than `defaultTheme.components`, but the `if`/`else if` then immediately overwrite that with `defaultTheme.components` since that's the default value.

packages/react/src/extend-theme.ts
TimKolberger: Moving this logic to a new function would ease the pain of backwards compat! There are pros and cons about this approach - let's chat about this :)

packages/react/src/extend-theme.ts
with-heart: I think I'm even more confused now lol, I think because I'm confused about what you mean by "right to left" and maybe FP terminology around `compose`. In FP, `compose` is RTL, meaning that it starts at the last function, executes it with the argument, and then takes that result and passes it to the function before it, and so on.

`compose(a, b, c)('hi') === a(b(c('hi')))`

The reverse (😉) of that is `pipe` aka `compose` + `reverse` which is LTR:

`pipe(a, b, c)('hi') === c(b(a('hi')))`

I don't know that I really have a preference one way or the other! It'll just be confusing to other FP-minded people if we're saying compose/RTL when we actually mean pipe/LTR 🙃

packages/react/src/theme-extensions/with-default-color-scheme.ts
TimKolberger: Oh, good catch 👀 💪 I did not see the default arg there. Thanks for pointing that out!

Looking forward to improve the API and code with you 😊 🎉

packages/react/src/extend-theme.ts
TimKolberger: 1. I removed the `reverse` call and replaced to compose function with `pipe` and moved this function to the `utils` package.
2. The override order is "Left To Right". I think I got confused there myself when I wrote the docs.

packages/react/src/theme-extensions/with-default-color-scheme.ts
TimKolberger: I removed the default value for the parameter `components` 👍

packages/react/src/extend-theme.ts
TimKolberger: The `baseTheme` is now optional for all use cases (with 1 or more overrides/extensions).
I updated the implementation of `extendTheme` to be more readable. With this in place I am not quite sure if another function would increase the maintainability. Let me know what you think :)

packages/react/src/theme-extensions/with-default-color-scheme.ts
TimKolberger: And I added your mentioned test case (I had to reverse the order of overrides due `pipe`)

with-heart (approved):

website/pages/docs/theming/customize-theme.mdx
```diff
@@ -244,6 +244,125 @@ None of these is strictly required to use Chakra - but we've learned some hard
 lessons on the "right" way and the "wrong" way to write styles. The above is our
 best suggestion on how to write style overrides and organize your custom theme.
 
+## Using Theme extensions
+
+<Badge fontSize="sm" colorScheme="teal" letterSpacing="wider">
+  v1.4.0
```
segunadebayo, line 250: Could we update this?

segunadebayo (approved):

PR #3519, opened 2021-03-08
Chakra Playground

anubra266:
## 📝 Description

>This PR adds a Chakra Playground to the docs

Follow it here:
https://chakra-play.vercel.app/

### DIFF
--- a/website/pages/playground/index.tsx
+++ b/website/pages/playground/index.tsx
@@ -1,0 +1,7 @@
+import React from "react"
+
+const Playground = () => {
+  return <div></div>
+}
+
+export default Playground


segunadebayo:
This sounds interesting @anubra266. If you can describe what you have in mind about the playground, that'll help a lot.

It'll also avoid the scenario where you've put in a lot of work and the PR seems to be up without getting merged. Early feedback/thoughts help a lot.

Ideally, I'd love to host the chakra playground in a different domain separate from the website but I'm happy to watch the progress.

Thanks man!

anubra266:
Okay, I'll move it to a different repo, but the idea is something like https://play.tailwindcss.com/ for Chakra UI.

segunadebayo:
That's interesting! Feel free to use the docs as the "testing" ground for this.

When it's fully formed, we'll look into how to make this happen!

PR #3529, opened 2021-03-09
feat: include text/layer styles in theme types

### LINKED ISSUE #3524: `chakra-cli tokens` doesn't generate textStyles type
# 🐛 Bug report

Types are not correctly generated by `chakra-cli tokens` for all theme keys. I've noticed this for `textStyles`, but believe it may be a generic issue for any theme property whose values are objects, if the [config file](https://github.com/chakra-ui/chakra-ui/blob/82426a03969c4c9cc61343636776e7202a6e9806/tooling/cli/src/command/tokens/config.ts) has not accounted for this. 

My high level understanding is that the [first inner loop of extractPropertyPaths](https://github.com/chakra-ui/chakra-ui/blob/82426a03969c4c9cc61343636776e7202a6e9806/tooling/cli/src/command/tokens/extract-property-paths.ts#L36) returns an empty array due to depth being 0, which means the outer loop has no `childKey`s to append, which means it exits without doing anything.

### From scratch
1. Define a theme for Chakra containing `textStyles: { 'large': { fontSize: 64  } }`. This can be applied to components as `<Box textStyle="large" />`
2. Use `chakra-cli tokens` command to generate types for the theme
3. Observe generated types (`node_modules/@chakra-ui/styled-system/dist/types/theming.types.d.ts`) contain `textStyles?: never` instead of any reference to textStyles in your theme. Observe Intellisense for `textStyle` prop is still just `string | undefined`.

### From repro sandbox (see below)
1. Download sandbox content
2. `yarn`
3. `yarn generate-theme` (or manually call `chakra-cli tokens src/theme.ts`)
4. Observe generated types (`node_modules/@chakra-ui/styled-system/dist/types/theming.types.d.ts`) contain `textStyles?: never` instead of any reference to textStyles in your theme. Observe Intellisense for `textStyle` prop is still just `string | undefined`.

## 💻 Link to reproduction

Note: You must download this sandbox and follow the repro steps above locally, since the CLI modifies node_modules directly.

https://codesandbox.io/s/young-rgb-hof4e?file=/package.json

## 🧐 Expected behavior

Generated types for `textStyles` should include the top-level keys from `theme.textStyles`, ie. `large`.

## 🧭 Possible Solution

Updated config for textStyles type generation in [config.ts](https://github.com/chakra-ui/chakra-ui/blob/82426a03969c4c9cc61343636776e7202a6e9806/tooling/cli/src/command/tokens/config.ts)? Might require a new config key (propertyPathExtractor or something) since none of the current config pieces can affect the failing stage. Or its own entire handler like colors have.

## 🌍 System information

|@chakra-ui/cli|1.2.1|
|@chakra-ui/react|1.0.0|
| Operating System | MacOS 11 |

callum-mellorreed-privitar:
Closes #3524

## 📝 Description

`textStyles` and `layerStyles` values were not provided by generated theme types. Now they are.

`textStyle` and `layerStyle` props are always typed as `string | undefined`, regardless of values defined in theme.

## 🚀 New behavior

`textStyle` and `layerStyle` props are typed from the default theme types or a union of possible string values based on the keys present within `theme.textStyles`. The default type changes from `string` to `string & {}` because TypeScript [cannot combine string literal and string](https://stackoverflow.com/a/61048124) as a union (this is already seen in `ThemingProps` for `variant`, `size`, `colorScheme`).

No, this will not prevent any existing supported values from being used.

However, because this requires simultaneous changes across 3 packages I think they would be internally breaking? Not clear on version strategy here, changeset applies minor to all packages involved for now.


tooling/cli/src/command/tokens/config.ts
```diff
@@ -7,6 +7,7 @@ export const themeKeyConfiguration: ThemeKeyOptions[] = [
   { key: "fonts" },
   { key: "fontSizes" },
   { key: "fontWeights" },
+  { key: "layerStyles" },
```
TimKolberger, line 10: This config file is only used by `extractPropertyPath`. Leaving `layerStyles` and `textStyles` in there could confuse other contributors. (Seems like I left `textStyles` in there, but it does nothing)

Would you kindly remove both entries?

TimKolberger (commented):
Great work @callum-mellorreed-privitar ✨ 

I added one comment. Then it looks good to go!

tooling/cli/src/command/tokens/config.ts
callum-mellorreed-privitar: Ah yeah of course, done!

with-heart (approved):

with-heart:
I think this is good to go but will wait for @TimKolberger approval :)

TimKolberger (approved):
Looks good to me! 👍

PR #3584, opened 2021-03-15
fix: aspect ratio SSR style tag display

isBatak:
## 📝 Description
On the SSR, `<style>` tag was displayed for a split second because of `display: "flex"` applied to all children.
A fix is to override the `display` property for style tags.

## 📝 Additional Information
You can reproduce this here
https://codesandbox.io/s/chakra-ssr-aspect-ratio-issue-qghp1

Preview (JS is disabled so only the HTML form server side is rendered)
diable-js

### DIFF
--- a/.changeset/few-wasps-judge.md
+++ b/.changeset/few-wasps-judge.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/layout": patch
+---
+
+fix: aspect ratio SSR style tag display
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks, @isBatak

I think a better solution would be to change `& > *` to `& > *:not(style)`. What do you think? Perhaps we could also include display none as well if it doesn't work.

isBatak:
> Thanks, @isBatak
> 
> I think a better solution would be to change `& > *` to `& > *:not(style)`. What do you think? Perhaps we could also include display none as well if it doesn't work.

I was thinking about `*:not` solution also but as there is already `"& > img, & > video":` thingy I did the simplest one.
I can try `:not` selector

PR #3615, opened 2021-03-18
fix: styled system css and css vars

### LINKED ISSUE #3501: Cannot override Heading fontSize if there is a textStyles value
# 🐛 Bug report

It a fontSize value is defined in the `textStyles` object of the Theme for a Heading component, it seems like we cannot override it by passing a fontSize prop to the Heading.

## 💥 Steps to reproduce

- add a textStyles object to the Theme, with at least a fontSize property. It can be a string or a responsive value.
- add a Heading component and pass it the corresponding textStyle key
- try to override this font size by adding a fontSize props (or using sx props)

## 💻 Link to reproduction

https://codesandbox.io/s/chakra-heading-font-size-pr14s

## 🧐 Expected behavior

- it should be possible to override the textStyles font size value

## 🧭 Possible Solution

- it seems like the breaking version is 1.3.1, according to the codesandbox

## 🌍 System information

| Chakra UI        |  1.3.3          |
| Browser          |    Chrome 88        |
| Operating System |      MacOS      |

## 📝 Additional information

- on the reproduction repo, overriding with `size` works, but not on my actual app (excepted when I pass both size and fontSize props); I don't understand why yet.

### LINKED ISSUE #3514: toCSSVar fails to create css vars for tokens that use nested paths
# 🐛 Bug report

`transition` tokens are missing from the `extractTokens` result. This happens because `pick` uses a `if ( key in obj )` check that does not work when the key is a nested property.

## 💥 Steps to reproduce

1. With the default theme, inspect `theme` and `tokens` variables at https://github.com/chakra-ui/chakra-ui/blob/82426a03969c4c9cc61343636776e7202a6e9806/packages/styled-system/src/css-var.ts#L96
2. The `theme` includes the `transition` property, but `"transition.duration", "transition.property", "transition.easing"` are missing from the `extractTokens` result, even though they are defined in the exported `tokens` array
3. This breaks CSS transitions for all components that use `transitionProperty`, `transitionTimingFunction` and `transitionDuration` values defined in the theme

## 🧐 Expected behavior

Components using `transitionProperty`, `transitionTimingFunction` and `transitionDuration` should be able to use the values that were defined in the theme.

## 🧭 Possible Solution

Replace the `key in object` check with `get(object, key) !== undefined`

## 🌍 System information

| Chakra UI        |   @chakra-ui/react 1.3.4 |
| Browser          |  Chrome 89.0.4389.72 |
| Operating System | macOS 11.1 |

## 📝 Additional information

My `transition.ts` foundation
```js
const transitionProperty = {
  common: "background-color, border-color, color, opacity, visibility",
};

const transitionTimingFunction = {
  "ease-out-qubic": "cubic-bezier(0.33, 1, 0.68, 1)",
  "ease-out-quart": "cubic-bezier(0.165, 0.84, 0.44, 1)",
};

const transitionDuration = {
  normal: "300ms",
};

const transition = {
  property: transitionProperty,
  easing: transitionTimingFunction,
  duration: transitionDuration,
};

export default transition;
```

Currently broken component with irrelevant props removed
```jsx
<Flex
  bgColor="primary.50"
  _hover={{ bgColor: "primary.100" }}
  transitionProperty="background-color"
  transitionTimingFunction="ease-out-quart"
  transitionDuration="normal"
 >
...
</Flex>
```

The styles that this component gets
Screen Shot 2021-03-07 at 5 05 02 PM

segunadebayo:
Closes #3501
Closes #3514

## 📝 Description

- Fix issue where responsive styles defined in `textStyles` not overridden when passing props.
- Fix issue where `toCSSVars` omitted the transition tokens

## 🚀 New behavior

- Styles defined in `layerStyles` and `textStyles` and apply can now be overridden with props.
- Tokens now include transition values in css custom properties.

No

### DIFF
--- a/.changeset/olive-walls-promise.md
+++ b/.changeset/olive-walls-promise.md
@@ -1,0 +1,8 @@
+---
+"@chakra-ui/styled-system": patch
+---
+
+Fix: issue where responsive styles defined in textstyles not overriden wiht
+props.
+
+Fix: issue where `toCSSVars` omitted the transition tokens
--- a/packages/styled-system/src/config/others.ts
+++ b/packages/styled-system/src/config/others.ts
@@ -32,6 +32,14 @@ const srFocusable = {
   whiteSpace: "normal",
 }
 
+const getWithPriority = (theme: any, key: any, styles: any) => {
+  const expanded = get(theme, key, {})
+  for (const prop in expanded) {
+    if (prop in styles) delete expanded[prop]
+  }
+  return expanded
+}
+
 export const others: Config = {
   animation: true,
   appearance: true,
@@ -58,15 +66,17 @@ export const others: Config = {
   },
   layerStyle: {
     processResult: true,
-    transform: (value, theme) => get(theme, `layerStyles.${value}`, {}),
+    transform: (value, theme, styles) =>
+      getWithPriority(theme, `layerStyles.${value}`, styles),
   },
   textStyle: {
     processResult: true,
-    transform: (value, theme) => get(theme, `textStyles.${value}`, {}),
+    transform: (value, theme, styles = {}) =>
+      getWithPriority(theme, `textStyles.${value}`, styles),
   },
   apply: {
     processResult: true,
-    transform: (value, theme) => get(theme, value, {}),
+    transform: (value, theme, styles) => getWithPriority(theme, value, styles),
   },
 }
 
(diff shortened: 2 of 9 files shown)


violabg:
I think this broken textStyles.
I just updated to version 1.4 and all my components that have textStyle, no longer receive styles from it

segunadebayo:
@violabg, I tested it now and it seems to work correctly.

- Can you try to clean your node_modules and try again?
- If it persists, send me a reproduction and I'm happy to dig deeper.

lazarnikolov94:
@violabg are you experiencing this problem locally in your project or in a CodeSandbox?

violabg:
I have deleted my node_modules and having problems with fontSize and fontWeight coming from textStyles.
I create a codeSandbox, ma locally is worst since only the color prop get passed to the component:

https://codesandbox.io/s/textstyle-demo-vzxyr?file=/src/App.tsx:240-245

violabg:
> https://codesandbox.io/s/textstyle-demo-vzxyr?file=/src/App.tsx:240-245

when it load the first time, you can see that the first two Boxes are styled correctly, while the Heading and Text are not.
If you change the text of one Box it breaks also for the Boxes, this is what I'm experiencing on my project, which is a localised Next.js project

violabg:
I was able to reproduce my problem by updating the codesanbox, with 

```js
const [title, setTitle] = useState("Box");

  useEffect(() => {
    setTitle("Box2");
  }, []);
```

PR #3619, opened 2021-03-18
docs: specify version range for peer deps

with-heart:
## 📝 Description

As of this morning, `framer-motion` is now at version 4. This is not a version that Chakra currently supports. Unfortunately, due to a combination of `peerDependencies` ranges being too wide in our packages and the installation instructions not specifying a version range, we're instructing people to always install the latest which is breaking fresh projects.

Users are instructed to install the latest version of our peer dependencies.

## 🚀 New behavior

Users are instructed to install specific major version ranges of our peer dependencies. This future-proofs projects, so that when one of our peer dependencies has a `major` version bump that we don't support, it won't break anyone's project.

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,12 +92,12 @@ link to your website.
 To use Chakra UI components, all you need to do is install the
 `@chakra-ui/react` package and its peer dependencies:
 
-```sh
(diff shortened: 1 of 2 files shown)


TimKolberger (approved):

with-heart:
Superseded by #3623

PR #3622, opened 2021-03-19
feat: export type ThemeComponentProps with theme property

TimKolberger:
## 📝 Description

Context: https://github.com/chakra-ui/chakra-ui/discussions/3568

Added typings for the `theme` prop in `ThemingPropsThunk` and export a standalone type `ThemeComponentProps`

```ts
import { ThemeComponentProps } from "@chakra-ui/react"
function baseStyle(props: ThemeComponentProps) {
  return {
    boxShadow: `0 1px 2px 0 rgba(0, 0, 0, 0.05) ${props.theme.colors.whiteAlpha[500]}`,
  }
}
```

Type is hidden in `ThemingPropsThunk`

## 🚀 New behavior

It is exported as standalone type.

No

### DIFF
--- a/packages/theme/src/theme.types.ts
+++ b/packages/theme/src/theme.types.ts
@@ -42,14 +42,14 @@ export interface ComponentDefaultProps
   extends Omit<ThemingProps, "styleConfig">,
     Dict {}
 
-export type ThemingPropsThunk<T> =
-  | T
-  | ((
-      props: Dict &
-        Omit<ThemingProps, "styleConfig"> & {
-          colorMode: ColorMode
-        },
-    ) => T)
+export interface ThemeComponentProps<Theme extends ChakraTheme = ChakraTheme>
+  extends Dict,
+    Omit<ThemingProps, "styleConfig"> {
+  colorMode: ColorMode
+  theme: Theme
+}
+
+export type ThemingPropsThunk<T> = T | ((props: ThemeComponentProps) => T)
 
 export interface SystemStyleObjectRecord {
   [key: string]: StyleObjectOrFn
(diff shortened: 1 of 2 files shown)


TheThirdRace:
A slight modification would cover most of my usecases:

```typescript
export interface ThemeComponentProps<Theme extends ChakraTheme = ChakraTheme>
  extends Dict,
    Omit<ThemingProps, "styleConfig"> {
  colorMode: ColorMode
  theme: Theme
}

export type ThemingPropsThunk<T> = T | ThemeComponentFunction<T>

export type ThemeComponentFunction<T> = (props: ThemeComponentProps) => T
```

It doesn't really cost anything and now we get flexibility.

I could now call a variant within another variant, like `outline` being called within `outlineWithBrand` for example. It's really  just because I can't cast the variant as a function if there's no type associated.

TimKolberger:
Thank you for you valuable input @TheThirdRace! ✨

packages/theme/src/theme.types.ts
```diff
@@ -42,14 +42,21 @@ export interface ComponentDefaultProps
...
-  | T
-  | ((
-      props: Dict &
-        Omit<ThemingProps, "styleConfig"> & {
-          colorMode: ColorMode
-        },
-    ) => T)
+export interface ThemeComponentProps<Theme extends ChakraTheme = ChakraTheme>
+  extends Dict,
+    Omit<ThemingProps, "styleConfig"> {
```
with-heart, line 47: Could this extend `ComponentDefaultProps` instead to simplify?

with-heart (approved):

packages/theme/src/theme.types.ts
TheThirdRace: If we take for granted that every default props are passed to the function, then yes it could.

Good catch @with-heart

PR #3623, opened 2021-03-19
feat: add support for framer-motion v4

### LINKED ISSUE #3618: Framer Motion v4 breaks Chakra
# 🐛 Bug report

## 💥 Steps to reproduce

1. Install a fresh instance of Chakra, including all of its required dependencies
2. Start the app
3. Notice it crash

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/chakra-framer-motion-v4-bug-4i6rq

## 📝 Additional information

Looking at the Framer Motion [changelog](https://github.com/framer/motion/blob/main/CHANGELOG.md), it appears to be related to the deprecation of `motion.custom()` in favor of `motion()`, which I believe is being tracked in #3459 .

with-heart:
Closes #3618

## 📝 Description

This PR makes a number of changes in order to add support for `framer-motion` v4 (released yesterday) while continuing to support the v3 range. This conditional support for both enables to merge this as a `minor` change rather than a `major` (which would be required if we were to _force_ users to upgrade to v4).

See my comment over here for a more detailed description of what's in this PR: https://github.com/chakra-ui/chakra-ui/issues/3618#issuecomment-802817719

`framer-motion` v4 breaks Chakra projects.

## 🚀 New behavior

`framer-motion` v4 no longer breaks Chakra projects. Users can use v3 or v4 without issue.

No

### DIFF
--- a/packages/menu/src/menu.tsx
+++ b/packages/menu/src/menu.tsx
@@ -12,7 +12,7 @@ import {
   useStyles,
 } from "@chakra-ui/system"
 import { cx, MaybeRenderProp, runIfFn, __DEV__ } from "@chakra-ui/utils"
-import { motion, Variants } from "framer-motion"
+import { CustomDomComponent, motion, Variants } from "framer-motion"
 import * as React from "react"
 import {
   MenuProvider,
@@ -141,7 +141,9 @@ const motionVariants: Variants = {
   },
 }
 
-const Motion = chakra(motion.div)
+const MotionDiv: CustomDomComponent<PropsOf<typeof chakra.div>> = motion(
+  chakra.div,
+)
 
 export const MenuList = forwardRef<MenuListProps, "div">((props, ref) => {
   const { rootProps, ...rest } = props
@@ -157,7 +159,7 @@ export const MenuList = forwardRef<MenuListProps, "div">((props, ref) => {
       {...positionerProps}
       __css={{ zIndex: props.zIndex ?? styles.list?.zIndex }}
     >
-      <Motion
+      <MotionDiv
         {...listProps}
         /**
          * We could call this on either `onAnimationComplete` or `onUpdate`.
--- a/packages/menu/src/menu.tsx
+++ b/packages/menu/src/menu.tsx
@@ -141,9 +141,11 @@ const motionVariants: Variants = {
   },
 }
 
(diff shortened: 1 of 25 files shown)


packages/menu/src/menu.tsx
```diff
@@ -141,7 +141,11 @@ const motionVariants: Variants = {
   },
 }
 
-const Motion = chakra(motion.div)
+// @future: only call `motion(chakra.div)` when we drop framer-motion v3 support
+const MotionDiv: CustomDomComponent<PropsOf<typeof chakra.div>> =
+  "custom" in motion
+    ? (motion as any).custom(chakra.div)
+    : (motion as any)(chakra.div)
```
TimKolberger, line 148: I think this change is not necessary, because it used the chakra factory, not `motion.custom`.

`chakra(motion.div)` and `motion(chakra.div)` are working differently, specifically with the `transition` prop.

afaik `chakra(motion.div)` does not pass `transition` on to the motion component, where as `motion(chakra.div)` does not pass transition to the chakra component.

TimKolberger (commented):
🚀  great work Mark!

packages/menu/src/menu.tsx
with-heart: Hmmmm, you might be right. This was the only generated component where I was receiving type errors, I think specifically for the types of `onDrag`. I'll re-investigate and see what needs done. Wanna mark as changes requested so this isn't merged till we're sure?

TimKolberger: I think those type errors are not related to the framer-motion update. The errors occurred with removing the any cast in this commit: https://github.com/chakra-ui/chakra-ui/commit/efd32c982cd58c1e53468743346a3b8f482b1ae3

I think your change is actually a viable solution for that, so I won't mark this as changes requested.
IMO this is good to go!

with-heart: Awesome! Want to give it an approval instead then? 😁 Or should we just merge it? I feel like this is fairly important due to the downstream issues it's causing.

segunadebayo: I think this change is needed.

segunadebayo (approved):
Great work here @with-heart 👌

PR #3636, opened 2021-03-21
fix: export of styles provider

TimKolberger:
## 📝 Description

StylesProvider is not defined in CSB - this is a try to fix this. 

`export const [StylesProvider, useStyles]`

## 🚀 New behavior

`export { StylesProvider, useStyles }`

No

## 📝 Additional Information

CSB deployment does not work for me, but I think this approach is worth a try.

### DIFF
--- a/.changeset/giant-seahorses-laugh.md
+++ b/.changeset/giant-seahorses-laugh.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/system": patch
+---
+
+- Fixed an issue where the StylesProvider export was not working in every
+  environment
--- a/packages/system/src/providers.tsx
+++ b/packages/system/src/providers.tsx
@@ -52,13 +52,12 @@ export function useTheme<T extends object = Dict>() {
   return theme as WithCSSVar<T>
 }
 
-export const [StylesProvider, useStyles] = createContext<
-  Dict<SystemStyleObject>
->({
+const [StylesProvider, useStyles] = createContext<Dict<SystemStyleObject>>({
   name: "StylesContext",
   errorMessage:
     "useStyles: `styles` is undefined. Seems you forgot to wrap the components in `<StylesProvider />` ",
 })
+export { StylesProvider, useStyles }
 
 /**
  * Applies styles defined in `theme.styles.global` globally


with-heart:
I still don't understand why this doesn't work. I've used the `export const [a, b] = thing` syntax so many times before with no issue. It even works fine in this CSB: https://codesandbox.io/s/busy-jackson-f9qlp?file=/src/array.js

with-heart:
You're right though, it's definitely not there. But it's still not in the CSB deploy built with this change :( https://codesandbox.io/s/create-react-app-ts-forked-65fng?file=/src/App.tsx

That doesn't make any sense.

TimKolberger:
Yeah, I find it really weird as well, but it is the only change I could find which relates to the StylesProvider.

I hope it fixes it, but as Mark said, the CSB deploy has still the same issue.

ljosberinn:
The following syntax breaks in CSB, probably due to transpilation reasons:
```js
export const [x,y] = something();
export const { x, y } = something();
```

The return value of `List` is [this](https://github.com/chakra-ui/chakra-ui/blob/main/packages/layout/src/list.tsx#L61):

```tsx
  return (
    <StylesProvider value={styles}>
      <chakra.ul
        ref={ref}
        listStyleType={styleType}
        listStylePosition={stylePosition}
        /**
         * We added this role to fix the Safari accessibility issue with list-style-type: none
         * @see https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html
         */
        role="list"
        __css={{ ...styles.container, ...spacingStyle }}
        {...rest}
      >
        {validChildren}
      </chakra.ul>
    </StylesProvider>
  )
  ```
  
Said `StylesProvider` is re-exported [here](https://github.com/chakra-ui/chakra-ui/blob/main/packages/system/src/index.ts#L7) from [here](https://github.com/chakra-ui/chakra-ui/blob/main/packages/system/src/providers.tsx#L55). Perfectly valid syntax so far.

However... in CSB, `system/src/index` becomes this:
image
image

Already in this preview we notice the absence of `StylesProvider`. Why is that?

image

It's clearly there, but....
image
there is no `exports.StylesProvider` happening. Only two occurrences.

So what changed? This is the still-working, older sandbox:

image

where `useStyles` and `StylesProvider` have separate `Object.defineProperty`.

I _think_ [this](https://github.com/chakra-ui/chakra-ui/commit/b3fbf8f75f8eaaa4a7976ef4dfafaebe32f0a957#diff-e48d5223b8321c56a59fe377995997da8bd9bc25b592df3c1aa87b02b4f70fdb) commit changed that behaviour.

TimKolberger:
@ljosberinn Nice work digging this out of the compiled JS!
Then there must be something wrong with the branch based CSB deployments. Maybe CSB does not handle transitive dependencies well, and installs the published version of `@chakra-ui/system` instead.

Is this a babel bug, or is CSB doing some magic there? 🤔

segunadebayo:
Thanks for that detailed investigation @ljosberinn.

I'll merge this update, for now, to see if it improves CSB and also because it was a refactor that we suspect might be the cause of the issue.

with-heart:
Want to link #3641 for a little more context around this.

CompuIves:
@ljosberinn This is an amazing repro, thanks a lot for this. We were indeed not handling `export const []` correctly, a fix PR is now opened here: https://github.com/codesandbox/codesandbox-client/pull/5606.

segunadebayo:
Awesome! Thank you @CompuIves for helping with this.

cc: @ljosberinn thanks for digging into this 👌

PR #3644, opened 2021-03-23
docs: Migration docs for obsolete `onClose` prop on <ModalCloseButton />

ansonkao:
## 📝 Description

Added migration docs for obsolete `onClose` prop on `<ModalCloseButton />`

If `onClose` is not specified on `<Modal />`, it throws an error. 

See comments here from inside the codebase: https://github.com/chakra-ui/chakra-ui/blob/main/packages/modal/src/modal.tsx#L495

### DIFF
--- a/website/pages/docs/migration.mdx
+++ b/website/pages/docs/migration.mdx
@@ -529,7 +529,7 @@ prop instead.
 - Only pass `size` values defined in the components theme. Hard-coded values
   will be ignored. Update the styles in `theme.components.Modal` to reflect your
   custom values.
-  
+
 - You can now disable focus trap by passing `trapFocus={false}`.
 
 - Modal comes with preset transitions to make it easy for you. You can remove
--- a/website/pages/docs/migration.mdx
+++ b/website/pages/docs/migration.mdx
@@ -522,10 +522,14 @@ prop instead.
 - Removed `preserveScrollBarGap` prop. We preserve scroll bar gap by default to
   prevent any layout shift.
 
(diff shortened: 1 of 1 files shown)


website/pages/docs/migration.mdx
```diff
@@ -522,6 +522,10 @@ prop instead.
 - Removed `preserveScrollBarGap` prop. We preserve scroll bar gap by default to
   prevent any layout shift.
 
+- `ModalCloseButton` now reads the `onClose` action from the `Modal` context - 
+  you no longer need to pass the `onClick` to it. Conversely, the `onClose` prop
```
segunadebayo, line 526: I don't think we ever required the user to pass `onClose` to the `ModalCloseButton` even in v0.x.

https://v0.chakra-ui.com/modal

I think we can omit this part. Let me know your thoughts.

ansonkao: In v0.x, it worked fine if `onClose` is specified in `<ModalCloseButton />` but not in `<Modal />`. But in v1.x+, it breaks. This one threw me for a loop until I dug into the source code! I guess this is a relic from v0.7.

```suggestion
- `ModalCloseButton` now reads the `onClose` action from the `Modal` context. 
   If you had passed your close callback to the `ModalCloseButton.onClick`, you
   need to migrate that callback to be passed into the `Modal.onClose` prop.
```

Would that read better?

segunadebayo: Yes, thank you

PR #3674, opened 2021-03-26
refactor: extract react utils from core utils

codebender828:
Closes chakra-ui/chakra-ui-vue-next#41

## 📝 Description

This pull request introduces an extraction of `react` utilities in the `@chakra-ui/utils` package in to their own package called `@chakra-ui/react-utils`.

This allows the `@chakra-ui/utils` package to be "framework independent", thus making `@chakra-ui/styled-system` framework independent as well, allowing it to be more conveniently shared across different frameworks for Chakra UI.

Currently `@chakra-ui/styled-system` depends on `react` and `react-dom`. 

The dependency graph is somewhat like this:
`@chakra-ui/vue-next -> @chakra-ui/styled-system -> @chakra-ui/utils -> react, react-dom`

**Why is this an issue:**
Since we're using `@chakra-ui/styled-system` in [`@chakra-ui/vue-next`](https) to prevent code duplication, it currently requires consumers to install `react` as a dependency as well into the project.

In order to prevent the above mentioned issue from happening, this PR removes the `react` and `react-dom` dependent utilities into a standalone `@chakra-ui/react-utils` package.

## 🚀 New behavior

No new behaviors. Just modified some deps :)

Nope :)

## 📝 Additional Information
N/A


packages/editable/src/editable.tsx
```diff
@@ -9,13 +9,8 @@ import {
...
   HTMLChakraProps,
 } from "@chakra-ui/system"
-import {
-  createContext,
-  cx,
-  MaybeRenderProp,
-  runIfFn,
-  __DEV__,
-} from "@chakra-ui/utils"
+import { cx, MaybeRenderProp, runIfFn, __DEV__ } from "@chakra-ui/utils"
```
codebender828, line 12: As I'm reviewing this, I can see that the definition of the `MaybeRenderProp` exported from `@chakra-ui/utils` is as such:

```ts
export type MaybeRenderProp<P> =
  | React.ReactNode
  | ((props: P) => React.ReactNode)
```

It depends on types from React. Since this is a type, should it also be moved to the `@chakra-ui/react-utils`? I only refactored direct modules, but I am not confident that it was sufficient. Happy to get feedback here as well.

packages/modal/src/use-modal.ts
```diff
@@ -1,5 +1,6 @@
 import { useIds } from "@chakra-ui/hooks"
-import { callAllHandlers, mergeRefs, PropGetter } from "@chakra-ui/utils"
+import { callAllHandlers, PropGetter } from "@chakra-ui/utils"
```
codebender828, line 2: Same question about exported types here as well 🤔

codebender828 (commented):
Hey @chakra-ui/react-team !

Thanks for taking the time to review this. I've managed to get all tests and builds passing on this extraction PR. 

However, I still need some confirmation from the React gurus on how to handle the types exported from `@chakra-ui/utils` that depend on `React`. Since they are types and not actual modules, should they also be moved to `@chakra-ui/react-utils`?

packages/accordion/package.json
```diff
@@ -60,6 +60,7 @@
     "@chakra-ui/descendant": "1.0.9",
     "@chakra-ui/hooks": "1.2.0",
     "@chakra-ui/icon": "1.1.3",
+    "@chakra-ui/react-utils": "1.0.0",
```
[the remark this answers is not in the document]
with-heart: Yeap!

packages/editable/src/editable.tsx
with-heart: Yeah, I think React-specific types should be moved over as well

packages/modal/src/use-modal.ts
with-heart: Yeap! Anything that's specific to React shouldn't belong in `utils` anymore, types and all

packages/react-utils/README.md
```diff
@@ -0,0 +1,11 @@
+# @chakra-ui/react-utils
+
+A Quick description of the component
```
with-heart, line 3: Not sure we should even include a `README.md` for this unless we plan on providing more of a description of what the purpose is

with-heart (changes requested):
Only marking this as "changes requested" to make sure we generate a `changesets` file that includes a `minor` version bump for all packages that the `"@chakra-ui/react-utils": "1.0.0"` line is added to :)

packages/editable/src/editable.tsx
segunadebayo: I agree!

packages/react-utils/README.md
codebender828: Good point. Will make changes for the README.md.

A simple one line description for the component is sufficient.

It's helpful for contributors who may want to add more react utilities.

packages/modal/src/use-modal.ts
codebender828: Thanks. Moved all react-based types to `react-utils` 👍

packages/react-utils/README.md
codebender828: Included a better description in the README of `@chakra-ui/utils`

segunadebayo:
Thanks for working on this @codebender828. I'll review and take care of the conflicts.

PR #3679, opened 2021-03-26
docs(component-style): default Chakra component + custom component styling

lazarnikolov94:
## 📝 Description

Refactored the Custom Style doc page to demonstrate styling of the built-in `Button` component. Also added instructions on how to style a custom Card component that's not part of Chakra UI.


website/pages/docs/theming/component-style.mdx
```diff
@@ -132,10 +132,60 @@ const theme = extendTheme({
...
+    flexDirection: "column",
+    background: "white",
+    alignItems: "center",
+    gap: 6
+  },
+  // Two variants: rounded and smooth
+  variants: {
+    rounded: {
+      padding: "32px",
+      borderRadius: "12px",
```
TimKolberger, line 155: What do you think about encouraging the reader to use theme tokens to keep their components in sync with the design system?

website/pages/docs/theming/component-style.mdx
```diff
@@ -159,32 +209,50 @@ passed, the `defaultProps` defined in the style config will be used.
...
-our app:
+And lastly - the fun part - let's use our custom Card component anywhere in our app:
 
 ```jsx live=false
 // 1. Using the default props defined in style config
 function Usage() {
-  return <Button>Click me</Button>
+  return (
+    <Card>
+      <Image src="/lin.png" rounded="full" w={32} h={32} boxShadow="md" />
```
TimKolberger, line 231: When the user does the copy-pasta from the documentation, the images won't load, because they are not present on their system. 
A fully qualified image url could provide a better user experience.

TimKolberger (commented):
Good read @lazarnikolov94! ✨ 
I added my two cents as comments :)

website/pages/docs/theming/component-style.mdx
lazarnikolov94: That makes total sense! Thanks @TimKolberger

website/pages/docs/theming/component-style.mdx
lazarnikolov94: Sounds good. I'll add the placeholder image in the `public` folder and set the url to `https://chakra-ui.com/eric.jpg` for example.

TimKolberger (approved):
I love the structure of this doc page! Every section builds upon the previous one, so as reader I really feel like learning something new with every section and I do understand why the previous sections matter. :sparkles:

Issue #1443, opened 2020-07-28
Yarn 2 (Plug n' Play) exposes dependency inconsistencies


beforan:
Yarn 2 (in default Plug n' Play mode, not using `node_modules`) is much stricter around package dependencies.

Trying to use chakra in such an environment causes the following sorts of errors:

```
Module not found: Error: @chakra-ui/number-input tried to access @chakra-ui/hooks, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

**Expected Behavior**

`Module not found` errors should not occur when building projects using Chakra in a Yarn 2 Plug n' Play environment.

**Link to minimal reproduction**
https://github.com/beforan/chrakra-ui-v1-yarn2

Sorry it's not super minimal (as a repo) but I think the environment is important, and useful to showcase the issue and fix.

It's a brand new CRA app, with all the starter stuff removed and chakra `1.0.0-rc.0` added, and that's all. It also has storybook.

**Steps to Reproduce**

1. Have yarn globally installed (`npm i -g yarn` is recommended by yarn2's docs, but i believe any yarn install is fine)
    - it will run the repo's local yarn 2 binary
1. Run `yarn` to install dependencies
1. Run any of the following to see the errors:
    - `yarn start`
    - `yarn build`
    - `yarn storybook`

`storybook` is useful here as it throws all the errors at once; `start` and `build` seem to do it one by one. Some difference in how storybook and react-scripts invoke webpack, i guess.

This repo also showcases the current workaround (as described [here](https://github.com/beforan/chrakra-ui-v1-yarn2)):

1. Add `packageExtensions` to `.yarnrc.yml` to tell yarn about the missing dependencies.
    - I have done this in this repo, but they are commented out. this should highlight which dependencies need adding.

It also shows some peer dependencies which are installed in the local project, so I'm not sure of the cause of that, but it may warrant investigation. These are also noted in the `.yarnrc.yml`

**Suggested solution(s)**

Add the missing dependencies to the relevant packages:

```
"@chakra-ui/css-reset@1.0.0-rc.0":
  dependencies:
    "@emotion/core": "^10.0.28"
"@chakra-ui/core@1.0.0-rc.0":
  dependencies:
    "@chakra-ui/transition": "1.0.0-rc.0"
"@chakra-ui/drawer@1.0.0-rc.0":
  dependencies:
    "@chakra-ui/utils": "1.0.0-rc.0"
[...]

PR #3715, opened 2021-04-01
fix: forward transitive react-dom peer dependency

with-heart:
Semi-related to #1443 

## 📝 Description

`react-dom` is currently a `peerDependency` for `@chakra-ui/portal`, `@chakra-ui/react-utils`, and `@chakra-ui/toast`. This means that any package of ours that depends on one of these packages must also list `react-dom` as a `peerDependency`, as [peer dependencies are not transitive](https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0) and aren't automatically inherited.

`react-dom` is not listed as a `peerDependency` n any of our packages that depend on our other packages that list it, causing resolution errors to be reported by more modern package managers like yarn2, npm7, etc.

Example from yarn2 ([reported here](https://discord.com/channels/660863154703695893/660865137137156146/827100463882305556)):

image

You can also see this reflected on the bundlephobia page for any of the affected packages, where `react-dom` is listed as a part of the package's composition rather than as a dependency. See the `@chakra-ui/react` page for an example: https://bundlephobia.com/result?p=@chakra-ui/react@1.4.1

## 🚀 New behavior

All of the packages that depend on `@chakra-ui/portal`, `@chakra-ui/react-utils`, or `@chakra-ui/toast` have had `react-dom` added as a peer dependency.

No


with-heart:
Also need to list them as `devDependencies` too to pass `manypkg` checks. The root `package.json` dependencies should be fine for this situation, but that functionality needs to be added still: https://github.com/Thinkmill/manypkg/issues/88

segunadebayo:
I'm planning on removing the `react-dom` dependency for `react-utils` because I have a better alternative to `flushSync` that we use in the code.

If it's okay, we can omit the peerDeps for packages that depend on `react-utils.`

with-heart:
@segunadebayo That's perfect! I'll remove those from this PR

PR #3725, opened 2021-04-02
docs(hooks): add use-popper page

navarroaxel:
## 📝 Description

Add a page for the `usePopper` hook in the doc site.

### DIFF
--- a/.changeset/curly-ghosts-worry.md
+++ b/.changeset/curly-ghosts-worry.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/docs": patch
+---
+
+Add docs for `usePopper` hook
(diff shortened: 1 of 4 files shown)


segunadebayo:
Hi @navarroaxel, we're careful not to publicly document all hooks due to the fact that they might change over time. I'm going to approve some but not all.

This `usePopper` hook is one of those. We'd like to keep this as an internal package, for now.

navarroaxel:
Hi @segunadebayo, I have some questions:

1. Could we keep this doc page with a warning? `Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use them at your own risk.`.
2. Could you tell me which hooks I can or I can´t create the doc page?

segunadebayo:
Yes, I'll go through all the hooks this week and add `internal` flag to some of them. I think that'll help

PR #3726, opened 2021-04-02
fix(transition): rest param

RadovanPelka:
Closes # 

## 📝 Description

> Support custom variants for transition components

> Not allowed custom transition variants

## 🚀 New behavior

> You can now rewrite chakra default transition variant

No, nice to have only

### DIFF
--- a/.changeset/great-grapes-flow.md
+++ b/.changeset/great-grapes-flow.md
@@ -1,5 +1,5 @@
 ---
-"@chakra-ui/transition": major
+"@chakra-ui/transition": minor
 ---
 
 Allow custom transition variants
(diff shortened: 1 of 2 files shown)


.changeset/great-grapes-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@chakra-ui/transition": major
```
segunadebayo, line 2: Kindly change this to a `minor`. I don't see it as a breaking change since it'll work as it used to, you're only making it possible to pass custom variants

RadovanPelka (commented):
Please review

RadovanPelka:
@segunadebayo 🙏

paescuj:
Nice! 😄 Has this been documented anywhere?

PR #3733, opened 2021-04-03
refactor: update popper for better ssr

### LINKED ISSUE #2481: Popover with width property set has positioning problem
# Bug report

## Describe the bug

Before rc.6 version I was able to set Popover's width (or minWidth) property to make it bigger. Using version rc.6 (or newer) there is a problem with popover positioning while width attribute is set.

## To reproduce

1. Open popover.
2. Popover has wrong position (is not aligned with its triggering element).

## Minimal reproduction

https://codesandbox.io/s/falling-microservice-oypkl?file=/src/index.js

## Expected behavior

Popover position should be recalculated correctly

## System information

- OS: macOS
- Version of @chakra-ui/react: 1.0.0

### LINKED ISSUE #3713: Popover: can't use popper's "offset" modifier in 1.4.x version
# 🐛 Bug report

I am unable to use the [popper's offset modifier](https://popper.js.org/docs/v2/modifiers/offset/) to add some offset to my popover since chakra's version `1.4.0`. It was working up until version `1.3.4`

## 💥 Steps to reproduce

1. Create a `<Popover />` instance with a `offset` modifier
2. Click on the popover's trigger
3. The popover has no offset

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/popover-offset-modifier-not-working-eqsnn?file=/src/index.js

The exact same code works as expected if you just change the chakra version to `1.3.4` or below (and the framer-motion version for compatibility):  https://codesandbox.io/s/popover-offset-modifier-working-rbk07?file=/src/index.js

## 🌍 System information

| Chakra UI        |      1.4.0 and above      |
| Browser          |     Chrome       |
| Operating System |   Mac          |

## 📝 Additional information

I have looked at the repo and I guess this could be related to #3583

segunadebayo:
Closes #2481 
Closes #3713

## 📝 Description

- Updates popper code to expose prop getters in a way that improves SSR,
- Update popper to use `min-width: max-content` so it can be sized according to its child content
- Use `ref` callback to detect the presence of popover header and body instead of use-effect

No

### DIFF
--- a/.changeset/eleven-frogs-kneel.md
+++ b/.changeset/eleven-frogs-kneel.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react-utils": patch
+---
+
+Update prop getter v2 type to take second parameter
--- a/.changeset/smart-turkeys-build.md
+++ b/.changeset/smart-turkeys-build.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/popper": minor
+---
+
+Add prop getters for popper and arrow for better ssr support
+
+Replace `utils` dependency with `react-utils`
--- a/.changeset/stale-falcons-divide.md
+++ b/.changeset/stale-falcons-divide.md
@@ -1,0 +1,8 @@
+---
+"@chakra-ui/theme": patch
+---
+
+**Popover**
+
+Fix: moved `maxW` from popover's `popper` to `content` to allow for better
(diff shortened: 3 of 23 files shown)


mandyGitH:
For anyone having the same problem, thread #4104 helped with my case

PR #3747, opened 2021-04-06
chore(storybook): upgrade to 6.2

with-heart:
## 📝 Description

This PR upgrades us to Storybook v6.2. Also widens the range of a few `@babel` dependencies to more closely match what Storybook requires.

Running/building Storybook throws an error complaining about "'loose' mode configuration must be the same" for a few plugins, but the error doesn't seem to cause the process to fail in any way. I spent more time than I care to admit trying to fix that error and this is the best I could come up with. Bonus points to anyone that can fix it.

No


TimKolberger:
I found an additional config at `.storybook/.babelrc`. Adding the `loose` option there fixes the error for me.

```json
{
  "presets": ["@babel/preset-env"],
  "plugins": [
    ["@babel/plugin-proposal-class-properties", { "loose": true }],
    ["@babel/plugin-proposal-private-methods", { "loose": true }]
  ]
}
```

with-heart:
This should be good to go

PR #3757, opened 2021-04-07
Fix/select icon rtl

### LINKED ISSUE #3755: Select icon should be at the inline-end for RTL
# 🐛 Bug report

The icon for `Select` is always on the right, regardless of LTR or RTL.

## 💥 Steps to reproduce

1. Go to the sandbox
2. Check the icon position

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/chakra-ui-select-rtl-icon-1kwbx

## 🧐 Expected behavior

The icon should be at the end (on the left for RTL, right for LTR)

image

## 🧭 Possible Solution

It should be placed at the inline-end, instead of right. 

## 🌍 System information

| Chakra UI        |      1.4.2      |
| Browser          |      Chrome      |
| Operating System |      Windows 10      |

estheragbaje:
Closes #3755

## 📝 Description

Updated RTL support such that we use `inlineStart` instead of `left` and `inlineEnd` instead of `right` for Select Component theme.

On RTL, the select icon was still showing up on the right instead of flipping to the left.

## 🚀 New behavior

On RTL, the select icon shows on the left. 

No

### DIFF
--- a/packages/theme/src/components/select.ts
+++ b/packages/theme/src/components/select.ts
@@ -19,7 +19,7 @@ function baseStyleField(props: Record<string, any>) {
 const baseStyleIcon = {
   width: "1.5rem",
   height: "100%",
-  right: "0.5rem",
+  insetEnd: "0.5rem",
   position: "relative",
   color: "currentColor",
   fontSize: "1.25rem",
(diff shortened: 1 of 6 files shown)


packages/theme/src/components/select.ts
```diff
@@ -19,7 +19,7 @@ function baseStyleField(props: Record<string, any>) {
 const baseStyleIcon = {
   width: "1.5rem",
   height: "100%",
-  right: "0.5rem",
+  insetEnd: "0.5rem",
```
manzoorwanijk, line 22: Shouldn't this be `insetInlineEnd`? 🤔

PR #3763, opened 2021-04-07
fix(tabs): do not unmount tab content when lazy

### LINKED ISSUE #3683: Tabs isLazy behavior
# 🚀 Feature request

Possibly change isLazy behavior to mount on first select and never unmount. If we want to keep the current behavior, consider adding a new prop.

I read through https://github.com/chakra-ui/chakra-ui/issues/1384 to understand the problem `isLazy` is trying to solve. There are two distinct features asked for in that issue:

- Do not mount all tabs at once, since they might do data loading.
- Mark individual tab panels as lazy. This was a later feature request, not added in the linked PR.

The solution to the problem above was to mount on select and unmount on a different tab being selected. I think this can be improved to further reduce data loading.

Currently a lazy tab will unmount whenever it's unselected, causing it to load data on being selected again. A true lazy tab will only mount once, on first selection. This mount-on-first-selection behavior makes the second feature request above unnecessary.

## ✅ Proposed solution or API

Pretty simple. Each tab panel has an internal flag (useRef) to keep track of whether it has _ever_ been selected.
- If it has not been selected and it's not active, render `null`
- Otherwise set the flag to true and render its children as normal.

## ↩️ Alternatives

If we don't want to change the `isLazy` behavior, we could add a new prop.

Can anyone think of a use case where the current behavior is preferred over the one I'm suggesting?

## 📝 Additional Information

N/A

stellarhoof:
fix(tabs): do not unmount tab content when lazy

Closes #3683

## 📝 Description

Tabs' isLazy behavior is now to mount on first select and never unmount.

Current behavior is to unmount on navigating away from a tab.

## 🚀 New behavior

Tabs' isLazy behavior is now to mount on first select and never unmount.

No

### DIFF
--- a/.changeset/hot-dodos-reflect.md
+++ b/.changeset/hot-dodos-reflect.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/tabs": patch
+---
+
+Do not unmount lazy tabs when unselected
--- a/packages/tabs/src/use-tabs.ts
+++ b/packages/tabs/src/use-tabs.ts
@@ -383,6 +383,9 @@ export function useTabPanels<P extends UseTabPanelsProps>(props: P) {
 export function useTabPanel(props: Dict) {
   const { isSelected, id, children, ...htmlProps } = props
   const { isLazy } = useTabsContext()
+  const hasBeenSelected = React.useRef(false)
+
+  if (isSelected) hasBeenSelected.current = true
 
   return {
     /**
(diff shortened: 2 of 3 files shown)


segunadebayo:
Hi @stellarhoof,

Want to update the test to match the new behavior?

stellarhoof:
@segunadebayo Done

segunadebayo:
Great work!

Sorry to bother you but if you could extend this update to the use-popover hook as well, that would be fantastic! It also uses the `isLazy prop.

stellarhoof:
@segunadebayo I'll make a different PR then since you merged this one.

PR #3780, opened 2021-04-09
docs(component-styles): use __css for useStyles return values

### LINKED ISSUE #3770: Conflict between variant/size and style props
# 🐛 Bug report

When using Chakra-UI with ReactJS, style props are being overwritten by variant/size props on custom components. The documentation suggests we must use the [useStyleConfig](https://chakra-ui.com/docs/theming/component-style#usestyleconfig-api) hook in order to have access to variant/size. I found two problems with this approach.
1. I have access to variant/size in chakra components without using the hook. I'm not sure if this is a bug or not. If not: what purpose does the hook serve?
2. When using the hook, other style props I might want to pass to my components when they are rendered are overwritten by variant/size styles. 

## 💥 Steps to reproduce

1. Create a react app and wrap it with a Chakra Provider and a custom theme`<ChakraProvider theme={custom} />` 
2. On the custom theme, create a custom style for the Button component. Define a variant with a custom font size. `myVariant: { fontSize: "2rem" }`
3. Creat a `<CustomButton />`component, as described by the [useStyleConfig](https://chakra-ui.com/docs/theming/component-style#usestyleconfig-api) API. 
3. Render side by side a chakra button and the custom button, both with the variant `myVariant` and the prop fontSize. 
```jsx
<Button variant="myVariant" fontSize="1rem">Button</Button>
<CustomButton variant="myVariant" fontSize="1rem">CustomButton</CustomButton>
```
4. Notice that the custom variant works for both the Button and the CustomButton, but the font size prop only works on the Button, but not on the CustomButton

## 💻 Link to reproduction

https://codesandbox.io/s/fervent-edison-rmlwh?file=/src/App.js

## 🧐 Expected behavior

- Style props should have precedence over variant/size styles
- Not sure if variant/size should work without the useStyleConfig API

## 🌍 System information

| @chakra-ui/react       |     1.4.1      |
| Google Chrome               |     89.0.4389.114 64bits
| Operating System |      Windows 10 Home     |       

## 📝 Additional information
Also, I'm currently using Chakra-UI on a work project. We are currently using the useStyleConfig API on custom buttons and checkboxes, but because of the bugs described above we are thinking of switching to out-of-the-box chakra components. Is there a reason we to keep using the useStyleConfig API?

TimKolberger:
Closes #3770

## 📝 Description

Updates the component style docs.

Docs promote the `sx` prop to pass theme styles to the component. This leads to some unexpected behaviour regarding overrides.

## 🚀 New behavior

Now the docs use the `__css` prop like all chakra components do.

No

### DIFF
--- a/website/pages/docs/theming/component-style.mdx
+++ b/website/pages/docs/theming/component-style.mdx
@@ -89,42 +89,43 @@ const Button = {
   baseStyle: {
     fontWeight: "bold",
     textTransform: "uppercase",
-    borderRadius: "base" // <-- border radius is same for all variants and sizes
+    borderRadius: "base", // <-- border radius is same for all variants and sizes
   },
   // Two sizes: sm and md
   sizes: {
     sm: {
       fontSize: "sm",
       px: 4, // <-- px is short for paddingLeft and paddingRight
-      py: 3  // <-- py is short for paddingTop and paddingBottom
+      py: 3, // <-- py is short for paddingTop and paddingBottom
     },
     md: {
       fontSize: "md",
       px: 6, // <-- these values are tokens from the design system
-      py: 4  // <-- these values are tokens from the design system
-    }
+      py: 4, // <-- these values are tokens from the design system
+    },
   },
   // Two variants: outline and solid
   variants: {
     outline: {
       border: "2px solid",
       borderColor: "purple.500",
-      color: "purple.500"
+      color: "purple.500",
     },
     solid: {
       bg: "purple.500",
-      color: "white"
-    }
+      color: "white",
+    },
   },
   // The default size and variant values
   defaultProps: {
     size: "md",
     variant: "outline",
-  }
+  },
 }
 ```
 
-Makes sense right? Now, let's update the theme to include this new component style.
+Makes sense right? Now, let's update the theme to include this new component
+style.
 
 ```jsx live=false
 import { extendTheme } from "@chakra-ui/react"
@@ -136,7 +137,9 @@ const theme = extendTheme({
 })
 ```
 
-**And that's it!** You can use your new Button along with its custom variants throughout your app. But what if we want to create a custom component that's not part of Chakra UI? Let's use the following design spec for a Card component:
+**And that's it!** You can use your new Button along with its custom variants
+throughout your app. But what if we want to create a custom component that's not
+part of Chakra UI? Let's use the following design spec for a Card component:
 
 <Img mt="10" mb="12" src="/card-spec.png" />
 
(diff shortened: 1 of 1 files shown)


website/pages/docs/theming/component-style.mdx
```diff
@@ -166,7 +166,7 @@ function Button(props) {
   const styles = useStyleConfig("Button", { size, variant })
 
   // 3. Pass the computed styles into the `sx` prop
```
segunadebayo, line 168:
```suggestion
  // 3. Pass the computed styles into the `__css` prop
```
 
We also need to explain what the heck the `__css` means to avoid confusion. We need to explain that it is similar in spirit to the `sx` prop but allows for prop overrides.

User can use either `sx` or `__css` prop depending on their "strictness" needs.

TimKolberger: Very true - What do you think about adding a section about the `__css` prop at [The sx prop](https://chakra-ui.com/docs/features/the-sx-prop). `__css` does the same as `sx`, it only has a different priority.

segunadebayo: I'd prefer we mention it in the theming docs only to avoid confusion. We can keep the `sx` prop in the docs and add a section for `Props Priority/Override` in the theming props that replace the `sx` with `__css`.

website/pages/docs/theming/component-style.mdx
```diff
@@ -210,29 +216,38 @@ The computed styles for the component based on `props` passed. If no `props` is
...
 
-  return (
-    <Box sx={styles} {...rest}>
-      {children}
-    </Box>
-  );
+  const styles = useStyleConfig("Card", { variant })
+
+  // Pass the computed styles into the `__css` prop
+  return <Box __css={styles} {...rest} />
```
aaarichter, line 227: Hi @segunadebayo 
I read this piece of code in the documentation and wondered what happened to `children`. Shouldn't it read:

```suggestion
  return <Box __css={styles} {...rest}>{children}</Box>
```

TimKolberger: jsx allows `children` in a prop notation and as  _real_ children.

```jsx
<div children="Hello" /> === <div>Hello</div>
```

Spreading all remaining props equals to the example above.

Please note that _real_ children have precedence over the prop notation:
  
```jsx
<div children="Hello">World</div> === <div>World</div>
```

aaarichter: Hey @TimKolberger 

thanks for your response. 
I agree that with the spreading of `{...rest}` on the component should pass `children` as props. 

However in line 222
```js
const { variant, children, ...rest } = props
```
`children` is destructed from `props` and is then not a part of `rest` object.

aaarichter: @TimKolberger https://github.com/chakra-ui/chakra-ui-docs/pull/284 😉

PR #3790, opened 2021-04-12
feat: added the ability to change the placeholder color of the Select component

lazarnikolov94:
## 📝 Description

Added a new `placeholderColor` prop in the `Select` component that can be used to define the placeholder color individually.

The placeholder color cannot be set. It can be controlled with the `color` property, but that also changes the color of the selected value, not the placeholder only.

## 🚀 New behavior

The placeholder gets the `placeholderColor` color, while the selected value gets the `color` color.

No. If no `placeholderColor` is being set the placeholder color will remain the default one.

### DIFF
--- a/.changeset/sharp-ties-film.md
+++ b/.changeset/sharp-ties-film.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/select": minor
+"@chakra-ui/docs": minor
+---
+
+Added a new placeholderColor prop in the Select component that can be used to
+define the placeholder color individually.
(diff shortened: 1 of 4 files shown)


segunadebayo:
Hi @lazarnikolov94,

Thanks for working on this. We want to be weary of adding implicit logic to our components.

This is something that can be added in userland if they need it. I'm of the opinion that a select should not look like an input (with placeholder color functionality), given that it's not technically a placeholder.

I'd recommend adding it in userland this way:

```jsx
function Example(){
 const [isPlaceholderSelected, setIsPlaceholderSelected] = useState(false);
 return (
    <Select
        color={isPlaceholderSelected ? "gray" : "red"}
        onChange={(event) => {
          setIsPlaceholderSelected(event.currentTarget.value === "");
          if (onChange) onChange(event);
        }}
      />
  )
}
```

Hope this makes sense. Thanks for working on this.

PR #3793, opened 2021-04-13
feat: embedded getting started video in docs

lazarnikolov94:
## 📝 Description

Embedding the YouTube video "Getting Started with Chakra UI" in the docs

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -34,6 +34,7 @@
     "react-icons": "^4.1.0",
     "react-live": "^2.2.3",
     "react-multi-ref": "1.0.0",
+    "react-player": "^2.9.0",
     "remark": "^13.0.0",
     "remark-autolink-headings": "^6.0.1",
     "remark-emoji": "^2.1.0",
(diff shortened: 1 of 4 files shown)


TimKolberger:
Nice work with the getting started video @lazarnikolov94! ✨ 

I think that it is a good idea to link to this video, but not as the first thing on the getting started page.
In the getting started section at `Next.js` seems like the best fit imo. 
What do you think?

lazarnikolov94:
> Nice work with the getting started video @lazarnikolov94! ✨
> 
> I think that it is a good idea to link to this video, but not as the first thing on the getting started page.
> In the getting started section at `Next.js` seems like the best fit imo.
> What do you think?

It does make sense. I can move it down.

lazarnikolov94:
> > Nice work with the getting started video @lazarnikolov94! ✨
> > I think that it is a good idea to link to this video, but not as the first thing on the getting started page.
> > In the getting started section at `Next.js` seems like the best fit imo.
> > What do you think?
> 
> It does make sense. I can move it down.

You can check it out now @TimKolberger

TimKolberger (approved):
Looks good to me ✨

PR #3833, opened 2021-04-16
docs (global styles) Correction of example config

mmmoli:
Closes nothing. I spotted the mistake and quickly fixed it without making an Issue first.

## 📝 Description

While defining global styles, I copied and pasted the example in the docs hoping to see them apply.

The Global Styles from the docs weren't having any effect. 

## 🚀 New behavior

Typo in the docs. Global styles need to be nested under a selector e.g. `html, body`.

No

## 📝 Additional Information

None


lazarnikolov94 (approved):
Good catch @mmmoli!

PR #3837, opened 2021-04-16
fix(components): components to swap sides (left/right) when theme direction='rtl'

mcha-dev:
## 📝 Description
Menu `placement` currently does not reflect theme direction (in other words, does not flip right/left when changing RTL/LTR).

Progress so far
- [x] Menu button
- [x] Slider
- [x] Drawer
- [x] Tooltip
- [x] Toast
- [x] Popover

## ⛳️ Current behavior
MenuList does not flip direction when theme's direction changes
[codesandbox example](https://codesandbox.io/s/goofy-mayer-pbmpt?file=/src/index.tsx)

## 🚀 New behavior
if direction is 'rtl' then flip `right/left|start/end`

No

## 📝 Additional Information
If this change is not required feel free to close it. 
If I'm missing something (i.e. not following a certain convention) feel free to correct or point required changes.


segunadebayo (changes requested):
Great work here @mcha-dev,

Some structural changes I'd make.

- Move `flipDirection` and its test to `packages/utils/rtl`
- Add changeset for `utils` package
- Use the `flipDirection` to `usePopper` hook and update `UsePopperProps` to take `direction : rtl | ltr` which will swap the placement based on this direction prop.
- Consider updating the `transformOrigin` modifier based on `direction`. This might not be needed thought but it's worth testing.

mcha-dev:
Thanks @segunadebayo I have made the changes you suggested. 

I looked into the `transformOrigin`. It's getting the `placement` as part of the state it receives (if I understood it correctly). So I think it's as you said, no change seems to be needed. 

Let me know if more change is needed.

mcha-dev:
Ops, forgot about the change sets. Will do them next

mcha-dev:
Do we also need a `changeset` for `menu` package?

packages/menu/src/use-menu.ts
```diff
@@ -18,6 +18,7 @@ import {
   getValidChildren,
   mergeRefs,
 } from "@chakra-ui/react-utils"
+import { useTheme } from "@chakra-ui/system"
```
segunadebayo, line 21: - useMenu shouldn't have to know about `@chakra-ui/system` or the theme. Please consider passing a `direction` prop to `useMenu`.

- Then you can read the direction from `useTheme` within the `Menu` component and pass it's value to the `useMenu` hook.

segunadebayo: Yes, you need a changeset for the `menu` package. Basically, every package you've changed needs a changeset.

mcha-dev:
Hi again @segunadebayo 

I noticed the same behavior in Slider & Drawer. [codesandbox](https://codesandbox.io/s/intelligent-varahamihira-1hxu4?file=/src/index.tsx) 

Should we 
- have an optional prop like `followThemDirection?: bool` if it's true, we flip right/left OR
- have additional `start` & `end` props, if users use `start/end` then we follow the theme direction, if they use `left/right` we don't

The first option seems easier but I don't know which one would fit with the convention here. What do you think?

segunadebayo:
Yes, we need to have the Slider, Drawer, Popover, Tooltip, and Toast RTL aware. Just like you did here.

Feel free to update them to use the same pattern as mentioned. I like the idea of adding a `start/end` option but I'd like us to do the flipping by default.

mcha-dev:
We can do the flipping internally but there might be scenarios in which users want `<Drawer />` (for example) to always be on the right _regardless of the theme direction._ 

> `start/end` fits more with CSS logical properties

mcha-dev:
Progress so far
- [x] Menu button
- [x] Slider
- [x] Drawer
- [ ] Tooltip
- [ ] Toast

mcha-dev:
@segunadebayo I'm working on the toaster, which is easy to do (to swap right/left positions under `rtl`) but we also need to swap the icon & close button. I cannot seem to find the content of the toast (the JSX/Icon/Button..).

toast

Also, when I work locally on a fork, I usually just `yarn build` it then install it into a project locally `yarn add file:/Users/mcha/Developer/personal/forks/chakra-ui/` but this doesn't seem to work with chakra, any idea how to build and test it (install it) locally?

mcha-dev:
Hi @segunadebayo, any more suggestions/work left to do on this?

segunadebayo:
Thanks for all your work on this @mcha-dev. I will review shortly

mcha-dev:
Hi @segunadebayo, have you had the chance to take a look at this? any comments?

with-heart:
@segunadebayo Could you take another look at this when you have a sec? I don't have enough experience with RTL/LTR switching to know what I'm looking for here.

mcha-dev:
Hi Mark, thanks for passing by. 

In a nutshell, all what this PR does is: 
1. Checks ChakraTheme direction `RTL | LTR`
2. Swaps the word 'right' with 'left' (& vice-verse) **if** the theme direction is `RTL`. _So that things are displayed on the proper side under RTL_

> Not trying to rush an approval here, just thought a simple explanation might help you better understand what's happening here.

.changeset/itchy-lions-provide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@chakra-ui/popper": patch
+"@chakra-ui/utils": patch
+---
+
+usePoper accepts direction prop now so it handles splacement for rtl languages
```
with-heart, line 6:
```suggestion
`usePopper` now accepts a `direction` prop so it can handle placement for RTL languages.
```

.changeset/nervous-pandas-search.md
```diff
@@ -0,0 +1,7 @@
+---
+"@chakra-ui/popover": patch
+"@chakra-ui/toast": patch
+"@chakra-ui/tooltip": patch
+---
+
+Flip right/left under RTL theme direction
```
with-heart, line 7:
```suggestion
Resolved an issue where the theme's `direction` wasn't handled. The component will now correctly flip between left/right based on `direction`.
```

.changeset/itchy-lions-provide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@chakra-ui/popper": patch
+"@chakra-ui/utils": patch
```
with-heart, line 3:
```suggestion
"@chakra-ui/popper": minor
```

This should be minor since it's adding new functionality to the hook rather than fixing a bug like the component packages. 

I'm gonna create a new changeset for `@chakra-ui/utils` because it should be more detailed about what new functions were added.

with-heart (changes requested):
REALLY appreciate you adding lots of tests here. Just a few nitpicks about the changesets and this is g2g from my view. Would still like Segun to take a look, just to make sure.

with-heart:
@allcontributors please add @mcha-dev for code and test

with-heart (approved):

segunadebayo:
Thanks for your hard work on this @mcha-dev.

I'll review this today.

segunadebayo (approved):

PR #3839, opened 2021-04-16
feat(tabs): make new isLazy behavior toggleable via unmountHiddenPanels prop

### LINKED ISSUE #3683: Tabs isLazy behavior

with-heart:
Closes #3683 

## 📝 Description

The behavior of the `isLazy` prop was broken in `@chakra-ui/tabs@1.3.2` due to a change in its default behavior. This change reverts the default behavior `isLazy` behavior to its old default, and allows toggling the newer behavior through the new `umountHiddenPanels` prop available on the `Tabs` component.

`isLazy` behavior was broken by `1.3.2`. `isLazy` leaves mounted the panels for any tabs that were previously selected.

## 🚀 New behavior

- `isLazy` default behavior restored to original pre-`1.3.2` behavior. When a tab is deselected, its panel is unmounted
- Added `unmountHiddenPanels` prop (defaults to `true`) to `Tabs`. This prop only works with `isLazy`, and allows the user to control whether tabs remain mounted or are unmounted (default behavior, `true`) when another tab is selected.

No, but it is breaking a previous breaking change by reverting it back to the old behavior.

## 📝 Additional Information

See #3763 for the previous change to this behavior

### DIFF
--- a/.changeset/perfect-beans-raise.md
+++ b/.changeset/perfect-beans-raise.md
@@ -1,0 +1,14 @@
+---
+"@chakra-ui/tabs": minor
+---
+
+This change restores the behavior of the `isLazy` prop (which was broken by
+`@chakra-ui/tabs@1.3.2`) and adds a new `unmountHiddenPanels` prop which
+configures the behavior of `isLazy`.
+
+If you'd like for your tab panel components to be unmounted when a different tab
+is selected, please continue to use `isLazy`. This is the default behavior.
+
+If you'd like for your tab panel components to remain mounted (but hidden) after
+when a different tab is selected, use `unmountHiddenPanels={false}` in
+combination with `isLazy`.
--- a/packages/tabs/src/use-tabs.ts
+++ b/packages/tabs/src/use-tabs.ts
@@ -18,6 +18,7 @@ import {
   EventKeyMap,
 } from "@chakra-ui/react-utils"
 import * as React from "react"
+import { shouldTabPanelRenderChildren } from "./utils"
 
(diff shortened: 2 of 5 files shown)


stellarhoof (commented):
Looks good 👍

TimKolberger (approved):

lazarnikolov94 (approved):

PR #3841, opened 2021-04-16
BREAKING(@chakra-ui/gatsby-plugin): correct fast refresh and theme behavior

LekoArts:
## 📝 Description

Hi!

This is essentially the same as https://github.com/system-ui/theme-ui/pull/1659 but for Chakra UI.
When looking at https://github.com/gatsbyjs/gatsby/issues/30387 and using Chakra UI in my own personal site I noticed that Fast Refresh didn't behave correctly when editing theme files. I had to manually refresh the page to see changes.

While working on this PR I followed the guide on https://chakra-ui.com/guides/integrations/with-gatsby and found that when I only provided a `theme.js` file I got this error:

2021_04_16_d056179d

I found https://github.com/vercel/next.js/issues/18941#issuecomment-736656326 and in my personal site I also use `extendTheme` + the guide for the manual setup at the bottom also suggests to use it. So I implemented it, too, in this PR.

- When editing theme files, e.g. `src/@chakra-ui/gatsby-plugin/theme.js` Fast Refresh doesn't hot reload changes and users need to manually refresh the page
- When using `src/@chakra-ui/gatsby-plugin/theme.js` without a `extendTheme` the page errors out

## 🚀 New behavior

Demo video of monkey-patched node_modules version of my changes:

https://user-images.githubusercontent.com/16143594/115039521-cfb40600-9ed0-11eb-9e48-f1c9a6bd0ac4.mp4

As you can see I now also was able to use a theme without manually using `extendTheme`

I'm not sure if it's breaking behavior if this case is true:

- The plugin uses `extendTheme()` for it's Provider
- The user uses an `extendTheme()` for their `src/@chakra-ui/gatsby-plugin/theme.js` file

Does a "double" usage of `extendTheme` harm/break anything?

## 📝 Additional Information

See https://github.com/system-ui/theme-ui/pull/1659 for an explanation about the _why_


tooling/gatsby-plugin/src/provider.js
```diff
@@ -0,0 +1,36 @@
+import * as React from "react"
+import {
+  ThemeProvider,
```
LekoArts, line 3: https://chakra-ui.com/guides/integrations/with-gatsby#chakra-providers lists a `ChakraProvider`. Does that make a difference?

tooling/gatsby-plugin/src/provider.js
```diff
@@ -0,0 +1,36 @@
...
+  ColorModeProvider,
+  CSSReset,
+  GlobalStyle,
+  PortalManager,
+  extendTheme,
+} from "@chakra-ui/react"
+import userTheme from "./theme"
+
+const theme = extendTheme(userTheme)
```
LekoArts, line 12: I also added this because https://chakra-ui.com/guides/integrations/with-gatsby#chakra-providers also shows this

tooling/gatsby-plugin/src/provider.js
with-heart: Whew, that page doesn't do a good job of providing any context around `ChakraProvider`. 

`ChakraProvider` essentially handles all of the things that the `content` variable on line 20 is doing. It'd probably make more sense for `WrapRootElement` to accept [the provider's props](https://chakra-ui.com/docs/getting-started#chakraprovider-props) so they could be spread onto the provider and users can directly override them. 

I'm not sure if that would make sense for `theme` or not though. I suppose with shadowing, it would still make sense for that to be handled through `./theme` rather than passed directly through a prop, right?

tooling/gatsby-plugin/src/provider.js
```diff
@@ -0,0 +1,36 @@
...
+  )
+  return (
+    <ThemeProvider theme={theme}>
+      {isUsingColorMode ? (
+        <ColorModeProvider options={theme.config}>{content}</ColorModeProvider>
+      ) : (
+        content
+      )}
+    </ThemeProvider>
+  )
```
with-heart, line 35: If `WrapRootElement` accepted the `ChakraProvider` props, this could become:

```suggestion
  return (
    <ChakraProvider theme={theme} {...props}>
      {element}
    </ChakraProvider>
  )
```

website/pages/guides/integrations/with-gatsby.mdx
```diff
@@ -69,17 +69,17 @@ export default theme
 Chakra exposes some context providers to give your website context of the theme,
 and color mode.
 
-Create the two files [`gatsby-browser.js`](https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#wrapRootElement) and [`gatsby-ssr.js`](https://www.gatsbyjs.com/docs/reference/config-files/gatsby-ssr/#wrapRootElement) at the root of your Gatsby project.
+Create the two files [`gatsby-browser.js`](https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#wrapRootElement) and [`gatsby-ssr.js`](https://www.gatsbyjs.com/docs/reference/config-files/gatsby-ssr/#wrapRootElement) at the root of your Gatsby project. Also create a `provider.js` that contains a React component that will be used in both files.
```
with-heart, line 72: It's been awhile since I've used `gatsby`, so trying to reason through this.

If the plugin already contains `gatsby-browser.js` and `gatsby-ssr.js` and the user can provide options via `gatsby-config.js`, would there be a need to do any of this? 

Could we maybe allow users to import the plugin's `WrapRootElement` if they wanted to render other things inside of it as part of the wrapper, so they wouldn't have to duplicate the effort of setting up `ChakraProvider` that we already do inside the plugin?

LekoArts: > If the plugin already contains gatsby-browser.js and gatsby-ssr.js and the user can provide options via gatsby-config.js, would there be a need to do any of this?

If the plugin exposes everything you can configure on `ChakraProvider` then there's not really the need to set up providers manually.

Gatsby merges the APIs together when multiple plugins use the same API so when a user uses `wrapRootElement` API in their e.g. `gatsby-ssr.js` file they don't need to import the plugin part or even care about it.

tooling/gatsby-plugin/src/provider.js
LekoArts: Sounds good, I'll incorporate it!

tooling/gatsby-plugin/src/provider.js
LekoArts: > I'm not sure if that would make sense for theme or not though. I suppose with shadowing, it would still make sense for that to be handled through ./theme rather than passed directly through a prop, right?

Yeah, shadowing is better here! With that you can hot reload the changes to the theme. If you would pass the theme via `gatsby-config` into the plugin options you'd need to restart the server every time 😬

LekoArts:
Thanks for the initial review! I'll address the changes today.

Can you comment on my question on "Is this a breaking change (Yes/No):"? :) Thanks!

tooling/gatsby-plugin/src/provider.js
LekoArts: Mhh, but the `ChakraProvider` doesn't have a `isUsingColorMode` prop/option. This would be a breaking change then

tooling/gatsby-plugin/gatsby-ssr.js
```diff
@@ -1,7 +1,19 @@
...
+import { WrapRootElement } from "./src/provider"
+import userTheme from "./src/theme"
+
+const theme = extendTheme(userTheme)
 
 export const onRenderBody = ({ setPreBodyComponents }) => {
-  setPreBodyComponents([<ColorModeScript key="chakra-ui-no-flash" />])
+  setPreBodyComponents([
+    <ColorModeScript
+      initialColorMode={theme.config.initialColorMode}
```
LekoArts, line 11: This was missing

tooling/gatsby-plugin/package.json
```diff
@@ -19,10 +19,10 @@
   },
   "peerDependencies": {
     "@chakra-ui/react": ">=1.0.0",
-    "gatsby": ">=2.29.3"
+    "gatsby": "^2.29.3 || ^3.0.0"
```
LekoArts, line 22: Set better peerDeps

tooling/gatsby-plugin/src/provider.js
```diff
@@ -0,0 +1,21 @@
+import * as React from "react"
+import { ChakraProvider, extendTheme } from "@chakra-ui/react"
+import userTheme from "./theme"
+
+const theme = extendTheme(userTheme)
+
+export const WrapRootElement = ({
```
LekoArts, line 7: The `isUsingColorMode` is obsolete really :/

tooling/gatsby-plugin/src/provider.js
TimKolberger: I'd recommend to move the `extendTheme` function call back into userland code.
There is a feature landing soon, which adds theme extensions and if a gatsby user wants to apply one of those, she would need to shadow this file.

tooling/gatsby-plugin/src/provider.js
TimKolberger: I don't mind a breaking change for this plugin 🤷  what do you think @chakra-ui/react-team

TimKolberger (commented):
Nice work @LekoArts! Thanks for taking care of the gatsby plugin 💖

tooling/gatsby-plugin/src/provider.js
LekoArts: @TimKolberger Ok, good to know! What about the bug I linked that was fixed with using `extendTheme()`?

If we go the route to move this to userland we'll need to update the docs.

tooling/gatsby-plugin/src/provider.js
with-heart: Yeah I'm totally fine with that!

tooling/gatsby-plugin/src/provider.js
TimKolberger: We should update the docs, because this approach lets the user control how the theme is created.
And because it previously worked only with `extendTheme` in `theme.js` I see it as a documentation flaw.

If a user wants to create its own application theme and provide a theme extension (in the near future), only the file `theme.js` needs to be shadowed.

LekoArts:
Ok, thanks for the input! I can work on this tomorrow

tooling/gatsby-plugin/src/provider.js
LekoArts: Done!

LekoArts:
Addressed the change requests, added `resetCSS` to the breaking changes (since we'll do a major anyways) and added two changesets

TimKolberger (approved):
Awesome :tada:

with-heart (approved):
Thanks @LekoArts! Especially love the detailed changesets, so awesome!

PR #3850, opened 2021-04-18
chore(css-vars): refactor css var

TimKolberger:
## 📝 Description

This PR includes a refactoring of `css-var.ts`, adds a helper function for creating and calculating css vars.

## 🚀 New behavior

The creation of css vars for a theme follows this pseudo code:

```ts
function createThemeVars(theme) {
  const tokens = extractTokens(theme)
  const cssVars = {}
  const cssMap = {}

  walkObject(tokens, (value, path) => {
    const key = path.join("-")
    cssVars[key] = value
  })

  return { cssVars, cssMap }
}
```

Create CSS vars:

```ts
const sliderWidth = cssVar("slider-width", "<fallback value>")
const { var, varRef } = sliderWidth
// var = --slider-width
// varRef = var(--slider-width, <fallback value>)
```

And calc CSS vars:

```ts
const translateX = calc.subtract(width, height).multiply(-1).toString()
```

No


packages/styled-system/src/create-theme-vars/create-theme-vars.ts
```diff
@@ -0,0 +1,97 @@
...
+      },
+    }
+  },
+  defaultHandler: (keys, value, options) => {
+    const lookupKey = keys.join(".")
+    const varKey = keys.join("-")
+
+    const { variable, reference } = cssVar(
+      varKey,
+      undefined, // TODO should we add a fallback value?
```
TimKolberger, line 80: Should we add a fallback value for the css vars?

segunadebayo: Yes, let's add a fallback value.

segunadebayo: I don't think we need to.

TimKolberger: Ok 👍  I'll remove the TODO comments

segunadebayo (approved):
Looks great, Tim!

Thanks for working on this!

PR #3854, opened 2021-04-18
feat: add new textDecoration style props

### LINKED ISSUE #3848: text-decoration-color does not work with the theme's colors
# 🐛 Bug report

Unable to pass in a theme's color into the `text-decoration-color` CSS property.

## 💥 Steps to reproduce

There are two ways to reproduce this. The first is by extending a theme:

`exampleTheme1` shows that `teal.300` can be passed to the `color` property. This is correct and functional behaviour. However, we are unable to do the same for the `text-decoration-color` property, which is incorrect behaviour.
```
const exampleTheme1 = extendTheme({
  layerStyles: {
    underline: {
      textDecorationLine: "underline",
      color: "teal.300",
      textDecorationThickness: "8px"
    }
  }
});
```

```
const exampleTheme2 = extendTheme({
  layerStyles: {
    underline: {
      textDecorationLine: "underline",
      textDecorationColor: "teal.300", // <-- instead of parsing "teal.300" into a color value, Chakra injects "teal.300" directly, which is an invalid CSS property
      textDecorationThickness: "8px"
    }
  }
});
```

Currently, the only way to pass a color to the `text-decoration-color` property is to pass in a hex value.
```
const exampleTheme3 = extendTheme({
  layerStyles: {
    underline: {
      textDecorationLine: "underline",
      textDecorationColor: "#4FD1C5",
      textDecorationThickness: "8px"
    }
  }
});
```

The same problem occurs when we want to pass inline styles into an element.
```
<chakra.span
  sx={{
    textDecorationThickness: "8px",
    textDecorationColor: "teal.300", // <--- this doesn't work either
    textDecorationLine: "underline"
  }}
>
  Second way
</chakra.span>
```

## 💻 Link to reproduction

https://codesandbox.io/s/small-bird-qr33s?file=/src/index.tsx:1139-1351

I have included the 3 different example themes so that you try them out.

## 🧐 Expected behavior

I expect to be able to pass in the theme's color into the `text-decoration-color` property in the following ways:
```
const exampleTheme2 = extendTheme({
  layerStyles: {
    underline: {
      textDecorationLine: "underline",
      textDecorationColor: "teal.300", // <---
      textDecorationThickness: "8px"
    }
  }
});
```
```
<chakra.span
  sx={{
    textDecorationThickness: "8px",
    textDecorationColor: "teal.300", // <---     
    textDecorationLine: "underline"
  }}
>
  Second way
</chakra.span>
```
## 🧭 Possible Solution

Sorry I did not go look through how Chakra converts a string into the color value

## 🌍 System information

| Chakra UI        |      1.5.1      |
| Browser          |     Firefox       |
| Operating System |    MacOS Big Sur Version 11.2.3       |

with-heart:
Closes #3848 

## 📝 Description

This changes adds style props for `textDecorationColor`, `textDecorationLine`, `textDecorationStyle`, and `textDecorationThickness`.

No

## 📝 Additional Information

I wasn't sure if I should do anything more fancy with `textDecorationLine`, `textDecorationStyle`, or `textDecorationThickness`. I noticed that `borderWidth` and `borderStyle` correspond to the theme keys for `borderWidths` and `borderStyles`, but I've never seen those used so wasn't sure if that was a pattern we should keep following or not.

### DIFF
--- a/packages/styled-system/src/config/typography.ts
+++ b/packages/styled-system/src/config/typography.ts
@@ -18,6 +18,10 @@ export const typography: Config = {
   whiteSpace: true,
   textDecoration: true,
   textDecor: { property: "textDecoration" },
+  textDecorationColor: t.colors("color"),
+  textDecorationLine: true,
+  textDecorationStyle: true,
+  textDecorationThickness: true,
   noOfLines: {
     static: {
       overflow: "hidden",
@@ -103,6 +107,22 @@ export interface TypographyProps {
    * The CSS `text-decoration` property
    */
   textDecor?: Token<CSS.Property.TextDecoration | number>
+  /**
+   * The CSS `text-decoration-color` property
+   */
+  textDecorationColor?: Token<CSS.Property.Color, "colors">
+  /**
+   * The CSS `text-decoration-line` property
+   */
+  textDecorationLine?: Token<CSS.Property.TextDecorationLine>
+  /**
+   * The CSS `text-decoration-style` property
+   */
+  textDecorationStyle?: Token<CSS.Property.TextDecorationStyle>
+  /**
+   * The CSS `text-decoration-thickness` property
+   */
+  textDecorationThickness?: Token<CSS.Property.TextDecorationThickness | number>
   /**
    * Used to visually truncate a text after a number of lines.
    */
--- a/.changeset/red-suns-join.md
+++ b/.changeset/red-suns-join.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/styled-system": minor
+---
+
+Added style props for `textDecorationColor`, `textDecorationLine`,
+`textDecorationStyle`, and `textDecorationThickness`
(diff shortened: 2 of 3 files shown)


packages/styled-system/src/config/typography.ts
```diff
@@ -18,6 +18,10 @@ export const typography: Config = {
   whiteSpace: true,
   textDecoration: true,
   textDecor: { property: "textDecoration" },
+  textDecorationColor: t.colors("color"),
```
segunadebayo, line 21:
```suggestion
  textDecorationColor: t.colors("textDecorationColor"),
```

packages/styled-system/src/config/typography.ts
```diff
@@ -18,6 +18,10 @@ export const typography: Config = {
...
+  textDecorationLine: true,
+  textDecorationStyle: true,
```
segunadebayo, line 23: I don't think we need these style props, to be honest, but since we already have `textDecorationColor`, we might want to leave them.

Another way to avoid this work is to use the css var: `textDecorationColor: var(--chakra-colors-green-300)`. I'll write some docs around this soon.

segunadebayo (changes requested):

segunadebayo:
Permit me to take on this PR and complete it. I'll open a PR shortly.

PR #3867, opened 2021-04-21
fix: Focus visible accessibility outline

### LINKED ISSUE #3449: Disable the accessibility outline for non-keyboard uses
Right now Chakra adds an outline (actually a `box-shadow`) to elements that receive focus, even when they are focused by clicking the mouse. For example:

image

This is to aid in keyboard accessibility, but it shouldn't happen when navigating via a mouse—check how popular websites handle this, including GitHub.

It looks like this was somehow integrated in `1.0` but the solution in https://github.com/chakra-ui/chakra-ui/issues/2016#issuecomment-691674038 leads to a broken link.

This behavior has been brought up before (eg. https://github.com/chakra-ui/chakra-ui/issues/2016, https://github.com/chakra-ui/chakra-ui/issues/708, https://github.com/chakra-ui/chakra-ui/issues/574) and the response has been to read [this Medium article](https://medium.com/@keeganfamouss/accessibility-on-demand-with-chakra-ui-and-focus-visible-19413b1bc6f9) and implement it. But this is something that 100% of users would need to do to, so it should be fixed in Chakra instead by default.

By keeping this broken default Chakra is actually encouraging devs to solve it incorrectly, using something like `box-shadow: none !important` regardless of navigation style. (eg. https://github.com/chakra-ui/chakra-ui/issues/708, https://github.com/chakra-ui/chakra-ui/issues/668)

The current default is actually going to result in worse accessibility.

mattwells19:
Closes #3449 

## 📝 Description

As described in the linked issue, it's undesirable to show the accessibility outline on components even when just selecting them. The outline should really only show when a keyboard user tabs over to it. One of the suggested ways to fix this was to only show the outline on `:focus-visible` instead of `:focus`.

The following components show a border when focused and when clicked:
 - `Accordion`
 - `Breadcrumb`
 - `Button`
 - `CloseButton`
 - `Editable`
 - `Link`
 - `Popover`
 - `Slider`
 - `Tabs`
 - `Tag`
 
These components on the website, though not consumable components, suffered the same problem.
 - The main search field
 - The Chakra "go back home" logo link
 - The linked heading in mdx components
 - The white Sponsor button on the landing page

## 🚀 New behavior

This PR changes all of the elements listed above to use `:focus-visible` to determine when to show the accessibility outline instead of using `:focus`.

No.

## 📝 Additional Information

The `Switch` and `Checkbox` components were left as. These components do not properly trigger the `:focus-within` selector when tabbed to. I assume this is because the element with the outline is not what's being tabbed to since it's just a show piece for the hidden input element.


segunadebayo:
It seems focus visible doesn't work on Safari: https://caniuse.com/?search=%3Afocus-visible.

Did you check this @mattwells19?

mattwells19:
> It seems focus visible doesn't work on Safari: https://caniuse.com/?search=%3Afocus-visible.
> 
> Did you check this @mattwells19?

Good catch! Unfortunately I don't have access to a Safari browser so I was not able to verify that.

As far as how to resolve this I can add a `:focus:not(:focus-visible)` in addition to the new `focus-visible`. The idea being that if focus-visible isn't supported, it'll fall back to just use focus like before.

What are your thoughts for this approach?

moritzjacobs:
What is the outlook of this being merged? :)

mattwells19:
> What is the outlook of this being merged? :)

Unfortunately, I haven't had the time to circle back to this to add the polyfill for Safari. I also still don't have access to a Safari browser to test those changes. If you, or anyone, has any recommendations for how to test that (without just buying a Mac) I'm open to them.

moritzjacobs:
> As far as how to resolve this I can add a `:focus:not(:focus-visible)` in addition to the new `focus-visible`. The idea being that if focus-visible isn't supported, it'll fall back to just use focus like before.

I have used [this](https://www.npmjs.com/package/focus-visible) polyfill a few times before and it works great, the caveat is, that since you cannot polyfill CSS selectors, it uses either a className or a `data-` attribute for applying the styles, so replacing `_focus` with `_focusVisible` is not enough, you would need to add some extra magic. I think this approach is somewhat complicated and probably not right for chakra, but I think an extra addon could be nice. Maybe I'll cook up a PoC this afternoon...

moritzjacobs:
(Disclaimer: this is my first "contribution" to chakra, so bear with me, if I get things wrong...)

I toyed around with adding the polyfill and you don't need to change much to make it work even in Safari:

`/packages/styled-system/src/pseudos.ts` => line 51: `  _focusVisible: "&[data-focus-visible-added]",`

(Not sure if `@chakra-ui/react` is the right place, let's pretend it is and then `yarn add focus-visible` there)
`/packages/react/src/chakra-provider.tsx` => line 2: `import "focus-visible/dist/focus-visible"`

Those changes should make your changes in this PR work in Safari, BUT:
- we're forcing the polyfill on every browser because we use the `data-focus-visible-added` selector approach and we can't keep `:focus-visible` as a fallback (Safari will just ignore the whole selector if we do)
- we lose noscript-support for keyboard nav
- we add a dependency
- we default to `:focus-visible` instead of `:focus` which shouldn't be assumed a welcome change by everyone, at least it should be configurable

moritzjacobs:
> @segunadebayo could you consider the proposal by @moritzjacobs when you have a sec?

I'm not really proposing it, I just have a similar use case as @mattwells19 and was interested what would make it work. To be honest, I'm somewhat torn if the polyfill should make it into the core like that, right now I'm leaning towards no.

TimKolberger:
Thank you for working on this issue @mattwells19!

I'd like to close this stale PR because the implemented solution does not work in Safari. 
Currently there are two viable options to achieve the intended behaviour:

1. Provide a custom theme
1. Use the package `focus-visible`

```
yarn add focus-visible
# or
npm install focus-visible


# at the root of your application
import "focus-visible/dist/focus-visible"
```

---

If adding this line does simplify the implementation in userland projects, I'd be happy to merge a PR:

>/packages/styled-system/src/pseudos.ts => line 51: _focusVisible: "&[data-focus-visible-added]",

moritzjacobs:
> If adding this line does simplify the implementation in userland projects, I'd be happy to merge a PR:
> 
> > /packages/styled-system/src/pseudos.ts => line 51: _focusVisible: "&[data-focus-visible-added]",

It would enable anyone to just install and use the polyfill without much overhead, that would be cool. What is the appropriate place for putting this in the docs? I'd be happy to write some documentation, if you want. @TimKolberger

Issue #3871, opened 2021-04-21
MenuButton as Button does not recieve type="button"


margalit:
# 🐛 Bug report

MenuButton as Button does not recieve type="button", even when passed in as a prop.

## 💥 Steps to reproduce

1. Create Menu with nested MenuButton
2. Add as={Button} to MenuButton
3. Inspect button

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/crazy-microservice-oizpx?file=/src/App.tsx

## 🧐 Expected behavior

MenuButtons rendered with Button receive type="button" or at least access can be set via  prop.

## 🧭 Possible Solution

Haven’t dug into it but we may be able to set the alternative case to true here: https://github.com/chakra-ui/chakra-ui/blob/117f2db20fbc7c261cdf9daa528a708eb191f6d6/packages/button/src/button.tsx#L124

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Chakra UI        |      1.5.2      |
| Browser          |     Firefox/Chrome       |
| Operating System |     OSX       |

## 📝 Additional information

This is breaking some of our forms as we render some Menus inside of them. These buttons trigger the submit function on the form when clicked.

huijiewei:
#3859

PR #3875, opened 2021-04-21
fix(button): #3859 fix button type as undefined

### LINKED ISSUE #3859: PopoverTrigger child Button type is undefined
# 🐛 Bug report

I use PopoverTrigger in form, but PopoverTrigger's child Button type is undefined
when click Trigger, form submit always trigger

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/hopeful-shamir-fcd4q?file=/src/App.tsx

## 🌍 System information

| Chakra UI        |  1.5.2          |
| Browser          |    Chrome 90        |
| Operating System |    macOS 11.2.3        |

antoniel:
Closes #3859 , #3871 

## 📝 Description

Fixes issue where button wasn't receiving type prop on first mount

Button does not receive type="button", even when passed in as a prop.

## 🚀 New behavior

If is a button, so doesn't have a as props should receive `type={type}` where default type is a button

No

### DIFF
--- a/packages/button/src/button.tsx
+++ b/packages/button/src/button.tsx
@@ -1,4 +1,3 @@
-import { mergeRefs } from "@chakra-ui/react-utils"
 import { Spinner } from "@chakra-ui/spinner"
 import {
   chakra,
@@ -118,17 +117,12 @@ export const Button = forwardRef<ButtonProps, "button">((props, ref) => {
     ...(!!group && { _focus }),
   }
 
-  const [isButton, setIsButton] = React.useState(!as)
-  const refCallback = React.useCallback((node: HTMLElement | null) => {
-    if (node?.tagName !== "BUTTON") {
-      setIsButton(false)
-    }
-  }, [])
+  const isButton = Boolean(!as)
 
   return (
     <chakra.button
       disabled={isDisabled || isLoading}
-      ref={mergeRefs(ref, refCallback)}
+      ref={ref}
       as={as}
       type={isButton ? type : undefined}
       data-active={dataAttr(isActive)}
--- a/packages/button/src/button.tsx
+++ b/packages/button/src/button.tsx
@@ -1,3 +1,4 @@
+import { mergeRefs } from "@chakra-ui/react-utils"
 import { Spinner } from "@chakra-ui/spinner"
 import {
   chakra,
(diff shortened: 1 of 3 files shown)


packages/button/src/button.tsx
```diff
@@ -118,17 +117,12 @@ export const Button = forwardRef<ButtonProps, "button">((props, ref) => {
     ...(!!group && { _focus }),
   }
 
-  const [isButton, setIsButton] = React.useState(!as)
-  const refCallback = React.useCallback((node: HTMLElement | null) => {
-    if (node?.tagName !== "BUTTON") {
-      setIsButton(false)
-    }
-  }, [])
+  const isButton = Boolean(!as)
```
TimKolberger, line 120: This solution is a bit too simple. The `as` prop could be a React Component which renders a button.
Checking the `node.tagName` is necessary.

The optional chaining operator `node?.tagName` returns `undefined | string`. If it is `undefined` we do not want to take any action, because we cannot be sure if it is not a button.

The previous logic with this change should to the trick:

```js
if (node && node.tagName !== "BUTTON") {
  setIsButton(false)
}
```

TimKolberger (changes requested):
Thanks for creating a PR for this @antoniel ✨

packages/button/src/button.tsx
antoniel: Can we use OR operator to check the case `as="button"`? On the pervious logic, if `as` exists `isButton = false` and there's no trigger to set it to true.

Something like: 
```tsx
const isButton = Boolean(!as) || Boolean(as === "button")
```

TimKolberger: Good catch, the initialState would be `false` if an `as` prop would be provided.

```js
const [isButton, setIsButton] = React.useState(!as || as === 'button')

// ...

if (node && node.tagName !== "BUTTON") { 
  setIsButton(false) 
}
```

What do you think?


I am getting curious: @segunadebayo do you have any insights why the `tagName` check was introduced?

antoniel: I've implemented and extend test cases, i'm also curious about tagName check seems unnecessary to me

aboveyunhai:
Sorry for @segunadebayo,  but the problem related to this one need to be address as quickly as possible. 
Since `type` is not correctly passed into button, it becomes very crucial inside a form. 
If your form has multiple buttons, any one of the buttons click will trig submit button hence submit the form.

packages/button/src/button.tsx
segunadebayo: I've updated the code to allow the user-defined `type` attribute to win. The reason we added a tagName check is to ensure we don't have `type="button"` to an element that isn't a button.

Technically, it is the work of the user to pass the correct type but since we're setting an internal default type of "button", it's up to us to ensure we don't add that to a non-button element.

Typical example: `<Button as="a">` should not have our internal `type=button` default attribute.

segunadebayo (approved):
With the recent update I made, this should be good to go.

TimKolberger (approved):

with-heart (approved):
I updated the changeset text slightly to be a little more clear for the changelog. 

Looks great though @antoniel! Thanks for taking the time to investigate this and work with us to resolve the bug.

with-heart:
@allcontributors please add @antoniel for code

PR #3881, opened 2021-04-21
fix(button): #3879 button type is preserved between render

### LINKED ISSUE #3879: Button type disappear when Button is re-rendered
# 🐛 Bug report

When you set the props "type" to a `<Button />` it's not preserved between re-render. It also happen with the default type="button"

## 💥 Steps to reproduce

1. Use a `<Button  />` inside a component that will re-render 
2. Inspect the button with the dev tools. See that the default `type="button"` is applied to the `<Button />`
3. Update the state in the parent component.
4. Inspect the button again. See that `type="button"` disappeared 

It's doing the same if you set a type explicitly 

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/chakra-ui-bug-button-ref-callback-79z4t

## 🧐 Expected behavior

The type should be preserved between render

## 🌍 System information

| Chakra UI        |     latest       |
| Browser          |      Chrome      |
| Operating System |     osx       |

## 📝 Additional information
CleanShot 2021-04-22 at 01 17 06

thierrylemoulec:
Closes #3879

## 📝 Description

Fix a problem where the type props for the `<Button />` was lost between renders

### DIFF
--- a/packages/button/src/button.tsx
+++ b/packages/button/src/button.tsx
@@ -120,7 +120,7 @@ export const Button = forwardRef<ButtonProps, "button">((props, ref) => {
 
   const [isButton, setIsButton] = React.useState(!as)
   const refCallback = React.useCallback((node: HTMLElement | null) => {
-    if (node?.tagName !== "BUTTON") {
+    if (node && node.tagName !== "BUTTON") {
       setIsButton(false)
     }
   }, [])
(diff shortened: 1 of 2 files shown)


thierrylemoulec:
Duplicate #3875

PR #3727, opened 2021-04-02
fix: lightMode flash issue

KristianRykkje:
Closes #  I couldn't find this issue.

## 📝 Description

> color-mode-script now sets the classname (light/dark) on the body.

## ⛳️ Current behavior (updates)

> When reloading the page after changed to dark mode it flashes light mode.
https://chakra-ui.com/docs/features/color-mode
image

## 🚀 New behavior

> Now the className (dark/light) should be set in the script before the "content".

## 💣 Is this a breaking change (Yes/No):
No

## 📝 Additional Information
This is my first PR to a open source project and would love some feedback on the way I made changes and maybe on some testing.

PR #3902, opened 2021-04-24
feat: migrate dark component styles to CSS using `_dark`  selector

jzxhuang:
Closes N/A (I can make an issue to track this if desired)

## 📝 Description

I've been tinkering around with Chakra's color mode and I saw some comments on migrating component styles to use CSS for dark theming rather than javaScript (`mode` and context). I took an initial stab at doing this to see what the changes would look like!

The idea is to change the component styles to rely on the `body` class `chakra-ui-dark` (selected via `_dark` pseudo prop). This will not only improve performance, but hopefully also help minimize the dark mode "flashing" that occurs. These initial PoC changes are based on:
* https://github.com/chakra-ui/chakra-ui/commit/ff7c36764650dc7f01957c417eae1ec8ce356495
* https://github.com/chakra-ui/chakra-ui/pull/3727#issuecomment-813310418

So far...
* I've implemented a quick PoC of what this would look like by migrating the styles in a single file (`Button`). I tried two different ways of nesting the `_dark` selectors, they have different impact on how a user would style a component using style props. I did my best to [illustrate the difference in a CodeSandbox](https://codesandbox.io/s/create-react-app-ts-forked-g1qq3?file=/src/App.tsx), would love for some feedback on what is preferred!
* I also adjusted `pseudoSelectors` to take in a prefix for the class name. I thought this might make it more customizable, for example if a user wanted to select `.myapp-dark` rather than `.chakra-ui-dark`. But if this isn't desired, can definitely revert this.
* I'd love some initial thoughts on the API and changes so far, if it looks ok I can maybe go ahead with migrating more of the styles!

Component themes have light/dark styling using JS (the `mode` function and React context from `ColorModeProvider`)

## 🚀 New behavior

Component themes use `_dark` pseudo prop (selects `body.chakra-dark`) for dark theme styling

No, but it would be a huge change as we would rewrite all the component theme files

## 📝 Additional Information

* Testing such a change would be pretty challenging, not sure if there's anything in place for theming — this might be a case where Storybook screenshots test could be super valuable? 

High-level todo list for this looks something like:

- [ ] Get feedback/verify API of `_dark` selector nesting [(relevant Sandbox)](https://codesandbox.io/s/create-react-app-ts-forked-g1qq3?file=/src/App.tsx)
- [ ] Get feedback/verify API of using `cssVarPrefix` in `pseudos` 
- [ ] Migrate all component styles
- [ ] Test test test everything
- [ ] Update docs

Support for `_dark` was added in https://github.com/chakra-ui/chakra-ui/commit/d189ae5020d5e72a252e0c59d681617c190bd0ab

### DIFF
--- a/packages/styled-system/src/pseudos.ts
+++ b/packages/styled-system/src/pseudos.ts
@@ -221,8 +221,26 @@ export const pseudoSelectors = {
    * this component or element.
    */
   _dark: ".dark &, [data-theme=dark] &",
+  /**
+   * Styles for when `.light` is applied to any parent of
+   * this component or element.
+   */
+  _light: ".light &, [data-theme=light] &",
 }
 
 export type Pseudos = typeof pseudoSelectors
+
+export const getPrefixedPseudoSelectors = (theme: any) => {
+  const prefix = theme?.config?.cssVarPrefix ?? ""
+  return {
+    ...pseudoSelectors,
+    _dark: `.${applyPrefix("dark", prefix)} &, [data-theme=dark] &`,
+    _light: `.${applyPrefix("light", prefix)} &, [data-theme=light] &`,
+  }
+}
+
+// TODO: extract this into an appropriate utils file
+const applyPrefix = (value: string, prefix = "", separator = "-") =>
+  `${[prefix, value].filter(Boolean).join(separator)}`
 
 export const pseudoPropNames = objectKeys(pseudoSelectors)
--- a/packages/button/stories/button.stories.tsx
+++ b/packages/button/stories/button.stories.tsx
@@ -35,6 +35,24 @@ export const basic = () => (
     <Button colorScheme="cyan">Button</Button>
     <Button colorScheme="orange">Button</Button>
     <Button colorScheme="yellow">Button</Button>
+
+    {/* The solid buttons were themed with `_dark` nested under other selectors */}
+    {/* 
+    If we pass the `_dark` prop, it overrides only `color` and `bg`, so we end up with the light theme values
+    Hover styling is kept because of the nested `_dark` styling in the `_hover` object
+    */}
+    <Button colorScheme="yellow" _dark={{ _hover: {} }}>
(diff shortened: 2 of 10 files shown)


packages/styled-system/src/pseudos.ts
```diff
@@ -221,8 +221,26 @@ export const pseudoSelectors = {
...
 export type Pseudos = typeof pseudoSelectors
 
+export const getPrefixedPseudoSelectors = (theme: any) => {
+  const prefix = theme?.config?.cssVarPrefix ?? ""
+  return {
+    ...pseudoSelectors,
+    _dark: `.${applyPrefix("dark", prefix)} &, [data-theme=dark] &`,
+    _light: `.${applyPrefix("light", prefix)} &, [data-theme=light] &`,
+  }
+}
```
jzxhuang, line 240: I thought it might be nice to let `_dark` target a custom name like `.my-app-dark`, but idk if this is really necessary.

For now I leveraged `cssVarPrefix` so it defaults  to `chakra-dark`. Previously it was `chakra-ui-dark` so this would actually be a breaking change. If we wanted to keep it backwards compatible we could just add a new property to `theme.config` like `colorModePrefix` instead of using `cssVarPrefix`.

TimKolberger:
Great work @jzxhuang! ✨ 

I really like this approach to handle dark mode!
Unfortunately rewriting the theme package would introduce a breaking change to all custom themes in the userland.

I think we should consider adding this API for the next chakra major version - in the future.
Thanks for providing this proof of concept!

TimKolberger:
Hi @jzxhuang, I am closing this PR as it won't make it into chakra v1 for know.

A similar selector is available in the styled-system, so you can use this feature in your own custom theme:

https://github.com/chakra-ui/chakra-ui/blob/main/packages/styled-system/src/pseudos.ts#L223

Thanks again for your time and opening this PR ✨

PR #3906, opened 2021-04-24
fix: export `usePopoverContext`

### LINKED ISSUE #3899: usePopoverContext is no longer exported
# 🐛 Bug report

I am attempting to upgrade from v1.2.1 to v1.5.2 and noticed that `usePopoverContext` is no longer exported, so I cannot use it to render a custom PopoverTriggerArrow.

My implementation looks something like:
```
export const PopoverTriggerArrow = forwardRef(function PopoverTriggerArrow(
  props: StyledComponentProps<>,
  ref
) {
  const { isOpen } = usePopoverContext();
  return (
    <Box as="span" ref={ref} {...props}>
      <Icon
        name={`faAngle${isOpen ? 'Up' : 'Down'}`}
        size="medium"
        color="inherit"
      />
    </Box>
  );
});
```

and this component is always a child of `PopoverTrigger` so it has access to the popover context.

I have attempted to solve this by instead using the `aria-expanded` prop that is passed down from `PopoverTrigger`, but the implementation quickly becomes ugly and prone to bugs.

Looks like this was caused by https://github.com/chakra-ui/chakra-ui/pull/3733/files#diff-90c4d042a18f8ae178e87611b561c7b86ddd6029b77d35b98b30f153dbe2a1f6L48

## 💥 Steps to reproduce

`import { usePopoverContext } from '@chakra-ui/react';` returns `undefined`

## 🧐 Expected behavior

`usePopoverContext` should be exported

## 🧭 Possible Solution

Revert this change so it is exported

joaorodrs:
Closes #3899

## 📝 Description

The fix was a simple line and could be done just exporting the `usePopoverContext` hook that was already imported in the modified file.

After [this changes](https://github.com/chakra-ui/chakra-ui/pull/3733/files#diff-90c4d042a18f8ae178e87611b561c7b86ddd6029b77d35b98b30f153dbe2a1f6L48), the `usePopoverContext` hook could not be imported from `@chakra-ui/react` anymore.

## 🚀 New behavior

Now should be possible to import the `usePopoverContext` hook from `@chakra-ui/react` normally.

No

## 📝 Additional Information

--

### DIFF
--- a/.changeset/forty-kids-exercise.md
+++ b/.changeset/forty-kids-exercise.md
@@ -1,5 +1,5 @@
 ---
-"@chakra-ui/popover": major
+"@chakra-ui/popover": minor
 ---
 
 Add an export for `usePopoverContext` hook
--- a/.changeset/forty-kids-exercise.md
+++ b/.changeset/forty-kids-exercise.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/popover": major
+---
+
+Add an export for `usePopoverContext` hook
--- a/packages/popover/src/popover.tsx
+++ b/packages/popover/src/popover.tsx
@@ -17,6 +17,8 @@ import { PopoverProvider, usePopoverContext } from "./popover-context"
 import { PopoverTransition, PopoverTransitionProps } from "./popover-transition"
 import { usePopover, UsePopoverProps } from "./use-popover"
 
+export { usePopoverContext }
+
 export interface PopoverProps extends UsePopoverProps, ThemingProps<"Popover"> {
   /**
    * The content of the popover. It is usually the `PopoverTrigger`,


.changeset/forty-kids-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@chakra-ui/popover": major
```
with-heart, line 2: This should be `minor` since it's only adding backwards compatible functionality.

with-heart (changes requested):
I don't mind adding this export back, but I did want to note that these hooks are something we've been discussing and wanted to make you aware of. 

We're not entirely sure we want these hooks to be considered part of our public api. They're mostly built to handle what we need them to handle internally for our components, and imo we haven't put nearly enough thought into imagining potential end-user use cases nor have we made the effort to document them.

Aside from the documented hooks like `useRadioGroup` and `useCheckboxGroup`, I think I'd personally prefer to leave them in a "use these at your own risk" status where we'll export them and you're welcome to use them, but if we make a breaking change to them, it's on the end user to figure out how to work around it. So sort of a semi-public api, where they're exported, but aren't yet fully supported.

Doesn't really relate directly to this PR, but just something to be aware of!

joaorodrs:
@with-heart Perfect! I actually realized that this hook isn't a "documentated" hook, but I wasn't really aware of this discussions. Thanks for the explanation

with-heart (approved):

PR #3910, opened 2021-04-24
docs(icons): copy icon to clipboard feature

### LINKED ISSUE #3889: docs: copy icons to clipboard
# 🚀 Feature request

Implement copy to clipboard behavior on [/docs/media-and-icons/icon#all-icons](https://chakra-ui.com/docs/media-and-icons/icon#all-icons) docs section 

It's an improvement in user experience. The component that displays each icon looks like a button, so an interaction to copy to the clipboard would be nice.
As a user, when I'm searching by a chakra icon on [/docs/media-and-icons/icon#all-icons](https://chakra-ui.com/docs/media-and-icons/icon#all-icons) I should be able to:
- Copy to clipboard the icon's import code
- Copy to clipboard the icon name

I'm not sure about what is the best approach.

## 📝 Additional Information

A reference for this feature is [systemuicons](https://systemuicons.com/) where user is able to copy all SVG code by one click.

joaorodrs:
Closes #3889 

## 📝 Description

The pull request contains code that is part of the website docs, more specifically from the icons list, at the [icons page](https://chakra-ui.com/docs/media-and-icons/icon#all-icons)

Nothing happens when the user clicks some icon

## 🚀 New behavior

When clicking on any icon at [icons page](https://chakra-ui.com/docs/media-and-icons/icon#all-icons), the icon name is copied to the clipboard

No

### DIFF
--- a/.changeset/honest-onions-look.md
+++ b/.changeset/honest-onions-look.md
@@ -2,4 +2,4 @@
 "@chakra-ui/docs": minor
 ---
 
-Add copy icon name to clipboard when clicking
+Clicking an icon in the icons list now copies the icon name to clipboard.
--- a/.changeset/honest-onions-look.md
+++ b/.changeset/honest-onions-look.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/docs": minor
+---
+
+Add copy icon name to clipboard when clicking
--- a/website/src/components/icons-list.tsx
+++ b/website/src/components/icons-list.tsx
@@ -64,6 +64,8 @@ const iconList = {
 }
 
 const IconsList = () => {
+  const toast = useToast()
+
   return (
     <Grid
       mt={7}
@@ -73,7 +75,6 @@ const IconsList = () => {
       {Object.keys(iconList).map((key, i) => {
         const Icon = iconList[key]
         const { onCopy } = useClipboard(`<${key} />`)
-        const toast = useToast()
 
         const onCopyIcon = () => {
           onCopy()
--- a/website/src/components/icons-list.tsx
+++ b/website/src/components/icons-list.tsx
@@ -1,5 +1,5 @@
 import React from "react"
-import { Grid, Flex, Text, useClipboard, useToast } from "@chakra-ui/react"
+import { Grid, Button, Text, useClipboard, useToast } from "@chakra-ui/react"
 import * as icons from "@chakra-ui/icons"
 
 const iconList = {
@@ -88,24 +88,27 @@ const IconsList = () => {
         }
 
         return (
-          <Flex
-            p={3}
+          <Button
+            pt={9}
+            pb={9}
             onClick={() => onCopyIcon()}
             key={i}
             align="center"
-            cursor="pointer"
-            transition="all 0.2s ease"
-            _hover={{ opacity: 0.7 }}
-            borderRadius="md"
-            borderWidth="1px"
+            variant="outline"
             flexDir="column"
             justify="center"
           >
             <Icon />
-            <Text mt={2} fontSize="sm" textAlign="center">
+            <Text
+              as="span"
+              mt={3}
+              fontSize="sm"
+              fontWeight="normal"
+              textAlign="center"
+            >
               {key}
             </Text>
-          </Flex>
+          </Button>
         )
       })}
     </Grid>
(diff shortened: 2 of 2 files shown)


website/src/components/icons-list.tsx
```diff
@@ -72,11 +74,28 @@ const IconsList = () => {
...
+
         return (
           <Flex
             p={3}
+            onClick={() => onCopyIcon()}
             key={i}
             align="center"
+            cursor="pointer"
+            transition="all 0.2s ease"
+            _hover={{ opacity: 0.7 }}
```
joaorodrs, line 98: Maybe I should remove this hover effect, unsure

website/src/components/icons-list.tsx
```diff
@@ -72,11 +74,28 @@ const IconsList = () => {
     >
       {Object.keys(iconList).map((key, i) => {
         const Icon = iconList[key]
+        const { onCopy } = useClipboard(key)
```
xiaohuoni, line 77: I think the correct text should be `<PhoneIcon />`,

joaorodrs: It was my intention but after duscussing (https://github.com/chakra-ui/chakra-ui/issues/3889#issuecomment-826138438) I think it's better to copy only the name, because the name you can copy directly in the `import {...` and on the tags

website/src/components/icons-list.tsx
TimKolberger: The current implementation is not very accessible. Let me point out a few things and how to fix them:

The `Flex` component renders a `<div>`.
A div has no meaning in terms of accessibility, but we want to make this one interactive.

There are two ways I know of:

1. Adding the correct aria-labels like `role="button"` and implementing keyboad navigation to handle focus `tabIndex={0}` and interaction (Did you notice that buttons fire their onClick handler when pressing "enter"?)
2. Using the html element `button` and do not use `p` tags inside of it. This lets the browser do all the magic mentioned above.

The second one sound quite simpler, doesn't it 😄 

```jsx
<Button // button
  onClick={() => console.log("clicked")}
  variant="outline"
  p="3"
  height="auto"
  align="center"
  flexDir="column"
  justify="center"
>
  <CopyIcon />
  <Text 
    as="span" // <-- no `<p>` tag
    mt={2} 
    fontSize="sm" 
    textAlign="center">
    CopyIcon
  </Text>
</Button>
```

TimKolberger (changes requested):
Thank you for implementing this feature ✨ 
I left some hints how to improve the accessibility of this feature.

anubra266:
I think the hardcoding here not a good approach,
Instead of 
```js
  {Object.keys(iconList).map((key, i) => {
        const Icon = iconList[key]
        const { onCopy } = useClipboard(key)

        return (
          ...
        )
      })}
    </Grid>
  )
}
```

You could do

```js
  {Object.entries(icons).map(([key, Icon], i) => {
        const { onCopy } = useClipboard(key)

        return (
          ...
        )
      })}
    </Grid>
  )
}
```

There will be no need for `iconList`
 And the whole code is shortened to

```js
import React from "react"
import { Grid, Button, Text, useClipboard, useToast } from "@chakra-ui/react"
import * as icons from "@chakra-ui/icons"

const IconsList = () => {
  const toast = useToast()

  return (
    <Grid
      mt={7}
      gap={5}
      templateColumns="repeat( auto-fit, minmax(150px, 1fr) )"
    >
      {Object.entries(icons).map(([key, Icon], i) => {
        const { onCopy } = useClipboard(key)

        const onCopyIcon = () => {
          onCopy()

          toast({
            title: `'${key}' copied to clipboard`,
            status: "success",
            duration: 2000,
            isClosable: false,
          })
        }

        return (
          <Button
            pt={9}
            pb={9}
            onClick={() => onCopyIcon()}
            key={i}
            align="center"
            variant="outline"
            flexDir="column"
            justify="center"
          >
            <Icon />
            <Text
              as="span"
              mt={3}
              fontSize="sm"
              fontWeight="normal"
              textAlign="center"
            >
              {key}
            </Text>
          </Button>
        )
      })}
    </Grid>
  )
}

export default IconsList

```

.changeset/honest-onions-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"@chakra-ui/docs": minor
+---
+
+Add copy icon name to clipboard when clicking
```
with-heart, line 5:
```suggestion
Clicking an icon in the icons list now copies the icon name to clipboard.
```

TimKolberger (approved):
🎉

PR #3956, opened 2021-05-04
chore: add preconstruct cli

with-heart:
## 📝 Description

This PR converts the repo to use [`preconstruct`](https://preconstruct.tools/) for the vast majority of our build needs. 

`preconstruct` greatly increases our build times (it was <3 minutes locally, and I think we can do even better than that eventually) and removes the need to use `yarn build` whatsoever on local as it links packages directly to their `src` folders. This is a pretty significant boost to our pipeline as a whole, so I'm pretty excited to be able to convert us over.

No probably but need to make sure. The `dist` folder output has a much different structure, but that shouldn't matter since it's resolved by fields in `package.json` files.

## 📝 Remaining Tasks

- [ ] Update `CONTRIBUTING.md` and remove old pre-dev requirements
- [ ] Remove `lerna` or at least get rid of the old `build`-related stuff we won't be using anymore
- [ ] Remove the `build`-related commands for individual packages that are handled by `preconstruct`
- [ ] Make sure pipelines are all passing
- [ ] Add changesets for changed packages to make sure new versions are published


package.json
```diff
@@ -16,6 +16,12 @@
       "**/@types/theme-ui__components"
     ]
   },
+  "preconstruct": {
+    "packages": [
+      "packages/*",
+      "tooling/cli"
```
with-heart, line 22: Since `cli` is the only package in this folder that's actually built, I didn't add a wildcare for the `tooling` directory. I'm wondering if `cli` should be moved into `packages` since it's an official part of the Chakra API.

package.json
```diff
@@ -28,7 +34,7 @@
     "fix:pkgs": "manypkg fix",
     "version:pkgs": "changeset version",
     "clean:emotion-types": "rm -rf node_modules/@emotion/core/types",
-    "postinstall": "yarn clean:emotion-types && yarn check:pkgs",
+    "postinstall": "yarn clean:emotion-types && yarn check:pkgs && preconstruct dev",
```
with-heart, line 37: `preconstruct dev` is the piece that connects everything together for local dev. For each package, it creates `dist` output that imports from the package's `src` entrypoint. No need to build anything locally! 

And since `postinstall` is connected to `yarn install`, it will automatically do this when the user installs packages. It constantly keeps the references up to date, and also will automatically be run when a user clones the repo and installs for the first time.

Big improvement in the contribution process imo

package.json
```diff
@@ -37,8 +43,8 @@
     "boot": "yarn prestart && yarn build",
     "start": "yarn prebuild && yarn storybook",
     "start:all": "yarn build && concurrently \"yarn docs dev\"  \"yarn storybook\" ",
-    "prebuild": "yarn pkg babel-plugin build",
-    "build": "lerna run build --ignore @chakra-ui/test-utils --no-private --stream",
+    "prebuild": "preconstruct dev && yarn pkg babel-plugin build && yarn pkg props-docs build",
```
with-heart, line 46: I added `yarn pkg props-docs build` here because we need to have that output built as part of the `build` command. I added `preconstruct dev` so that `props-docs` can reliably access the `propNames` export from `@chakra-ui/styled-system` when it runs.

packages/accordion/package.json
```diff
@@ -18,14 +18,14 @@
...
+  "types": "dist/chakra-ui-accordion.cjs.d.ts",
+  "typings": "dist/chakra-ui-accordion.cjs.d.ts",
   "exports": {
     ".": {
-      "require": "./dist/cjs/index.js",
-      "default": "./dist/esm/index.js"
+      "require": "./dist/chakra-ui-accordion.cjs.js",
+      "default": "./dist/chakra-ui-accordion.esm.js"
     }
   },
```
with-heart, line 30: You'll see a similar change for each package. Since `preconstruct` build things a little differently, I had to update the references here. `preconstruct` automatically updates the `main` and `module` fields, but doesn't handle the rest of the fields (yet) so I had to build a script to automatically update the rest of them. This shouldn't ever need to be run  again, but I added it in `scripts/fix-types-and-typings.ts` for the time being just in case.

tooling/cli/src/utils/init-cli.ts
```diff
@@ -2,7 +2,7 @@ import checkNode from "cli-check-node"
 import welcome from "cli-welcome"
 import updateNotifier from "update-notifier"
 import unhandledError from "cli-handle-unhandled"
-import * as pkgJSON from "../../package.json"
```
with-heart, line 5: I had to update this because TypeScript was complaining that the `description` and `version` fields weren't exported. Since `package.json` is a single object, there's not anything to collect with `*` anyway.

with-heart:
omg this also means we can rearrange the pipeline steps order, so we can lint+test first before spending a bunch of time needlessly building if it's just going to fail one of those steps anyway.

babel.config.js
```diff
@@ -21,7 +25,7 @@ module.exports = function (api) {
   ]
 
   const plugins = [
-    "@chakra-ui/babel-plugin",
+    transformPureCalls,
```
with-heart, line 28: `@chakra-ui/babel-plugin` was causing issues with build and local dev. 

If I added it as a `preconstruct` package, build would fail because it wasn't necessarily built before the other packages since it isn't technically a dependency of anything else.

If I didn't add it, we'd need to guarantee that it was built before `preconstruct dev` was run, which meant adding it to `postinstall`, which meant a slightly annoying little delay anytime you installed modules.

So rather than add it as a dependency to a bunch of things or add it to `postinstall`... I deleted the whole package and inlined it. 😲

packages/utils/src/walk-object.ts
```diff
@@ -1,4 +1,4 @@
-import { isArray, isObject } from "@chakra-ui/utils"
+import { isArray, isObject } from "./assertion"
```
with-heart, line 1: `preconstruct` doesn't like packages importing from themselves without having themselves as dependencies 😆 

That was an unnecessary import anyway, so fixed it here.

package.json
Andarist: Maybe worth moving the `preconstruct dev` before `yarn check:pkgs`? If the check fails you will at least have those preconstruct links already working.

packages/utils/src/walk-object.ts
Andarist: Could you report this as a bug to preconstruct? I think this works now in node by default, according to the resolution algorithm: https://nodejs.org/api/modules.html#modules_all_together , It looks like `LOAD_PACKAGE_SELF` is allowing that explicitly.

I recall having some problems with that in the past and the previous version of the algorithm did not have this explicit step. However, I can't see why `LOAD_NODE_MODULES` would not allow this in the old algorithm so maybe this has always worked?

Either way - I think it makes sense to support this in preconstruct so we should fix this.

with-heart:
Working on fixing the current pipeline failures over here: https://github.com/preconstruct/preconstruct/pull/393#issuecomment-832546671

segunadebayo:
Thanks for working on this @with-heart. I used most of the ideas here to create a working version in the `build/preconstruct` branch.

Tim mentioned that this might be a potential breaking change for those importing modules from `@chakra-ui/react/dist/**` so we can merge that branch in when doing a v2 release.

Please check it out and update as needed.

Appreciate the work here.

PR #3959, opened 2021-05-05
chore(prerelease): configure changesets

TimKolberger:
## 📝 Description

This PR configures changesets to create pre releases.
The branch `next` behaves like the main branch, due the configuration in `.github/workflows/release.yml`.

1. Every push to the `next` branch triggers changesets
2. It creates/updates the "Version Packages" PR targeting `next`
3. Merging the PR releases a new pre release

**We need to keep the next branch manually up to date to the changes in `main`**


segunadebayo:
Great work @TimKolberger. I think we need an equivalent Github action for this as well.

When we push to `next`, it should do a pre-release.

TimKolberger:
Thanks @segunadebayo! I think this should already be configured with this line:
https://github.com/chakra-ui/chakra-ui/blob/main/.github/workflows/release.yml#L11

with-heart:
Wow, this took way less work than I was expecting! Nice dude!

PR #3961, opened 2021-05-05
fix: docs theming component style missing field

greatgumz:
Closes # 

## 📝 Description

`fields` is missing from baseStyle object property which is needed to override base styling per component

baseStyle object property does not work as designed

## 🚀 New behavior

cleans up required object property

No

### DIFF
--- a/website/pages/docs/theming/component-style.mdx
+++ b/website/pages/docs/theming/component-style.mdx
@@ -66,7 +66,9 @@ The basic API for styling a single part component is:
 ```jsx live=false
 export default {
   // Styles for the base style
-  baseStyle: {},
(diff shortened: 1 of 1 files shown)


cschroeter:
Hey buddy, thanks for your PR.

But I'm afraid the changes your propose here are invalid. `field` is not a default property in every component.
In the `<Input />` component you will find a field property you can style, for example:

```tsx 
export const Input: ComponentStyleConfig = {
  baseStyle: {
    field: {
      _placeholder: {
        opacity: 1,
        color: 'gray.500',
      },
    },
  },
  // ...
  ```
  
  But most of the other components do not have that.

PR #3976, opened 2021-05-09
chore: speed up test runs

with-heart:
## 📝 Description

This PR greatly improves `test` speed. 

- dropped `lerna run` in favor of a single `jest` process
- changed `test` script to `jest --maxWorkers=50%` which makes local runs even faster
- created `test:ci` script which runs `jest --runInBand`

Local tests take multiple minutes to run. CI tests takes 6-8 minutes to run.

## 🚀 New behavior

Local tests take less than a minute to run. CI tests take 1-3 minutes to run.

No

### DIFF
--- a/packages/styled-system/tests/interpolation.test.ts
+++ b/packages/styled-system/tests/interpolation.test.ts
@@ -9,7 +9,7 @@ test("should handle array interpolations", () => {
     xl: "70em",
   })
 
-  // @ts-expect-error "&" is technically disallowed
+  // @ts-ignore
   const result = css({ "&": [{ bg: "red" }, { bg: "green" }] })(
     toCSSVar({
       breakpoints: customBreakpoints,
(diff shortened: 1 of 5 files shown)


TimKolberger (approved):
Awesome 🎉  this is a real time saver!

packages/styled-system/tests/interpolation.test.ts
```diff
@@ -9,7 +9,7 @@ test("should handle array interpolations", () => {
     xl: "70em",
   })
 
-  // @ts-expect-error "&" is technically disallowed
```
with-heart, line 12: For some reason VSC shows this as a type error if I remove the line, but `tsc --noEmit` reports it as an unused line if I leave it in place. 

Same thing is happening in #3978 and I have no idea why, so just added `@ts-ignore` to turn it off completely.

segunadebayo (approved):
Amazing! Thank you for taking the time to improve this @with-heart

PR #3978, opened 2021-05-09
chore: reconfigure lint+tsc for speed improvements

with-heart:
## 📝 Description

This PR accomplishes a few things:

- Replaces `lint:types` with `type-check`
- `type-check` executes a single `tsc` process and checks all files in the project
	- Greatly increases speed of type-checking (now only ~40s locally)
	- Expands coverage to include all TypeScript files in the project (including `website`) rather than only `src` directories
- `lint` executes a single `eslint` process and checks all files in the project
	- Greatly increases speed of linting (now only ~45s locally)
	- Expands coverage to include all files in the project (including `website`) rather than only `src` directories

Overall, this gives us much greater coverage on type-checking and linting, and should greatly increase the speed of both in local dev env and in the pipelines.

No


with-heart:
Before:
image

After:
image

👀

paambaati:
@with-heart If you're trying to optimize TypeScript type-checks, might I recommend you also add the [`incremental`](https://www.typescriptlang.org/tsconfig/#incremental) option? This is useful in local runs; `tsc` creates a local cache file for the first run, and reuses it on consecutive runs. This makes successive `tsc` runs quite fast.

with-heart:
Currently trying to figure out why the deployment build is failing with a type error that type-check and local build don't notice. 

Reached out to Twitter for help: https://twitter.com/grow_love/status/1394636390349119488 🤞

mattwells19:
> Currently trying to figure out why the deployment build is failing with a type error that type-check and local build don't notice.
> 
> Reached out to Twitter for help: https://twitter.com/grow_love/status/1394636390349119488

Next JS is saying that `results[active]` is of type:
image

According to this type, `results[active]` is an array, so you would need to index it before accessing the `url` property. It doesn't look like you changed that implementation. Has `match-sorter` changed it's implementation at all?

segunadebayo:
Hey Mark @with-heart, I pulled changes from here into the `build/preconstruct` branch.

Please take a look

PR #4026, opened 2021-05-18
fix(popper): perf regression for lazy init

dodas:
## 📝 Description

Fixes regression of https://github.com/chakra-ui/chakra-ui/pull/3343

Add `isOpen` option to `usePopper` hook. 
The `popper.js` instance will not be created until this option is true.
`Menu`, `Popover` and `Tooltip` components now use this option,
so the `popper.js` instance is created only once the popper is open.
This should significantly improve render and scroll performance.

`popper.js` instance is created for every `Menu`/`Popper`/`Tooltip` instance, regardless of whether it is open.

## 🚀 New behavior

`popper.js` instance is only created once the popper is actually needed for positioning.

No

### DIFF
--- a/packages/popper/src/use-popper.ts
+++ b/packages/popper/src/use-popper.ts
@@ -13,7 +13,7 @@ import { cssVars, getEventListenerOptions } from "./utils"
 export type { Placement }
 
 export interface UsePopperProps {
-  isOpen?: boolean
+  enabled?: boolean
   /**
    * The main and cross-axis offset to displace popper element
    * from its reference element.
@@ -99,7 +99,7 @@ export type ArrowCSSVarProps = {
 
 export function usePopper(props: UsePopperProps = {}) {
   const {
-    isOpen = true,
+    enabled = true,
     modifiers = [],
     placement: placementProp = "bottom",
     strategy = "absolute",
@@ -120,7 +120,7 @@ export function usePopper(props: UsePopperProps = {}) {
   const cleanup = useRef(() => {})
 
   const setupPopper = useCallback(() => {
-    if (!isOpen || !reference.current || !popper.current) return
+    if (!enabled || !reference.current || !popper.current) return
 
     // If popper instance exists, destroy it so we can create a new one
     cleanup.current?.()
@@ -167,7 +167,7 @@ export function usePopper(props: UsePopperProps = {}) {
 
     cleanup.current = instance.current.destroy
   }, [
-    isOpen,
+    enabled,
     placementProp,
     modifiers,
     matchWidth,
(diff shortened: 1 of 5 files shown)


packages/popper/src/use-popper.ts
```diff
@@ -13,6 +13,7 @@ import { cssVars, getEventListenerOptions } from "./utils"
 export type { Placement }
 
 export interface UsePopperProps {
+  isEnabled?: boolean
```
cschroeter, line 16: Why not `isOpen?: boolean` ?
In all the consumers of `usePopper` you map between `isOpen` to `isEnabled` - so isn't  `isOpen` a natural fit?

packages/popper/src/use-popper.ts
```diff
@@ -98,6 +99,7 @@ export type ArrowCSSVarProps = {
 
 export function usePopper(props: UsePopperProps = {}) {
   const {
+    isEnabled = true,
```
cschroeter, line 102: What is the benefit in terms of performance when setting this by default to true?

A simple tooltip like this
```tsx 
<Tooltip label="Hey, I'm here!" aria-label="A tooltip">
  Hover me
</Tooltip>
```

will still create a popper instance, or?

cschroeter (changes requested):

packages/popper/src/use-popper.ts
dodas: Hi Christian!

I was deciding between the two. The `popper` could be used for positioning things that are not open/close ([popper.js homepage examples](https://popper.js.org/)), so I thought `isEnabled` is more accurate. I don't mind renaming it if you prefer.

packages/popper/src/use-popper.ts
dodas: I am passing `isEnabled: isOpen` to `usePopper` in the `useTooltip` hook, so the `popper.js` instance will not be created until the tooltip is actually visible.

dodas:
I would like to clarify that in the app I am working on we are currently patching chakra-ui to include these changes, as it significantly improves render and scroll performance on pages with lot of components that use `usePopper` hook.

packages/popper/src/use-popper.ts
cschroeter: Ah it is not the `isOpenProp` lol my bad - i was like wtf how should this work :D

packages/popper/src/use-popper.ts
cschroeter: in my opinion `isOpen` is actually more accurate than `isEnabled` 

an example - you could add `usePopper` within a component so at that point `usePopper` is kind of enabled - but not necessarly open.

If you have no objections I would go with `isOpen` and would merge it afterwards

segunadebayo:
I thought `enabled` was a better name and is consistent with the Popperjs API and how we name props in the hooks package.

PR #4038, opened 2021-05-19
fix(media-query): fix an issue with useBreakpointValue

### LINKED ISSUE #3693: Support `useBreakpoint` wihin IFrames
```tsx
// use-breakpoint.ts  line: 65
mediaQueries.forEach(({ query, ...breakpoint }) => {
  // Here: allow to provide users own implementation of matchMedia
  const mediaQuery = window.matchMedia(query)
   // etc. ..
})  
``` 

Material-UI has support for custom matchMedia implemented in https://github.com/mui-org/material-ui/pull/18741/files

cschroeter:
If one had used useBreakpointValue within an IFrame,
the root window element was used to determine the breakpoint.
Now the window from the current context is used.

Closes #3693

### DIFF
--- a/.changeset/new-seas-attack.md
+++ b/.changeset/new-seas-attack.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/media-query": patch
+---
+
+fix(media-query): fix an issue with useBreakpointValue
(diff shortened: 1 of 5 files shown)


segunadebayo (changes requested):
Looks great!

Please add the `react-env` package to the `package.json` of the media-query package.

Thanks

PR #4052, opened 2021-05-20
fix: Fix type export in styled-system package

helious:
## 📝 Description

> This fixes the esm build for the styled-system package; currently the esm build is trying to export `WithCSSVar` from an empty `types.js` file.

> When including `@chakra-ui/styled-system` `rollup` is not able to build as the `WithCSSVar` `type` is trying to be re-exported in the esm build.

## 🚀 New behavior

> When including `@chakra-ui/styled-system` `rollup` is able to build.


cschroeter:
Thanks for the PR

PR #4069, opened 2021-05-22
fix: (proposition) fixing the Collapse transition unmountOnExit issue

### LINKED ISSUE #4039: Collapse Transition not working with `unMountOnExit`
# 🐛 Bug report

The collapse transition isn't working correctly on version 1.6.2

## 💥 Steps to reproduce

1. Go to the below sandbox
2. Collapse animation that displays a chart is working (version 1.6.0)
3. Switch to 1.6.2
4. Collapse animation that displays a chart is not displaying the chart animation anymore

## 💻 Link to reproduction

https://codesandbox.io/s/chakra-collapse-wxi75?file=/src/App.js

## 🧐 Expected behavior

The collapse functionality should work the same as version 1.6.0

lazarnikolov94:
Closes #4039

## 📝 Description
As @tpck reported in #4039, the Collapse transition with `unmountOnExit` doesn't work in the provided bar chart example. Upon further investigation I found out that [this line](https://github.com/chakra-ui/chakra-ui/blob/main/packages/transition/src/collapse.tsx#L128) might be causing the issue. If we don't override the `display` property the animation executes smoothly. The rest of the examples in the Storybook are also working correctly even when this line is removed. @segunadebayo why do we set the `display` property to `block` or `none` depending on the `startingHeight`? Is there a specific test case for this that I need to test?

The enter animation doesn't happen.

## 🚀 New behavior
The enter animation runs smoothly.

Not sure.


lazarnikolov94:
Here's a more complex example with the CodeSandbox build from this PR: https://codesandbox.io/s/chakra-collapse-forked-pbed6?file=/src/App.js

segunadebayo:
@lazarnikolov94 If the collapse has a child that is focusable, say it contains a button, link, etc. When the collapse is closed, you should be able `Tab` to inside its content. That's the reason we use the `display` attribute.

packages/transition/package.json
```diff
@@ -59,7 +59,9 @@
     "react": ">=16.8.6"
   },
   "devDependencies": {
+    "chart.js": "^3.2.1",
```
segunadebayo, line 62: Please add this to the root package.json not the dev-dependency here.

PR #4075, opened 2021-05-23
feature/wizard

### LINKED ISSUE #2836: Add Steps component
# 🚀 Feature request

> Please describe your request in one or two sentences.

It would be useful to have some form of Steps (Stepper) component to show navigation with a certain amount of steps.

> Please provide valid reason(s) why this should added to Chakra UI.
>
> If this feaure is related to a problem you've noticed. Mention it as well

I think this is quite common use case and should be part of the library.

## ✅ Proposed solution or API

> Please provide code snippets, gists, or links to the ideal design or API

I really like the Ant Design version of [Steps component](https://ant.design/components/steps/). It would be really nice to have at least some light version of it.

jeanverster:
Closes #2836

## 📝 Description

I have had this branch open for a long time now and never really had much time to work on it, I finally managed to get around to completing it! So I figured I would open a PR and get some feedback.

This PR implements a Wizard / Stepper component (used the Wizard naming convention so as to not clash with other components which have `Stepper` in the name)

I've taken cues mainly from the Ant Design `Steps` component, so the functionality is similar to that. I also really liked the ability to render children inside the `Step` component from the material UI Stepper, so I included that too.

In terms of a11y, I wasn't quite sure what is required in terms of stepper / wizard components, did some googling but didn't really manage to find much. So I have added some `aria` attributes to the disabled / active steps, but any further feedback here would be appreciated.

This would be my first contribution to Chakra, so some implementation details might need tweaking, but would be happy to action.

May-23-2021 12-56-38

Currently, there is no Stepper / Wizard component which ships with Chakra UI

## 🚀 New behavior

This PR would add functionality for a Stepper component out the box

No

## 📝 Additional Information

Apologies about the linting noise, I'm not sure why my editor decided to lint some files I didn't even work on. Also, I have `organizeImports` set to true in my editor, so it would have caused some of the imports to be shuffled around, if this is an issue I'd be happy to revert those changes.


cschroeter:
Hi @jeanverster,

thanks for submitting this PR. Before we can look at the PR in detail, I would like to ask you to undo all style changes like formatting etc. Especially with a big PR like this, it's so incredibly hard to focus on the essentials.

jeanverster:
Hi @cschroeter - no problem, I will revert formatting changes and update 👌🏼  Thanks!

cschroeter:
Thanks for updating the PR. Having 21 files instead of 35 is a nice thing to have :P

Before we go into details lets talk about the component API design first. 
Couple of weeks ago I've created a Stepper for Chakra UI Pro (https://pro.chakra-ui.com/components/application-ui/steps/vertical-steps) - so I'm happy to see that someone is willing to bring this to Chakra UI Core


The API design in the Pro component is a bit different than yours:


```tsx
const { nextStep, prevStep, reset, activeStep } = useSteps({ initialStep: 0 })

return (
  <Steps activeStep={activeStep}>
    <Step title="Select campaign settings">
     <StepContent>
       <Button  onClick={prevStep}>Back</Button>
       <Button onClick={nextStep}>Next</Button>
     </StepContent>
    </Step>
  </Steps>
)
```

Regarding the public API, I have added a few comments directly to the source code. 

In general I've some mixed feelings about the component name wizard. I could not find any major UI framework that uses the name `<Wizard />`so in terms of consistency and expectation, we should use the same or a very similar name. Thoughts on that?

packages/wizard/src/wizard.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ThemingProps,
+  useMultiStyleConfig,
+} from "@chakra-ui/system"
+import { cx } from "@chakra-ui/utils"
+import * as React from "react"
+
+export interface WizardProps extends HTMLChakraProps<"div">, ThemingProps {
+  activeStep: number
+  orientation?: "vertical" | "horizontal"
+  labelOrientation?: "vertical" | "horizontal"
```
cschroeter, line 16: Is there a situation where the overall orientation is different than the label orientation? I could not find an example in the storybook so I would drop support for label orientation in this PR to keep things simple

packages/wizard/src/wizard.tsx
```diff
@@ -0,0 +1,83 @@
...
+  isLoading?: boolean
```
cschroeter, line 17: `isLoading` and `isError` is something to avoid in component API design in general for [more insights](https://kentcdodds.com/blog/stop-using-isloading-booleans).

TL;DR `isLoading` and `isError` are mutual exclusive (can not occur at the same time) so the state of a step should be reflected as an enum instead.


But long story short I would not support this anyway (at start).

I would suggest to start things off with `activeStep: number` and `orientaiton: "vertical" | "horizontal"`

packages/wizard/src/wizard-step.tsx
```diff
@@ -0,0 +1,206 @@
...
+import { darken, lighten, mode } from "@chakra-ui/theme-tools"
+import { Collapse } from "@chakra-ui/transition"
+import { AnimatePresence, motion } from "framer-motion"
+import * as React from "react"
+import { WizardConnector } from "./wizard-connector"
+
+const AnimatedCheck = motion(CheckIcon)
+const AnimatedCloseIcon = motion(CloseIcon)
+const AnimatedSpan = motion(chakra.span)
+export interface WizardStepProps extends HTMLChakraProps<"div"> {
```
cschroeter, line 20: I thing here is the biggest difference from my initial idea about a Stepper component. In your proposal your are only allowed to have a text (description) as content for your Step ( WizardStep). This is in my opinion not flexible enough.

cschroeter (changes requested):

jeanverster:
Hey @cschroeter - thanks for the feedback here! I was initially thinking about including some kinda hook to encapsulate all the logic required to use the component, I am happy to include this if it will make things simpler for the users. I definitely agree about the name "Wizard" 😅  I initially wanted to called it "Stepper" but decided on Wizard based on some feedback from an issue I opened ages ago.. I am happy to rename it to "Steps" to keep in line with other UI frameworks 👍🏼  Think it would make more sense.

packages/wizard/src/wizard-step.tsx
jeanverster: So the `description` prop would only be for a "sub title" to the actual step title. How I have it currently is that the user can pass child components to the `Step` component which would render as step content (only in vertical orientation, though). I would be happy to make it more explicit like how you have it in the pro version api and export a `StepContent` component which would handle the collapsing of content when the user switches between steps

packages/wizard/src/wizard.tsx
jeanverster: Cool, sounds good I will remove these states for now 👍🏼

segunadebayo:
Great work @jeanverster. From the look of this PR, it seems like a good fit for a recipe instead.

Feel free to create this in your own repo, even publish it. Then, we'll add a link to it in the docs so others can find it as well.

jeanverster:
Thanks @segunadebayo 🙏🏼  I would be happy to do the work to get this ready to include in the chakra core offering, even simplify it a whole lot and make it just a basic horizontal / vertical stepper with no bells and whistles 🙂  I think it would be a good addition! But am happy either way. Also just to clarify, if we do decide to go the recipe route, then would that essentially just go under the "guides" section?

segunadebayo:
Yes, it'll go into the guide or resources section. For now, I'll prefer you create a CodeSandbox or publish it in your NPM.

Happy to consider this in the future.

PR #4102, opened 2021-05-27
fix(tabs): render tab indicator at the correct pos

### LINKED ISSUE #2040: TabIndicator is not calculated correctly when used wihtin a Drawer

### LINKED ISSUE #2440: the `borderImage` prop isn't supported
Chakra doesn't seem to recognize the `borderImage` prop for CSS.

```jsx
<Box
  width="100px"
  height="100px"
  border="1px"
  borderImage="linear-gradient(to bottom, red, blue) 1 100%"
/>
```

But if you use `style={{ borderImage: ... }}` it works fine.

cschroeter:
closes #2040

### DIFF
--- a/packages/tabs/package.json
+++ b/packages/tabs/package.json
@@ -68,6 +68,7 @@
     "@chakra-ui/utils": "1.8.0"
   },
   "devDependencies": {
+    "@chakra-ui/modal": "1.8.7",
     "@chakra-ui/system": "1.6.7",
     "react": "^17.0.1"
   },
--- a/packages/tabs/package.json
+++ b/packages/tabs/package.json
@@ -68,7 +68,6 @@
     "@chakra-ui/utils": "1.8.0"
   },
   "devDependencies": {
-    "@chakra-ui/modal": "1.8.7",
     "@chakra-ui/system": "1.6.7",
     "react": "^17.0.1"
   },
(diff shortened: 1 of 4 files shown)


segunadebayo (commented):
Thanks for working on this.

Can you add a test or storybook to show this scenario? I'd love to check it out.

cschroeter:
@segunadebayo  done

packages/tabs/package.json
```diff
@@ -68,6 +68,7 @@
     "@chakra-ui/utils": "1.8.0"
   },
   "devDependencies": {
+    "@chakra-ui/modal": "1.8.7",
```
segunadebayo, line 71: Please remove this. I understand it's intended to be used in the stories but you can simply import from the `@chakra-ui/modal` without adding it. 

Since the `dist/` of the packages are hoisted to the root `node_modules`, there's nothing to worry about.

```suggestion
    "@chakra-ui/modal": "1.8.7",
```

PR #4103, opened 2021-05-27
fix(style): remove outline transitions

### LINKED ISSUE #4041: After version 1.6.2, the ring style will appear when the button blur
# 🐛 Bug report

After version `1.6.2`, the ring style (outline) will appear when the button blur (Chrome browser, used [`focus-visible`](https://chakra-ui.com/docs/migration#css-reset) package).

## 💥 Steps to reproduce

1. Go to [CodeSandbox](https://codesandbox.io/s/chakra-ui-button-blur-jnvfh?file=/src/App.tsx)
2. Click the button
3. Click anywhere except the button

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/chakra-ui-button-blur-jnvfh?file=/src/App.tsx

## 🌍 System information

| Chakra UI        |1.6.2|
|`focus-visible`| 5.2.0 |
| Browser          |Chrome 90.0.4430.212 |
| Operating System |macOS 11.3.1 |

with-heart:
Closes #4041

## 📝 Description

This change allows `focus-visible` to work correctly. Previously these
components used `transition: all` which caused the focus outline to
briefly flash and fade out when a component with no outline due to
`focus-visible` was blurred.

No

## 📝 Additional Information

For some reason the changes to `theme` weren't showing up in local Storybook. I wasn't able to figure out why, so we need to check that the `transition` styles are correct in the deployed csb before merging.


packages/button/stories/button.stories.tsx
```diff
@@ -11,6 +11,7 @@ import { FaFacebook, FaTwitter } from "react-icons/fa"
 import { MdBuild, MdCall } from "react-icons/md"
 import { BeatLoader } from "react-spinners"
 import { Button, ButtonGroup, IconButton } from "../src"
+import "focus-visible"
```
cschroeter, line 14: I don't think this is needed. It has the same issue without this dependency

cschroeter (commented):
LGTM

packages/button/stories/button.stories.tsx
with-heart: Whoops, didn't mean to include this! Had added it just for testing locally with Storybook. I'll remove that

segunadebayo (commented):
Looks good to me.

Can we consider using the transition tokens instead? So this feels more coherent with the theme?

```jsx
transitionProperty: "common",
transitionDuration: "normal"
```

What do you think @with-heart?

with-heart:
@segunadebayo Love that idea. Didn't even realize those were a thing!

with-heart:
@segunadebayo ❤️

PR #4107, opened 2021-05-28
Update color-mode.mdx

vineet8588:
Add a typescript  explanation for changing the theme config in color mode page.

## 📝 Description

Add an explanation in the color-mode docs page for extending the theme config in typescript

If no type for the config object is described, then an error is thrown in typescript

## 🚀 New behavior

No error after specifying the type

No

### DIFF
--- a/website/pages/docs/features/color-mode.mdx
+++ b/website/pages/docs/features/color-mode.mdx
@@ -49,6 +49,26 @@ const theme = extendTheme({ config })
 export default theme
 ```
 
+For typescript, you need to explicitly describe the theme config type as `ThemeConfig` object.
+
+```jsx live=false
+// theme.tsx
+
+// 1. import `extendTheme` function
+import { extendTheme, ThemeConfig } from "@chakra-ui/react"
+
+// 2. Add your color mode config
+const config : ThemeConfig = {
+  initialColorMode: "light",
+  useSystemColorMode: false,
+}
+
+// 3. extend the theme
+const theme = extendTheme({ config })
+
+export default theme
+```
+
 > Remember to pass your custom `theme` to the `ChakraProvider`, otherwise your
 > color mode config won't be taken into consideration.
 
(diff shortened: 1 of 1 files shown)


with-heart:
Documenting this is good, but I feel like we should loosen the types here because you shouldn't need to do anything special to declare your theme.

anubra266:
@with-heart, this might be necessary.

 image

website/pages/docs/features/color-mode.mdx
```diff
@@ -49,6 +49,26 @@ const theme = extendTheme({ config })
 export default theme
 ```
 
+For typescript, you need to explicitly describe the theme config type as `ThemeConfig` object.
+
+```jsx live=false
+// theme.tsx
```
anubra266, line 55: Use `theme.ts`

DecliningLotus:
I agree with @with-heart that the types should be loosened or updated to support setups like below, because needing to override it manually is unnecessary.

image

We should have a PR that updates the types rather than promote a workaround through the documentation.

anubra266:
Yeah...I agree

with-heart:
cc @TimKolberger

TimKolberger:
We only allow the values `light` and `dark` for the chakra colorMode. 
The TypeScript types enforce this configuration - so from an API perspective this is correct.

TypeScript ist hard - but I think **we are not helping the user on their learning path by loosening the types of chakra-ui**.

The shortest fix for this is declaring the string a [literal type](https://www.typescriptlang.org/docs/handbook/literal-types.html) with `as const`.

```ts
const config = {  
  initialColorMode: "light" as const,  
  useSystemColorMode: false,
}
```

I think we should update the docs either with the `as const` or the proposed solution in this PR.

website/pages/docs/features/color-mode.mdx
vineet8588: Ok , I'll update this PR as soon as I figure out how

vineet8588: Ok

vineet8588 (commented):
Replace theme.tsx with theme.ts

PR #4113, opened 2021-05-29
docs: removed extra dot in checkbox onChange desc

lazarnikolov94:
## 📝 Description
The `onChange` prop's description in Checkbox has an extra dot


with-heart (approved):
Wunderbar!

PR #4135, opened 2021-06-01
Feat/expand aria current pseudos

kristian240:
Closes # 

## 📝 Description

I added new pseudo prop that can be used for selecting element set as `aria-current="step"`.

To select that element you need to do this `&[aria-current="step"]`.

## 🚀 New behavior

It would be awesome if this can be achievable same as current page: `_activePage`.

New pseudo selector would be `_activeStep`

No, not a breaking change.

## 📝 Additional Information

I updated the docs as well 😄

### DIFF
--- a/packages/styled-system/src/pseudos.ts
+++ b/packages/styled-system/src/pseudos.ts
@@ -164,6 +164,11 @@ export const pseudoSelectors = {
    */
   _activeLink: "&[aria-current=page]",
   /**
+   * Used to style the active link in a navigation
+   * Styles for CSS Selector `&[aria-current=step]`
+   */
+  _activeStep: "&[aria-current=step]",
+  /**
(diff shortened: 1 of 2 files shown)


packages/styled-system/src/pseudos.ts
```diff
@@ -163,6 +163,11 @@ export const pseudoSelectors = {
    * Styles for CSS Selector `&[aria-current=page]`
    */
   _activeLink: "&[aria-current=page]",
+  /**
+   * Used to style the active link in a navigation
```
tim-ethitrade, line 167:
```suggestion
   * Used to style the current step within a process
```

with-heart (approved):
This LGTM if we merge @tim-ethitrade's suggestion!

PR #4144, opened 2021-06-02
docs(404): new 404 page

anubra266:
A more detailed 404 page with a return home button


website/pages/404.tsx
```diff
@@ -1,12 +1,41 @@
...
+        spacing="4"
+        as="section"
+        mt={["20", null, "40"]}
+        textAlign="center"
+      >
+        <Heading>404 | Page Not Found</Heading>
+        <Text fontSize={{ md: "xl" }}>
+          You just hit a route that doesn&#39;t exist... the sadness.😢
+        </Text>
+        <NextLink href="/" passHref>
```
segunadebayo, line 26: When you use `passHref`, please add `as="a"` to the Button so it works as a link not a button.

PR #4151, opened 2021-06-02
feat(input): add new input parts

gifaeriyanto:
Closes # 

## 📝 Description

> I've added input parts to support theming for `InputGroup`, `InputLeftElement` and `InputRightElement`.

> Currently, we can't customize the theme of `InputGroup`, `InputLeftElement` and `InputRightElement` because the parts is unavailable.

## 🚀 New behavior

> We able to customize the theme of `InputGroup`, `InputLeftElement` and `InputRightElement`.

### DIFF
--- a/packages/input/src/input-element.tsx
+++ b/packages/input/src/input-element.tsx
@@ -30,12 +30,14 @@ const InputElement = forwardRef<InputElementProps, "div">((props, ref) => {
   const input: any = styles.field
 
   const attr = placement === "left" ? "insetStart" : "insetEnd"
+  const part = placement === "left" ? "leftelement" : "rightelement"
 
   const elementStyles: SystemStyleObject = {
     [attr]: "0",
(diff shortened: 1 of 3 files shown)


segunadebayo:
Hi @gifaeriyanto,

We've thought about adding left and right elements to the theme but we don't think it is necessary because the left and right elements can be anything (a button, an icon button, an icon, or even a select component).

Due to these variations, we won't be adding these parts. Feel free to extend that on your end.

Thanks for understanding.

gifaeriyanto:
Hi @segunadebayo , thank you for letting me know.

vinterdo:
Hello, 
This PR is exactly what is missing in my use case. I want to customize my input groups to have a border when their input is focused via focus-within. I want the border to wrap both the left and right addon, so I can't do that via theming either addon or input themselves. Similarly somebody may want to add eg. shadow under the whole group. Please reconsider merging this PR

PR #4166, opened 2021-06-04
fix: consistent id generation

### LINKED ISSUE #4044: Accordion renders invalid IDs on SSR (Next.js)
# 🐛 Bug report

The Accordion component renders invalid IDs on SSR. Instead of counting the IDs for the accordion button and panel, the SSR-output shows `id="accordion-button-undefined"` and `id="accordion-panel-undefined"`.   
After the first render the IDs change to their correct value, like `id="accordion-button-1"`.

## 💥 Steps to reproduce

1. Open Sandbox in an new window
2. Disable JavaScript to see SSR output
3. Open the element inspector
4. Inspect the DOM elements of `<AcordionButton>` and `<AccordionPanel>` 

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/accordion-nextjs-issue-wrqym?file=/pages/index.tsx
Open Next.js page in new window: https://wrqym.sse.codesandbox.io/

## 🧐 Expected behavior

The IDs should be incremented correctly on client __and__ server. 

## 🌍 System information

| Chakra UI        |   1.6.2         |
| Browser          |   Chrome 90.0.4430.212 (arm64)  |
| Operating System |    macOS (Catalina)        |

## 📝 Additional information

Example SSR output of one accordion item

```html
<div className="chakra-accordion__item css-17mg6aq">
  <h2>
    <button
      type="button"
      id="accordion-button-undefined"
      aria-expanded="false"
      aria-controls="accordion-panel-undefined"
      className="chakra-accordion__button css-1rboc9n"
    >
      <div className="css-1eziwv">Section 1</div>
      <svg
        viewBox="0 0 24 24"
        focusable="false"
        className="chakra-icon chakra-accordion__icon css-186l2rg"
        aria-hidden="true"
      >
        <path
          fill="currentColor"
          d="M16.59 8.59L12 13.17 7.41 8.59 6 10l6 6 6-6z"
        ></path>
      </svg>
    </button>
  </h2>
  <div
    className="chakra-collapse"
    style="overflow:hidden;display:none;opacity:0;height:0px"
  >
    <div
      role="region"
      id="accordion-panel-undefined"
      aria-labelledby="accordion-button-undefined"
      className="chakra-accordion__panel css-1hb5ode"
    >
      ...
    </div>
  </div>
</div>
```

segunadebayo:
Closes #4044

## 📝 Description

This PR introduces a consistent API to generate isomorphic Ids on the client and server.

Our current approach (with reach/auto-id) generates an `undefined` id on the server and resolves to an `id` on the client. The only way to get a consistent id is by telling the consumer to specify an `id`.

## 🚀 New behavior

Consistent id on server and client. Big thanks to the work done by the React Aria team, I basically leveraged their SSRProvider model.

No

## 📝 Additional Information

To get this working, I had to render an `IdProvider` to the set of providers in `ChakraProvider`

### DIFF
--- a/examples/next-js/package.json
+++ b/examples/next-js/package.json
@@ -3,7 +3,7 @@
   "version": "1.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "rm -rf .next && next dev",
     "build": "next build",
     "start": "next start"
   },
--- a/packages/hooks/src/use-id.ts
+++ b/packages/hooks/src/use-id.ts
@@ -1,35 +1,42 @@
+// This implementation is heavily inspired by react-aria's implementation
+
 import * as React from "react"
-import { useSafeLayoutEffect } from "./use-safe-layout-effect"
 
-/**
- * Credit: https://github.com/reach/reach-ui/blob/develop/packages/auto-id/src/index.tsx
- */
-let handoffComplete = false
-let id = 0
-const genId = () => ++id
+type IdContextValue = {
+  prefix: number
+  current: number
+}
 
-/**
- * Reack hook to generate unique id
- *
- * @param idProp the external id passed from the user
- * @param prefix prefix to append before the id
- */
-export function useId(idProp?: string, prefix?: string) {
-  const initialId = idProp || (handoffComplete ? genId() : null)
-  const [uid, setUid] = React.useState(initialId)
+const defaultIdContext: IdContextValue = {
+  prefix: Math.round(Math.random() * 10000000000),
+  current: 0,
+}
 
-  useSafeLayoutEffect(() => {
-    if (uid === null) setUid(genId())
-  }, [])
+const IdContext = React.createContext<IdContextValue>(defaultIdContext)
 
-  React.useEffect(() => {
-    if (handoffComplete === false) {
-      handoffComplete = true
-    }
-  }, [])
+export const IdProvider: React.FC = React.memo(({ children }) => {
+  const currentContext = React.useContext(IdContext)
+  const isRoot = currentContext === defaultIdContext
+  const context: IdContextValue = React.useMemo(
+    () => ({
+      prefix: isRoot ? 0 : ++currentContext.prefix,
+      current: 0,
+    }),
+    [isRoot, currentContext],
+  )
 
-  const id = uid != null ? uid.toString() : undefined
-  return (prefix ? `${prefix}-${id}` : id) as string
+  return React.createElement(IdContext.Provider, { value: context }, children)
+})
+
+export function useId(idProp?: string, prefix?: string): string {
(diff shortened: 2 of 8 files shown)


examples/next-js/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "rm -rf .next && next dev",
```
cschroeter, line 6: You want to keep that? If so I would suggest using `predev: 'rimraf .next'`

packages/react/src/chakra-provider.tsx
```diff
@@ -72,8 +73,14 @@ export const ChakraProvider = (props: ChakraProviderProps) => {
     cssVarsRoot,
   } = props
 
-  return (
+  const _children = (
```
cschroeter, line 76: If you want to conditionally wrap I can highly recommend sth like

```
const ConditionalWrapper = ({ condition, wrapper, children }) => 
  condition ? wrapper(children) : children;
  ```
  
  
  ```tsx
<ConditionalWrapper
    condition={portalZIndex}
    wrapper={(children) => <PortalManager zIndex={portalZIndex}>{children}</PortalManager>}
>
    <EnvironmentProvider environment={environment}>{children}</EnvironmentProvider>
</ConditionalWrapper>
 ```

File: examples/next-js/package.json (hunk 1 above)
Comment (segunadebayo) [reply examples/next-js/package.json:6]: Good catch, I'll remove that.

File: packages/react/src/chakra-provider.tsx (hunk 2 above)
Comment (segunadebayo) [reply packages/react/src/chakra-provider.tsx:76]: That looks interesting. Not a fan of those wrappers but I'll consider it.

Thanks for sharing 🙏

**Comment by cschroeter** [comment]
LGTM tough <3

File: packages/hooks/src/use-id.ts [hunk 3]
```diff
@@ -49,5 +56,27 @@ export function useId(idProp?: string, prefix?: string) {
...
+ * Used to generate an id, and after render, check if that id is rendered so we know
+ * if we can use it in places such as `aria-labelledby`.
+ *
+ * @param partId - The unique id for the component part
+ *
+ * @example
+ * const { ref, id } = useOptionalPart<HTMLInputElement>(`${id}-label`)
+ */
+export function useOptionalPart<T = any>(partId: string) {
+  const [id, setId] = React.useState<string | null>()
```
with-heart, line 74:
```suggestion
  const [id, setId] = React.useState<string | null>(null)
```

with-heart, line 74:
```suggestion
  const [id, setId] = React.useState<string | null>(null)
```

with-heart (approved):

PR #4180, opened 2021-06-07
fix: initial position issue with popover or menu

### LINKED ISSUE #4109: Popover add unwanted scroll because of PopoverContent position
# 🐛 Bug report

Hello,

I am currently working with `Popover` and I found a bug.

When a `Popover` has not been opened yet, the `PopoverContent` is misplaced and can cause unwanted horizontal scroll.
Once it has been opened, the `PopoverContent` is placed correctly and everything is normal.

## 💥 Steps to reproduce

Use a `Popover` at the right of the window and a scroll will happen.

## 💻 Link to reproduction

See https://codesandbox.io/s/vigorous-dijkstra-d1s2k, the red block is the `PopoverContent` and as you can see, I've set the `Popover` placement in left but even so, the content is in bottom-end. 
Hover the button and the content will be placed correctly.

## 🧐 Expected behavior

PopoverContent should have the correct position event when it's not been opened.

## 🧭 Possible Solution

I found a workaround by setting `isLazy` to the `Popover` or by setting the `PopoverContent` in a `Portal` but I feel like it's a hacky fix 🤔 

## 🌍 System information

| Chakra UI        |      1.6.3      |
| Browser          |      Chrome and Firefox (at least)      |
| Operating System |        Mac OS Big Sur    |

### LINKED ISSUE #4186: Prevent Menu dropdown from overflowing and causing a scrollbar on bottom
# 🐛 Bug report

When using a dropdown menu in a nav element and pushing it to the right with a `<Spacer/>` the dropdown menu causes overflow and forces a scrollbar to appear on the bottom of the page.
This goes away after pressing the Menu and having the dropdown menu actually render.

## 💥 Steps to reproduce

1. Go to sandbox
2. see scrollbar on bottom
3. press the avatar on top right
4. see scrollbar disappear after dropdown renders

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/competent-chaum-ds0pw?file=/src/index.js

## 🧐 Expected behavior

I expect the dropdown menu not to overflow and cause a scrollbar

## 🧭 Possible Solution

No idea to be honest.

## 🌍 System information

| Chakra UI        |        1.6.3    |
| Browser          |        chrome    |
| Operating System |      windows 10      |

segunadebayo:
Closes #4109
Closes #4186

## 📝 Description

This PR makes it possible for users to choose if they want to position the menu and popover components on mount or when it's open.

We only position these components when it's open and it can cause an overflow on the page

I'd consider the current behavior a breaking change from the previous release but now we're in a better place. It's up to the user to choose which strategy they want.

### DIFF
--- a/.changeset/lucky-taxis-shave.md
+++ b/.changeset/lucky-taxis-shave.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/menu": minor
+"@chakra-ui/popover": minor
+---
+
+Added `computePositionOnMount` prop to allow positioning the popover or menu
+before initial open
--- a/.changeset/ten-turkeys-relax.md
+++ b/.changeset/ten-turkeys-relax.md
@@ -2,4 +2,6 @@
 "@chakra-ui/popper": patch
 ---
 
-Wrap force update within a function to prevent null scenarios
+- Wrap force update within a function to prevent null scenarios
+- Add default `inset` value to prevent overflow in scenarios where `enabled` is
+  `false` (i.e. when the popper is not visible)
--- a/.changeset/ten-turkeys-relax.md
+++ b/.changeset/ten-turkeys-relax.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/popper": patch
+---
+
+Wrap force update within a function to prevent null scenarios
--- a/packages/menu/src/use-menu.ts
+++ b/packages/menu/src/use-menu.ts
@@ -102,6 +102,13 @@ export interface UseMenuProps extends UsePopperProps, UseDisclosureProps {
    * @default "unmount"
    */
   lazyBehavior?: LazyBehavior
+  /**
+   * If `true`, the menu will be positioned when it mounts
+   * (even if it's not open)
+   *
+   * @default false
+   */
+  computePositionOnMount?: boolean
 }
 
 /**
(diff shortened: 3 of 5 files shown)


cschroeter (changes requested):
I would like to bring up tree points that I think are worth to talk about - if you don't mind. So lets get started :)

a) Since `enabled` is supported in Popper.js isn't that a bug in Popper.js itself?
b) From  a user perspective it is not clear that using `computePositionOnMount` will come with some performance costs.

c) 
For me the Menu component API getting to complex and also exposes to much implementation detail.
. I would expect that an `<Menu isLazy />` will not compute position on mount and also not initialize Popper.js where non lazy Menus, Tooltips etc will just eagerly do that. Would love to hear your thoughts on that

segunadebayo:
Thanks for the review @cschroeter. 

Here are my thoughts (we already discussed this 😃):

- This isn't related to `usePopper` because it does what we need it to do. It only exposes an `enabled` prop to allow control for when popper.js kicks in.

- I've added some notes around `computePositionOnMount` to indicate the perf cost that comes with that. Thanks for adding that.

- `isLazy` is a tricky subject but here's the state of my mind about it.

**Lazy rendering**: `isLazy` controls whether the "content" of the menu or popover is rendered ahead of time or when it's open.

**Lazy positioning**: It's a bit different but related to the above. It controls whether the popover or menu should be positioned ahead of time or when it's open.

It makes sense that `isLazy` should toggle both of them but I'm not convinced that this won't cause breaking changes. I'm quite wary of breaking changes at the moment.

I don't think the API is getting complex or we're exposing internal details. I think this is a valid use-case where the user needs to decide how they want to both position the menu/popover and render the content.

That said, I don't expect any new prop to be added to these components in a while as I believe it covers most use-cases.

PR #4191, opened 2021-06-07
chore: add directory to all package.json repository metadata

forivall:
Closes # 

## 📝 Description

Adds the "directory" metadata field to the "repository" section of all of the `package.json` files. From [npm docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository):

> If the `package.json` for your package is not in the root directory (for example if it is part of a monorepo), you can specify the directory in which it lives:

N/A

## 🚀 New behavior

N/A

No.

## 📝 Additional Information

I just ran the zsh oneliner `for p in packages/*(/); jq '.repository.directory = $dir' --arg dir $p $p/package.json | awk 'BEGIN{RS="";getline<"-";print>ARGV[1]}' $p/package.json`, and manually updated the "react" package, since it was just a string.


cschroeter:
LGTM

@forivall I would be interested to know if there is a key benefit to this

forivall:
When using "npm repo" and click on the repo link on npmjs.com, and other tools, it goes straight to the source directory if the package, instead of the root. So it's a little nicer for checking out the source code, which is what I was doing (as co-workers are using some of your utils)

PR #4199, opened 2021-06-09
docs: update tooltip docs with arrow styling information

adamklepacz:
Update tooltip docs with arrow styling information.

Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior

NO.

## 📝 Additional Information

@segunadebayo When I was trying to customize the Tooltip - the information about styling up the arrow was missing. No info on docs, stackoverflow etc. It was hard to figure out how to change the color of an arrow. Then I found the way to do it by digging into the code of the component,  which is not so obvious to every other developer. I think it would be super nice to add this information to the docs as some other people may struggle with it while extending the default tooltip styles.

### DIFF
--- a/website/pages/docs/overlay/tooltip.mdx
+++ b/website/pages/docs/overlay/tooltip.mdx
@@ -160,6 +160,25 @@ Using the `placement` prop you can adjust where your tooltip will be displayed.
 </VStack>
 ```
 
+## Styling arrow
+
+The API for changing the background of an arrow is:
+
+> Note ⚠️: If you're styling the arrow element, it does not accept Chakra UI 
+> color values. You have to put the color as an explicit css value.
+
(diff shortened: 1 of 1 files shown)


adamklepacz:
@segunadebayo  When I was trying to customize the Tooltip - the information about styling up the arrow was missing. No info on docs, stackoverflow etc. It was hard to figure out how to change the color of an arrow. Then I found the way to do it by digging into the code of the component, which is not so obvious to every other developer. I think it would be super nice to add this information to the docs as some other people may struggle with it while extending the default tooltip styles.

segunadebayo:
This feels more like a bug to me than a documentation update. I'm going to take a look at it soon

cschroeter:

PR #4211, opened 2021-06-11
fix(menu): fixes menu typeahead

tim-stasse:
## 📝 Description

While not yet released, 245a164 introduced a bug that breaks the Menu components typeahead search functionality. This change adds a fix that also [presumably] keeps true to the original intent of 245a164.

The Menu typeahead does not work at all.

## 🚀 New behavior

Reintroduces the pre-[245a164](https://github.com/chakra-ui/chakra-ui/commit/245a164f6058e96986b2354017d96816b5d336e9) behaviour for Menu typeahead.

No

## 📝 Additional Information

No issue was created or associates with this PR because 245a164 hasn't been released yet. Please inform if doing so before merging the PR is preferred.

### DIFF
--- a/.changeset/brown-radios-rhyme.md
+++ b/.changeset/brown-radios-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/menu": patch
+---
+
+Fixed menu typeahead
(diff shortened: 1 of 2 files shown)


segunadebayo:
Good catch @tim-stasse!

PR #4222, opened 2021-06-12
Update color-mode.mdx (Docs)

LFerronato:
Add update suggestion for website documentation, to avoid white flash

## 📝 Additional Information
I believe there are no side effects. It just adds a suggestion in the documents to avoid white flashes when reloading the page (for NextJS, but I believe it works for pure React too)


LFerronato:
i will re-write

PR #4240, opened 2021-06-16
fix(toast): update use-toast to not return new object each render

### LINKED ISSUE #4139: Updated toast should retain original options.
# 🚀 Feature request

I think updating a toast should retain the previous options.

When I create a toast, say with `error` status, When updating the toast, I must specify the status again, or it resets to default, which is info. 
When I create a toast, I might just want to update the description. I shouldn't have to specify the status and the title all over again.

## ✅ Proposed solution or API
See how it is done [here](https://fkhadra.github.io/react-toastify/update-toast)

### LINKED ISSUE #4230: [Toast] useToast() returns a new object each time
# 🐛 Bug report

This issue seems to have returned: https://github.com/chakra-ui/chakra-ui/issues/1448

Each time we call `useToast` hook, a new object is returned.

## 💥 Steps to reproduce

Same as previous issue: https://github.com/chakra-ui/chakra-ui/issues/1448

## 💻 Link to reproduction

Same as previous issue: https://github.com/chakra-ui/chakra-ui/issues/1448

## 🧐 Expected behavior

The object returned by `useToast` hook should always be from the same reference.

## 🧭 Possible Solution

 I think perhaps the `useChakra()` hook on this line (https://github.com/chakra-ui/chakra-ui/blob/ebda07095bffd9b3135c5d19803a3a08397b78ef/packages/toast/src/use-toast.tsx#L191) is returning new object references each time which is causing the `useCallback` to have no effect.

## 🌍 System information

| Chakra UI        |   1.6.3         |
| Browser          |   Chrome         |
| Operating System |    macOS latest        |

rossmoody:
Closes #4230 
Closes #4139

## 📝 Description

I took a stab at fixing the behavior outlined in the issue. I'm unsure if this is the best solve but some of the investigation I did might help us get there if this isn't it. 

I created a [CodeSandbox similar to the original issue](https://codesandbox.io/s/chakra-toast-4230-cn8rb?file=/src/Component.tsx) to reproduce the new issue and noticed that only the toasts that get passed an argument were re-rendering on state updates.

I think what's happening is the `useMemo` has the `options` argument as a dependency and so it triggers a re-render on any state update because the options being passed are considered new.

## 🚀 New behavior

The proposed solve is to store the `options` arguments as a scoped state variable that the `useMemo` depends on instead.

No


feross:
Thanks for taking a stab at this @rossmoody. Didn't know you were a Chakra UI user :)

rossmoody:
All about Chakra! I found it from one of your tweets @feross  and have been consistently impressed ever since. 

Updated the Sandbox link which I had wrong and there is a possibility this might address or conflict some consideration of this issue as well: https://github.com/chakra-ui/chakra-ui/issues/4139

packages/toast/src/use-toast.tsx
```diff
@@ -189,16 +189,19 @@ export function createStandaloneToast({
  */
 export function useToast(options?: UseToastOptions) {
   const { theme, setColorMode, toggleColorMode, colorMode } = useChakra()
+
+  const [toastOptions] = React.useState(options)
```
kevin940726, line 193: Why not `useRef`? Also do we allow the default options to be changed after rendered?

anubra266:
@rossmoody It does fix #4139, so you could update your description to also close that issue when it's merged.

anubra266 (approved):

anubra266:
cc @segunadebayo

packages/toast/src/use-toast.tsx
```diff
@@ -202,15 +202,20 @@ export function createStandaloneToast({
...
+  const toastOptions = React.useRef(options)
+
+  return React.useMemo(
+    () =>
+      createStandaloneToast({
+        theme,
+        colorMode,
+        setColorMode,
+        toggleColorMode,
+        defaultOptions: toastOptions,
```
segunadebayo, line 215:
```suggestion
        defaultOptions: toastOptions.current,
```

packages/toast/src/use-toast.tsx
```diff
@@ -202,15 +202,20 @@ export function createStandaloneToast({
...
+      }),
+    [theme, setColorMode, toggleColorMode, colorMode, toastOptions],
```
segunadebayo, line 217:
```suggestion
    [theme, setColorMode, toggleColorMode, colorMode],
```

segunadebayo (changes requested):
Thanks for working on this @rossmoody.

Kindly see my comments below. I think we can also leverage the `useLiveRef` hook for this use-case. See the `hooks` package to learn how it works.

anubra266:
> Thanks for working on this @rossmoody.
> 
> Kindly see my comments below. I think we can also leverage the `useLiveRef` hook for this use-case. See the `hooks` package to learn how it works.

Hi @segunadebayo I don't think we have a `useLiveRef` hook

packages/toast/src/use-toast.tsx
rossmoody: I pulled this branch down and built it again with `useLatestRef` hook. Typescript yelled at me if I omitted this from the dependency array.

anubra266 (approved):

TimKolberger (approved):

TimKolberger:
Closed in favor of #5493 because source repository does not exist anymore and I could not add a changeset.

PR #4253, opened 2021-06-18
fix discrepancy in main theme entrypoint filename

wgiddens:
## 📝 Description
Fixes a small discrepancy in the docs. The directory structure recommends the main theme entrypoint be called index.js but the comment calls it theme.js.

No.

### DIFF
--- a/website/pages/docs/theming/customize-theme.mdx
+++ b/website/pages/docs/theming/customize-theme.mdx
@@ -215,7 +215,7 @@ This way, you can structure your main theme entrypoint file to be much cleaner,
 like this:
 
 ```jsx live=false
-// theme.js
+// theme/index.js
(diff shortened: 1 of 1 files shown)


TimKolberger:
Thank you for opening a PR to improve chakra-ui!

I think the `theme.ts` comment helps the reader to understand that this code is supposed to be somewhere else than in the `index.ts` of their application.

Replacing it with `index.ts` would make this ambiguous, so I would let it be as it is. 

What do you think?

wgiddens:
The thing that threw me off was the directory structure recommendation listed just above it named the file index.js.  

What about changing the comment to `theme/index.js`?

TimKolberger:
`theme/index.js` looks good to me 👍

wgiddens:
It is switched over to `theme/index.js`

PR #4264, opened 2021-06-21
fix(select): background bars in select menu - firefox #4194

### LINKED ISSUE #4194: Select dropdown has white bars above and below the list of options
# 🐛 Bug report

On the select drop down in dark mode there are white bars above and below the options, this is not seen in lightmode as its white. (doesn't happen in chrome, but does in firefox). It is also worth nothing that chrome does not have gaps above and below the list of options like firefox does. I also am having trouble inspecting element as the dropdown closes every time i try to select that element.

## 💥 Steps to reproduce

Visit the [docs](https://chakra-ui.com/docs/form/select)

## 🧐 Expected behavior

Bars to be the same colour as the rest of the background or remove bars like it is on chrome.

## 🌍 System information

| Chakra UI        |       latest     |
| Browser          |        firefox    |
| Operating System |      windows 10      |

## 📝 Additional information

image

image

mah51:
Closes #4194

## 📝 Description

Fixes issue with select dropdowns having a white bar in dark mode, I wasn't sure where to put prop so I put it in the select file, but if you tell me where to set this I can change it. Thanks1

white bar in background of dropdown (windows 10 firefox latest)

## 🚀 New behavior

Adds set background color for select instead of default inherit attribute.

No

### DIFF
--- a/packages/select/src/select.tsx
+++ b/packages/select/src/select.tsx
@@ -9,6 +9,7 @@ import {
   ThemingProps,
   useMultiStyleConfig,
   HTMLChakraProps,
+  useColorModeValue,
 } from "@chakra-ui/system"
 import { cx, mergeWith, split, __DEV__ } from "@chakra-ui/utils"
 import * as React from "react"
@@ -30,6 +31,7 @@ export const SelectField = forwardRef<SelectFieldProps, "select">(
         {...ownProps}
         ref={ref}
         className={cx("chakra-select", className)}
+        bg={useColorModeValue("white", "gray.700")}
       >
         {placeholder && <option value="">{placeholder}</option>}
         {children}


packages/select/src/select.tsx
```diff
@@ -30,6 +31,7 @@ export const SelectField = forwardRef<SelectFieldProps, "select">(
         {...ownProps}
         ref={ref}
         className={cx("chakra-select", className)}
+        bg={useColorModeValue("white", "gray.700")}
```
segunadebayo, line 34: Please move this change to the [select's theme](https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/select.ts)

segunadebayo (changes requested):
Thanks for this fix, please see my comment below.

packages/select/src/select.tsx
mah51: Hi, Ive had a look into this, but due to the nature of the select component I am struggling to inspect the html associated with it. Also I think this might be a different problem to what I originally thought, as the background of the Select bar itself is meant to be transparent as on chrome but on firefox for some reason it has extra space above and below the options which show the white background of the default html page.

segunadebayo:
Thanks for attempting this @mah51.

I've added the fix for this in a separate PR.

PR #4268, opened 2021-06-22
fix(use-breakpoint): avoid re-rendering when using `useBreakpoint`

### LINKED ISSUE #4267: defaultBreakpoint still causes re-render
# 🐛 Bug report

When not using `defaultBreakpoint` in `useBreakpoint`, the initial value was always `undefined` and then when a breakpoint could be found, it was updated to the appropriate value. This causes a re-render, which makes a lot of sense.

However, with the support for `defaultBreakpoint`, it feels very unfortunate that it would still cause a re-render, even if the found breakpoint is the same as `defaultBreakpoint`.

## 💻 Link to reproduction

https://codesandbox.io/s/quirky-dijkstra-zgngc?file=/src/index.tsx

As you can see, the `App` is rendered twice. If you remove the argument to `useBreakpoint`, we will get `undefined` and then `md`, which **makes a lot more sense**. But now, when the `defaultBreakpoint` already matches the breakpoint that Chakra finds later, we shouldn't need to re-render.

## 🧐 Expected behavior

I expect no re-render if the same breakpoint is found.

## 🧭 Possible Solution

Looking at [the code](https://github.com/chakra-ui/chakra-ui/blob/bb0356f10e14886a19072b68f9e4999c6a3bfc1d/packages/media-query/src/use-breakpoint.ts), it looks like we are storing an object in the state. This object will be changed to a new object whenever the `useEffect` goes through.

It also looks like we only actually need the `currentBreakpoint?.breakpoint`, so why not store this value directly in the state? Then it would be strings and `useState` wouldn't trigger a re-render, since the values are identical.

## 🌍 System information

| Chakra UI        | `1.6.4`    |
| Browser          | `Chrome 90.0.4430.93` |
| Operating System | `Ubuntu 20.04.2` |

## 📝 Additional information

None.

anton-johansson:
Closes #4267 

## 📝 Description

Avoids re-rendering when using `useBreakpoint` together with `defaultBreakpoint`.

Since we were storing an object in `useState`, React could not see the new value as identical, and triggered a re-render.

## 🚀 New behavior

We're now only storing the actual `string`, which React can see as identical, and will avoid re-render. We were not using the other values on the object anyway.

No, I wouldn't say so. It will of course change the behavior, but should not do so in any breaking way.

## 📝 Additional Information

**Note:** I could not get the development mode working, so I haven't been able to try this locally. I want to be able to link my project to my local Chakra UI, but I couldn't find any links after running `yarn boot`. This is why I chose a "Draft" PR.

### DIFF
--- a/packages/media-query/src/use-breakpoint.ts
+++ b/packages/media-query/src/use-breakpoint.ts
@@ -42,18 +42,18 @@ export function useBreakpoint(defaultBreakpoint?: string) {
     )
 
     if (mediaQuery) {
-      const { query, ...breakpoint } = mediaQuery
+      const { breakpoint } = mediaQuery
       return breakpoint
     }
 
     return undefined
   })
 
-  const current = currentBreakpoint?.breakpoint
+  const current = currentBreakpoint
 
   const update = React.useCallback(
-    (query: MediaQueryList, breakpoint: Breakpoint) => {
-      if (query.matches && current !== breakpoint.breakpoint) {
+    (query: MediaQueryList, { breakpoint }: Breakpoint) => {
+      if (query.matches && current !== breakpoint) {
         setCurrentBreakpoint(breakpoint)
       }
     },


anton-johansson:
This fix does not seem to solve the problem. Debugging a bit, it seems like `setCurrentBreakpoint` isn't even called. I don't know what's triggering the re-render...

Any ideas?

segunadebayo:
Thanks for taking a look at this.

This may need more investigation before we can come up with a solution. For now, I'm going to close this.

PR #4292, opened 2021-06-27
docs: add /apply-theme route with customizing theme feature

tomchentw:
## 📝 Description

This PR adds the `/apply-theme` route to the official v1 docs site. It allows user to try customizing theme directly on the v1 docs site.

## 🚀 New behavior

This is fairly useful for the users to experiment with Chakra-UI. It also provides a way for organization users to create a public Gist that hosts their design system, saving their needs to create a forked Chakra-UI documentation site.

JPEG-1

To get started, the user has two different ways to interact with this route:

1. Import from a public gist URL:

gist-import

2. Try from a Live Editor

live-editor

Furthermore, it features storing the Gist Id or the code string in the `localStorage`, giving the flexibility of visiting the docs site.

3. Storing Gist Id in localStorage

stored-gist

4. Storing the code string in localStorage

stored-code

Last, when importing the Gist ID, it stores the Gist ID  to the query param so that you could share the URL to someone else directly:

```
https://chakra-ui.com/apply-theme?fromGistId=989ad340001061726bf2c0734d3739cf
```

5. Read from URL

gits-from-url

No

### DIFF
--- a/website/pages/apply-theme.tsx
+++ b/website/pages/apply-theme.tsx
@@ -18,7 +18,7 @@ import {
 import PageContainer from "components/page-container"
 import Sidebar from "components/sidebar/sidebar"
 import { getRoutes } from "layouts/mdx"
-import _ from "lodash"
+import { last, omit } from "lodash"
 import { extension } from "theme"
 import { useContext, useCallback, useEffect, useState } from "react"
 import {
@@ -47,7 +47,7 @@ function GistTheme({ onChangeGistId }) {
           url: { value },
         },
       } = event
-      const fromGistId: string = _.last(value.split("/"))
+      const fromGistId: string = last(value.split("/"))
       if (!isValidGistId(fromGistId)) {
         return
       }
@@ -182,7 +182,7 @@ function ApplyTheme() {
     })
     replace({
       pathname: route,
-      query: _.omit(query, ["fromGistId"]),
+      query: omit(query, ["fromGistId"]),
     })
   }, [code, replace, route, query])
 
--- a/website/pages/apply-theme.tsx
+++ b/website/pages/apply-theme.tsx
@@ -169,7 +169,7 @@ function ApplyTheme() {
   )
   const setThemeConfig = useContext(ThemeConfigContext)
   useEffect(() => {
-    if (!isValidGistId(query.fromGistId)) {
+    if (!isValidGistId(query?.fromGistId)) {
       return
     }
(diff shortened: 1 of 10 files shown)


website/pages/apply-theme.tsx
```diff
@@ -0,0 +1,239 @@
...
+  Text,
+  Link,
+  extendTheme,
+  useClipboard,
+  useToast,
+} from "@chakra-ui/react"
+import PageContainer from "components/page-container"
+import Sidebar from "components/sidebar/sidebar"
+import { getRoutes } from "layouts/mdx"
+import _ from "lodash"
```
ljosberinn, line 21: please use `import { last } from 'lodash' `  or `import last from 'lodash.last'` as you'd otherwise import the entirety of lodash

website/src/components/customizable-theme/index.tsx
```diff
@@ -0,0 +1,229 @@
+import { useRouter } from "next/router"
+import _ from "lodash"
```
ljosberinn, line 2: same here :)

ljosberinn:
Thank you, this is a great idea!

with-heart:
Hi @tomchentw! I'd like to echo @ljosberinn here and say that this is a really cool idea. I'd like to take responsibility for thoroughly reviewing this and helping shape this into a solid feature for the site.

Busy this morning but I will have time either later today or tomorrow to dig in. Thanks for working on this!

website/pages/apply-theme.tsx
tomchentw: @ljosberinn Got it and I will update it. Forgot that we don't have `babel-plugin-lodash` setup here.

I noticed that there are some other places of `website` that `import _ from "lodash"` directly so I assume it's okay do to so. I could fix those places in another PR for sure :)

JPEG-1-1

tomchentw: Switched to `import last from 'lodash/last'`

tomchentw: FYI, I've created a PR to take care of existing imports: https://github.com/chakra-ui/chakra-ui/pull/4318

segunadebayo:
This is an awesome idea. I think it makes sense to have it be a separate app or playground so that users can play with it in isolation. I don't think it's something we want to add to the docs directly.

Please consider setting this up in a different URL and we can link to it in the docs.

Thanks for taking the time to work on this.

PR #4300, opened 2021-06-28
fix(dependency): sharp on m1

### LINKED ISSUE #4299: Repo cannot build on m1 Macs
The repo is unable to build on M1 Macs because the `sharp` package cannot build.

anubra266:
Fixes #4299

The repo cannot run on m1 Macs because the present version of the `sharp` package depends on homebrew to build. But homebrew is not fully supported on m1 macs yet. So there is a new `sharp` version that fixes this. 
This PR updates sharp.

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,7 +163,7 @@
     "react-table": "7.6.2",
     "react-test-renderer": "^17.0.1",
     "rimraf": "3.0.2",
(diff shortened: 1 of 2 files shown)


anubra266:
@with-heart 🙏

with-heart (approved):

PR #4317, opened 2021-07-02
feat: enable root element to be themed

### LINKED ISSUE #4270: Can't add custom variant to FormControl component
# 🐛 Bug report

When trying to add a custom variant to the `FormControl` component, it doesn't seem to work. I originally tried adding the variant under the `FormControl` component key. Looking at the `FormControl` component however it is actually in the theme as `Form` instead, https://github.com/chakra-ui/chakra-ui/blob/main/packages/form-control/src/form-control.tsx#L220 and https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/form.ts.

## 💥 Steps to reproduce

1. Extend the Chakra theme and add a custom variant for `Form`
2. Add the variant prop to the FormControl component
3. See styles not take affect

## 💻 Link to reproduction

https://codesandbox.io/s/chakra-theme-form-control-50btm?file=/src/input.js
CodeSandbox reproduction: https://your-codesandbox-link-goes-here.io

## 🧐 Expected behavior

I expected the `FormControl` component to honor the variant styles defined in the theme.

## 🌍 System information

| Chakra UI        |     1.6.4       |
| Browser          |      Google Chrome Version 91.0.4472.114 (Official Build) (x86_64)      |
| Operating System |       MacOS Big Sur     |

bhishp:
Closes #4270

## 📝 Description

> Support variants on FormControl by using the the form styles from the theme. Changed the FormControl to look at the styles object rather than being embedded within the component

> Variants cannot be set

## 🚀 New behavior

> Variants can be set to override FormControl styling

No

### DIFF
--- a/packages/form-control/tests/form-control.test.tsx
+++ b/packages/form-control/tests/form-control.test.tsx
@@ -238,7 +238,7 @@ test("inherits variant styles", async () => {
               Form: {
                 variants: {
                   "floating-label": {
-                    "form-control": {
+                    formControl: {
                       bg: "white",
                       border: "1px",
                       borderColor: "#ccc",
--- a/packages/form-control/tests/form-control.test.tsx
+++ b/packages/form-control/tests/form-control.test.tsx
@@ -1,6 +1,7 @@
 import { chakra, forwardRef, PropsOf } from "@chakra-ui/system"
 import { fireEvent, render, screen, testA11y } from "@chakra-ui/test-utils"
 import * as React from "react"
+import { ChakraProvider, extendTheme } from "../../react"
 import {
   FormControl,
   FormControlOptions,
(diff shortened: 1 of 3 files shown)


packages/theme/src/components/form.ts
```diff
@@ -2,7 +2,7 @@ import { mode } from "@chakra-ui/theme-tools"
 
 type Dict = Record<string, any>
 
-const parts = ["requiredIndicator", "helperText"]
+const parts = ["formControl", "requiredIndicator", "helperText"]
```
with-heart, line 5:
```suggestion
const parts = ["container", "requiredIndicator", "helperText"]
```

We _mostly_ have a pattern of calling the root element a `container`. Could you update this to follow that pattern?

packages/theme/src/components/form.ts
```diff
@@ -21,6 +21,10 @@ function baseStyleHelperText(props: Dict) {
 }
 
 const baseStyle = (props: Dict) => ({
+  formControl: {
```
with-heart, line 24:
```suggestion
  container: {
```

with-heart (changes requested):
We like the idea of making this themeable, but we need to follow the existing naming pattern for the theme key.

bhishp:
Made those suggested changes, thanks for feedback @with-heart

packages/theme/src/components/form.ts
bhishp: 👍🏽

with-heart (approved):

with-heart:
Thanks @bhishp!

@allcontributors please add @bhishp for code

with-heart:
I made a few changes to this PR:

- Updated the title to `feat: enable root element to be themed`
	- The original issue is a bug report but I would consider this a new feature.
	- I like to follow most of the guidelines set out in [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/). In this case, I edited it to follow [rule #2](https://chris.beams.io/posts/git-commit/#limit-50) and [rule #5](https://chris.beams.io/posts/git-commit/#imperative).
- Added a changeset file (`.changeset/sharp-cooks-clap.md`)
	- Set the version bump as `minor` for both `@chakra-ui/form-control` and `@chakra-ui/theme` since this is new functionality for both.
	- Added a summary explaining _what_ changed and _why_ the change was made along with an example of how the user can utilize the change.

packages/form-control/tests/form-control.test.tsx
```diff
@@ -222,3 +223,95 @@ test("has the correct data attributes", async () => {
...
+          })}
+        >
+          {children}
+        </ChakraProvider>
+      ),
+    },
+  )
+
+  const formControl = screen.getByTestId("control")
+  expect(formControl).toMatchInlineSnapshot(`
```
with-heart, line 265: This will fail fairly often because the generated snapshot doesn't match _only_ the variant styles, but also includes styles from `baseStyle` and other pieces of the theming system. 

How do you feel about removing this test completely? We don't test any other components in this way.

bhishp: That is a good point regarding the other parts of the theming system. Yes, I am happy to remove this completely, I was following the advice from the contributing guide, "All commits that fix bugs or add features need a test" but I did also notice that these sort of tests do no exist elsewhere.

with-heart: Sorry about that! That advice is true for many things, but "All" is probably not the right word.

PR #4345, opened 2021-07-07
docs: apply custom theme from a Gist

tomchentw:
## 📝 Description

I've extracted the key feature from #4292 : 

> Apply custom theme from a Gist

## 🚀 New behavior

It starts with the newly added icon on the header. It renders a `<Popover />` with an `<input />` asking for the Gist URL.

JPEG-0

The user should provide a Gist URL. It should contain one CommonJS module defined the custom theme. For example,
https://gist.github.com/tomchentw/989ad340001061726bf2c0734d3739cf

JPEG-1

## 🚀 New behavior

We could paste the Gist URL to the input and hit `Apply`. Since the custom theme above changes the `xs` size of the `<Button />`, It updates the icon size as expected:

e449df7251828d00a1188f234a9c7180

A static image:

JPEG-2

No

## 📝 Additional Information

Let me know if this is easier to review, @with-heart . We could ship the `localStorage` feature or `live editing feature later on with more thorough thinking.

### DIFF
--- a/website/pages/_app.tsx
+++ b/website/pages/_app.tsx
@@ -1,9 +1,9 @@
 import { ChakraProvider } from "@chakra-ui/react"
+import { CustomizableThemeProvider } from "components/customizable-theme"
 import FontFace from "components/font-face"
 import { DefaultSeo } from "next-seo"
 import Head from "next/head"
 import React from "react"
-import theme from "theme"
 import { getSeo } from "utils/seo"
 
 const App = ({ Component, pageProps }) => {
@@ -27,9 +27,9 @@ const App = ({ Component, pageProps }) => {
         )}
       </Head>
       <DefaultSeo {...seo} />
-      <ChakraProvider theme={theme}>
+      <CustomizableThemeProvider>
         <Component {...pageProps} />
-      </ChakraProvider>
+      </CustomizableThemeProvider>
       <FontFace />
     </>
   )
--- a/website/src/components/customizable-theme/gist-popover.tsx
+++ b/website/src/components/customizable-theme/gist-popover.tsx
@@ -17,10 +17,15 @@ import {
   Button,
   Text,
 } from "@chakra-ui/react"
-import { useState, ReactElement } from "react"
-import { getFromGistId, isValidGistId } from "./helpers"
+import { useState, useContext, ReactElement } from "react"
+import {
+  getFromGistId,
+  isValidGistId,
+  CustomizableThemeContext,
+} from "./helpers"
 
 function GistPopover({ icon }) {
+  const updateThemeFromFile = useContext(CustomizableThemeContext)
   const [state, setState] = useState({ isLoading: false, error: false })
   function onSubmit(event) {
     event.preventDefault()
@@ -38,7 +43,7 @@ function GistPopover({ icon }) {
     setState((it) => ({ ...it, isLoading: true }))
     getFromGistId(fromGistId).then(
       (it) => {
-        console.log("Success", it)
+        updateThemeFromFile(it)
         setState((it) => ({ ...it, isLoading: false }))
       },
       (error) => {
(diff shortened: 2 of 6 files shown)


with-heart:
Hi @tomchentw! You're really crushing it with your PRs related to this functionality and I really love the work you've been doing. After @segunadebayo and I discussed this yesterday on our [Maintaining Chakra UI livestream](https://youtu.be/0TLxtE_GCbM?t=176), we feel like this functionality would make more sense as its own standalone app rather than being embedded in the docs. 

In my view, this is for a few reasons:

1. This functionality feels awkward in the docs site. Since the user's theme isn't built for the purpose of styling the Chakra docs, the output would be inconsistent and potentially confusing for users.
2. We've always wanted to have a "theme explorer" app that provides users with a way to customize their theme and see immediate results, so they can explore their theme and experience changes in output as they happen. 
3. The docs site isn't built to support this functionality. We use a lot of custom styling that isn't connected to the theme—and while we could fix this—I would hate for a user's custom theme to then break the docs site in subtle ways that they might not notice.

Would you be interested in working on a project to build a fully-fledged theme explorer type of app instead?

segunadebayo:
Totally agree with @with-heart, this should be a standalone app, perhaps you can call it "Theme Explorer".

Kindly consider this and let us know what you think (here or via Discord).

In the meantime, we'll close this

anubra266:
Hi @tomchentw I'm working on Chakra Play, though I've not done much on it recently, College troubles.🙂
But I'll love if we could work on it together, this being a part of it. The user can view his chakra code live, customize the theme with code, and also import his theme from a gist.

tomchentw:
## Intentions

@with-heart @segunadebayo I'd like to share my intentions of creating #4292 and #4345 :
> Having a customized Chakra-UI docs site for _YOUR COMPANY/ORGANIZATION_

Imagine one organization who's leveraging Chakra-UI. They want to customize theme with a few values: `sizes`, `spacings`, `colors`. It does take efforts to host a custom Chakra-UI docs site, and not to mention they have to keep up with the upstream branch updates. 

If the docs site allows to import theme from gist, they could simply setup something like:

1. https://chakra-ui.organization.com -> points to `https://chakra-ui.com/apply-theme?fromGistId=989ad340001061726bf2c0734d3739cf`
2. Whoever in the organization visit that page will automatically get the organization specific Chakra-UI docs site.

Want to use a different `red` color palette and wondering how it affects `<Button />` or `error toast`? No problem. The UI change could be shown directly to non-engineering departments and get immediate feedbacks.

To prevent confusion, we could add a top-banner indicating that a custom theme is currently applied like I did in #4292 or [GIF here](https://user-images.githubusercontent.com/922234/123544227-e4d8cb00-d784-11eb-8220-f49b33933421.gif):
JPEG-1

Since the changes are all running locally, the user could always reset to the default theme.

Perhaps the real question here: **how would you setup a customized Chakra-UI docs site for your organization?**

## Theme Explorer

A theme explorer is definitely useful for a playground situation. When I was reading the [customize-theme](https://chakra-ui.com/docs/theming/customize-theme) page in the first place, I'm already imagine having a theme explorer right at that page so that I could try it out directly. Maybe this is one approach we could take from here?

On the other hand, I imagine a theme explorer would look similar to a CodeSandbox, leaving this playground more engineering-centered approach compared to the docs site. Plus, the best of the Chakra-UI docs site lies within its massive examples (with various variants and real-world usages). I would imaging it takes efforts to leverage (import) these examples as a separate theme explorer App.

Given these, I'm still convinced having a customized theme feature directly in the docs site adds value to it. :)

tomchentw:
@anubra266 I'm impressed with the [choc-autocomplete](https://github.com/anubra266/choc-autocomplete) package. Haven't had a chance to try it but looking forward to! I also checked [Chakra Play](https://chakra-play.tk/) you mentioned here and yeah I'm open to see how it could play together!

anubra266:
@tomchentw I definitely love this idea. It'll be awesome on [Choc UI](https://choc-ui.tech) too.

PR #4353, opened 2021-07-09
fix(button): take left and right icons into consideration for button loading state

dsumer:
## 📝 Description

This is a fix so that the left and right icon of a button also stay in the DOM while the button isLoading to avoid jumping button elements.

Currently, the left and right icons of a button don't get rendered when the button is in the loading state, which leads to less width of the button, leading to unwanted layout shifts.

## 🚀 New behavior

The left and right icons also get rendered when the button is in the loading state and will be hidden with opacity 0 like the rest of the children.

No

### DIFF
--- a/.changeset/cold-guests-hug.md
+++ b/.changeset/cold-guests-hug.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/button": patch
+---
+
+Resolved an issue where a `Button` in loading state didn't consider the width of
+`leftIcon` and `rightIcon`, resulting in layout shifts when the button leaves
+the loading state. Buttons now render with the same width regardless of state.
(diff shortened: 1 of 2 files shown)


with-heart:
Current: https://codesandbox.io/s/create-react-app-ts-forked-su1wr?file=/src/App.tsx
After this PR: https://codesandbox.io/s/create-react-app-ts-forked-kbfxe?file=/src/App.tsx

with-heart (approved):

with-heart:
I added a changeset file (`.changeset/cold-guests-hug.md`) so `@chakra-ui/button` will receive a version bump for this change and will generate a `CHANGELOG` entry.

with-heart (approved):

PR #4405, opened 2021-07-18
chore: change rm -rf to rimraf for a better cross-platform experience

samarmohan:
## 📝 Description

> `rm -rf` does not work on Windows so users can not install dependencies amongst other stuff. `rimraf` on the other hand is cross-platform so it works everywhere and achieves the same goal.

> `rm -rf` is used to delete files, it does not work on Windows machines because `rm -rf` is a *nix only command..

## 🚀 New behavior

> Use `rimraf` to delete files

## 📝 Additional Information
N/A


with-heart (approved):

with-heart:
Thanks @samarmohan!

@allcontributors please add @samarmohan for code

PR #4409, opened 2021-07-19
fix/editable-cannot-click-when-empty

### LINKED ISSUE #4403: Editable component can't be clicked when empty
### Description

When an Editable is empty, I can't click on it to change the value

### Link to Reproduction

https://codesandbox.io/s/jovial-shadow-1ctmc

### Steps to reproduce

1. Click the Editable
2. Delete the text
3. Get out of focus (Press enter, click somewhere else, etc)

### Chakra UI Version

1.6.5

### Browser

Google Chrome 91

### Operating System

- [ ] macOS
- [ ] Windows
- [X] Linux

### Additional Information

You can also recreate it in the [documentation page](https://chakra-ui.com/docs/form/editable)

me doing that

anubra266:
Fixes #4403

**Problem:** An editable's Input takes focus for editing when the preview is clicked. But the preview is a simple span, so it's size is dependent on the length of the text in it. When the editable is empty and has no value, there's no text in the preview's span, so it's size is diminished, making it impossible to click on.

**Solution:** The approach I've taken is to set the `width` of the Preview to `full` so the `Preview` span takes the length of the root editable component, making it available for a click. The user can set a `minWidth` or `width` on the `Editable` component that define the boundaries of the  `Preview` component.


with-heart:
Nice dude! This seems like a solid fix!

with-heart (approved):
Nice dude! This seems like a solid fix!

with-heart:
I updated the changeset to add a bunch of additional context around the change

anubra266:
Thanks for your review @with-heart

yutaosawa81:
I encountered the bug just yesterday and tried the same way that adds `w="full"` to EditablePreview. It is now clickable but seems to be hard to click the EditablePreview with zero height and only padding.

image

I think we can fix it by settting `height: 30px` (replace it to actually EditableInput height) to Editable and `height: 100%` to EditablePreview. Could you fix this as well? Thank you

packages/editable/src/use-editable.ts
```diff
@@ -212,6 +212,7 @@ export function useEditable(props: UseEditableProps = {}) {
         ref: mergeRefs(ref, previewRef),
         children: isValueEmpty ? placeholder : value,
         hidden: isEditing,
+        w: "full",
```
segunadebayo, line 215: Thanks for your work here @anubra266.

I usually consider hooks like this to be agnostic of Chakra UI styling. In this case, I'll move the `w=full` to the `EditablePreview` component and pass it as a prop there.

segunadebayo (changes requested):

segunadebayo:
After taking a look at this, I think this can be a user-land fix. Ideally, the user wants to specify a `placeholder` when the input is empty to prevent this.

Adding `width:100%` is making an assumption that it'll always take up the full-width when it's in the preview mode, and that doesn't feel right to me. There's also a potential (visual) breaking change possibility.

I'm going to close this and recommend that users apply a `placeholder` instead.

PR #4432, opened 2021-07-22
Fix/toast centering

### LINKED ISSUE #4387: Toast is not centered when width exceeds 560px
### Description

When width exceeds 560px, I expect the toast it to still be centered. However, it seems to be aligned off-centered, regardless of width.

### Link to Reproduction

Easier to demonstrate in Chakra docs

### Steps to reproduce

1. Go to https://chakra-ui.com/docs/feedback/toast#custom-component, scroll down to `Custom Component`
2. Change the code to 
```
function CustomToastExample() {
  const toast = useToast({duration: 50000000})
  return (
    <Button
      onClick={() =>
        toast({
          position: "top",
          render: () => (
            <Box color="white" p={3} w="560px" bg="blue.500">
              Hello World
            </Box>
          ),
        })
      }
    >
      Show Toast
    </Button>
  )
}
```
3. Play around with the width that is originally set to `560px`
image

### Chakra UI Version

1.6.4

### Browser

Firefox, Chrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

seaerchin:
Closes #4387 

## 📝 Description

When users set a fixed-width on their custom toast, this causes their toast to not be centered. This is because we set a fixed `maxWidth` property on the wrapper component around the toast, which results in the invisible wrapper being centered correctly, but the visible custom component overflowing the parent (the wrapper), making it seem off-centered. 

A more detailed explanation of why this arises can be found [here](https://github.com/chakra-ui/chakra-ui/issues/4387#issuecomment-884419618)

Fixes the centering issue so that when the `maxWidth` property is set on the hook, the invisible wrapper has the correct `maxWidth`, allowing correct alignment of the visible component.

## 🚀 New behavior

Visible toast component is centered correctly when parent component has enough space to accommodate the component. This can now be done through using the exposed `maxWidth` property on the hook

No.

## 📝 Additional Information
Inbuild chakra toast does not have this issue as we set the `maxWidth` to be 100%, allowing for automatic wrapping. The choice was made to expose the property on the hook rather than changing the maxWidth on the `ReachAlert` itself so that users are able to customize for their use case (eg: accommodate for mobile view so that maxWidth should be ~400 px etc).

### DIFF
--- a/packages/toast/src/toast.tsx
+++ b/packages/toast/src/toast.tsx
@@ -65,6 +65,7 @@ export const Toast: React.FC<ToastProps> = (props) => {
     message,
     onCloseComplete,
     onRequestRemove,
+    maxWidth = 560, 
     requestClose = false,
     position = "bottom",
     duration = 5000,
@@ -118,7 +119,7 @@ export const Toast: React.FC<ToastProps> = (props) => {
         className="chakra-toast__inner"
         style={{
           pointerEvents: "auto",
-          maxWidth: 560,
+          maxWidth,
           minWidth: 300,
           margin: "0.5rem",
         }}
--- a/.changeset/itchy-mangos-hunt.md
+++ b/.changeset/itchy-mangos-hunt.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/toast": patch
+---
+
+Toasts have been updated to fix centering when the width of the toast exceeds
+560 pixels
--- a/packages/toast/src/toast-manager.tsx
+++ b/packages/toast/src/toast-manager.tsx
@@ -28,7 +28,7 @@ type State = { [K in ToastPosition]: ToastOptions[] }
 type CreateToastOptions = Partial<
   Pick<
     ToastOptions,
-    "status" | "duration" | "position" | "id" | "onCloseComplete"
+    "status" | "duration" | "position" | "id" | "onCloseComplete" | "maxWidth"
   >
 >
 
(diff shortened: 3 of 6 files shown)


packages/toast/src/toast.tsx
```diff
@@ -65,6 +65,7 @@ export const Toast: React.FC<ToastProps> = (props) => {
     message,
     onCloseComplete,
     onRequestRemove,
+    maxWidth = 560,
```
segunadebayo, line 68: This looks interesting to me. Instead of exposing just `maxWidth`, I would expose `style` so the user can handle any other edge-case unrelated to `maxWidth`.

What do you think?

segunadebayo (changes requested):

packages/toast/src/toast.tsx
seaerchin: my reason for not doing so was actually really simple - i didn't want to assume that `minWidth`/`margin` should be customizable. due to my limited experience with frontend code, i wasn't sure if there were other considerations beyond [modal styling](https://css-tricks.com/considerations-styling-modal/#dealing-with-width) and wanted to err on the safe side. 

that being said, if you foresee a usecase for allowing customisable `minWidth` and `margin`, i would be more than happy to make the appropriate changes!

seaerchin: @segunadebayo just double checking here - should i make the change and expose `style`?

segunadebayo: Yes please.

cschroeter:

seaerchin:
@cschroeter Apologies for this, got busy with schoolwork and completely missed out on this PR. i think #4922 aims to solve this issue as well and it seems to be active!

PR #4435, opened 2021-07-22
fix: radio styling

### LINKED ISSUE #4295: Checkbox & Radio style props are not applied consistently
### Description

When styling `Checkbox` & `Radio` components using the same style props, I expected a similar visual appearance but it's currently not the case.

### Link to Reproduction

https://codesandbox.io/s/eloquent-hoover-g4s8f

### Steps to reproduce

By adding the `background` or `borderColor` props to the `Checkbox` & `Radio` components.

### Chakra UI Version

1.6.4

### Browser

Chrome 91

### Operating System

- [x] macOS
- [x] Windows
- [x] Linux

### Additional Information

In the case of `Checkbox`, the style props are properly applied to the `container` part of the component. While with `Radio`, styles are applied to the `control` directly.

seaerchin:
Closes #4295

## 📝 Description
`Radio` styling was being applied to the checkbox rather than the container. This behaviour was different from `Checkbox` (the chakra checkbox component), where the styling was applied to the container. 

As the radio is an extremely small element and its colours can be adjusted through usage of the `colorScheme` prop, it makes more sense for the behaviour to be brought in line with `Checkbox`, where styling is applied to the container as a whole.

This was done through making the props being applied to the container take `styleProps` (which is a strict superset of the old `layoutProps`). This still preserves behaviour where additional `htmlProps` is being applied to the checkbox (of the `Radio`)

Styling for `Radio` is being applied onto the checkbox element rather than the container element, which is different from what `Checkbox` establishes and is also counter-intuitive. 

## 🚀 New behavior
This PR changes the internals of `Radio` so that `styleProps` are also applied rather than just `layoutProps`

No

## 📝 Additional Information
-

### DIFF
--- a/packages/radio/stories/radio.stories.tsx
+++ b/packages/radio/stories/radio.stories.tsx
@@ -1,6 +1,13 @@
 import * as React from "react"
 import { chakra } from "@chakra-ui/system"
-import { Stack, Wrap, SimpleGrid, Container, WrapItem } from "@chakra-ui/layout"
+import {
+  Stack,
+  Wrap,
+  SimpleGrid,
+  Container,
+  WrapItem,
+  Box,
+} from "@chakra-ui/layout"
 import {
   useRadio,
   Radio,
@@ -23,7 +30,17 @@ export const Readonly = () => (
     I'm a readonly radio
   </Radio>
 )
-
+export const StyledRadio = () => (
+  <Radio
+    background="gray.100"
+    border="1px"
+    borderColor="gray.300"
+    colorScheme="gray"
+    padding={4}
+  >
+    Radio
+  </Radio>
+)
 export const WithSizes = () => {
   const sizes = ["sm", "md", "lg"]
 
--- a/packages/radio/stories/radio.stories.tsx
+++ b/packages/radio/stories/radio.stories.tsx
@@ -1,13 +1,6 @@
 import * as React from "react"
 import { chakra } from "@chakra-ui/system"
-import {
-  Stack,
-  Wrap,
-  SimpleGrid,
-  Container,
-  WrapItem,
-  Box,
-} from "@chakra-ui/layout"
+import { Stack, Wrap, SimpleGrid, Container, WrapItem } from "@chakra-ui/layout"
 import {
   useRadio,
   Radio,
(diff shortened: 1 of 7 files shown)


seaerchin:
not too sure why the upstream doesn't have these commits but i can rebase and fix if needed

simontaisne (commented):
Thanks for the PR @seaerchin. The code makes sense to me. Anything missing before we can merge this one @segunadebayo?

seaerchin:
Hi @simontaisne, apologies for the absence as i got caught up with schoolwork!! Let me know if there's anything needed! i'll rebase off the current head ref!

simontaisne:
No need to apologize, thank you for your work! Lmk when it's good to review 👍

seaerchin:
@simontaisne should be good for review!

packages/radio/stories/radio.stories.tsx
```diff
@@ -23,7 +23,17 @@ export const Readonly = () => (
     I'm a readonly radio
   </Radio>
 )
-
+export const StyledRadio = () => (
```
simontaisne, line 26: Could we maybe have a similar story for `Checkbox`?

simontaisne (approved):
Looking good! Thanks for the fix @seaerchin 👍

Not sure about the changes on the members & sponsors though?

packages/radio/stories/radio.stories.tsx
seaerchin: added! thanks for the suggestion!

seaerchin:
@simontaisne unsure why that happened with the sponsors actually as i rebased and took the original repo's ref all the way; i rebased and dropped the erroneous commits so it should be good now!

simontaisne:
Hi @segunadebayo @TimKolberger @ljosberinn could we please get a review from a maintainer here :)

anubra266:
cc @TimKolberger

segunadebayo:
Thanks for working on this @seaerchin.

My sincere apologies for the delayed response. I consider this a breaking change given how long ago this PR/Issue was raised. We'll add it to the next major release.

seaerchin:
If it's breaking solely because of age, I'm happy to rebase off the current HEAD ref! Lmk if I need to do anything else beyond that @segunadebayo

PR #4437, opened 2021-07-22
chore: remove code that was added as a workaround for concurrent mode…

### LINKED ISSUE #4259: Code should be removed?
The code at

https://github.com/chakra-ui/chakra-ui/blob/db46840efea41d66c4d7c8f59af7d944dfb0cbd9/packages/radio/src/use-radio.ts#L211-L221

should probably be removed since https://github.com/facebook/react/issues/18591 was closed earlier this year.

cc @with-heart because you appear to be subscribed to that issue.

Toxiapo:
Closes #4259

## 📝 Description

This PR removes code that was originally added as a workaround for React concurrent mode issue. The issue has been addressed [here](https://github.com/facebook/react/issues/18591).

N/A
## 🚀 New behavior
N/A
No.

## 📝 Additional Information
N/A

### DIFF
--- a/packages/number-input/src/use-number-input.ts
+++ b/packages/number-input/src/use-number-input.ts
@@ -452,7 +452,7 @@ export function useNumberInput(props: UseNumberInputProps = {}) {
       autoCorrect: "off",
       onChange: callAllHandlers(props.onChange, onChange),
       onKeyDown: callAllHandlers(props.onKeyDown, onKeyDown),
-      onFocus: callAllHandlers(props.onFocus, onFocusProp, () => setFocused.on),
+      onFocus: callAllHandlers(props.onFocus, onFocusProp, setFocused.on),
       onBlur: callAllHandlers(props.onBlur, onBlurProp, onInputBlur),
     }),
     [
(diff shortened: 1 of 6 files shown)


packages/number-input/src/use-number-input.ts
```diff
@@ -453,9 +452,7 @@ export function useNumberInput(props: UseNumberInputProps = {}) {
       autoCorrect: "off",
       onChange: callAllHandlers(props.onChange, onChange),
       onKeyDown: callAllHandlers(props.onKeyDown, onKeyDown),
-      onFocus: callAllHandlers(props.onFocus, onFocusProp, () =>
-        scheduleMicrotask(setFocused.on),
-      ),
+      onFocus: callAllHandlers(props.onFocus, onFocusProp, () => setFocused.on),
```
TimKolberger, line 455: I think we need to pass the function definition here, not a new callback which returns the setter.

```suggestion
      onFocus: callAllHandlers(props.onFocus, onFocusProp, setFocused.on),
```

TimKolberger (changes requested):
Thanks for cleaning to codebase of those workarounds 💖 
I left one comment - the rest looks good to me!

packages/number-input/src/use-number-input.ts
Toxiapo: Oh yes, thank you for catching that!

TimKolberger (approved):
LGTM ✨

PR #4442, opened 2021-07-23
Fix/tooltip close

### LINKED ISSUE #3659: [1.4.1] Tooltips with new Identity and openDelay dont hide after first hover
# 🐛 Bug report

Tooltips with `openDelay` that get a new identity will open and not close after the **first time** the mouse hovers over the trigger, if the mouse leaves before the tooltip opens

## 💥 Steps to reproduce

1. Have a tooltip with `openDelay` > 0
2. Have it re-mount with a new identity
3. Hover over the tooltip-trigger
4. Leave the trigger before the tooltip opens
5. See that after the configured delay, the tooltip opens and stays open

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/hardcore-water-jvxlj?file=/src/App.js:387-390

## 🧐 Expected behavior

The tooltip should neither open, nor stay open if i leave the trigger before it opens

## 🧭 Possible Solution

Some debugging showed, that the mouseleave-event isn't triggered in this case. Maybe the ref/domnode provided here has the wrong value?
https://github.com/chakra-ui/chakra-ui/blob/1d85e472918346f1eb4ba7aed5291b9e7254e1f8/packages/tooltip/src/use-tooltip.ts#L155

## 🌍 System information

| Chakra UI        |      1.4.1      |
| Browser          |     Chrome  89 & Firefox 86     |
| Operating System |      macOS 11.2.3      |

## 📝 Additional information

This did not happen on chakra 1.3.x

Here is a video of the error happening in the provided code-sandbox:

https://user-images.githubusercontent.com/20770029/112314138-f6ec2e80-8ca8-11eb-8af0-c0f4d024e142.mp4

### Update
Using the automatically created Build from #4442 Fixes the Issue in the initial Codesandbox:
https://codesandbox.io/s/affectionate-mccarthy-bpmjf?file=/src/index.js

5ebastianMeier:
Closes #3659 

## 📝 Description

When tooltips are used with `openDelay` they behave inconsistent and sometimes don't disappear until you hover them a second time.

Currently the element inside the trigger ref is directly passed to `useEventListener` which is null on initial render.
As a result, the event listener for `mouseleave` is not being applied on first render.

## 🚀 New behavior

With the changes in this PR, the element inside the trigger ref is not directly passed, but passed inside a callback that still points to the trigger ref.

Tooltips should now close as intended.

No.

### DIFF
--- a/.changeset/grumpy-dodos-turn.md
+++ b/.changeset/grumpy-dodos-turn.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/tooltip": patch
+---
+
+Fix tooltips not closing when openDelay is set
--- a/packages/tooltip/src/use-tooltip.ts
+++ b/packages/tooltip/src/use-tooltip.ts
@@ -159,7 +159,7 @@ export function useTooltip(props: UseTooltipProps = {}) {
    * React regarding the onMouseLeave polyfill.
    * @see https://github.com/facebook/react/issues/11972
    */
-  useEventListener("mouseleave", closeWithDelay, ref.current)
+  useEventListener("mouseleave", closeWithDelay, () => ref.current)
 
   const getTriggerProps: PropGetter = React.useCallback(
     (props = {}, _ref = null) => {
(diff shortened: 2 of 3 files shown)


segunadebayo:
This is great! Thanks for fixing it @5ebastianMeier

segunadebayo (approved):

PR #4443, opened 2021-07-23
feat(editable): add editable textarea element

### LINKED ISSUE #1622: Request to add EditableTextArea to [Editable] component
**What problem does this feature solve?** A clear and concise description of the
To make multiline texts Editable, we need to have a EditableTextArea, the same way we have EditableInput. I think it would be make Editable even more useful.

**What does the proposed API look like?** Describe how you propose to solve the
Same as EditableInput!

heozeop:
Editable textarea element to handle multi-line input within editable context. 

Closes #1622 

## 📝 Description

>  Add EditableTextare element to handle multiline input with editable context

## 🚀 New behavior

> new line input is possible with editable context.

### Use union type for inputRef

  ```typescript
// editable/src/use-editable.ts
  ...
  const [value, setValue] = useControllableState({
    defaultValue: defaultValue || "",
    value: valueProp,
    onChange: onChangeProp,
  })
  ...
  const inputRef = useRef<HTMLInputElement | HTMLTextAreaElement>(null)
```
  Because the editable context has one controllable state as default, 
  use union type for input to maintain compatibility with the above condition.

### split test suite for editable textarea

  Despite of the similarity of `editable.test.tsx` and `editableTextarea.test.tsx`, which is the test file in `editable pakage`,
  I choose to split test suite to minimize lines of suite.

### DIFF
--- a/packages/editable/src/editable.tsx
+++ b/packages/editable/src/editable.tsx
@@ -181,7 +181,7 @@ export const EditableTextarea = forwardRef<EditableTextareaProps, "textarea">(
         __css={{
           outline: 0,
           ...commonStyles,
-          ...styles,
+          ...styles.textarea,
         }}
         className={_className}
       />
--- a/packages/editable/src/editable.tsx
+++ b/packages/editable/src/editable.tsx
@@ -8,6 +8,7 @@ import {
   useMultiStyleConfig,
   useStyles,
   HTMLChakraProps,
+  useStyleConfig,
 } from "@chakra-ui/system"
 import { cx, runIfFn, __DEV__ } from "@chakra-ui/utils"
 import { createContext, MaybeRenderProp } from "@chakra-ui/react-utils"
@@ -160,6 +161,41 @@ if (__DEV__) {
   EditableInput.displayName = "EditableInput"
 }
 
+export interface EditableTextareaProps extends HTMLChakraProps<"textarea"> {}
+
+/**
+ * EditableTextarea
+ *
+ * The textarea used in the `edit` mode
+ */
+export const EditableTextarea = forwardRef<EditableTextareaProps, "textarea">(
+  (props, ref) => {
+    const { getTextareaProps } = useEditableContext()
+    const styles = useStyleConfig("Textarea", props)
+
+    const textareaProps = getTextareaProps(
+      props,
+      ref,
+    ) as HTMLChakraProps<"textarea">
+    const _className = cx("chakra-editable__textarea", props.className)
+
+    return (
+      <chakra.textarea
+        {...textareaProps}
+        __css={{
+          outline: 0,
+          ...commonStyles,
+          ...styles,
+        }}
+        className={_className}
+      />
+    )
+  },
+)
+
+if (__DEV__) {
+  EditableTextarea.displayName = "EditableTextarea"
+}
 /**
  * React hook use to gain access to the editable state and actions.
  */
--- a/packages/editable/src/editable.tsx
+++ b/packages/editable/src/editable.tsx
@@ -140,7 +140,7 @@ export const EditableInput = forwardRef<EditableInputProps, "input">(
     const { getInputProps } = useEditableContext()
     const styles = useStyles()
 
-    const inputProps = getInputProps(props, ref) as HTMLChakraProps<"input">
+    const inputProps = getInputProps(props, ref)
     const _className = cx("chakra-editable__input", props.className)
 
     return (
@@ -173,10 +173,7 @@ export const EditableTextarea = forwardRef<EditableTextareaProps, "textarea">(
     const { getTextareaProps } = useEditableContext()
     const styles = useStyleConfig("Textarea", props)
 
-    const textareaProps = getTextareaProps(
-      props,
-      ref,
-    ) as HTMLChakraProps<"textarea">
+    const textareaProps = getTextareaProps(props, ref)
     const _className = cx("chakra-editable__textarea", props.className)
 
     return (
@@ -200,13 +197,8 @@ if (__DEV__) {
  * React hook use to gain access to the editable state and actions.
  */
 export function useEditableState() {
-  const {
-    isEditing,
-    onSubmit,
-    onCancel,
-    onEdit,
-    isDisabled,
-  } = useEditableContext()
+  const { isEditing, onSubmit, onCancel, onEdit, isDisabled } =
+    useEditableContext()
 
   return {
     isEditing,
--- a/packages/editable/src/editable.tsx
+++ b/packages/editable/src/editable.tsx
@@ -8,7 +8,6 @@ import {
   useMultiStyleConfig,
   useStyles,
   HTMLChakraProps,
-  useStyleConfig,
 } from "@chakra-ui/system"
 import { cx, runIfFn, __DEV__ } from "@chakra-ui/utils"
 import { createContext, MaybeRenderProp } from "@chakra-ui/react-utils"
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by smkhalsa** [comment]
This is a feature that we'd like to use. Is there anything holding this up?

File: packages/editable/src/use-editable.ts [hunk 1]
```diff
@@ -159,7 +159,7 @@ export function useEditable(props: UseEditableProps = {}) {
   }, [value, onSubmitProp])
 
   const onChange = useCallback(
-    (event: React.ChangeEvent<HTMLInputElement>) => {
+    (event: React.ChangeEvent<HTMLInputElement & HTMLTextAreaElement>) => {
```
takethefake, line 162: This could be misleading. The element type should somehow be inferred from the used `Element` and then either be `HTMLInputElement` or `HTMLTextAreaElement`

takethefake (changes requested):
Hej @heozeop thanks for your Contribution.

Your code looks good to me in general, but it feels like we're mixing things up in this by making `use-editable` work with `HTMLInputElement`  and `HTMLTextAreaElement` i would rather make `use-editable` generic and as well split up `editable` into two components `editable-input` and `editable-textarea` which use the generic version of `use-editable` with their respective `Element`-Type.

Splitting this component into two would also clarify the props in the documentation since the `Textarea`-props aren't displayed their although they could be used like this. 

This is just my opinion, before doing any refactoring I would wait for one of the core members of the chakra team to give some input. @segunadebayo @with-heart @TimKolberger

packages/editable/src/use-editable.ts
heozeop: First of all, Really thank you for review about subtle issue with nice manner.

Well, after a month of waiting other review,  I finally try to refactor this issue.
I wanna explain the reason I use union type as the solution on commit [f1bcf2c](https://github.com/chakra-ui/chakra-ui/pull/4443/commits/f1bcf2cd8d1765172665fda16af74a08c101bdc2)

1. On component `Editable`, there is no way to know which element will be used with the context. 
2. Because the props of editable were abstracted with just string input, so I think that it is reasonable not to reveal element type of change event.

cpvalente:
I was wondering if there is a timeline for reviewing the PR with the new changes? It would be fantastic to be able to use this

mSigson:
Hoping this feature will be merged and deployed soon - I would have immediate use for this on a current project for a client.

Thank you @heozeop and @takethefake for your time and efforts put towards implementing this change!

packages/editable/src/editable.tsx
```diff
@@ -160,6 +161,41 @@ if (__DEV__) {
...
+ */
+export const EditableTextarea = forwardRef<EditableTextareaProps, "textarea">(
+  (props, ref) => {
+    const { getTextareaProps } = useEditableContext()
+    const styles = useStyleConfig("Textarea", props)
+
+    const textareaProps = getTextareaProps(
+      props,
+      ref,
+    ) as HTMLChakraProps<"textarea">
```
TimKolberger, line 179: Can we move the type casting to the function `getTextareaProps`?
I think this should simplify the usage and the type safety of `getTextareaProps` and `getInputProps`.

e.g.

```ts
function getTextareaProps(): HTMLChakraProps<"textarea"> {}
function getInputProps(): HTMLChakraProps<"input"> {}
```

or if TS yells at you

```ts
function getTextareaProps() {
  return ({}) as HTMLChakraProps<"textarea">
}

Comment (heozeop) [reply packages/editable/src/editable.tsx:179]: Thank you for your good idea!
I think you idea is reasonable enough to make change on it.
I may work on it next week.
Thanks again!

Comment (heozeop) [reply packages/editable/src/editable.tsx]: I made changes on new commit!
I added default HTMLTypes on PropGetter!

**Review by TimKolberger — APPROVED** [review]

**Comment by heozeop** [comment]
@takethefake @with-heart @TimKolberger
Hi! guys!
I just resolved the error below
-  Type of 'height' attribute of html element is incompatible with chakra's 'height'

If you have time to look at it, please!

Thank you very much!

**Comment by heozeop** [comment]
# situation
- I started working on this PR about a year ago.
- To pass the checklist, I merged main branch.
- Due to error on main branch (like storybook button error), yarn lint always failed.
- So I decided to revert merge commit and force push
- The result is that incompatible with main error! (you can find [here](https://github.com/chakra-ui/chakra-ui/pull/4443/checks?check_run_id=5066226702))

# Please share your opinion
I think I have two options.
1. fix all errors on main branch.
2. create new pr that start from current main with same content.

Do you have any idea? Please share your idea! thanks!

**Comment by TimKolberger** [comment]
The lint errors on main are fixed for now 🎉 
I feel you - long running feature branches are a pain to work with.

My personal preference is to rebase feature branches onto main to prevent big merge commits.
This can be a pain as well if there are many merge conflicts and commits.

This PR adds a great feature which I'd love to merge when it's ready.

**Comment by heozeop** [comment]
Thank you very much @TimKolberger!

Thanks for your comment and commit, I'm really relieved.

As you recommended, I rebased to chakra-ui:main. 

Thank you again!

**Comment by macnibblet** [comment]
Any chance we might see this merged soon?

**Comment by heozeop** [comment]
Hello! @TimKolberger @with-heart! Can you merge this PR or review the change?

File: packages/editable/src/editable.tsx [hunk 3]
```diff
@@ -160,17 +161,44 @@ if (__DEV__) {
...
+
+/**
+ * EditableTextarea
+ *
+ * The textarea used in the `edit` mode
+ */
+export const EditableTextarea = forwardRef<EditableTextareaProps, "textarea">(
+  (props, ref) => {
+    const { getTextareaProps } = useEditableContext()
+    const styles = useStyleConfig("Textarea", props)
```
TimKolberger, line 174: This differs from the implementation of EditableInput. Editable is a MultiPartStyle component and we are using the key `input` there.
Without this change there would be no way to style the EditableTextarea independently of the Textarea component.

Kindly use the useStyles hook and access the `textArea` property. We may have to add a default styling in the theme.

TimKolberger (changes requested):
The code looks good! I added a comment regarding the theme styles, my apologies to not catching this earlier.

packages/editable/src/editable.tsx
heozeop: Thanks for kind review @TimKolberger !

TimKolberger:
No worries 😊
Have you tried building the repository before running the tests? The main branch just ran successfully on my machine with `yarn && yarn build && yarn test`

TimKolberger:
Hey @heozeop!

I really appreciate the work you put into this 💖 

Needless to say that this PR was open for far too long 😪 

Kindly allow me to bring this over the finish line:
1. Added the `textarea` element to the `anatomy` package to allow theme styles to be added
1. Reused the input styles in the theme for the `textarea` component, which seems like a good starting point for further customization
1.  Updated the changelog to address the target the users of Chakra UI, not the maintainers/contributors
1. and rebased onto the main branch

@all-contributors add @heozeop for code

TimKolberger (approved):

kk21:
I am not sure if something changed for using controlled value (React useState) of Editable.

`<Editable value={name} onChange={(value) => setName(value)}>`

Now setting a new value doesnt seem to propagate into the "inner" value of Editable.
Pressing "Esc" key in edit mode after making some temporary edit will also change the value when it shld not.

TimKolberger:
Hi @kk21,

Kindly open an issue with a reproduction in CodeSandbox. It is hard to verify an issue without actual code.

I don't see any changes regarding the value handling in this PR, it is still using the useControllableState hook - a reproduction would help a lot.

PR #4444, opened 2021-07-23
fix(skeleton): don't animate when skeleton was previously loaded

### LINKED ISSUE #4394: Skeleton animating unnecessarily
### Description

Changing the state of a component with a loaded skeleton causes the skeleton to flash.

### Link to Reproduction

https://codesandbox.io/s/upbeat-sea-u0pjv?file=/src/index.tsx

### Steps to reproduce

1. Click 'Toggle state update'
2. Text flashes

### Chakra UI Version

1.6.5

### Browser

Firefox 89

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Example:
Screen Recording 2021-07-15 at 4 03 31 PM

Only happens on first state change.

codetheweb:
Closes #4394.

## 📝 Description

https://github.com/chakra-ui/chakra-ui/pull/2775 fixed one bug but introduced another. 

If a component has a child Skeleton that's loaded (`<Skeleton isLoaded/>`) and the parent component re-renders, the Skeleton would animate because it was no longer the first render.

## 🚀 New behavior

It now needs to meet a stronger condition to animate when loaded. If `isLoaded=true` and it's the first render **or** the skeleton was already loaded in the previous render, no animation will happen.

No.

## 📝 Additional Information

Test was added. Was failing before my fix.

### DIFF
--- a/.changeset/curly-frogs-kick.md
+++ b/.changeset/curly-frogs-kick.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/skeleton": patch
+---
+
+fix(skeleton): don't animate when skeleton was previously loaded
--- a/packages/skeleton/src/skeleton.tsx
+++ b/packages/skeleton/src/skeleton.tsx
@@ -8,6 +8,7 @@ import {
   useStyleConfig,
   HTMLChakraProps,
 } from "@chakra-ui/system"
+import { usePrevious } from "@chakra-ui/hooks"
 import { cx, __DEV__ } from "@chakra-ui/utils"
 import * as React from "react"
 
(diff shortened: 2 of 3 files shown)


codetheweb:
Anything I need to do on my side to move this forward? Would be nice to get this patch in since I'm deploying an application soon that makes heavy use of Skeletons. :)

codetheweb:
For anyone else who needs this fix ASAP:

I published this branch as `@codetheweb/chakra-ui-skeleton-patch`. You can use it by modifying your `package.json`:

```json
{
  ...
  "resolutions": {
    "@chakra-ui/skeleton": "npm:@codetheweb/chakra-ui-skeleton-patch@*"
  }
  ...
}
```

PR #4458, opened 2021-07-25
fix(utils): use correct window for instanceof checks

### LINKED ISSUE #4457: Various issues when rendering into a different window
### Description

When rendering chakra components into a different window object (e.g. popup via window.open), I expect the components to behave correctly as if I rendered them into the current window. For example, `<Menu>` should close when clicking outside or selecting an item.

### Link to Reproduction

https://codesandbox.io/s/gifted-cartwright-km537?file=/src/index.js

### Steps to reproduce

1. Go to https://codesandbox.io/s/gifted-cartwright-km537?file=/src/index.js
2. Click on "Render in Popup"
3. In the popup, click the "Actions" menu. Make sure the original window is visible on your screen otherwise it won't open.
4. Try to close the menu by selecting an item or clicking outside

Note that the menu in the original window works as expected.

### Chakra UI Version

1.6.5

### Browser

Chrome 92

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

We're using chakra-ui in a multi-window electron app, which renders into different windows using React portals. Chakra portals, styling etc. worked out of the box 👏 but some components like Menu misbehave.

I already tracked it down to `instanceof Element` checks (and similar), which fail if the element belongs to a different window. Will open a PR soon.

By the way, in this use case, there's also an issue around `window.requestAnimationFrame` usage, which is why the Menu doesn't open when the original window is covered/minimized. Ideally, the timers would also be called on the `window` object that belongs to the respective component, but these calls may be deeper inside framer dependencies.

agundermann:
Closes #4457

## 📝 Description

Use constructors like `Element` for `instanceof` checks from the `window` object that the current node belongs to.

When rendered into a different window, some components currently misbehave (e.g. `<Menu>` doesn't close when clicking an item or clicking outside) because `instanceof Element` (and similar) checks fail, because the elements inherit the `Element` from a the other window object.

Also see #4457 .

## 🚀 New behavior

All `instanceof SomeContrusctor` checks now use `instanceof win.SomeConstructor` instead, where `win` is the window object that the current node or event belongs to.

No

## 📝 Additional Information

Didn't add a lot of tests, since JSDOM doesn't support `window.open`, and also since most of the code wasn't explicitly tested. Let me know if you need more tests or have any ideas how to better test this.

### DIFF
--- a/packages/utils/src/dom.ts
+++ b/packages/utils/src/dom.ts
@@ -1,13 +1,35 @@
 import { Booleanish, EventKeys } from "./types"
 
-export function getOwnerWindow(node?: HTMLElement | null): Window {
-  return node instanceof Element
+export function isElement(el: any): el is Element {
+  return (
+    el != null &&
+    typeof el == "object" &&
+    "nodeType" in el &&
+    el.nodeType === Node.ELEMENT_NODE
+  )
+}
+
+export function isHTMLElement(el: any): el is HTMLElement {
+  if (!isElement(el)) {
+    return false
+  }
+
+  const win = el.ownerDocument.defaultView ?? window
+  return el instanceof win.HTMLElement
+}
+
+export function getOwnerWindow(node?: Element | null): typeof globalThis {
+  return isElement(node)
     ? getOwnerDocument(node)?.defaultView ?? window
     : window
 }
 
-export function getOwnerDocument(node?: HTMLElement | null): Document {
-  return node instanceof Element ? node.ownerDocument ?? document : document
+export function getOwnerDocument(node?: Element | null): Document {
+  return isElement(node) ? node.ownerDocument ?? document : document
+}
+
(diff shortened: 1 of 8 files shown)


packages/utils/src/dom.ts
```diff
@@ -1,13 +1,35 @@
...
     : window
 }
 
-export function getOwnerDocument(node?: HTMLElement | null): Document {
-  return node instanceof Element ? node.ownerDocument ?? document : document
+export function getOwnerDocument(node?: Element | null): Document {
+  return isElement(node) ? node.ownerDocument ?? document : document
+}
+
+export function getEventWindow(event: Event): typeof globalThis {
```
agundermann, line 31: Couldn't use `Window` here, because some properties like `PointerEvent` are missing.

cschroeter:
LGTM

TimKolberger (approved):

PR #4473, opened 2021-07-27
feat(number-input): custom format

### LINKED ISSUE #4474: [NumberInput] Custom format
### Description

I'd like to propose a custom number format feature for the `NumberInput` that meets advanced localization requirements.

Currently there are ways to customize the format, but these are not flexible enough to use a format that differs from the Javascript floating point representation.

While trying to localize the `NumberInput` I noticed that there's still a hard-coded burden that is not easy to overcome.

In my example the component is used in a german-speaking environment and the localization requires floating point numbers to be represented with a comma decimal separator.

Although this is the use case I'm dealing with right now, I'm aware there might be more use cases in other locales I didn't even think of yet.

Currently there are two ways that looked like they could help me to achieve my goal.

#### Using the approach described in the docs
First I tried this approach: https://chakra-ui.com/docs/form/number-input#formatting-and-parsing-the-value

However, a `console.log` revealed that any comma input was discarded even before the `onChange` event was emitted.

#### Using the `pattern` prop
Before I had a look at the source code I made the assumption that the regex defined here might also be used internally to actually parse the value.

The standard pattern is `"[0-9]*(.[0-9]+)?"` so I just replaced the `.` with a `,` and tried it with this pattern: `"[0-9]*(,[0-9]+)?"`.
But when I pressed comma inside the input, nothing happened.

#### Why is a comma input discarded?

When the `onKeyDown`-Event of the `NumberInput` is handled, the first step is to validate whether or not the key input is considered valid - if not, it is prevented.

```
/**
* only allow valid numeric keys
*/
if (!isValidNumericKeyboardEvent(event)) {
    event.preventDefault()
}
```

In detail this checks the following:

```
const FLOATING_POINT_REGEX = /^[Ee0-9+\-.]$/

/**
 * Determine if a character is a DOM floating point character
 * @see https://www.w3.org/TR/2012/WD-html-markup-20120329/datatypes.html#common.data.float
 */
export function isFloatingPointNumericCharacter(character: string) {
  return FLOATING_POINT_REGEX.test(character)
}

/**
 * Determine if the event is a valid numeric keyboard event.
 * We use this so we can prevent non-number characters in the input
 */
export function isValidNumericKeyboardEvent(event: React.KeyboardEvent) {
  if (event.key == null) return true

  const isModifierKey = event.ctrlKey || event.altKey || event.metaKey

  if (isModifierKey) {
    return true
  }

  const isSingleCharacterKey = event.key.length === 1

  if (!isSingleCharacterKey) {
    return true
  }

  return isFloatingPointNumericCharacter(event.key)
}
```

As you see above, every single character is currently validated against the hard-coded floating point regex.

This regex is certainly useful as it ensures that only characters that `parseFloat` can parse are entered.

On the other hand every customization you try to do is limited by this as well.

#### NumberInput implementation details

The `useNumberInput` hook uses another package called `counter` which encapsulates the whole number state management.

Internally `useCounter` uses this parse function, which removes any unparsable content from the value:
```
function parse(value: StringOrNumber) {
  return parseFloat(value.toString().replace(/[^\w.-]+/g, ""))
}
```
Without this replacement `parseFloat` would only parse the part of the value that came before any invalid character.

E.g.: `parseFloat('1.23example45')` would only return `1.23` but with the additional replacement is returns `1.2345`.

The regex doesn't cover the edge cases of `parseFloat` precisely, but I guess in this case it is intended to increase the convenience somehow.

So as a takeaway any customization we apply on the value representation should be converted to the regular floating point representation **before** passing it to `useCounter`.

### Proposed Solution or API

An optional new prop called `charPattern` can be used to define a regular expression that the key of the `onKeyDown`-Event is matched against.

You can look at it as a complement to the `pattern` property, but instead of defining a pattern that the string as a whole is matched against, we can specify one that each entered character is matched against.

This gives developers an opt-in to custom formats that is suitable to enable the comma input scenario described above.

Unfortunately this also means that characters can be entered that `parseFloat` does not parse.

To overcome this problem we need a bidirectional conversion from the custom format to the format `parseFloat` understands.

The property `parseValue` takes care of the conversion from custom format to `parseFloat` format.

The property `formatValue` takes care of converting the `parseFloat` format to the custom format.

Both properties are required if the `charPattern` prop is used.

This bidirectional conversion might also be a nice opportunity to create an explicit API surface for the use case described in https://chakra-ui.com/docs/form/number-input#formatting-and-parsing-the-value.

In TypeScript we can define an optional extension to express that as soon as you use one of the three new props, you also need the other two:
```
type UnusedCustomFormat = {
  charPattern?: never;
  parseValue?: never;
  formatValue?: never;
}

type UsedCustomFormat = {
  charPattern: RegExp;
  parseValue: (value: string) => string;
  formatValue: (value: string) => string;
}

type OptionalCustomFormat = UsedCustomFormat | UnusedCustomFormat;

// Valid
const test1: OptionalCustomFormat = {
};
// Error: Type is missing the following properties from type `UsedCustomFormat`: parseValue, formatValue
const test2: OptionalCustomFormat = {
  charPattern: /^[Ee0-9+\-.]$/
};
```

Edit: just noticed that declaring this as an optional `type` based extension seems to increase the complexity and might lead to TypeScript errors. I'll try to come up with a better solution later, so please ignore this aspect for the moment.

I opened up a Draft PR with an example implementation of this approach here:
https://github.com/chakra-ui/chakra-ui/pull/4473

### Alternatives

_No response_

#### Decimal separators
Here's an article that touches on several aspects of the topic and gives a nice overview of format usage across countries worldwide:
https://en.wikipedia.org/wiki/Decimal_separator
#### Previous discussions on the topic
https://github.com/chakra-ui/chakra-ui/issues/2461
https://github.com/chakra-ui/chakra-ui/issues/438

5ebastianMeier:
Closes #4474

## 📝 Description

> Add a brief description

## 🚀 New behavior


nikolicstjepan:
Hi, thanks for getting this done, and thanks to all contributors for making Chakra awesome. I was wondering when this might be merged? @5ebastianMeier

5ebastianMeier:
@nikolicstjepan I'm already using this in our codebase for some weeks now.
Theoretically this could be merged right away.
The way the props for this feature are represented in TypeScript might deserve another thought (see https://github.com/chakra-ui/chakra-ui/issues/4474).

segunadebayo:
Thanks for this PR @5ebastianMeier. I'm consolidating all number-input related issues here #4982.

5ebastianMeier:
@segunadebayo would have been nice to squash merge it into the new branch so that it still counts as a contribution, but I'm glad to see it getting merged anyway.

PR #4511, opened 2021-08-03
fix: type definition for apply style prop

jrolfs:
## 📝 Description

Support for responsive values in the `apply` prop was added [here](https://github.com/chakra-ui/chakra-ui/pull/3245), however, overlapping type definitions prevent this from being accurately reflected in TypeScript.

```tsx
// Type error, expects `string` for `apply`;
<Text apply={{ sm: 'styles.h3', lg: 'styles.h4' }}>
```

## 🚀 New behavior

```tsx
// No type error, expects `ResponsiveValue<string>`
<Text apply={{ sm: 'styles.h3', lg: 'styles.h4' }}>
```

No.

### DIFF
--- a/.changeset/ninety-suits-camp.md
+++ b/.changeset/ninety-suits-camp.md
@@ -1,0 +1,14 @@
+---
+"@chakra-ui/system": patch
+---
+
+Fix type definitions for `apply` prop.
+
+The `apply` prop supports responsive styles:
+```tsx
+// Before: type error, expects `string` for `apply`
+<Text apply={{ sm: 'styles.h3', lg: 'styles.h4' }}>
+
+// After: no type error, expects `ResponsiveValue<string>` for `apply`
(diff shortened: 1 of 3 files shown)


jrolfs:
Yeah, so this came out of a bit of a @jrolfs + Chakra odyssey that started with #1028. I came out with a workable interim solution for responsive `size` in my use case, thanks to `apply`. That said, I think I'm pretty well-positioned to consider implementing responsive `size`, `variant` etc, but I'd be curious whether y'all would consider a PR before the next major. It might be a little tricky to make it non-breaking with all of the style object merging precedence but I'm willing to take a crack at it.

TimKolberger (approved):

segunadebayo:
Thanks for this @jrolfs.

Regarding the responsive variant and size, we have in mind to add those to the next major. Feel free to take a tab at it in a draft PR. I'm curious to see how you approach it.

PR #4526, opened 2021-08-05
feat(modal): replace react-remove-scroll-bar with tua-body-scroll-lock

takethefake:
Closes # 

## 📝 Description

> Add a brief description

Fiddled around with `tua-body-scroll-lock` to check if it is possible to migrate from `react-remove-scroll-bar` to it.

Currently added a flag to `modal.ts` which decides which lib is used for scroll locking. 

Added `react-remove-scroll-bar` to keep the possibility to remove the scroll bar if we want to change to `tua-body-scroll-lock`

Didn't found an easy way to reimplement `allowPinchZoom` but didn't even get how a change in this property changes the behavior.

### Current drawback why we shouldn't use `tua-body-scroll-lock`:

- Added a Story `BlockScrollOnMount` to the modal stories. When you open the modal and scroll down to the bottom with the mouse wheel without moving your mouse and then try to scroll up again the modal content does not change. Scrolling the modal content only works after moving your mouse again. 

Didn't get why this happens.

This behavior doesn't occur with `react-remove-scroll-bar`

## 🚀 New behavior

Yes for now, since there ain't an alternative for `allowPinchZoom` yet in here.

### DIFF
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -68,7 +68,9 @@
     "@chakra-ui/transition": "1.3.3",
     "@chakra-ui/utils": "1.8.1",
     "aria-hidden": "^1.1.1",
-    "react-remove-scroll": "2.4.1"
+    "react-remove-scroll": "2.4.1",
+    "react-remove-scroll-bar": "^2.2.0",
+    "tua-body-scroll-lock": "^1.2.1"
   },
   "devDependencies": {
     "@chakra-ui/system": "1.7.1",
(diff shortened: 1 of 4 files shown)


segunadebayo:
This is an interesting investigation, Daniel. Thanks for taking the time to create this and share the drawback.

segunadebayo:
Given that this was an experiment, I'm going to close this one out.

Maybe we can consider using another package in the future or build our own.

PR #4570, opened 2021-08-14
refactor: Make theme packages framework agnostic

### LINKED ISSUE #4530: Make @chakra-ui/theme framework agnostic
### Description

The **@chakra-ui/theme** provides the default theme for theming API. 
Most of the codes are not related to the UI framework. They can be used anywhere like Vue, Angular... 

For now, it requires `@chakra-ui/system` which is tightened to React, for:
- Types definition in `theme.type.ts` 
- A re-exported utils `keyframes` in `skeleton.ts`

### Proposed Solution or API

- For Type definitions, we can introduce new packages for types only, such as `@chakra-ui/system.types` to support the Typescript version, and this package can be framework agnostic as well (which may benefit downstream typescript packages).
- The `keyframes`:
   -  Can be fixed by import it from `@emotion/css`, this may not be ideal for  **React** version, which is recommended to use `@emotion/react`
   -  Or we **try to import it from both packages** `@emotion/css` or `@emotion/react` and raise a meaningful error on the missing case (most of the time, one of these packages has already been required by the component library).
   - Then we can safely re-export it in `@chakra/theme-tools` packages

### Additional Information

Although, this will require efforts to export a shared `system.types` packages and somehow, quite unintuitive for try to import `keyframes` from both emotion packages, but make `theme` package framework-agnostic will speed up the development of the downstream package (Vue, Angular version), unify the theming API and enlarge the Chakra ecosystem, which is really worth doing.

simpletrontdip:
Closes #4530 

## 📝 Description
Theme and theme-tools logic is not related to UI framework, so they should be agnostic, this provides a huge benefit for the downstream packages (e.g. Vue, Angular). We don't have to replicate it anymore
The **theme** and **theme-tools** packages are not framework-agnostic because they are depending on **system**. There 2 reasons for it:
- Type definitions import
- Emotion keyframes utils

This PR tries to address that, so these packages can be shared and benefit our ecosystem.
The dependency graph will be more concise and intuitive, all packages should depend on **styled-system** and **theme** packages have no relation with **system** package.

- ColorMode, ColorModeOptions are defined in `color-mode` package
- ThemingProps is inside `system` itself
- **keyframes** is loaded directly from `emotion/react`

## 🚀 New behavior
- ColorMode, ColorModeOptions, ThemingProps are now in **styled-system**, can be shared across versions
- ~~**keyframes** will be loaded either from `@emotion/react` (optimized for React) or `@emotion/css` which is also a framework-agnostic(this is quite hacky 🗡️🗡️ )~~ or we have to inject it from StyleFn (quite complex and break things). Please let me know if we can improve this 👯 👯  
- ***Update***: Now we add a new type `CssSerializer` (defined in `styled-system`) as placeholder in `ThemeComponentProps` and inject `emotion` as a serializer for factory function `_styled` and style config util `useStyleConfig` in `system`. This way, we can validate _serializer injection in the StyleFn, enforce the dependency.
```ts
// styled-system/system.types.ts
export interface CssSerializer {
  css: (...args: any) => Dict
  keyframes: (...args: any) => String
  injectGlobal?: (...args: any) => any
}


// theme/theme.types.ts
export interface ThemeComponentProps<Theme extends ChakraTheme = ChakraTheme>
  extends Dict,
    Omit<ThemingProps, "styleConfig"> {
  colorMode: ColorMode
  theme: Theme
  _serializer: CssSerializer
}

// inject in factory function and useStyleConfig
import * as emotion from "@emotion/react"
const mergedProps = mergeWith(
  {
    theme,
    colorMode,
    _serializer: emotion as CssSerializer,
  },
  ...
)
const finalBaseStyle = runIfFn(baseStyle, mergedProps)
  
// And use it in style function
const baseStyle = (props: ThemeComponentProps) => {
  const { _serializer,  } = props
  const fade = _serializer.keyframes(...)
  return {
    // ...
    animation: `${speed}s linear infinite alternate ${fade}`,
  }
}  
```

No

## 📝 Additional Information

- For version number update, I'm not so familiar with it, feel free to update as needed ^^
- To fix the race condition while building packages, I have added some commands to the root package
   - `build:types`, `build:cjs`, `build:esm` are now in the root, they execute appropriate child packages command in `lerna`'s order 
   - `postbuild`, `prebuild` has been pulled up to the root, similarly.


packages/theme/src/components/skeleton.ts
```diff
@@ -1,6 +1,20 @@
...
+} catch {
+  // eslint-disable-next-line global-require
+  emotion = require('@emotion/css')
+}
+
+if(!emotion) {
+  throw Error('Either @emotion/react (for React) or @emotion/css (for others) must be installed')
+}
+
+const { keyframes } = emotion
```
segunadebayo, line 16: I might need to think about this a bit more.

An alternative approach is to put the keyframe definition in the theme as an object and translate that to emotion's keyframe at the system or styled-system level.

simpletrontdip: @segunadebayo  Or should we introduce a generalized version of it? 
Something like `CssSerializer`, some theme components may need `keyframes`, or `css`, `injectGlobal`
```ts
interface CssSerializer {
  css: (...args: any) => Dict,
  keyframes: (...args: any) => String,
  injectGlobal: (...args: any) => any
}

type StyleResolverProps = SystemStyleObject & {
  __css?: SystemStyleObject
  sx?: SystemStyleObject
  theme: any
  serializer: CssSerializer
  css?: CSSObject
}
```
And inject that serializer to `props` at the `system` level?
```ts
import emotion from "@emotion/react"
...
export const toCSSObject: GetStyleObject = ({ baseStyle }) => (props) => {
  const { theme, css: cssProp, __css, sx, ...rest } = props
  const styleProps = objectFilter(rest, (_, prop) => isStyleProp(prop))
  const finalBaseStyle = runIfFn(baseStyle, {
    serializer: emotion,
    ...props,
  })
  // ... other logics
}  
```
And use them in style function:
```ts
// skeleton.ts
const baseStyle = (props: Record<string, any>) => {
  // ...
  const {
    startColor = defaultStartColor,
    endColor = defaultEndColor,
    speed,
    theme,
    serializer,
  } = props
  // ...
  const fadeAnimation = serializer.keyframes({
    from: { borderColor: startColor, background: startColor },
    to: { borderColor: endColor, background: endColor },
  })
  
  return {
    opacity: 0.7,
    borderRadius: "2px",
    borderColor: start,
    background: end,
    animation: `${speed}s linear infinite alternate ${fadeAnimation}`,
  }
}  
```
Will this work, do I miss something?

segunadebayo: This looks interesting. I'll take a look at the code to see how to make this work.

simpletrontdip: Actually, I've updated this PR to the proposed approach, you can check it now @segunadebayo

cschroeter:

PR #4594, opened 2021-08-19
(proposal) refactor: move ChakraProvider to its own package

feychenie:
## 📝 Description

**TL;DR:** *no change to `@chakra-ui/react`, just allow to use `ChackraProvider` without having to install the whole lib.*

Chakra's core "system" is an amazing evolution of the now abandoned [styled-system](https://github.com/styled-system/styled-system), and it would be awesome to be able to use it to build completely custom component libraries without having to install the whole original component set and theme. 

System things and helpers are already available as separate packages, with the notable exception of the crucially needed `ChakraProvider`: it is only available through `@chakra-ui/react`, with the whole lib and theme as dependancies.

Tree shaking and other build-level things would of course remove the burden from the final bundle, but this is more about not having to think about it. (and maybe opening to more possible usages 😉 )

Here, only `ChakraProvider` has been moved to a separate package, as -in my current understanding- all theme helpers are for overriding Chakra's existing theme, but it is not included in the standalone provider anymore. This is maybe shortsighted?

Left to do/figure:
- [ ] Readme
- [ ] Typescript: the `theme` prop should not be optional anymore in the new package, but still be optional when importing `ChakraProvider` from `@chakra-ui/react`.
- [ ] Typescript: `theme` type should ensure the presence of the `config` key with color mode and css variables prefix

You can only get the `ChakraProvider` via `@chakra-ui/react` but this would also include the default theme and the whole set of components.

## 🚀 New behavior

You can now _also_ import the `ChakraProvider` via `@chakra-ui/provider`, but you will _have to provide the **whole** theme_ for your components.

There should be no breaking change for current library users.


packages/provider/src/chakra-provider.tsx
```diff
@@ -0,0 +1,99 @@
...
+   * based on where `ChakraProvider` is rendered.
+   */
+  environment?: EnvironmentProviderProps["environment"]
+}
+
+/**
+ * The global provider that must be added to make all Chakra components
+ * work correctly
+ */
+export const ChakraProvider: React.FC<ChakraProviderProps> = (props) => {
```
feychenie, line 64: Changed the type to `React.FC` her in order to allow setting `defaultProps` in `@chakra-ui/react`

itsmelion:
I like it, and would go beyond.
I tried to publish a reusable component built with chakra, and unfortunately I'm obligated to wrap it in ChakraProvider, and sadly it sets a bunch of things on global css. body tags etc making it harder to use chakra to make publishable components, without colliding with some other pre-existing css, or even chakra itself.

I would segregate the concerns of Context(functionality/state) and the Theming, allowing to also opt-out global css.

packages/react/src/chakra-provider.tsx
```diff
@@ -1,100 +1,8 @@
...
 import defaultTheme from "@chakra-ui/theme"
-import { Dict } from "@chakra-ui/utils"
-import {
-  EnvironmentProvider,
-  EnvironmentProviderProps,
-} from "@chakra-ui/react-env"
-import * as React from "react"
-import { IdProvider } from "@chakra-ui/hooks"
 
-export interface ChakraProviderProps
```
itsmelion, line 19: I just experienced a 'breaking change' cuz i cant import `ChakraProviderProps` from `@chakra-ui/react` anymore.
i had to pull from the  `@chakra-ui/provider` pkg.
If that's intended, or desired I'm ok with it, else, it needs to be re-exported here.

PR #4603, opened 2021-08-21
docs: update tabs doc by passing the ref to <StyledTab>.

yukiyohure:
Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior

## 📝 Additional Information
@segunadebayo  The explanation of the section says "Because TabList needs to know the order of the children, we use cloneElement to pass state internally.", but the ref is not passed to any component in the example <CustomTab>.

### DIFF
--- a/website/pages/docs/disclosure/tabs.mdx
+++ b/website/pages/docs/disclosure/tabs.mdx
@@ -410,7 +410,7 @@ function CustomTabs() {
     const styles = useStyles()
 
     return (
-      <StyledTab __css={styles.tab} {...tabProps}>
+      <StyledTab ref={ref} __css={styles.tab} {...tabProps}>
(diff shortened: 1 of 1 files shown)


website/pages/docs/disclosure/tabs.mdx
```diff
@@ -410,7 +410,7 @@ function CustomTabs() {
     const styles = useStyles()
 
     return (
-      <StyledTab __css={styles.tab} {...tabProps}>
+      <StyledTab ref={ref} __css={styles.tab} {...tabProps}>
```
segunadebayo, line 413: The `ref` here should be passed to the `useTabs` hook `useTabs({...props, ref})` because we do some merging of all user-defined and internal refs.

PR #4605, opened 2021-08-22
fix: add correct variant type to UseToastOptions

LarsOlt:
## 📝 Description

Corrected the TS type for the useToast hook options according to the [docs](https://chakra-ui.com/docs/feedback/toast#variants)

`UseToastOptions.variant?: string | undefined`

## 🚀 New behavior

`UseToastOptions.variant?: "subtle" | "solid" | "left-accent" | "top-accent" | undefined`

No

### DIFF
--- a/packages/toast/src/use-toast.tsx
+++ b/packages/toast/src/use-toast.tsx
@@ -53,7 +53,7 @@ export interface UseToastOptions {
   /**
    * The alert component `variant` to use
    */
-  variant?: string
+  variant?: "subtle"| "solid"| "left-accent" | "top-accent"
   /**
    * The status of the toast.
    */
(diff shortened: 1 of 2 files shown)


packages/toast/src/use-toast.tsx
```diff
@@ -53,7 +53,7 @@ export interface UseToastOptions {
   /**
    * The alert component `variant` to use
    */
-  variant?: string
+  variant?: "subtle"| "solid"| "left-accent" | "top-accent"
```
segunadebayo, line 56: This is quite strict and assumes that the user doesn't customize the theme to include more variants.

```suggestion
  variant?: "subtle"| "solid"| "left-accent" | "top-accent" | (string & {})
```

Use this walkaround to allow custom variant values.

PR #4606, opened 2021-08-22
perf: remove useless useMemo

feychenie:
## 📝 Description

The useMemo in useStyleConfig can't return any memoized result because of a dependancy on the locally created object `mergedProps`, so I figured it's better to just remove it as it is misleading (and can actually _cost_ a tiny bit of perf).

No

### DIFF
--- a/packages/system/src/use-style-config.ts
+++ b/packages/system/src/use-style-config.ts
@@ -7,7 +7,7 @@ import {
   omit,
   Dict,
 } from "@chakra-ui/utils"
-import { useMemo, useRef } from "react"
+import { useRef } from "react"
 import isEqual from "react-fast-compare"
 import { useChakra } from "./hooks"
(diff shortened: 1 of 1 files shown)


packages/system/src/use-style-config.ts
```diff
@@ -43,37 +43,35 @@ export function useStyleConfig(themeKey: any, props: any = {}, opts: any = {}) {
...
-      if (!isStyleEqual) {
-        stylesRef.current = styles
-      }
+    if (!isStyleEqual) {
+      stylesRef.current = styles
     }
+  }
 
-    return stylesRef.current
-  }, [styleConfig, mergedProps, opts?.isMultiPart])
```
feychenie, line 76: `mergedProps` here is always a new object, preventing memoization

segunadebayo: Good catch!

PR #4630, opened 2021-08-27
fix(list): ensure margin start is themeable

South-Paw:
## 📝 Description

Sets list's margin start in the component's theme rather than on the component so that the margin starts can changed via theme overrides.

List margin start is not currently themeable as how it is applied on [L88](https://github.com/chakra-ui/chakra-ui/blob/main/packages/layout/src/list.tsx#L88) and [L99](https://github.com/chakra-ui/chakra-ui/blob/main/packages/layout/src/list.tsx#L99) means that on [L72-73](https://github.com/chakra-ui/chakra-ui/blob/main/packages/layout/src/list.tsx#L72-L73) it will override any component theme styles.

## 🚀 New behavior

Just moves the margin start value to the theme, no change except it should now be themeable as intended 😄

No, it shouldn't be breaking.

## 📝 Additional Information

I can't get the project installed due to node-gyp errors so I haven't been able to do all the steps in contributing properly sorry 😢


South-Paw:
resolved conflicts with #4611

takethefake:
Hej @South-Paw thanks for your contribution! 

Could you add a changeset to your commit which describes the changes you've made? In this case, it should resolve in `patch` change for `List` run `yarn changeset` to get this done. See https://github.com/atlassian/changesets for further information.

South-Paw:
Hi @takethefake,

I had trouble installing the project on my local due to some node-gyp errors which stopped me from running the changeset command. I'll have another go at getting it working tomorrow afternoon

PR #4224, opened 2021-06-13
(fix) useSystemColorMode behavior between system and app choices

LFerronato:
remove initial call of "listener()"
and change addListener (deprecated) to addEventListener
and change removeListener (deprecated) to removeEventListener
[I couldn't make all the necessary changes]

Closes # 

if use <code>useSystemColorMode=true</code> and change color mode inside de application, when refresh the application, the color mode will be reset to the system option.

## ⛳️ Current behavior (updates)

> Please describe the current behavior that you are modifying
if:
* system: light
* app: changed to dark
* if refresh page, will be light.

## 🚀 New behavior

> Please describe the behavior or changes this PR adds
preference is always the last option (either in the system or in the app)

## 💣 Is this a breaking change (Yes/No):
Actually, I do not know

## 📝 Additional Information

I recorded a video to demonstrate this, and my solution.
https://www.youtube.com/watch?v=tjVQwK97hGk
But I couldn't implement everything directly in the chakra code

I hope to help

PR #4634, opened 2021-08-28
Better handling of `system` as initial color mode

SkySails:
Sort of similar to #4224, but with a somewhat different approach

## 📝 Description

I wanted to add the possibility to use the system preferred color scheme as a value for
`initialColorMode`, while still respecting a user's previous choice.

When the `system` value is used for the `<ColorModeScript />`s `initialColorMode` prop, the color mode will _always_ initially be set to the color mode that matches the system preference. 

However, this may not be what a user expects after manually selecting a color mode through the UI. A user is more likely to expect the color mode to remain in the configuration that he/she left it in.

Additionally, while I am not 100% sure if intended or not (or if I missed something), there seemed to be a mismatch of the types between the `ColorModeProvider` and `ColorModeScript`  components,  where the latter would allow the `system` option while the first did not (at least not explicitly).

## 🚀 New behavior

As long as a user does not manually select a color mode through a website interaction, the color mode will change when the system preference changes (in real-time). When the page is reloaded, it will default to the system preference just like before.

However, if the color mode is manipulated during the visit, localStorage would be updated and upon the next visit/page reload, the color mode would no longer default to the system preference, but instead use the previously selected mode.

In order to revert back to using the system preference as default, one just needs to remove the localStorage entry.

No.

## 📝 Additional Information

While reading #4224, and before trying my own approach, I thought a lot about whether this change is wanted by a majority (and thus is motivated as is), or if the change should be mapped to a standalone configuration property.

As it currently stands, the changes that this PR brings to the `system` option would make it _partially_ incompatible with the `useSystemColorMode` option, as setting it to `true` would **not** cause the website to **load** using the system's preference if the color mode was manually updated before. It would still cause the application to listen for and adapt to changes in the preference, though.

I thought about the `useSystemColorMode` option as a directive for the application to "listen for updates to the color preference and update the color mode when it is changed", making the `initialColorMode` option feel like a better fit for this implementation.

With all this said, I'd be more than happy to discuss this new usage of the `initialColorMode` option. There might of course be factors that I failed to take into account, and perhaps this could be made more intuitive with another option in the theme config.

### DIFF
--- a/packages/color-mode/src/color-mode-provider.tsx
+++ b/packages/color-mode/src/color-mode-provider.tsx
@@ -9,10 +9,11 @@ import {
 } from "./color-mode.utils"
 import { localStorageManager, StorageManager } from "./storage-manager"
 
-export type { ColorMode }
+type ConfigColorMode = ColorMode | "system" | undefined
+export type { ColorMode, ConfigColorMode }
 
 export interface ColorModeOptions {
-  initialColorMode?: ColorMode
+  initialColorMode?: ConfigColorMode
   useSystemColorMode?: boolean
 }
 
@@ -59,6 +60,8 @@ export function ColorModeProvider(props: ColorModeProviderProps) {
     colorModeManager = localStorageManager,
   } = props
 
+  const defaultColorMode = initialColorMode === "dark" ? "dark" : "light"
+
   /**
    * Only attempt to retrieve if we're on the server. Else this will result
    * in a hydration mismatch warning and partially invalid visuals.
@@ -67,8 +70,8 @@ export function ColorModeProvider(props: ColorModeProviderProps) {
    */
   const [colorMode, rawSetColorMode] = React.useState<ColorMode | undefined>(
     colorModeManager.type === "cookie"
-      ? colorModeManager.get(initialColorMode)
-      : initialColorMode,
+      ? colorModeManager.get(defaultColorMode)
+      : defaultColorMode,
   )
 
   React.useEffect(() => {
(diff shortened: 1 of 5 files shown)


crubier:
This PR is exactly how I expect Color mode to behave. Right now without this PR, the "system" Color mode just seems to not work at all in my app.

takethefake (approved):
Hej @SkySails, thanks for your contribution! 

I've checked your PR and it seems solid to me! Therefore I would give it a go!

SkySails:
Hej @takethefake! Thank you for taking a look, I'm happy to help!

takethefake:
@segunadebayo how do we want to continue with this PR?

packages/color-mode/src/color-mode-provider.tsx
```diff
@@ -59,6 +60,8 @@ export function ColorModeProvider(props: ColorModeProviderProps) {
     colorModeManager = localStorageManager,
   } = props
 
+  const defaultColorMode = initialColorMode === "dark" ? "dark" : "light"
```
Robbie-Cook, line 63: Dis look sus

packages/color-mode/src/color-mode-provider.tsx
```diff
@@ -83,14 +86,16 @@ export function ColorModeProvider(props: ColorModeProviderProps) {
      */
     if (isBrowser && colorModeManager.type === "localStorage") {
       const mode = useSystemColorMode
-        ? getColorScheme(initialColorMode)
-        : root.get() || colorModeManager.get()
+        ? getColorScheme(defaultColorMode)
```
Robbie-Cook, line 89: Actually, here is an area to look at

jnsdls:
I'm reasonably sure this should have been a breaking change, here's the situation it led me into:
- next project
- whole project is always supposed to be `"light"` mode (no darkmode implementation yet)
- have the theme set to `"light"` and am not touching anything with colormode script or local storage at all
- my system (macos) is set to dark mode

after this PR merged & the new release included it that leads to the following scenario:
- the colormode in the theme gets completely ignored (since it's never in local storage)
=> my app is now behaving in darkmode because of my system preferences, even though it's set to `"light"` and `useSystemColorMode` is explicitly false

for now I worked around it by wrapping the entire dom tree in a `<LightMode>` switch, but ... definitely feels like a breaking change happened here. (since i was basically forced into system color mode)

PR #4642, opened 2021-08-30
fix(checkbox): pass tabIndex to checkbox-input if provided via props

### LINKED ISSUE #4641: [Checkbox] can't make checkbox truly unfocusable
### Description

When providing the `isFocusable`-prop to a checkbox it's still tabbable via keyboard and the `::focus` styles are applied. Only in combination with `isDisabled` the checkbox is not focusable.

### Link to Reproduction

Not needed

### Steps to reproduce

_No response_

### Chakra UI Version

1.6.6

### Browser

_No response_

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

Suggested Change:

if the user adds `tabIndex` to the `Checkbox`-component pass the tab index to `getInputProps` to truly make a checkbox not focusable, otherwise, only the label isn't focusable. 

Alternative: A more invasive way would be to set `tabIndex=-1` inside `Checkbox` if `isFocusable===true` but this might be a breaking change for some users.

takethefake:
Closes #4641 

## 📝 Description

Currently `isFocusable=false` does not make a component truly non-focusable. If the User provides a `tabIndex=-1` to Checkbox, the `tabIndex` is only applied to the `CheckboxLabel` and never reaches the `input`. 

## 🚀 New behavior

If the user provides a `tabIndex` to `Checkbox` the `tabIndex` is passed down to `getInputProps` which gives the user the ability to truly make a Checkbox non-focusable.

No


packages/checkbox/src/checkbox.tsx
```diff
@@ -150,7 +150,10 @@ export const Checkbox = forwardRef<CheckboxProps, "input">((props, ref) => {
       className={cx("chakra-checkbox", className)}
       {...getRootProps()}
     >
-      <input className="chakra-checkbox__input" {...getInputProps({}, ref)} />
+      <input
+        className="chakra-checkbox__input"
+        {...getInputProps({ tabIndex: props.tabIndex ?? undefined }, ref)}
```
segunadebayo, line 155: I think you can destructure `tabIndex` so it doesn't get duplicated in the `rootProps` and `inputProps`.

Maybe a better place to handle this is in the `useCheckbox` hook itself?

PR #4678, opened 2021-09-07
feat: multi-thumb slider

### LINKED ISSUE #2642: Range Slider
# Feature request

## Is your feature request related to a problem? Please describe.

There is no way to select an upper and lower value (range) using the slider, which is useful for a lot of applications like a price filter.

## Describe the solution you'd like

Add a new component called Range Slider (or an additional prop which transforms the existing slider) which has two SliderThumbs. One controls the upper limit and the other the lower limit. The Thumbs can't cross over.

The value & defaultValue will accept an array of values `[0, 100]` & the onChange events will also return an array of values.

## Describe alternatives you've considered

Antd have a nice Range Slider.

luigiminardim:
Closes #2642

Create a multi-thumb slider that can be a range slider.

> Add a brief description

The current slider available by Chakra-UI has just one thumb that holds a single value.

## 🚀 New behavior

The multi-thumb slider has an API that allows the developers to use as many thumbs as needed. This behavior allows developers to create range sliders and other types of multi-thumb sliders. 

No.

## 📝 Additional Information

This PR adds a new package in the Chakra-UI library.

### DIFF
--- a/packages/multi-slider/README.md
+++ b/packages/multi-slider/README.md
@@ -1,0 +1,75 @@
+# @chakra-ui/multi-slider
+
+The Multi-Thumb Slider is used to allow users to make selections from a range of
+values.
+
+Multi-THumb Sliders reflect a range of values along a bar, from which users may
+select varius values. They are ideal for picking price filters or other type of
+filters.
+
+## Installation
+
+```sh
+yarn add @chakra-ui/multi-slider
+
+# or
+
+npm i @chakra-ui/multi-slider
+```
+
+## Import components
+
+```js
+import {
+  MultiSlider,
+  MultiSliderTrack,
+  MultiSliderFilledTrack,
+  MultiSliderThumb,
+} from "@chakra-ui/react"
+```
+
+## Usage
+
+```jsx
+<MultiSlider>
+  <MultiSliderTrack>
+    <MultiSliderFilledTrack startThumbKey={0} endThumbKey={1} />
+  </MultiSliderTrack>
+  <MultiSliderThumb thumbKey={0} aria-label="slider 0" defaultValue={10} />
+  <MultiSliderThumb thumbKey={1} aria-label="slider 1" defaultValue={15} />
+</MultiSlider>
+```
+
+### Changing the slider color
+
+```jsx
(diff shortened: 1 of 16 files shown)


palegz (approved):

segunadebayo:
Hi @luigiminardim,

First off, this is amazing! Thanks for taking the time to understand the system and add the multi-thumb slider component. I like the approach you took. Some thoughts based on this PR.

- We're trying to cut down on the number of packages we create to reduce complexity. I'd recommend moving most of the code into the slider package since they belong together (technically)
- Most of the logic/functions can be shared between the slider and multi-thumb slider. This is based on the first point I mentioned above. If both the slider and multi-thumb slider live together, we can ensure they use very similar utilities.
- The pattern you used seems to be inconsistent with the API of `useSlider` and we need to ensure they're as consistent as possible. You have a couple of `use*` hooks that return the prop getters, but the `useSlider` hook returns the getters directly without wrapping them in a `use*` hook.

Since this is the most requested feature, I'm going to refine some ideas in this PR and open a draft PR that addressed most of these points.

luigiminardim:
Hi @segunadebayo, 

First of all, thanks for the feedback!
I appreciate the points you have noticed. I do think that the slider and the multi-thumb slider can be placed in the same package sharing the same code, by the way, the next step I would take is building the simple slider using the multi-thumb slider.  The reason I didn`t take this approach is because of the way I was thinking about the multi-thumb slider API, which I was concerned would cause a breaking change in the traditional slider.

Another note I would like to take is that I think my implementation misses some accessibility functionality since the user has to press in the exact location of the thumb to move it. I think I could implement using the pan gesture and get the closest thumb to move around.

I can`t wait to see the refined ideas of your PR. I hope that my sharing was useful for the development of the library and the community.

segunadebayo:
Good points @luigiminardim. Here's the PR that includes most of the ideas in your PR plus the accessibility areas you missed.

https://github.com/chakra-ui/chakra-ui/pull/4740

In the same spirit, I'm going to close this one in favor of completing that PR with docs, tests, and storybook.

I really appreciate your thoughts and work on this feature. Thank you, man!

PR #4703, opened 2021-09-10
fix(number-input): spinner getting stuck in useNumberInput button props

### LINKED ISSUE #4646: Holding decrement button on useNumberInput causes new inputs to count down to min value
### Description

Holding decrement button with a min value until the button is disabled on useNumberInput causes new inputs to count down to the min value 

### Link to Reproduction

https://codesandbox.io/s/use-number-input-bug-oxyy6?file=/src/index.js

### Steps to reproduce

1. Set a `min` value for the useNumberInput
2. Enter a number slightly greater than the min value in the input field
3. Click and hold the decrement button until the current input value has reached the min value and the decrement button is disabled
4. Type in another number greater than the min value
5. Watch the current input value decrement without any user input

### Chakra UI Version

1.6.7

### Browser

Google Chrome 92.0.4515.159 

### Operating System

- [ ] macOS
- [ ] Windows
- [X] Linux

### Additional Information

It appears when hovering the mouse over any of the buttons while the value is auto decrementing appears to stop the bug.

xotea:
Closes #4646 

## 📝 Description

Using `useNumberInput` and setting a `min` and/or `max` value and a chakra `<Button />` with `getIncrementButtonProps` or `getDecrementButtonProps` the spinner gets stuck if you reach the limit and move the mouse off the now disabled button while still holding left click in latest Chromium-based browsers. In Firefox it's fine.

## 🚀 New behavior

Adding an `onPointerLeave` event handler to the existing ones seems to fix the problem. This was also recommended [in this React issue](https://github.com/facebook/react/issues/18753) as a solution.

The `onPointerUp` handler is also needed if the mouse doesn't leave the button while editing the input field value.

No

### DIFF
--- a/packages/number-input/src/use-number-input.ts
+++ b/packages/number-input/src/use-number-input.ts
@@ -377,6 +377,7 @@ export function useNumberInput(props: UseNumberInputProps = {}) {
         onMouseLeave: callAllHandlers(props.onMouseUp, spinner.stop),
         onTouchEnd: callAllHandlers(props.onTouchEnd, spinner.stop),
         onPointerLeave: callAllHandlers(props.onPointerLeave, spinner.stop),
+        onPointerUp: callAllHandlers(props.onPointerUp, spinner.stop),
         disabled,
         "aria-disabled": ariaAttr(disabled),
       }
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for working on this @xotea.

I took some ideas from your PR and merged them into a single PR to address multiple number-input issues

See #4982

PR #4734, opened 2021-09-19
feat(popover): manual trigger mode

### LINKED ISSUE #4733: feat(popover): manual trigger mode
Hello Chakra folks!

I have a use case where I need to have a `Popover` where neither clicking nor hovering on the trigger should toggle the visibility of the popover.
Similar to the [Controlled Usage example](https://chakra-ui.com/docs/overlay/popover#controlled-usage), however I want the `Popover Target` button to not close the popover.
In my case the `PopoverTrigger` children is an input and I want to be able to click inside it without closing the popover.

My current workaround:
```tsx
<Popover
    isOpen={isOpen}
    onClose={onClose}
    closeOnBlur={false}
    // Passing anything else than `click` or `hover` works, however TS types do not accept it.
    trigger={"manual" as any}
>
    ...
</Popover>
```

### Proposed Solution or API

Add a `manual` mode for `trigger`. Or should it be called `none` instead?

It seems like just adding this value to the possible `TRIGGER` values makes this work, since Chakra doesn't bind any event listeners in that case.

_____
Thanks!

dodas:
Closes #4733

## 📝 Description

The `trigger` can now be set to `manual` in case you don't want the `PopoverTrigger` to close the popover on click, but rather serve just as an reference point for positioning.

No way to disable hover/click triggering the popover.

## 🚀 New behavior

The `trigger` can now be set to `manual` in order to not bind any mouse/keyboard to the trigger for fully controlled usage.

No

### DIFF
--- a/.changeset/sharp-mirrors-jog.md
+++ b/.changeset/sharp-mirrors-jog.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/popover": minor
+---
+
+The `trigger` can now be set to `manual` in order to not bind any mouse/keyboard
+to the trigger for fully controlled usage.
--- a/packages/popover/src/use-popover.ts
+++ b/packages/popover/src/use-popover.ts
@@ -21,6 +21,7 @@ import { RefObject, useCallback, useEffect, useRef, useState } from "react"
 const TRIGGER = {
   click: "click",
   hover: "hover",
+  manual: "manual",
 } as const
 
 export interface UsePopoverProps extends UsePopperProps {
@@ -87,6 +88,9 @@ export interface UsePopoverProps extends UsePopperProps {
    *
    * `click` - means the popover will open on click or
    * press `Enter` to `Space` on keyboard
+   *
+   * `manual` - means the popover won't automatically open or close
+   * on any trigger interaction.
    */
   trigger?: keyof typeof TRIGGER
   openDelay?: number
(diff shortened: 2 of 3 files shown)


segunadebayo:
That looks interesting @dodas. I'm not sure I like the idea of using "manual" to solve this use-case.

What do you think about providing an explicit `PopoverAnchor` or `PopoverReference` component that wraps the component you'd like to use as the positioning reference?

glocore:
Just a thought: In a controlled usage scenario, clicking on the `PopoverTrigger` does not open the Popover, so shouldn't that be the case for closing as well? Rather than adding a config option for this, wouldn't it make sense to have the default behaviour be such that clicking on the `PopoverTrigger` has no effect for controlled usage?

dodas:
Hey Segun!

I like the idea of `PopoverAnchor`. I have a working version of that. I will put it up in a new PR so we can discuss it there.

@glocore in controlled usage, clicking `PopoverTrigger` does call `onOpen`/`onClose` on the Popover, so if you are listening to those, your popover will get both opened and closed.
In my case, I want the popover to be closable from `usePopoverConext`, so I need to listen to `onClose`, but I don't want it to be triggered on the trigger click/hover.

glocore:
Yup, I later realised that the PopoverAnchor idea made sense and hence deleted my comment!

anubra266:
Hi @dodas are you still working on this?

chloe-schoreisz:
Hello @dodas ! I'd love to see this implemented, as I'm trying to implement an Input with autofill by using the popover :) are you still implementing this ?

dodas:
Hey @anubra266  and @chloe-schoreisz 
I wasn't paying attention to this lately, but Chloe's use-case seems very legit and this would help a lot with it.
I will try to push it soon, just need to clean it up a bit.

dodas:
I just pushed https://github.com/chakra-ui/chakra-ui/pull/5044, please comment there.

anubra266:
@dodas can we close this PR and link the new one to issue #4733

PR #4736, opened 2021-09-19
Fix 3884 text styles not applied correctly

### LINKED ISSUE #3884: textStyles defined inside extendTheme aren’t applied correctly
# 🐛 Bug report

Styles defined via `textStyles` are not applied correctly although they’re set within `extendTheme`.

## 💥 Steps to reproduce

1. Go to my [CodeSandbox](https://codesandbox.io/s/textstyle-problems-ibu9h?file=/src/App.tsx:176-186)
2. Inspect Heading, Button, Text and the body element with the developer tools inside the browser
3. Have a look at the styles set by Chakra UI and compare them

### textStyles which should be applied
image

### body styles
image
✅ `font-size` from `textStyles` is used.
💥 `font-family` and `line-height` aren’t applied by `textStyles`.

### Heading styles
image
💥 Neither `font-family` nor `font-size` and `line-height` are applied by `textStyles`.

### Button styles
image
✅ `font-family` from `textStyles` is used.
💥 `font-size` and `line-height` aren’t applied by `textStyles`.

### Text styles
image
✅ All styles set by `textStyles` are used.  

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/textstyle-problems-ibu9h?file=/src/App.tsx:176-186

## 🧐 Expected behavior

Styles like `font-family`, `font-size` and `line-height` defined in `textStyles` should be applied when passed through `extendTheme` according to [this comment](https://github.com/chakra-ui/chakra-ui/issues/3658#issuecomment-807299129). This behaviour should be consistent and reliable when used on all elements (even `<body />`).

## 🧭 Possible Solution

–

## 🌍 System information

| Chakra UI        |     1.5.2       |
| Operating System |      macOS      |

## 📝 Additional information

–

primos63:
Closes #3884 

## 📝 Description

> Remove change from PR #3615 which caused this issue.
> Modified `toCSSObject` to correctly handle merging styles in precedence order, __css, baseStyle, other (apply, layerStyles and textStyles), styleProps, sx and css. Each style has its CSS generated and merged with the accumulated CSS object.

> If a component already has a property for which a `textStyle` property would overwrite, the update is skipped.

## 🚀 New behavior

> Each style property is processed in precedence order. For each style the CSS for the style is generated, the CSS is then merged with an accumulator for the final computed CSS. This ensures that properties, including responsive ones, have the correct values.


dsa:
Is there an ETA on this? I'm noticing this problem everywhere when trying to apply theme styles to built-in components (like Heading, FormLabel, etc.).

packages/system/src/system.ts
```diff
@@ -40,16 +46,31 @@ interface GetStyleObject {
...
+  //   other (apply, layerStyles, textStyles)
+  //   styleProps
+  //   sx
+  //   css
+  stylePrecedence.push({ ...__css })
+  stylePrecedence.push(Object.assign({}, runIfFn(baseStyle, props)))
+  stylePrecedence.push(objectFilter(rest, (_, prop) => isOtherProp(prop)))
+
+  const styleProps = objectFilter(
+    rest,
```
segunadebayo, line 63: As we discussed during the call, let's see how we can do both `objectFilter` in a single call. If need be, we can write a custom `objectFilter`. Here's what a sample implementation can look like

```jsx
const { styleProp, otherProp, domProps } = splitProps(rest)
```

packages/system/tests/style-resolver.test.ts
```diff
@@ -74,7 +74,7 @@ test("should resolve styles correctly", () => {
         "@media screen and (min-width: 30em)": Object {
           "--chakra-line-clamp": 4,
           "fontSize": "23px",
-          "letterSpacing": "0.2px",
+          "letterSpacing": "50px",
```
segunadebayo, line 77: Good catch here. Thank you 🎉

packages/system/src/system.ts
```diff
@@ -40,16 +46,31 @@ interface GetStyleObject {
...
+  const stylePrecedence = []
+
+  // Styles are applied by precedence (lowest to highest)
+  //   __css
+  //   baseStyle
+  //   other (apply, layerStyles, textStyles)
+  //   styleProps
+  //   sx
+  //   css
+  stylePrecedence.push({ ...__css })
```
segunadebayo, line 58: Could we do without spreading the object and just push to the `stylePrecedence` array directly?

segunadebayo (commented):
This looks great! @primos63.

Let's address the comments and we should be good to go

segunadebayo:
Closing this in favor of #5253. I'll review that PR tomorrow

Issue #2190, opened 2020-10-07
storybook chakra-ui themes addon


hems:
# Feature request

## Is your feature request related to a problem? Please describe.

Currently, there is no easy way of switching chakra-ui theme on Storybook

## Describe the solution you'd like

A implementation similar to https://github.com/echoulen/storybook-addon-styled-component-theme 

## Describe alternatives you've considered

Forking https://github.com/echoulen/storybook-addon-styled-component-theme and hacking a chakra-ui version of the same addon

## Additional context

I'm not entirely sure if the chakra-ui team is planning to release such addon for Storybook and if forking the styled-component repo would be the best way of going about it

with-heart:
That's an interesting idea. We could certainly create an addon to enable that but with our current capacity, that's pretty low priority. If anyone would like to take the time to investigate though and attempt to implement something similar, we can see how we can integrate that into the Chakra UI tooling ecosystem.

hems:
> That's an interesting idea. We could certainly create an addon 

I believe you wouldn't need to create one from scratch, you could just fork the addon i mentioned here:
https://github.com/echoulen/storybook-addon-styled-component-theme

> to enable that but with our current team capacity, that'll have to be low priority. 

I would say someone with TypeScript knowledge and good understanding of Chakra would be able to fork and change that addon to work with chakra in less than 2 hours?

> If anyone would like to take the time to investigate though and attempt to implement something similar

I'm new to typescript and very new to chakra ( and even more knew to the @next version ), so i'm still wrapping my head around the differences in the implementation `styled-components` ( currently used on my application ) to the way theme-ing works  on chakra, i'm still slightly confused but the new documentation is actually clarifying a lot of my questions.

From my understanding We would have to fork that addon i mentioned and then replace the `<ThemeProvider>` from `styled-compoents` tag by the `<ChakraProvider>` and then use the input from the addon to set the `initialColorMode`  as described on this [help page](https://next.chakra-ui.com/docs/features/color-mode#change-color-mode-behavior)??

hems:
Actually it seems that library already accepts a `CustomThemeProvider` so we could potentially just write a "CustomThemeProvider" and then use that to write the `<ChakraProvider>` tag without having to  change the addon, the only thing is the library uses `styled-components` as a dependency because the default theme provider is the one from `styled-components`

hems:
One issue that is happening on my application is that if i use

```js
const config = {
  useSystemColorMode: false,
[...]

PR #4749, opened 2021-09-22
feat: storybook theme selector, which stays persistent trough the stories

maapteh:
## 📝 Description

This is work in progress. We can now select the color mode (or with small mod also a complete other theme) and let it stay persistent trough all the stories. So we can ditch setting colours everywhere and only use them when really applicable. It makes it nice to see components in their own selectable theme.

cut

Storybook now just shows variants but in a defined color.

## 🚀 New behavior

You select the color scheme, and all stories with no predefined scheme will show that scheme: teal/gray/telegram/messenger/etc

See: https://chakra-ui-storybook-n27abd8jl-chakra-ui.vercel.app/?path=/story/button--with-variants

No

## 📝 Additional Information

I just started with one story since i dont know if this is something you want :) So in Buttons variants i removed the Teal schema defined on the button. Then i added all colours including the brand ones, which you can now select. The selected scheme is in the url so when you share something it stays in that schema.

solves #2190

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -6,6 +6,7 @@ import {
   useColorMode,
   useColorModeValue,
   theme,
+  withDefaultColorScheme
 } from "@chakra-ui/react"
 import { StoryContext } from "@storybook/react"
 import * as React from "react"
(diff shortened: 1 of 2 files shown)


segunadebayo:
Hi @maapteh,

This looks like a great solution but I think there's a lot of value in seeing all the possible color schemes at the same time, instead of toggling each time.

Thank you for making a POC before going too far. I like the idea but I'm not sure if we want to add this into the Storybook.

Thanks for understanding.

PR #4757, opened 2021-09-24
chore(system): export domElements

Enteleform:
## 🚀 New behavior

Exports `domElements` and `DOMElements` from `@chakra-ui/system` so they can be used in userland.

No


segunadebayo:
Hi @Enteleform,

What's the use-case for exporting these utilities?

Enteleform:
> Hi @Enteleform,
> 
> What's the use-case for exporting these utilities?

Hello,

I'm using this for a dev sandbox so that I can access all `Chakra` dom elements & components globally, in order to facilitate rapid prototyping without managing imports.

For the components (`Modal`, `ModalContent`, etc.), I have to instantiate the global references individually due to a lack of an iterable container for said components.  
(*[ChakraComponent.ts](https://gist.github.com/Enteleform/6acaae2fb7ae61942a376cb22a1b553c), [ChakraComponent.d.ts](https://gist.github.com/Enteleform/410800e355fc3724f0343b6711082e32)*)

However, `domElements` makes this a bit more concise/maintainable, for the `.ts` implementation of dom elements at least:

### @ App Entry:
```ts
import {domElements} from "@chakra-ui/system"
import {chakra     } from "@chakra-ui/react"

globalThis.$ = chakra.div

for(const element of domElements)
  {globalThis[`$${element}`] = chakra[element]}
```

### Elsewhere
```jsx
<$ backgroundColor="gray.800">
  <$span color="red.500" > Foo </$span>
  <$span color="blue.500"> Bar </$span>
</$>
```

segunadebayo:
Hi @Enteleform,

I see what you mean and it looks interesting. I don't think this use case is worth exposing those modules to be public. This would mean that altering any element from that array will be a breaking change. I don't think it's the right decision to make for now.

Since it's for prototyping purposes, feel free to copy out the array to your project. I totally understand that it might not be convenient.

PR #4758, opened 2021-09-25
refactor(utils): add ref like typing

AmagiDDmxh:
Closes # 

## 📝 Description

> Add a ref like typing for use

## 🚀 New behavior

> Better typing

NO

## 📝 Additional Information

Tested using a simple case, example codes are
```
const aRef = createRef()
isRefObject(aRef)
```

Without throw any typing errors


AmagiDDmxh:
Closed because it is supposed to be `any`

PR #4791, opened 2021-10-05
refactor: update media query to use no deprecated functions

lukasbals:
Closes # 

## 📝 Description

update media query to use `addEventListener` and `removeEventListener`
instead of the deprecated `addListener` and `removeListener` functions.
https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener

The used functions in the `media-query` package are deprecated.

## 🚀 New behavior

The used functions in the `media-query` package are not deprecated.

No.

### DIFF
--- a/.changeset/sharp-snakes-camp.md
+++ b/.changeset/sharp-snakes-camp.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/media-query": minor
+---
+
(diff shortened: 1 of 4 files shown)


takethefake:
Hej @lukasbals thanks you very much for your contribution to this.

But since #4807 fixes this and the general behavior of the `useMediaQuery`-hook I'll have to close this PR in favor of the other PR.

PR #4803, opened 2021-10-07
fix(range-slider): issue when thumbs are stacked

### LINKED ISSUE #4800: Can't change Range Slider values when slider thumbs are stacked to the left.
### Description

When using the Range Slider component I expect to move the thumbs to change range values but if the thumbs are stacked to the left I can't move or drag neither of the thumbs.

### Link to Reproduction

https://codesandbox.io/s/range-slider-7sj2f

### Steps to reproduce

1. Slide all the thumbs to the left so they are stacked.
2. Try to move the thumb.

### Chakra UI Version

1.6.9

### Browser

_No response_

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

_No response_

segunadebayo:
Closes #4800

## 📝 Description

This PR fixes an issue where, when slider thumbs are scrubbed to have the same value, you can't change the value of any of the thumbs afterward. This is more obvious when you scrub both values to the `min`.

## 🚀 New behavior

- You can change the value by clicking anyway on the track
- To prevent overlapping thumbs, you can use the `minStepsBetweenThumbs` prop

No

## 📝 Additional Information

- [x] Ensure the Range-Slider or PanSession works within an iframe
- [x] Fix issue where Slider Thumb stacking on each other

### DIFF
--- a/.changeset/perfect-dogs-compare.md
+++ b/.changeset/perfect-dogs-compare.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/hooks": patch
+---
+
+Forward `threshold` options from `usePanSession` to `PanSession` class
--- a/packages/hooks/src/use-pan-gesture.ts
+++ b/packages/hooks/src/use-pan-gesture.ts
@@ -15,6 +15,7 @@ export interface UsePanGestureProps {
   onPanEnd?: PanEventHandler
   onPanSessionStart?: PanEventHandler
   onPanSessionEnd?: PanEventHandler
+  threshold?: number
 }
 
 export function usePanGesture(
@@ -27,6 +28,7 @@ export function usePanGesture(
     onPanEnd,
     onPanSessionStart,
     onPanSessionEnd,
+    threshold,
   } = props
 
   const hasPanEvents = Boolean(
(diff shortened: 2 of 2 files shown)


Brennvo:
I was looking up how other multi-thumb sliders work. Most allow the dual-thumbs to actually [cross over each other](https://dequeuniversity.com/library/aria/slider-multirange), but I personally think this is less than ideal since most developers will have to end up validating negative values and such.  Another solution is to not allow the thumb markers to ever be placed on top of each other, which is easy to achieve when using [triangles as the slider thumbs](https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/multithumb-slider.html), but I think that goes against the Chakra design system. After some more digging, [I came across this multi-thumb slider using circle thumbs that do not allow crossing](https://demos.jquerymobile.com/1.3.2/widgets/sliders/rangeslider.html). I think this is pretty cool — if your mouse is moving right, then the right thumb moves. Move the mouse left, and the left thumb moves. One thing to note is that the tab order of the thumbs will always stay the same, which is a requirement per the [WAI-ARIA guidelines](https://www.w3.org/TR/wai-aria-practices-1.1/#slidertwothumb):

> The tab order remains constant regardless of thumb value and visual position within the slider. For example, if the value of a thumb changes such that it moves past one of the other thumbs, the tab order does not change.

Is this something you think is feasible? I can also assist if this sounds like a good plan.

PR #4817, opened 2021-10-10
feat(tooltip): support `motionPreset`

### LINKED ISSUE #4722: [Motion] Unify `motionPreset` across components
At the moment, only the `Modal` component uses the `motionPreset` prop to allow users to change the motion type of the modal, or even disable it.

We want to support this prop in all motion related components, namely `Popover` and `Tooltip`.

**Benefits**
- For those who don't want the animations, they can disable it by passing `motionPreset="none"`
- For those who want to try different motion/transitions, they can use `motionPreset="slideIn" | "scale" | "fade"`

jeferson-sb:
Partially Closes #4722 

## 📝 Description

Add support for `motionPreset` to be either `none`, `slideInTop` and `scale` on Tooltip component.

Current the Tooltip component only has one transition which is scale.

## 🚀 New behavior

You can try different transitions or disable them with the `motionPreset` prop.

No

## 📝 Additional Information

tooltip

### DIFF
--- a/.changeset/purple-readers-decide.md
+++ b/.changeset/purple-readers-decide.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/tooltip": patch
+"@chakra-ui/docs": patch
+---
+
+- Add `motionPreset` prop to tooltip to be able to change default transition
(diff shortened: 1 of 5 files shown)


jeferson-sb:
@joernroeder yeah the tooltip will always come from the top on every placement, but we can think on a better transition too...

segunadebayo:
After some thought, we won't be going forward with this PR.

We're starting to migrate out of `framer-motion` and make it opt-in for the users.

Issue #4686, opened 2021-09-08
[modal] `closeOnEsc=true` not always respected with `closeOnOverlayClick=false`


Andarist:
### Description

When `closeOnOverlayClick=false` is used and modal loses the focus when that overlay is clicked then closing on escape stops working.

### Link to Reproduction

https://chakra-ui.com/docs/overlay/modal#close-modal-on-overlay-click

### Steps to reproduce

1. Go to https://chakra-ui.com/docs/overlay/modal#close-modal-on-overlay-click
2. Click on 'Open modal'
3. Click on the overlay
4. Press ESC

Actual: modal stays opened
Expected: modal should get hidden

Possible solution: "trap" focus in the modal by focusing it again when it blurs in this scenario.

### Chakra UI Version

latest

### Browser

Chrome 93

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

PR #4823, opened 2021-10-12
fix: modal focus and `closeOnEsc`

antoniel:
Closes https://github.com/chakra-ui/chakra-ui/issues/4686 

## 📝 Description

Fixes wrong behavior, where if the user focus the overlay, the model will not closes when pressing `esc`, only when focus the content again

if closeOnEsc is equas to true, the modal should close once user press escape key

## 🚀 New behavior

Adds `tabIndex={-1}` on div that has the listener to keyPress

Nope


antoniel:
For sure @segunadebayo, but I can't do it by today I'm out of time, probably I can add it tomorrow, If this is a bug that needs to be release asap we can proceed with this PR and then backs in another PR to adds tests to it

PR #4835, opened 2021-10-13
fix(hooks): id is inconsistent when in strict mode

### LINKED ISSUE #4328: Warning: Prop `id` did not match. Server: "accordion-button-3" Client: "accordion-button-6"
### Description

Whenever I use an accordion with NextJs, I get mismatches for ids.

### Link to Reproduction

https://codesandbox.io/s/snowy-thunder-gxsfl?file=/pages/index.tsx:0-1603

### Steps to reproduce

First thing first, I can't reproduce it with the codesandbox. I think it's running `next start` while you only see the errors with `next dev`...

I still linked a codesandbox with something that will reproduce the problem when ran in local.

To reproduce, I simply create a new page, let's say "accordion.tsx" with this code taken directly from the documentation:

```typescript
import {
  Accordion,
  AccordionButton,
  AccordionIcon,
  AccordionItem,
  AccordionPanel,
  Box,
  Heading
} from "@chakra-ui/react";
import React from "react";

export default function IndexPage() {
  return (
    <article>
      <Heading>Welcome to Chakra UI example</Heading>
      <Accordion allowToggle>
        <AccordionItem>
          <h2>
            <AccordionButton as="button" id="ttr">
              <Box flex="1" textAlign="left">
                Section 1 title
              </Box>
              <AccordionIcon />
            </AccordionButton>
          </h2>
          <AccordionPanel pb={4}>
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
            eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim
            ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
            aliquip ex ea commodo consequat.
          </AccordionPanel>
        </AccordionItem>

        <AccordionItem>
          <h2>
            <AccordionButton>
              <Box flex="1" textAlign="left">
                Section 2 title
              </Box>
              <AccordionIcon />
            </AccordionButton>
          </h2>
          <AccordionPanel pb={4}>
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
            eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim
            ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
            aliquip ex ea commodo consequat.
          </AccordionPanel>
        </AccordionItem>
      </Accordion>
    </article>
  );
}
```

Load the page, look at the console and I get:

```
Warning: Prop `id` did not match. Server: "accordion-button-3" Client: "accordion-button-6"
```

This is new from Chakra `1.6.4`.

Everything was working fine in my project with `1.6.3`.

When I migrated to `1.6.4`, I saw there had been a big modification for generating `id`s so they could better match between server and client. (supposedly, I never had the problem before so I have no idea why it was "fixed")

I looked at the modified file and it was a huge change. I tried it as is, but I couldn't make it work at all in my project because **with `1.6.4` pretty much all ids were a mismatch now...**

Since I was using Chakra's `useId` hook for my own custom components, I created a new `useId` with the old code and removed the import from Chakra. Everything went back to normal.

I didn't use any accordion at the time, so I never saw any other problem. But obviously, anything that would use Chakra `1.6.4` `useId` will behave incorrectly with NextJs, as we can see with `Accordion`.

I figured someone would eventually create an issue because the console is completely red when used with NextJs in SSG... Turns out I'm the first to report it after all :( 

### Chakra UI Version

1.6.4

### Browser

MS Edge 91.0.864.64 (latest at the moment)

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

I tried passing my own `id` to the accordion button, but it's simply ignored by Chakra implementation, it overrides it with it's own automatic `id`. This should be looked at, if I could somehow pass my own `id`, I could easily patch in the old version to generate `id`s, the one that worked 100% with NextJs.

Screenshot of a simple page with only the accordion on it
image

Screenshot of the console for that same page
image

Screenshot of React Dev Tools on the Accordion Button (id = "ttr")
image

Screenshot of React Dev Tools on the Accordion Button sub component "button" (id = "accordion-button-6")
image

I have no idea why on the server the id is `accordion-button-3` while on the client it's `accordion-button-6`.

My repository is private, but I could invite someone from the staff to debug it if necessary.

If you require anything else, don't hesitate to ask.

Regards,

TheThirdRace

igorwessel:
Closes #4328

## 📝 Description

We were changing (generating) a new id in useMemo, however when used with strict mode, 
he double-invoking the function passed inside useMemo to precisely check side effects possibly placed in the wrong places.

## 🚀 New behavior

We fix the side effect that we were doing in useMemo moving to useEffect, where we can have side effects because it occurs in the commit phase.

```js
export function useId(idProp?: string, prefix?: string): string {
  const context = React.useContext(IdContext)
  return React.useMemo(
    () =>
      idProp ||
      [prefix, context.prefix, ++context.current].filter(Boolean).join("-"), 
	/* ++context.current, This in strict mode is a side effect that you cannot put here */

    // eslint-disable-next-line react-hooks/exhaustive-deps
    [idProp, prefix],
  )
}
```

```js
export function useId(idProp?: string, prefix?: string): string {
  const context = React.useContext(IdContext)
  const [id, setId] = React.useState(context.current)

  React.useEffect(() => {
    setId(genId(context))
  }, [context])

  return React.useMemo(
    () => idProp || [prefix, context.prefix, id].filter(Boolean).join("-"),
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [idProp, prefix, id],
  )
}
```

No


igorwessel:
I will add tests for this hook based on the current implementation.

igorwessel:
A possible solution to strict mode would be to use useState to store the ID and update the ID when the component has been rendered

```js
export function useId(idProp?: string, prefix?: string): string {
  const seed = useUIDSeed()

  const [id, setId] = React.useState("")

  React.useEffect(() => {
    setId(seed("").replace("val-", ""))
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [])

  return React.useMemo(() => idProp ?? [prefix, id].filter(Boolean).join("-"), [id, prefix, idProp])
}
```

That way we can force the component to render again to stay in sync, I ended up sending the tests and forgot that I didn't send these updates.

anton-johansson:
@igorwessel, what is the status of this PR? It would be great to get rid of those inconsistencies and warnings. :)

igorwessel:
> @igorwessel, what is the status of this PR? It would be great to get rid of those inconsistencies and warnings. :)

I don't think there's anything else to add, Now I just resolved the conflicts with the main branch.

segunadebayo:
Great work here @igorwessel. I really appreciate the thought and tests in this PR 💖

Happy New Year 😊

PR #4881, opened 2021-10-19
docs: separates npm and yarn commands im Nextjs docs so they can be c…

zealousAnemone:
…opied individually by users

Closes # 

## 📝 Description

Small change to NextJS guide, makes it easier to copy shell commands

Currently, commands for npm and yarn are grouped together in one code block so if you hit "copy", it copies both.

## 🚀 New behavior

npm and yarn scripts are separated

No


segunadebayo:
This looks great @zealousAnemone, but I don't think it's necessary. It distorts the UI and I think users can already copy the commands even if it's super-convenient in some cases.

Thank you for working on this 💖

PR #4952, opened 2021-10-29
docs: update color-mode.mdx

akshitarora921:
## 📝 Description

>Added missing import statement in react ColorModeScript example

### DIFF
--- a/website/pages/docs/features/color-mode.mdx
+++ b/website/pages/docs/features/color-mode.mdx
@@ -119,6 +119,7 @@ file.
 ```jsx live=false ln={10}
 // index.js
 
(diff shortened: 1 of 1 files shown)


website/pages/docs/features/color-mode.mdx
```diff
@@ -119,6 +119,7 @@ file.
 ```jsx live=false ln={10}
```
segunadebayo, line 119: I assume you'd need to update the `ln={10}` since that is used to highlight a specific area in the code.

akshitarora921: yes, I will do the change. Thanks for pointing it out.

PR #4956, opened 2021-10-30
build: use microbundle to build packages

### LINKED ISSUE #4937: Issue with Next.js 12.0.0
### Description

With the release of Next.js 12.0.0 which now supports ESM modules, it breaks when using with Chakra-ui

### Link to Reproduction

No need for a codesandbox

### Steps to reproduce

Follow this thread: https://github.com/vercel/next.js/issues/30330
This comment: https://github.com/vercel/next.js/issues/30330#issuecomment-952847838 explains the issue and how to fix it. 

### Chakra UI Version

1.6.9

### Browser

Brave Version 1.31.87 Chromium: 95.0.4638.54 (Official Build) (x86_64)

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

cschroeter:
Closes #4937 

## 📝 Description

Use Microbundle to bundle packages and use it to fix issues when using Chakra UI
in Nextjs 12+

Won't work in NextJS 12

## 🚀 New behavior

Works in NextJS 12+

No


segunadebayo:
This looks great! I tested it and it works.

The only downside I see is that it takes a longer time to build the modules compared to the previous babel CLI setup. In my mind, I'd like us to hit the sweet spot between "working for Next.js 12" and "faster build times".

Not sure how we can achieve this but it's worth discussing so we don't merge this and then switch the build system in a couple of weeks again. Next.js released a `canary` that seems to fix the issue (I haven't confirmed) so maybe our users upgrade.

What are your thoughts @TimKolberger @with-heart?

cschroeter:
> The only downside I see is that it takes a longer time to build the modules compared to the previous babel CLI setup.

That's true but it also does more like bundling and compressing. So its like comparing apple and pears :P

> Not sure how we can achieve this but it's worth discussing so we don't merge this and then switch the build system in a couple of weeks again. 

Why would we switch again? ESBuild and SWC does not help when it comes down to creating types which is the step that takes the longest.

> Next.js released a `canary` that seems to fix the issue (I haven't confirmed) so maybe our users upgrade.

It's not a fix it just sets the `esmExternals: false` by default for dependencies that are not built correctly. And this `esmExternals` means CJS so no threeshaing or any perfomance optimizations.

Please note that this PR also address a couple of other issues with the current Chakra Build Setup

cschroeter:
Disabling compression brings down the build time from 362s to 296s (~20%) on my old 2017 MacBook.

TimKolberger:
Switching to bundled packages will introduce breaking changes to our users. Currently the user land is allowed to import internal modules, just because the dist files are not bundled.
It is not our intention to allow this, but it is what it is. 
Because of this we stalled with the introduction of preconstruct.

Is there no alternative to fix the next 12 issue without a bundler?

segunadebayo:
This sounds reasonable @cschroeter. Thank you for the detailed report on this.

> favor build time over a correctly configured package setup.

I don't mean to suggest that we're favoring build time over a consistent package setup. I mean, in addition to the benefits we get from microbundle, I'd love to include a faster build time.

Because the build systems don't change frequently, I'd love a thorough look into other alternatives before making the decision on this one. I can set up a test for ESBuild + TSC to achieve the same and we'll compare the results.

Let's compare the options based on the numbers and choose the best one.

segunadebayo:
@TimKolberger, I'd like us to get out of that situation as well.

If we can solve that with extensive documentation on how to resolve the breaking change (if it happens), I'd prefer that.

We've been stuck on the current setup for this reason and not might be a good time to get to the bottom of this.

>Is there no alternative to fix the next 12 issues without a bundler?

Next.js will publish a patch with a temporary fix (fallback to CommonJS). In the end, we need to adjust our build system to output the correct ESM output.

cschroeter:
> 
> 
> Switching to bundled packages will introduce breaking changes to our users. Currently the user land is allowed to import internal modules, just because the dist files are not bundled. It is not our intention to allow this, but it is what it is. Because of this we stalled with the introduction of preconstruct.
> 
> Is there no alternative to fix the next 12 issue without a bundler?

Can you give a small code example what the user can do atm and what he can't do after this would be merged? I dont understand it tbh

cschroeter:
> Because the build systems don't change frequently, I'd love a thorough look into other alternatives before making the decision on this one. I can set up a test for (ESBuild + TSC, and Preconstruct) to achieve the same and we'll compare the results.

We already did this in my current company.. And introducing ESBuild (GoLang) or SWC (Rust) is pointless in my opinion. The only thing those tools help is when tranpiling to cjs and esm. If you run` build:cjs `or` build:esm` for example in button package - you can see that this transpiles within just 1.5s so the benefits would be small.

What takes the most time is `build:types` which is around 8.5s. Neither ESBuild nor SWC help us here since you have to transpile this using `tsc`.

The guy from SWC already workin on an TypeScript compiler in Rust (https://github.com/swc-project/swc/issues/571) but this is not yet done.

cschroeter:
By the way has anyone even tested if three shaking works with Chakra UI at the moment ? @segunadebayo @TimKolberger 
Because what I've seen doing this - I have my doubts.

segunadebayo:
Some resources I found along the way:

https://medium.com/digio-australia/migrating-an-npm-package-to-use-es-modules-d59877963d61
https://web.dev/publish-modern-javascript/
https://github.com/prateekbh/babel-esm-plugin

segunadebayo:
> Can you give a small code example of what the user can do atm and what he can't do after this would be merged? I don't understand it tbh

@cschroeter I think what Tim is referring to is:

```jsx
import { useOutsideClick } from "@chakra-ui/hooks/dist/useOutsideClick"
```

segunadebayo:
> By the way has anyone even tested if three shaking works with Chakra UI at the moment? @segunadebayo @TimKolberger Because of what I've seen doing this - I have my doubts.

Yes, I have at my previous company (tradeling.com). It works pretty reliably

cschroeter:
> ```js
> import { useOutsideClick } from "@chakra-ui/hooks/dist/useOutsideClick"
> ```

Ok understood. But did we indicate anywhere in the documentation that this is how it should be imported?
And anyway who would prefer `import { useOutsideClick } from "@chakra-ui/hooks/dist/useOutsideClick"` over `import { useOutsideClick } from "@chakra-ui/hooks"` :smile:

Not sure if I would talk about breaking chagnes if someone intentionally imported "internal" modules but thats just my 2cents

segunadebayo:
Hahaha. I totally understand that hence the reason I'm not too concerned about it 😃.

I just tried the same test with the `preconstruct` branch and it builds in 158.18s (no compression) compared to 332.43s for this branch.

segunadebayo:
It seems the build and test workflows are failing. Anything we can do to get it passing?

cschroeter:
> Hahaha. I totally understand that hence the reason I'm not too concerned about it 😃.
> 
> I just tried the same test with the `preconstruct` branch and it builds in 158.18s (no compression) compared to 332.43s for this branch.


Can you verify that `node -e "import('@chakra-ui/button')"` works with the preconstruct setup?

cschroeter:
> It seems the build and test workflows are failing. Anything we can do to get it passing?

Yeah I think I know whats the matter.Can fix it later

TimKolberger:
> Not sure if I would talk about breaking chagnes if someone intentionally imported "internal" modules but thats just my 2cents

Please don't get me wrong here, I'd love us to ship bundles. Just want to point out that this change will have an impact on our users beyond fixing a compatibility issue with Next 12.

> If we can solve that with extensive documentation on how to resolve the breaking change (if it happens), I'd prefer that.

If a user used an internal function in their codebase, there is no fix for that, besides copy-paste the previously required function to their own codebase. This is valid, but it feels odd to push this change out in patch version.

What do you think about bumping it a major version?

We are using preconstruct in all beta packages currently in development. IMO we should keep it consistent to keep it easy for us. I added `"type": "module"` in one of the next repositories to the package.json and was able to require it with `node -e "import('@chakra-ui/X')"` after it was bundled with preconstruct.

cschroeter:
@TimKolberger I would also not mind bumping a major version.

But I'm curious what is the benefit of Preconstruct over Microbundle? I read the docs briefly but I could not tell. Also it looks like Preconstruct is not actively mantained. https://github.com/preconstruct/preconstruct/graphs/contributors

TimKolberger:
The difference I see is that preconstruct is optimized for mono repos, where as microbundle is great for single packages.

With microbundle we need to define for each package how it needs to be built. (Define the npm script `build` in every package).

With preconstruct you define it globally, e.g. run "yarn start" in the repo root and all packages are getting built and watched without spinning your laptop fans on max 😆 
This reduces the complexity of this repository and is more efficient when watching _n_ packages at the same time.

> Also it looks like Preconstruct is not actively mantained.

Latest release is <1 month ago, seems ok to me.

MrHBS:
@TimKolberger There are beta packages?

segunadebayo:
After doing some extensive research with both systems Preconstruct and Microbundle. I'm more inclined to go with Preconstruct for the ease and speed it offers compared to microbundle.

Here's a PR from the Theme UI team that switches from Microbundle to Preconstruct to get the benefits of speed while ensuring that the code is bundled correctly. See https://github.com/system-ui/theme-ui/pull/1423.

The team at Reach UI also moved to Preconstruct from a custom babel setup (like we do). See https://github.com/reach/reach-ui/commit/3d1c2d7627162f86a509e879828303cd5752b171

This doesn't mean to imply that we're doing the same just because others are doing so, but I've tested both with our codebase and I find that Preconstruct scales better for our project. I'll update the Preconstruct branch and open a PR.

I appreciate your work on this @cschroeter 💖

PS: This was a pretty challenging decision to make but we need to move past this quickly

PR #4963, opened 2021-11-01
fix: reduce bundle size of framer motion

### LINKED ISSUE #4958: Investigate bloated bundle size for Checkbox
### Description

When I add a Checkbox to my App, I don't expect the bundle size to increase by 100kb

### Link to Reproduction

https://codesandbox.io/s/hungry-water-uhrnv?file=/src/index.tsx

### Steps to reproduce

`yarn create react-app my-app --template typescript`

Delete all the bloatware and than add Chakra UI

`arn add yarn add @chakra-ui/react @emotion/react@^11 @emotion/styled@^11 framer-motion@^4`

```tsx
import { Button, ChakraProvider } from "@chakra-ui/react";
import React from "react";

function App() {
  return (
    <ChakraProvider>
      <Button>Hello World</Button>
    </ChakraProvider>
  );
}

export default App;
```
**291K** 2.747da5cd.chunk.js

```tsx
import { Button, ChakraProvider, Checkbox } from "@chakra-ui/react";
import React from "react";

function App() {
  return (
    <ChakraProvider>
      <Button>Hello World</Button>
      <Checkbox>Hi</Checkbox>
    </ChakraProvider>
  );
}

export default App;
```
**391K** 2.eb16164b.chunk.js

### Chakra UI Version

Latest

### Browser

-

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

_No response_

noyanyan:
Closes #4958

## 📝 Description
Chakra UI uses framer motion in some components, and once one of  that component is used, the bundle size increases by 100kb.
For example, if you use only Button and Checkbox, the bundle size will be 100kb (gzip:31kb) larger than if you use Button. This prevents the user from using Chakra UI casually.

Fortunately, Framer Motion provides an official bundle size reduction method; by using motion's alternative component `m` and feature packages, we can reduce the bundle size up to 18kb.

Using checkbox (or other some components) increases bundle size by 100kb(gzip:31kb) . This means that 21% of the total bundle code is framer motion.

## 🚀 New behavior
The increase in bundle size can be kept to 18kb(gzipped).
There were several options for bundle size reduction. What do you think about the following issues?

### Limited or relatively many features
`m` does not have an animation function, so you will need to load the animation function separately. There are two features provided:
> `domAnimation`: This provides support for animations, variants, exit animations, and tap/hover/focus gestures. (~13kb)

> `domMax`: This provides support for all of the above, plus pan/drag gestures and layout animations. (~18kb)
`domAnimation`

`domAnimation` has less features, but the current component works fine as far as I can confirm. User-extended code might use features which are not in `domAnimation`, but this should not be a problem since Chakra Doc requires that `motion` be imported.
https://chakra-ui.com/guides/integrations/with-framer

### Loading Features
For bundle size, still we can 2 options.

To summarize, synchronous loading will increase the main bundle size, and lazy loading will increase the amount of code. But, I am not familiar with what happens when the library uses lazy loading internally.

**domAnimation or domMax, and Async or sync loading. What do you think?**

Maybe yes. For users who have been using the framer motion property without enclosing it in `motion`, this may cause the animation to stop working.

### DIFF
--- a/.changeset/flat-shoes-yell.md
+++ b/.changeset/flat-shoes-yell.md
@@ -1,0 +1,14 @@
+---
+"@chakra-ui/accordion": major
+"@chakra-ui/checkbox": major
+"@chakra-ui/menu": major
+"@chakra-ui/popover": major
+"@chakra-ui/popper": major
+"@chakra-ui/system": major
+"@chakra-ui/toast": major
+"@chakra-ui/tooltip": major
+"@chakra-ui/transition": major
+"@chakra-ui/docs": major
+---
+
+Reduce framer-motion bundle size
--- a/packages/accordion/stories/accordion.stories.tsx
+++ b/packages/accordion/stories/accordion.stories.tsx
@@ -169,6 +169,7 @@ const data = [
 ]
 
 export function Bug_2160() {
+  // see: https://github.com/chakra-ui/chakra-ui/issues/2160
   const inputRef = React.useRef<HTMLInputElement>()
   const [displayData, setDisplayData] = React.useState(data)
   const [filter, setFilter] = React.useState("")
@@ -194,6 +195,10 @@ export function Bug_2160() {
 
   return (
     <chakra.div padding={4}>
+      <Text>
+        Accordion (when used as filterable list of items) is not stealing focus
+        from input field.
+      </Text>
       <chakra.div mt={3} mb={12}>
         <chakra.input
           ref={inputRef}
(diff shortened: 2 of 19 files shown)


frysztak:
Well, I already tried ;) #3488

I think the real question here is -- does Checkbox really need to use framer-motion? I understand that it's convenient, but pulling entire library of this size just for a scale animation seems like a lot

noyanyan:
Thanks for pointing that out!
I hope your PR will be merged. I'll keep the PR open just to make more people aware of this issue.

noyanyan:
There is no longer any reason to keep this PR open, as Core team will consider optimizing this in the next major release.
Close this.
https://github.com/chakra-ui/chakra-ui/issues/4958#issuecomment-978057542

PR #4460, opened 2021-07-26
Fix/parse gradient

primos63:
Correct parseGradient function so that it correctly handles CSS functions.

Closes #4433

## 📝 Description

Using the `linear` function with a CSS function as a stop causes invalid CSS to be generated. Therefore, no gradient is applied to the element. If corresponding parameters are used in the `linear-gradient` CSS function, the gradient is correctly applied. The current `parseGradient` function only accounts for `stops` as values and not as functions.

## ⛳️ Current behavior (updates)

Applying a gradient using `linear` in the following JSX
```jsx
<Heading bgGradient="linear(to-r, green.200, pink.500 calc(20px + 20px))">
  Chakra-UI: Create accessible React apps with speed
</Heading>
```
results in the generation of invalid CSS 
```css 
background-image: linear-gradient(to right, var(--chakra-colors-green-200), var(--chakra-colors-pink-500) calc(20px);};};
```

While applying a gradient using `linear-gradient`
```jsx
<Heading
  bgImage="linear-gradient(
    to right, 
    var(--chakra-colors-green-200)), 
    var(--chakra-colors-pink-500 calc(20px + 20px))"
>
  Chakra-UI: Create accessible React apps with speed
</Heading>
```
results in CSS generation of 
```css
background-image: linear-gradient(to right, var(--chakra-colors-green-200), var(--chakra-colors-pink-500) calc(20px + 20px));
```

## 🚀 New behavior

Using the `linear` with the `calc` function will now produce the correct CSS output.
```css
background-image: linear-gradient(to right, var(--chakra-colors-green-200), var(--chakra-colors-pink-500) calc(20px + 20px));
```

PR #4970, opened 2021-11-01
fix: gradient parser when a position is specified

giuseppelt:
## 📝 Description

Fixed the gradient parsing logic when a position is specified

Actually parsing result:
-  `radial(circle at center, #000, #FFF)` produces `radial-gradient(circle at, center, #000, #FFF)` <-- See extra comma
-  `radial(circle at 50% 50%, #000, #FFF)` produces `radial-gradient(circle at, 50%, 50%, #000, #FFF)` <-- See extra comma

## 🚀 New behavior

Fixed the parsing logic.
Added a couple of test to secure the implementation.

No

## 📝 Additional Information

I just flatten the array before joining it. I could use `.flat` method but is not supported on older browser/node versions.

### DIFF
--- a/packages/styled-system/src/utils/parse-gradient.ts
+++ b/packages/styled-system/src/utils/parse-gradient.ts
@@ -64,7 +64,12 @@ export function parseGradient(value: string | null | undefined, theme: Dict) {
     // else, get and transform the color token or css value
     const key = `colors.${_color}`
     const color = key in theme.__cssMap ? theme.__cssMap[key].varRef : _color
-    return _stopOrFunc ? [color, _stopOrFunc].join(" ") : color
+    return _stopOrFunc
+      ? [
+          color,
+          ...(Array.isArray(_stopOrFunc) ? _stopOrFunc : [_stopOrFunc]),
+        ].join(" ")
+      : color
   })
 
   return `${_type}(${_values.join(", ")})`
--- a/.changeset/funny-laws-relate.md
+++ b/.changeset/funny-laws-relate.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/styled-system": patch
+---
+
+fix gradient parser when a position is specified
--- a/packages/styled-system/tests/parse-gradient.test.ts
+++ b/packages/styled-system/tests/parse-gradient.test.ts
@@ -115,6 +115,21 @@ describe("linear gradient", () => {
   })
 })
 
+describe("radial gradient", () => {
+  test("should parse gradient with a named position", () => {
+    expect(
+      parseGradient("radial(circle at center, #fff, #000)", theme),
+    ).toMatchInlineSnapshot(`"radial-gradient(circle at center, #fff, #000)"`)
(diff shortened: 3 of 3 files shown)


packages/styled-system/src/utils/parse-gradient.ts
```diff
@@ -64,7 +64,12 @@ export function parseGradient(value: string | null | undefined, theme: Dict) {
     // else, get and transform the color token or css value
     const key = `colors.${_color}`
     const color = key in theme.__cssMap ? theme.__cssMap[key].varRef : _color
-    return _stopOrFunc ? [color, _stopOrFunc].join(" ") : color
+    return _stopOrFunc
```
sohamsshah, line 67: Can we add a brief comment that explains what this operation does?

primos63: This change will revert and break what I did in #4460 - Correct parseGradient function so that it correctly handles CSS functions.

giuseppelt: As I stated in the summary, that operation simply flattens the array.
The fix will not break #4460, as my implementation passes all tests, including the ones you added. Instead #4460 introduced the problem.

`radial(farthest-side at 50% 50%, black, white)` produces `radial-gradient(farthest-side at,50%,50%, black, white)` <-- additional commas.
This happens because `_stopOrFunc` is not a string nor a func, but it's an array. The evaluations:
```js
color => 'farthest-corner';
_stopOrFunc => [ 'at', '50%', '50%' ];

[color, _stopOrFunc].join(" ") => ['farthest-corner', [ 'at', '50%', '50%' ]].join(" ") => "farthest-corner at,50%,50%]"
```
that because JS auto flattens the inner array with the default comma separator (you can check the docs).

So what I did, is simply check if the `_stopOrFunc` is an array. And if it is, I flatten it to
`['farthest-corner, 'at', '50%', '50%']` <-- NO INNER array
so the join operation actually works.

primos63: Have you tried to access the CSB that was created for this? It doesn't work for me. There are usually two rows, but here there's only one. This is the typical display:

image

 The CSB is having a problem loading `https://unpkg.com/@chakra-ui/provider@1.6.10`. Version 1.6.9 is there, but not 1.6.10. Also there's a `@chakra-ui/machine` in the package.json. Any idea how that got there?

giuseppelt: Sorry, don't know what you referring to.
I just fixed the function. I didn't touch anything else.

giuseppelt:
Can this be merged?

primos63:
Something happened to the build for this PR causing the CodeSandbox to fail. I extracted, built and tested the code locally. Looks good to me.

primos63 (approved):
Code looks good to me.

giuseppelt:
@sohamsshah can we merge this?

Issue #4914, opened 2021-10-25
Change how Color Mode behaves: allow automatic setting


Manc:
### Description

I've been digging into the code because I had trouble setting up Color Mode intuitively. In my opinion the following should generally be the default behaviour in 2021:

1. When a user visits the website for the first time, the system setting should be applied automatically (e.g. dark mode at night).
2. When the system's color mode changes, all the user's open app windows and websites should be expected to change to the new color mode.
3. The user may have the ability to lock in a different preferred color mode per website if there is an option for that on the website. This preference should then be persisted.
4. If the user has locked in a preferred color mode, they should still have the option to revert that back to "automatic". The persisted preference (e.g. in local storage) should then be deleted.

This is how it works on GitHub, https://developer.apple.com, and many other sites. It totally makes sense (to me).

<img width="1008" alt="Screenshot of the Footer of the Apple website" src="https://user-images.githubusercontent.com/811578/138649794-153d642a-affb-45d1-849a-2d5c24135bf6.png">

### Problem Statement/Justification

To my understanding of the Chakra UI code, this is not possible. I believe there's one central flaw: There is no differentiation between *current* color mode and *user preferred* color mode.

Currently, even if the config is set to consider the system's color mode, it may correctly detect the mode, but will then immediately persist that mode to local storage. Later, when the system mode changes, the website's mode will not change, because the previous mode has been persisted and will only ever change when the user click's, for example, the sun or moon button. Unless the user manually deletes the mode in local storage, the mode will never change back to "automatic", which should be the default in the most cases by far.

### Proposed Solution or API

`setColorMode` (of the `useColorMode()` hook) currently accepts a value of type `any`. It should accept `"light" | "dark" | "system"`.

[`localStorageManager`](https://github.com/chakra-ui/chakra-ui/blob/f77fd9999ec1105cf846cd830019d2c3ba5a0f4e/packages/color-mode/src/storage-manager.ts#L18) should only store the value if it is `"light" || "dark"` and *delete* the currently stored value if it is `"system"`.

A key problem is this bit ([reference](https://github.com/chakra-ui/chakra-ui/blob/f77fd9999ec1105cf846cd830019d2c3ba5a0f4e/packages/color-mode/src/color-mode-provider.tsx#L117)):

    if (useSystemColorMode) {
        removeListener = addListener(setColorMode)
    }

What happens here: Whenever the system color mode changes, set and persist the mode to local storage – forever.

[...]

PR #4989, opened 2021-11-04
fix(color-mode): fix colormode in system-view

### LINKED ISSUE #4987: Chakra 1.6.12 now features color mode issues (now dark by default, even though docs promise light by default. Also, overrides appear not to work)
### Description

Docs tell me that Chakra will be light color mode: https://chakra-ui.com/docs/features/color-mode#setup, but the default is your system color mode. Overrides appear to not work

### Link to Reproduction

https://codesandbox.io/s/dank-cloud-rymzh?file=/src/App.tsx

### Steps to reproduce

1. Install latest Chakra (1.6.12)
2. Set your system color mode to 'dark'
3. Wrap something in a `<ChakraProvider/>`
4. Notice that it is all black now, instead of light: 
image

Even if you set: 
```
const config = {
  initialColorMode: "light",
  useSystemColorMode: false,
}
```
 this now does not override the color mode. **It always takes your system color mode**

### Chakra UI Version

1.6.12

### Browser

All

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

_No response_

takethefake:
Closes #4987

## 📝 Description

Fix the behavior of color-mode when in system mode 

## 🚀 New behavior

Fixed color mode behavior priority in the following order:

- if `useSystemColorMode` is true system-color will be used as default - initial
  colormode is the fallback if system color mode isn't resolved

- if `--chakra-ui-color-mode` is defined through e.g. `ColorModeScript` this
  will be used

- if `colorModeManager` = `localStorage` and a value is defined for
  `chakra-ui-color-mode` this will be used

- if `initialColorMode` = `system` system-color will be used as default -
  initial colormode is the fallback if system color mode isn't resolved

- if `initialColorMode` = `'light'|'dark'` the corresponding value will be used

No

### DIFF
--- a/.changeset/wild-bikes-sneeze.md
+++ b/.changeset/wild-bikes-sneeze.md
@@ -1,0 +1,19 @@
+---
+"@chakra-ui/color-mode": patch
+---
+
+Fixed color mode behavior priority in the following order:
+
+- if `useSystemColorMode` is true system-color will be used as default - initial
+  colormode is the fallback if system color mode isn't resolved
+
+- if `--chakra-ui-color-mode` is defined through e.g. `ColorModeScript` this
+  will be used
+
+- if `colorModeManager` = `localStorage` and a value is defined for
+  `chakra-ui-color-mode` this will be used
+
+- if `initialColorMode` = `system` system-color will be used as default -
+  initial colormode is the fallback if system color mode isn't resolved
+
+- if `initialColorMode` = `'light'|'dark'` the corresponding value will be used
--- a/packages/color-mode/src/color-mode-provider.tsx
+++ b/packages/color-mode/src/color-mode-provider.tsx
@@ -83,22 +83,39 @@ export function ColorModeProvider(props: ColorModeProviderProps) {
      * reasons, do so after hydration.
      *
      * Priority:
-     * - system color mode
-     * - defined value on <ColorModeScript />, if present
-     * - previously stored value
+     * - if `useSystemColorMode` is true system-color will be used as default - initial
+     * colormode is the fallback if system color mode isn't resolved
+     *
+     * - if `--chakra-ui-color-mode` is defined through e.g. `ColorModeScript` this
+     * will be used
+     *
+     * - if `colorModeManager` = `localStorage` and a value is defined for
+     * `chakra-ui-color-mode` this will be used
+     *
+     * - if `initialColorMode` = `system` system-color will be used as default -
+     * initial colormode is the fallback if system color mode isn't resolved
+     *
+     * - if `initialColorMode` = `'light'|'dark'` the corresponding value will be used
      */
     if (isBrowser && colorModeManager.type === "localStorage") {
-      const mode = useSystemColorMode
-        ? getColorScheme(defaultColorMode)
-        : root.get() ||
-          colorModeManager.get() ||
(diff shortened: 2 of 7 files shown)


SkySails:
Thank you for following up on this issue so quickly @takethefake! I tried these changes out and they seem to work like I had intended it to from the beginning, so I'm happy!

Before you beat me to it, I also made a fix (https://github.com/SkySails/chakra-ui/commit/6934ce016b460f0706349a7d20c4017e094472ce) with a slightly different approach based on the thoughts in #4914 , namely by separating the _preference_ from the state. The result is probably exactly the same, but there _is_ something I am missing in this PR -  a way of setting the colorMode to "system" programmatically. Since I separated preference/state, in my case it was just a matter of calling `setColorMode("system")`.

Do you have any thoughts about this? Perhaps not important enough since you can reset it by manually removing the localStorage entry?

takethefake:
Hey @SkySails thanks for reaching out!

this sounds like a nice addition but I would rather focus on your suggestion in #4914 and focus on some misbehavior that somehow occurs in bundling as I may have figured out.

Currently, e.g. our CodeSandbox does use `system`-color by default, which can't be overwritten by `initialColorMode`

I tried to track this down with several approaches:

1. I wrote a bunch of tests that should verify the expected behavior of color-mode. Thats why i'm quite sure there ain't a logical fail here.

2. I've moved the files from the `color-mode`-package inside csb and noticed that the behavior changes rather if use the built `color-mode`-package or the runtime version (check out the csb [here](https://codesandbox.io/s/color-mode-csb-error-j6fu1?file=/src/App.tsx)

3. I've created a React project, which is based on [examples/create-react-app](https://github.com/chakra-ui/chakra-ui/tree/main/examples/create-react-app) [here](https://github.com/takethefake/color-mode-error) with the current version of the built from csb as packages. 

- in Dev mode everything works fine
- `y build && serve -s build` also works fine
- the deployed version on [vercel](https://color-mode-error.vercel.app/) has the same error as we have in csb

I'm not sure why this happens and will look into the bundled version tomorrow. I'm happy for any advice or ideas that you folks may have. @segunadebayo

SkySails:
Well done on the tests! I understand, one thing at a time is more reasonable.

I tried both the Vercel-hosted version and your repository and I am unable to toggle light/dark-mode using the button. I also played around with the rendering settings (namely the `prefers-color-scheme` setting) and I confirm that it hard-defaults to the system color scheme.

I will have a look at the implementation later to see if I can spot anything!

takethefake:
After diving deeper into this topic today I've checked the modules that are used in codesandbox and vercel and noticed, that somehow the wrong version was used for `@chakra-ui/color-mode`. After somehow the vercel version worked I did a redeploy of the codesandbox of this version and now everything seems to work. 

Might have been a hickup in the vercel space?

cschroeter:
LGTM

I would like to see some documenation on the website, explaining the hierachy in which the color mode is resolved

takethefake:
@cschroeter added some docs to this and the new docs repo

TimKolberger (approved):
Awesome that this feature has tests now  🚀 ✨  
Great job @takethefake 
LGTM!

aaronmcadam:
Really nice job on the tests, they're really comprehensive! 👏

PR #5010, opened 2021-11-05
fix(select): allow icon spacing to be customized via theme

### LINKED ISSUE #4938: Cannot customize field paddingEnd in Select component
### Description

When I apply `paddingEnd` to the `field` part in the `sizes` section of the `Select` theme, I expect it to apply to the `select` element, but the `select` element's `paddingEnd` is always set to `2rem`.

### Link to Reproduction

https://codesandbox.io/s/chakra-select--icon-spacing-nc8em?file=/src/index.tsx

### Steps to reproduce

Notice that it is impossible to customize the spacing between the icon and the field in the `Select` component via theming `paddingEnd` on the `field` part.

### Chakra UI Version

1.6.10

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

It's possible to work around this by specifying a `padding` value with short-hand syntax (e.g: `padding: 10px 20px 40px 20px`), but you lose theme/token awareness when doing so.

The `paddingEnd` value is hard-coded here: https://github.com/chakra-ui/chakra-ui/blob/445fc8b15bd1155a53ac95c5c1ba74d671062355/packages/select/src/select.tsx#L133-L136.

I found this related issue: https://github.com/chakra-ui/chakra-ui/issues/1146, but the solution mentioned there does not make sense given the current implementation.

jrolfs:
Closes #4938

## 📝 Description

Allow the spacing between the `field` and `icon` parts of the `Select` component to be applied from the theme.

The spacing is currently hard-coded as `2rem` in the `Select` component here: https://github.com/chakra-ui/chakra-ui/blob/445fc8b15bd1155a53ac95c5c1ba74d671062355/packages/select/src/select.tsx#L133-L136.

## 🚀 New behavior

Allow said spacing to be applied in the `Select` component's theme, while preserving the current default behavior as best as possible.

No

### DIFF
--- a/.changeset/funny-steaks-heal.md
+++ b/.changeset/funny-steaks-heal.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/select": patch
+"@chakra-ui/theme": patch
+---
+
+Fixed an styling issue where it was not possible to customize the icon spacing of the `Select` component.
--- a/packages/select/src/select.tsx
+++ b/packages/select/src/select.tsx
@@ -130,10 +130,11 @@ export const Select = forwardRef<SelectProps, "select">((props, ref) => {
     color,
   }
 
-  const fieldStyles: SystemStyleObject = mergeWith({}, styles.field, {
-    paddingEnd: "2rem",
-    _focus: { zIndex: "unset" },
-  })
+  const fieldStyles: SystemStyleObject = mergeWith(
+    { paddingEnd: "2rem" },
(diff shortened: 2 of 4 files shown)


TimKolberger (approved):
Thank you for working on this 💖 LGTM!

anubra266:
@segunadebayo can we merge this?

PR #5012, opened 2021-11-06
chore: remove plops

MrHBS:
I dont think they are used


segunadebayo:
Hey @MrHBS,

We use plop to scaffold new packages for faster development. They're definitely useful

PR #5019, opened 2021-11-07
docs: accessibility fixes for Search button

scottaohara:
The button to open the search text field had a `role=search` on it, as well as an `aria-label`.

[`role=search` is a type of landmark](https://www.w3.org/TR/wai-aria-1.2/#search), and not a role that indicates it is actionable.    Removing this role allows the control to be correctly exposed as a button element.

The `aria-label="Search the docs"` on the button suppresses the announcement of its child content, making it so that the key commands to open the search modal will not be announced.  

As "Search the docs" is already child content of the button (line 67), removing the `aria-label` allows the name of this button to still be announced correctly, and the additional "press ... to search" content will now be announced.

If you would like the search landmark to still be used / exposed, it must be an element that wraps this button, and not used on the button itself.

Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior

### DIFF
--- a/website/src/components/algolia-search.tsx
+++ b/website/src/components/algolia-search.tsx
@@ -44,7 +44,6 @@ export const SearchButton = React.forwardRef(
       <chakra.button
         flex="1"
         type="button"
-        role="search"
         mx="6"
         ref={ref}
         lineHeight="1.2"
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this PR @scottaohara. I admire your work in the accessibility space and it means a lot to see you contribute to Chakra UI 😊

scottaohara:
thanks @segunadebayo. just a quick thing I came across in doing some larger queries of use of `role=search` on the web, so thought it'd be more helpful to just make the change than spin up an issue.

Issue #4981, opened 2021-11-02
RangeSlider throw error - Cannot read properties of undefined


weiliao715:
### Description

When I using RangSlider with 3 thumbs, an TypeError is thrown when two of the thumbs are stacked. 

### Link to Reproduction

https://codesandbox.io/s/charkra-ui-rangeslider-0sjiz?file=/src/index.js

### Steps to reproduce

1. Create RangSlider with 3 thumbs. The index 0 and index 1 are stacked.

![image](https://user-images.githubusercontent.com/54960477/139886149-d275ed31-8099-44a3-8255-c7a9add3a275.png)

2. Move the thumb at index 2

3. See error

![image](https://user-images.githubusercontent.com/54960477/139899260-f5bdfca4-296c-4660-bc01-ae42d9390c9b.png)

### Chakra UI Version

1.6.10

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

I investigated the use-range-slider file and add few console log

![2](https://user-images.githubusercontent.com/54960477/139900088-afc0624f-8fda-4eae-8297-005d609b3804.PNG)

I discovered the error came from index exceeds the range. Suppose I have only index 0, 1, 2, we should not get to index 3.

![1](https://user-images.githubusercontent.com/54960477/139887959-69324c55-3851-4baf-911a-0f9e008c8e04.PNG)

Below function in use-range-slider  causes the index exceeds the range.

![3](https://user-images.githubusercontent.com/54960477/139900292-febe3a98-8699-4dfd-adcd-5a5949d26944.PNG)

anubra266:
Hi @weiliao715 do you still have this issue? Can't see any logged errors in your codesandbox reproduction.

weiliao715:
Hi @anubra266 the issue still persists. You need to move the thumb on the right side to trigger the error. Make sure you click on the left half of the circular thumb.

weiliao715:
@anubra266 sorry, click on the right half of the circular thumb

weiliao715:
@anubra266, @burhanuday thanks for helping me out with this error.

I found another problem related to range slider, and I opened a new error #5041

PR #5020, opened 2021-11-07
fix: range slider out of bounds error

burhanuday:
Closes https://github.com/chakra-ui/chakra-ui/issues/4981

## 📝 Description

fixed the bug in RangeSlider where an index out of bounds error would occur in
case of stacked thumbs

The logic to check if a thumb was stacked or not was incorrect. 

It would previously only check for duplicate values and if a duplicate value existed, it would consider that the current thumb was stacked

## 🚀 New behavior

The check has been changed to find if duplicates of the current closest thumbs are present

No

### DIFF
--- a/packages/slider/src/use-range-slider.ts
+++ b/packages/slider/src/use-range-slider.ts
@@ -347,9 +347,14 @@ export function useRangeSlider(props: UseRangeSliderProps) {
   const onPanSessionStart = (event: AnyPointerEvent) => {
     const pointValue = getValueFromPointer(event) || 0
     const distances = value.map((val) => Math.abs(val - pointValue))
-    const isThumbStacked = new Set(distances).size !== distances.length
     const closest = Math.min(...distances)
     let index = distances.indexOf(closest)
+
+    // check if the clicked thumb is stacked by checking if there are multiple
+    // thumbs at the same distance
+    const isThumbStacked =
+      distances.filter((distance) => distance === closest).length > 1
+
     // when two thumbs are stacked and the user clicks at a point larger than
     // their values, pick the next closest thumb
     if (isThumbStacked && pointValue > value[index]) {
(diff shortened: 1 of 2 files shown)


packages/slider/src/use-range-slider.ts
```diff
@@ -347,9 +347,14 @@ export function useRangeSlider(props: UseRangeSliderProps) {
   const onPanSessionStart = (event: AnyPointerEvent) => {
     const pointValue = getValueFromPointer(event) || 0
     const distances = value.map((val) => Math.abs(val - pointValue))
-    const isThumbStacked = new Set(distances).size !== distances.length
     const closest = Math.min(...distances)
     let index = distances.indexOf(closest)
+
+    // check if the clicked thumb is stacked by checking if there are multiple
+    // thumbs at the same distance
+    const isThumbStacked =
```
anubra266, line 355:
```suggestion
const isThumbStacked = distances.some((distance) => distance === closest)
```

packages/slider/src/use-range-slider.ts
```diff
@@ -347,9 +347,14 @@ export function useRangeSlider(props: UseRangeSliderProps) {
...
+      distances.filter((distance) => distance === closest).length > 1
```
anubra266, line 356:
```suggestion
const isThumbStacked = distances.some((distance) => distance === closest)
```

burhanuday: `.some` will return `true` even for 1 occurence. Need to check if multiple Thumbs are in the same location. In this case, `.some` will always return true

anubra266: Oh...we need more than one occurence. Cool then. Thanks for working on this

anubra266 (approved):

PR #5044, opened 2021-11-09
feat(popover): PopoverAnchor

### LINKED ISSUE #4733: feat(popover): manual trigger mode

dodas:
## 📝 Description

Follow up on https://github.com/chakra-ui/chakra-ui/pull/4734
Closes #4733

Demo: an Input with a popover

https://user-images.githubusercontent.com/14360171/141017574-e91c3e52-88ff-4229-a2d1-a42731c9a50d.mov

### DIFF
--- a/.changeset/twelve-trainers-eat.md
+++ b/.changeset/twelve-trainers-eat.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/popover": minor
+---
+
+Add `PopoverAnchor` component which allows you to set the Popover reference
+point without acting as a trigger.
--- a/packages/popover/src/popover.tsx
+++ b/packages/popover/src/popover.tsx
@@ -61,6 +61,22 @@ if (__DEV__) {
 }
 
 /**
+ * PopoverAnchor is element that is used as the positioning reference
+ * for the popover.
+ */
+export const PopoverAnchor: React.FC = (props) => {
+  // enforce a single child
+  const child: any = React.Children.only(props.children)
+  const { getAnchorProps } = usePopoverContext()
+
+  return React.cloneElement(child, getAnchorProps(child.props, child.ref))
+}
+
+if (__DEV__) {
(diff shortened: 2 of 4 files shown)


dodas:
I feel like the code around `anchorRef`/`maybeReferenceRef` is a bit unreadable. However, I think it's quite concise and does exactly what we need.

Any suggestions on how we can improve it are welcome!

dodas:
@segunadebayo
did you have a chance to look?
any feedback is welcome!

segunadebayo (approved):
I think this looks good @dodas. Great work 👌

Please add some documentation around this feature, and I'll get it merged.

bielas:
We would like to build simple autocomplete component with an Input element as a popover trigger. Trigger should happen on input click. When popover opens we should be able to type our autocomplete text without focusing on input once again - on popover open, focus should stay on input which at the same time is a trigger. The solution we took is to put initialFocusRef on the input element. However it creates one problem - popover doesnt close on clicking outside the whole component but on clicking again on the trigger - input element. How to achieve that, how to fix that problem?

PR #5050, opened 2021-11-10
feat: add chromatic visual regression tests 

isBatak:
## 📝 Description

This PR adds github action which deploys storybook views to chromatic and tests visual regression.

No visual regression tests

## 🚀 New behavior

Each commit will run visual regression test on chromatic service

No

## Additional notes

I don't have access to this repo, so someone will have to register this repo on chromatic.com to gain project token and then add it to the GH secrets (CHROMATIC_PROJECT_TOKEN)


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Build packages
+        run: yarn build
+
+        # 👇 Adds Chromatic as a step in the workflow
+      - name: Publish to Chromatic
+        uses: chromaui/action@v1
+        # Chromatic GitHub Action options
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          # 👇 Chromatic projectToken, refer to the manage page to obtain it.
```
burhanuday, line 50: do checkout the following options in the docs:
```
onlyChanged: true
exitOnceUploaded: true
```

first one will run visual diffing tests only on changed components. since chromatic limits visual diff tests for free projects, we don't want to run those if components haven't changed

`exitOnceUploaded: true` will terminate the action once components are sent to chromatic and will save compute time used by github actions

read here: [chromatic docs](https://www.chromatic.com/docs/github-actions#available-options)

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,52 @@
...
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile && yarn bootstrap
+
+      - name: Build packages
+        run: yarn build
```
burhanuday, line 42: My knowledge on Chakra's tooling is limited. Can you check if we need to run `build` again because afaik storybook also runs the build command for stories

burhanuday (commented):
Did you run the `lint`/`prettier` command for the whole project? might be the reason why you have so many extra changes

.github/workflows/chromatic.yml
isBatak: @burhanuday there is no `onlyChanged` option in the docs.
Are you talking about `exitZeroOnChanges`? It's `true` by default.

.github/workflows/chromatic.yml
isBatak: @burhanuday Yes it does, but this is monorepo managed by manypk, so all storybook dependencies has to be built first. 
For example, Button stories depends on `@chakra-ui/styled-system` which is built and managed by manypk, not storybook. 
I guess there is a way to setup webpack alias for all packages and tell Storybook to handle the transpilation and bundling of the `.ts` source files... but I'm not sure that's a good idea... there might be some differences.

This is how I understand the current setup, maybe I'm wrong. @segunadebayo can you confirm this?

.github/workflows/chromatic.yml
burhanuday: @isBatak I was referring to their TurboSnap feature here - https://www.chromatic.com/docs/turbosnap#enable

isBatak: I guess we need to update Storybook to v6.2 then

isBatak: Also in this phase I'm concerned with heuristic of change detection and false positives 
https://www.chromatic.com/docs/turbosnap#how-it-works

burhanuday: Hmm. I have been using chromatic with this enabled for a couple of months. I haven't faced the issue. But in any case, let's not enable it if they don't recommend doing so

TimKolberger:
Thanks for taking the time to create this PR @isBatak 💖 

I think chromatic would be useful, but is currently an overhead we do not want to maintain currently.
Gladly we now have an example implementation in this PR if we consider using chromatic.

Issue #5041, opened 2021-11-09
RangeSlider - thumbs are trapped when 3 thumbs are stacked


weiliao715:
### Description

When 3 of the thumbs in RangerSlider are stacked, none of them can be moved

### Link to Reproduction

https://codesandbox.io/s/rangeslider-3-stacked-thumbs-vce5h?file=/src/index.js

### Steps to reproduce

1. Go to the CodeSandbox above
2. You will see a ranger slider with 4 thumbs.

![image](https://user-images.githubusercontent.com/54960477/140981054-2f2bb210-6040-4b0d-8a0a-23b568afb2a4.png)

3. Try to move the left side thumbs

![image](https://user-images.githubusercontent.com/54960477/140981265-25e684d4-9390-4bb0-ab2a-55834176aeda.png)

4. All 3 thumbs are trapped there

### Chakra UI Version

1.6.12

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

I think the problem come from the index shown as below (use-range-slider.js)

![image](https://user-images.githubusercontent.com/54960477/140982080-05135960-1295-4b95-862d-99e097c35884.png)

Suppose index 0, 1, 2 thumbs are stacked, the "indexOf" function return index 0 in this case. And the later "index++" increases the index to 1. The range of index 1 thumb is limited by index 0 and index 2 thumbs. Therefore, none of them can be moved. I believe we want to get to index 2 in this case instead of a simple "index++"

PR #5081, opened 2021-11-13
fix: thumb does not move in case of stacked thumbs

burhanuday:
Closes https://github.com/chakra-ui/chakra-ui/issues/5041

## 📝 Description

fixed a bug where thumb would not move in case of stacked thumbs

The thumb does not move because the incorrect thumb is selected

## 🚀 New behavior

Fixed the index of thumb being selected

No

### DIFF
--- a/.changeset/happy-pants-tickle.md
+++ b/.changeset/happy-pants-tickle.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/slider": patch
+---
+
+fixed a bug where a thumb would not move in case of stacked thumbs
--- a/packages/slider/src/range-slider.tsx
+++ b/packages/slider/src/range-slider.tsx
@@ -23,14 +23,12 @@ interface RangeSliderContext
   name?: string | string[]
 }
 
-const [
-  RangeSliderProvider,
-  useRangeSliderContext,
-] = createContext<RangeSliderContext>({
-  name: "SliderContext",
-  errorMessage:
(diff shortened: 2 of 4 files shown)


segunadebayo:
Hi @burhanuday,

Can we write a test to cover this so we don't introduce regressions in the future?

Thank you.

burhanuday:
@segunadebayo not sure on how to emulate click on the `RangeSliderFilledTrack` component. We don't have any tests emulating clicks to start off of

burhanuday:
@segunadebayo i found the following issue - https://github.com/testing-library/testing-library-docs/issues/495 . It demonstrates an example of testing a slider with RTL but the author also states that it will only work in the browser since JSDOM does not support `getBoundingClientRect`. I dont think we can reliably test mouse interactions like drag on the slider. All our current tests on `RangeSlider` only test the keyboard related functionality. If possible, can you tag someone from the Chakra team who can help me with this?

weiliao715 (changes requested):
The start index may not be 0
Line 363 should be "index = index + thumbsAtPosition.length - 1"

weiliao715:
@segunadebayo thanks for merging.

Please check the line363 in use-range-slider.js, I think we should replace "index = thumbsAtPosition.length - 1" with "index = index + thumbsAtPosition.length - 1". It is an index increment instead of override the index value.

burhanuday:
Yes. @weiliao715 is correct. I'll make a seperate PR for this since this has been merged.

weiliao715:
@burhanuday Please create a PR to resolve this issue, thanks.

PR #5100, opened 2021-11-16
fix: prop-docs package

segunadebayo:
## 📝 Description

This PR fixes a build issue with the prop-docs package and adds types for the output files.

Based on the last release, there's currently no `dist` output for this package. This is an unintended side-effect from the migration to preconstruct.

## 🚀 New behavior

Added `yarn build:prop-docs` to the CI after running the build command so we can get the correct build artifacts for that package

No

## 📝 Additional Information

- Updated the website prop-table to be more typesafe
- The build script is quite slow and expect that we can improve this. It takes as long as `yarn preconstruct build` which is strange but understandable, perhaps it's related to the slowness of `ts-node`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,8 +49,8 @@ jobs:
       - name: Build packages
         run: yarn build
 
-      # - name: Build CRA templates
-      #   run: yarn build:templates ../cra-templates
+      - name: Build prop docs
+        run: yarn build:prop-docs
 
       - name: Run tests
         run: yarn test:ci
--- a/tooling/props-docs/src/build.ts
+++ b/tooling/props-docs/src/build.ts
@@ -188,7 +188,7 @@ ${esmPropExports}`,
 
 function writeTypes(componentInfo: ComponentInfo[]) {
   const typeExports = componentInfo
-    .map(({ exportName }) => `export declare const ${exportName}:  PropDoc`)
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -49,8 +49,8 @@ jobs:
       - name: Build packages
         run: yarn build
 
-      # - name: Build CRA templates
-      #   run: yarn build:templates ../cra-templates
+      - name: Build prop docs
+        run: yarn build:prop-docs
```
TimKolberger, line 53: What do you think about adding this to a `postbuild` script in the root package.json?

tooling/props-docs/src/build.ts
```diff
@@ -193,6 +186,48 @@ ${esmPropExports}`,
   )
 }
 
+function writeTypes(componentInfo: ComponentInfo[]) {
+  const typeExports = componentInfo
+    .map(({ exportName }) => `export declare const ${exportName}:  PropDoc`)
```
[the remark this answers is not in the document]
TimKolberger: Awesome idea to add types to this package 👍

.github/workflows/release.yml
segunadebayo: I'm trying to reduce the frequency of running this script due to its slowness. Adding it to the `postbuild` script will make it run for every workflow action (PR and Release) when we only need this in the release workflow.

TimKolberger (approved):

PR #5129, opened 2021-11-21
fix(icon): auto assign key for each path element if createIcon prop path is array

### LINKED ISSUE #5127: Getting "Each child in a list should have a unique "key" prop." warning when using createIcon with multiple paths
### Description

When i use createIcon function with multiple paths i get "Each child in a list should have a unique "key" prop." warning

### Link to Reproduction

https://codesandbox.io/s/exciting-drake-ch4k0?file=/src/index.tsx

### Steps to reproduce

_No response_

### Chakra UI Version

1.6.12

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

zgid123:
## 📝 Description

> Assign key for each path if we pass array of paths for createIcon

> We will need to pass key for each path manually

```ts
import { createIcon } from '@chakra-ui/icons';

export const HeartIcon = createIcon({
  displayName: 'HeartIcon',
  path: [
    <path key={0} stroke='none' d='M0 0h24v24H0z' fill='none' />,
    <path
      key={1}
      d='M19.5 13.572l-7.5 7.428l-7.5 -7.428m0 0a5 5 0 1 1 7.5 -6.566a5 5 0 1 1 7.5 6.572'
    />,
  ],
});
```

## 🚀 New behavior

> createIcon will assign key automatically for each path, we don't need to assign when passing to the prop

```ts
import { createIcon } from '@chakra-ui/icons';

import { defaultProps } from './common';

export const HeartIcon = createIcon({
  displayName: 'HeartIcon',
  defaultProps,
  path: [
    <path stroke='none' d='M0 0h24v24H0z' fill='none' />,
    <path
      d='M19.5 13.572l-7.5 7.428l-7.5 -7.428m0 0a5 5 0 1 1 7.5 -6.566a5 5 0 1 1 7.5 6.572'
    />,
  ],
});
```

## 📝 Additional Information

This will fix #5127

### DIFF
--- a/packages/icon/src/create-icon.tsx
+++ b/packages/icon/src/create-icon.tsx
@@ -32,14 +32,14 @@ export function createIcon(options: CreateIconOptions) {
   const {
     viewBox = "0 0 24 24",
     d: pathDefinition,
-    path,
     displayName,
     defaultProps = {},
   } = options
+  const path = React.Children.toArray(options.path)
 
   const Comp = forwardRef<IconProps, "svg">((props, ref) => (
     <Icon ref={ref} viewBox={viewBox} {...defaultProps} {...props}>
-      {path ?? <path fill="currentColor" d={pathDefinition} />}
+      {path.length ? path : <path fill="currentColor" d={pathDefinition} />}
     </Icon>
   ))
 
--- a/.changeset/young-guests-bow.md
+++ b/.changeset/young-guests-bow.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/icon": major
+---
+
+Minor change: auto assign key when passing array of paths to createIcon
--- a/packages/icon/stories/icon.stories.tsx
+++ b/packages/icon/stories/icon.stories.tsx
@@ -1,6 +1,6 @@
 import * as React from "react"
 import { Md3DRotation } from "react-icons/md"
-import { Icon, IconProps } from "../src"
+import { Icon, IconProps, createIcon } from "../src"
 
 export default {
   title: "Components / Media and Icons / Icon",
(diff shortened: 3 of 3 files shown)


packages/icon/src/create-icon.tsx
```diff
@@ -32,10 +33,14 @@ export function createIcon(options: CreateIconOptions) {
...
     viewBox = "0 0 24 24",
     d: pathDefinition,
-    path,
     displayName,
     defaultProps = {},
   } = options
+  let path = options.path
+
+  if (Array.isArray(path)) {
+    path = path.map((p) => React.cloneElement(p, { key: uuid() }))
```
ljosberinn, line 42: I'd refrain from adding an additional dependency for this for these reasons:
- the `path` property of the icon is highly unlikely to change. Using the array index ( `path.map((p, index) => cloneElement(p, { key: index }))` is sufficient here
- although unlikely, if `createIcon` is called multiple times, e.g. because its within the body of a rerendering component, `uuid` would re-execute and create a _new_ key for the _same_ path, which defeats the purpose of using a key in the first place.

ljosberinn:
Hi @zgid123 - thank you for the PR, this is a great idea! I've added a comment above :)

zgid123:
@ljosberinn updated

packages/icon/src/create-icon.tsx
```diff
@@ -32,10 +32,14 @@ export function createIcon(options: CreateIconOptions) {
...
     d: pathDefinition,
-    path,
     displayName,
     defaultProps = {},
   } = options
+  let path = options.path
+
+  if (Array.isArray(path)) {
+    path = path.map((p, index) => React.cloneElement(p, { key: index }))
+  }
```
segunadebayo, line 42: I think we can use React's internals to resolve this. Let's use `React.Children.toArray` instead.

```suggestion
  const path = React.Children.toArray(options.path)
```

segunadebayo (changes requested):
Thanks for working on this @zgid123.

- See my comment below.
- Please add a changeset by running `yarn changeset`

zgid123:
updated

packages/icon/src/create-icon.tsx
ljosberinn: while it is an array now, the mapping to add the key appears to be missing :)

ljosberinn: interesting, TIL [code](https://github.com/facebook/react/blob/cae635054e17a6f107a39d328649137b83f25972/packages/react/src/ReactChildren.js#L112) - looks like this is good to go then!

zgid123:
Sorry for spaming this PR. I just updated the code again

I changed the code from
```ts
{path ?? <path fill="currentColor" d={pathDefinition} />}
```
to
```ts
{path.length ? path : <path fill="currentColor" d={pathDefinition} />}
```

Because for old logic, `path` can be `undefined`, so if path is undefined, we will render `path` tag.

But for current logic, path is always an array, if we don't pass `path`, it will be empty array instead of `undefined`, so I change to that logic, if we pass `path` to the property, it will render path, else will render `path` tag

PR #5130, opened 2021-11-21
fix: drawer content ref returns null

georgekrax:
The ref prop of the <DrawerContent /> component returns null. The issue is with this component though, because that is where the ref is passed. I guess the issue is with the order of props passed, as they are overwritten with the spread {...rest}

Closes # 

## 📝 Description

The `<DrawerContent />` component `ref` props returns null

## 🚀 New behavior

No

### DIFF
--- a/.changeset/tasty-hounds-buy.md
+++ b/.changeset/tasty-hounds-buy.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/transition": patch
+---
+
+Fixed issue where the `ref` of `Slider` returns `null` due to prop override
--- a/packages/transition/src/slide.tsx
+++ b/packages/transition/src/slide.tsx
@@ -93,7 +93,6 @@ export const Slide = React.forwardRef<HTMLDivElement, SlideProps>(
       <AnimatePresence custom={custom}>
         {show && (
           <motion.div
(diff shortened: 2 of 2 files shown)


georgekrax:
Here is a [**correct** reproduction](https://discord.com/channels/660863154703695893/660865137137156146/911923079469490197) of what the component should do, provided by @primos63

segunadebayo:
CleanShot 2021-12-02 at 19 38 14@2x

Hi @georgekrax,

I just checked this and I don't think it's a bug. the `ref` to the content is only available when the drawer is open.
https://codesandbox.io/s/jolly-snow-b9s41?file=/src/App.tsx

georgekrax:
> CleanShot 2021-12-02 at 19 38 14@2x
> 
> Hi @georgekrax,
> 
> I just checked this and I don't think it's a bug. the `ref` to the content is only available when the drawer is open.
> https://codesandbox.io/s/jolly-snow-b9s41?file=/src/App.tsx

 @segunadebayo Yes, because you use the **correct** reproduction, which imports the fixed `<Drawer />`. Please import the component from `@chakra-ui/react`, and you will see that it always returns null.   Please revisit this pull request, and see the whole Discord discussion.

PR #5133, opened 2021-11-22
docs: update storybook url to enable composition

yoannfleurydev:
## 📝 Description

Hi Chakra UI team 😊 
This PR update the Storybook URL to the correct Storybook deployment URL so the automatic composition of Storybook works out of the box.

Using (automatic) composition, [we have the following result](https://start-ui.vercel.app/storybook) (check the bottom of the nav)

image

This is because the Storybook deployed on Netlify is old.

## 🚀 New behavior

Using (automatic) composition with the new URL, we should have the following result 

image

I think this change could impact users with Storybook v5 

### Before the merge of this PR

Users using Storybook v6 can do the following to their `main.js` file to enable composition

```js
  refs: {
    'chakra-ui': {
      title: 'Chakra UI',
      url: 'https://storybook.chakra-ui.com',
    },
    '@chakra-ui/react': { disable: true },
  },
```

### After the merge of this PR

I guess users using older Storybook can do the following to their `main.js` file to enable composition

```js
  refs: {
    'chakra-ui': {
      title: 'Chakra UI',
      url: 'https://chakra-ui.netlify.app',
    },
    '@chakra-ui/react': { disable: true },
  },
```

I tried to explain everything I can, feel free to reach me on Discord (I'm on the server **YoannFleuryDev#0001**) or on this PR if you want to discuss about it.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -91,7 +91,7 @@
   ],
   "storybook": {
     "title": "Chakra UI",
-    "url": "https://chakra-ui.netlify.app"
+    "url": "https://storybook.chakra-ui.com"
   },
   "devDependencies": {
     "@emotion/react": "^11.4.0",
--- a/.changeset/long-dancers-vanish.md
+++ b/.changeset/long-dancers-vanish.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react": patch
+---
+
+Update storybook url configuration for `@chakra-ui/react` to
+[https://storybook.chakra-ui.com](https://storybook.chakra-ui.com)


packages/react/package.json
```diff
@@ -91,7 +91,7 @@
   ],
   "storybook": {
     "title": "Chakra UI",
-    "url": "https://chakra-ui.netlify.app"
+    "url": "https://chakra-ui-storybook.vercel.app"
```
TimKolberger, line 94: What do you think about using the subdomain of chakra ui?

```suggestion
    "url": "https://storybook.chakra-ui.com"
```

TimKolberger (commented):
Awesome @yoannfleurydev! We had the same idea 😆

packages/react/package.json
yoannfleurydev: Oh yeah, I didn't know about the subdomain! That's better to avoid future changes! 👍🏻 

Awesome, I'll change this line with your suggestion.

yoannfleurydev:
I did the change but now the tests are failing, kind of strange, I'm rebased on the upstream main and I did not change any code.

TimKolberger (approved):
🚀

PR #5135, opened 2021-11-22
Fix use media query missing matches dependency

### LINKED ISSUE #5112: useMediaQuery not working -- using stale state value
### Description

When I resize the screen to be smaller, then larger, it continues to think it's small

### Link to Reproduction

🥲

### Description of the bug

At this line https://github.com/chakra-ui/chakra-ui/blob/a0de717c98e574e73f8a444f576fc738907230ab/packages/media-query/src/use-media-query.ts#L38

We're comparing `matches` to `currentMatches`, this is where the bug is. Because we do not supply `matches` as a dependency for the `useSafeLayoutEffects` in https://github.com/chakra-ui/chakra-ui/blob/a0de717c98e574e73f8a444f576fc738907230ab/packages/media-query/src/use-media-query.ts#L53

So we keep a reference to a function that has a stale reference to matches, therefore it keeps using the original value instead of the latest one.

There's two ways to fix it:

- Simply add `matches` to the dependencies array for the `useSafeLayoutEffect` hook
- Use a ref as a secondary storage measure to store `matches` and use that in comparison, `useRef` returns a stable value, so we do not need to add it to the dependencies.

I chose the latter in my patch because I did not read about `useSafeLayoutEffect` and do not know what excessive re-cals in it will cause

### Patch

Here's the patched one I'm using, the only change is the use of `useRef`

```tsx
import { useEnvironment } from '@chakra-ui/react-env'
import { isBrowser } from '@chakra-ui/utils'
import * as React from 'react'

const useSafeLayoutEffect = isBrowser ? React.useLayoutEffect : React.useEffect

// NOTE: This hook is a PATCHED version of the "useMediaQuery" hook
// that ships with @chakra-ui/react. The difference/fix we have is that
// we're storing the "matches" value in a ref for comparison, therefore
// fixing a bug by not using stale value generated at first-run time.

/**
 * React hook that tracks state of a CSS media query
 *
 * @param query the media query to match
 */
export default function useMediaQuery(query: string | string[]): boolean[] {
    const env = useEnvironment()
    const queries = Array.isArray(query) ? query : [query]
    const isSupported = isBrowser && 'matchMedia' in env.window

    const [matches, setMatches] = React.useState(
        queries.map(queryItem => (isSupported ? !!env.window.matchMedia(queryItem).matches : false))
    )
    const matchesRef = React.useRef(matches)

    useSafeLayoutEffect(() => {
        if (!isSupported) return undefined

        const mediaQueryList = queries.map(queryItem => env.window.matchMedia(queryItem))

        const listenerList = mediaQueryList.map(() => {
            const listener = () => {
                const isEqual = (prev: boolean[], curr: boolean[]) =>
                    prev.length === curr.length && prev.every((elem, idx) => elem === curr[idx])

                const currentMatches = mediaQueryList.map(mediaQuery => mediaQuery.matches)

                if (!isEqual(matchesRef.current, currentMatches)) {
                    matchesRef.current = currentMatches
                    setMatches(currentMatches)
                }
            }

            env.window.addEventListener('resize', listener)

            return listener
        })

        return () => {
            mediaQueryList.forEach((_, index) => {
                env.window.removeEventListener('resize', listenerList[index])
            })
        }
    }, [query])

    return matches
}
```

### Chakra UI Version

1.7.1

### Browser

Google Chrome 95

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Let me know if I can explain this further, hope the patch is helpful!

primos63:
Closes #5112 

## 📝 Description

> Corrected and optimized useMediaQuery hook. Created a new matchMedia mock that allows window resize events to be translated into the appropriate Media Query List Event. 

> useMediaQuery currently relies on resize events which makes the routine inefficient. It also unloads and reloads the media query event listeners which causes it to miss events. Previous implementations relied on an equality check between previous matches and current matches.

## 🚀 New behavior

> useMediaQuery now relies on the Media Query List Event delivered to the listener when a media change occurs. This eliminates the need for the equality check. Listening to the 'change' event on the Media Query List Object is more performant as the callback is only invoked when a specified media query is matched. There are no useEffect dependencies so the listeners are added and removed when the component is loaded and unloaded. No more missed events.
There is a new matchMedia mock which allows the testing of the useMediaQuery hook.


TimKolberger (commented):
This looks awesome @primos63 ✨

VinSpee:
This is a pretty critical issue for our team, and idea when it might go out in a release?

PR #5136, opened 2021-11-22
fix: allow comma to be a decimal separator in NumberInput

randex:
Doesn't close but helps with #4474 

## 📝 Description

Not every country has a dot (.) as a decimal separator, [some](https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma) are using comma (,) for this. In these countries it makes `NumberInput` unusable on iOS, because the digital keyboard only contains a comma which doesn't pass a regex test in `isFloatingPointNumericCharacter` resulting in comma not being allowed in the field.

image

Comma is not allowed in a decimal number in NumberInput.

## 🚀 New behavior

Comma is allowed in a decimal number in NumberInput, allowing to use this field on iOS in [regions](https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma) where comma is the default decimal separator. 

No

### DIFF
--- a/.changeset/eight-hornets-decide.md
+++ b/.changeset/eight-hornets-decide.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/number-input": patch
+---
+
+Allow comma to be a decimal separator in NumberInput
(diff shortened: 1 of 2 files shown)


nikolicstjepan:
Any update on this?

segunadebayo:
Thanks for working on this @randex. I just merged a PR that fixes this as well as some other issues in the NumberInput component.

We'll release an update shortly

Issue #5119, opened 2021-11-19
Confusing error "useStyles: `styles` is undefined" from a simple mistake


Southclaws:
### Description

If you wrap a ListItem in anything other than OrderedList/UnorderedList, you get a confusing useStyles ContextError

### Link to Reproduction

https://codesandbox.io/s/gracious-breeze-lwfur

### Steps to reproduce

Shove a ListItem inside another element with an as="ol"/as="ul" and you get this error.

I'm sure there's some logical reason why ListItem components _must_ be children of OrderedList/UnorderedList components and it doesn't work with a simple `as` on some other generic component like Box or Flex. But the error you get seems completely unrelated:

```
useStyles: `styles` is undefined. Seems you forgot to wrap the components in `<StylesProvider />` 
```

It makes you think "how does the rest of the app work then if StylesProvider isn't set" so it's certainly confusing and I've hit this twice recently (just enough time between both occurances for me to forget what the cause was!)

### Chakra UI Version

1.7.0

### Browser

Seems to be irrelevant to browsers, occurs in all

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Just a QoL thing, a better error that points to the actual cause and solution would be great, especially for newcomers to Chakra!

Patrick-Ullrich:
This happens because of the `useStyles` hook that allows for the `ListItem` to get its styles without having to go to the theme directly. 

### List: 
https://github.com/chakra-ui/chakra-ui/blob/1c866cd2f6c6469a481c72883dcd4e90e13eebbd/packages/layout/src/list.tsx#L62

### ListItem
https://github.com/chakra-ui/chakra-ui/blob/1c866cd2f6c6469a481c72883dcd4e90e13eebbd/packages/layout/src/list.tsx#L115

### Error Message from Context
https://github.com/chakra-ui/chakra-ui/blob/f77fd9999ec1105cf846cd830019d2c3ba5a0f4e/packages/system/src/providers.tsx#L51

### createContext
https://github.com/chakra-ui/chakra-ui/blob/f77fd9999ec1105cf846cd830019d2c3ba5a0f4e/packages/react-utils/src/context.ts#L26

## Potential Solution:
Could localize the `createContext` call:
```
const [StylesProvider, useStyles] = createContext<Dict<SystemStyleObject>>({
  name: "StylesContext",
  errorMessage:
[...]

PR #5137, opened 2021-11-22
fix(@chakra-ui/layout): style provider in list with specific error me…

Patrick-Ullrich:
Closes # https://github.com/chakra-ui/chakra-ui/issues/5119

## 📝 Description

Instead of using the shared `StylesProvider` from `@chakra-ui/system`, we are creating the Context directly in `List` to give a more relevant error message.

`list.tsx` is using `StylesProvider` and `useStyles` from `@chakra-ui/system` which has a generic error message.

## 🚀 New behavior

`list.tsx` creates the context directly to pass in a more concise error message.

No

## 📝 Additional Information

More components are using StylesProvider and if we are moving forward with this solution, we should refactor all components to use this pattern to give a better error message.

We could also investigate wrapping 
```const [StylesProvider, useStyles] = createContext<Dict<SystemStyleObject>>({
  name: "StylesContext",
  errorMessage:
    "useStyles: `styles` is undefined. Seems you forgot to wrap the components in a `<*List />` ",
})```
into a function that takes component name and calls `createContext` for us.

Looking for input.

### DIFF
--- a/.changeset/tricky-waves-serve.md
+++ b/.changeset/tricky-waves-serve.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/layout": minor
+---
+
+To give a better error message, list is creating the context directly to pass in
+a more clear error message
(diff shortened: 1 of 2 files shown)


Patrick-Ullrich:
What can we do to get this merged?

PR #5152, opened 2021-11-25
useBreakpointValue return type

guillaumervls:
Hi and thanks for the great work on Chakra UI !

We can simplify the code here because `useBreakpoint` cannot return `undefined`. That's also more convenient this way so we don't have to deal with the `undefined` case when using this hook.

Tell me if I'm missing something here.

Cheers !


dodas:
hey @guillaumervls 
I believe the return value of `useBreakpoint` / `useBreakpointValue` can be `undefined` during server side rendering, where we don't have access to the media query APIs.
The breakpoint(value) can only be determined after it's been mounted in the browser, so the first render will get `undefined`. You can use the second parameter of `useBreakpointValue` – the `defaultBreakpoint` to avoid this.

guillaumervls:
Thanks @dodas for your reply. Yes indeed. Sorry for the inconvenience.

PR #5195, opened 2021-12-05
fix(TS): #5194 Disallow props from outside the prop interface

### LINKED ISSUE #5194: [TS] Chakra currently allows any prop to be passed to component
### Description

When I try to pass a random prop, like "someRandomProp" to a component, I expect it to throw a TS error

### Link to Reproduction

https://codesandbox.io/s/admiring-gould-4wicm?file=/src/index.tsx

### Steps to reproduce

1. Render any chakra component
2. Pass to it a prop that does not exist in its prop interface (such as 'someRandomProp')
3. Observe that there is no type error thrown

### Chakra UI Version

1.7.2

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Interestingly enough, I found that when we do this, we do get the error:

```javascript
// Properly displays a typescript error
const three = <_Chakra.Button as='button' test />;
```

This implies to me that we have all the infra setup for getting this working, there may just be a small tweak we need to make to get it all flowing through.  I think we just need to set a default prop in `ComponentWithAs` -- let me put up a PR and see what you think!

andrewmtam:
Closes #5194  

## 📝 Description

Disallow props from outside the prop interface

// This should throw a type error
<Chakra.Button test/>

## 🚀 New behavior

// Now it does!
<Chakra.Button test/>

Possibly... but in a good way.  If people have been passing invalid props to their components, now there will be type-safety around it.

### DIFF
--- a/.changeset/sweet-adults-refuse.md
+++ b/.changeset/sweet-adults-refuse.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/system": patch
+---
+
+Disallow props that do not exist in the prop interface
--- a/packages/system/src/system.types.tsx
+++ b/packages/system/src/system.types.tsx
@@ -75,7 +75,7 @@ export type MergeWithAs<
   }
 
 export type ComponentWithAs<Component extends As, Props extends object = {}> = {
-  <AsComponent extends As>(
+  <AsComponent extends As = Component>(
     props: MergeWithAs<
       React.ComponentProps<Component>,
       React.ComponentProps<AsComponent>,
--- a/packages/system/tests/forward-ref.test.tsx
+++ b/packages/system/tests/forward-ref.test.tsx
@@ -1,0 +1,26 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import * as React from "react"
+import { chakra, forwardRef } from ".."
+
+/**
+ * These tests should fail while type checking
+ * if the typings for `ChakraComponent` change
+ * and create a type regression
+ */
+describe("`forward-ref` prop typings", () => {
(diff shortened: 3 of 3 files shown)


andrewmtam:
It looks like the test in the github action failed, but I also don't think I changed anything that would cause this test to fail.  Think it's something that a rerun might just fix?

andrewmtam:
(sorry I closed this erroneously x_x)

TimKolberger (approved):
Great work! Thank you for this PR @andrewmtam ✨ 

This could lead so some errors in userland projects, but IMO the good type of errors ✌️ because they will show which typings are missing in the component props interface.

LGTM 🚀

segunadebayo:
This fix looks good at first glance but I'm not sure if it'll play nicely with the dreaded `as` prop feature. I haven't tested it though

```jsx
// I should get props of "a" and "button"
<chakra.button as="a" href="..." />
```

TimKolberger:
I confirmed that it does not interfere with the as prop typings with this test:

Due the current test setup with swc the test files are not typechecked while running tests (Ill open an issue for this), but good old Editor TS integration passes.

```ts
// packages/system/tests/as-prop.test.tsx

it("should enforce required props for the chakra factory", () => {
  // @ts-expect-error Property 'thisIsARequiredProp' is missing
  const missingPropComp = <chakra.div as={CompWithRequired} />

  // make jest happy
  expect(true).toBe(true)
})
```

andrewmtam:
Awesome, thanks for checking into that for me. Let me know if there's anything else I can do here to help out

PR #5199, opened 2021-12-06
Control whether Tooltip can be closed with Esc key

### LINKED ISSUE #5156: Tooltip keydown listener causing performance issues
### Description

Using a lot of tooltips in a page causes a many keydown handlers to be registered on document via `useEventListener("keydown", onKeyDown)`. Typing then causes CPU spikes and ultimately feels sluggish and freezes due to large amount of listeners being handled.

`useEventListener` takes optional `env` parameter, which defaults to `document` when not provided. This makes it impossible to prevent Tooltip from subscribing keydown event handlers, in case the "Close on Esc" behavior is not desired. 

### Proposed Solution or API

Don't subscribe event listener if `env` is not defined. Add a prop `closeOnEsc` to control the behavior in Tooltip component.  

Proposed `useEventListener` effect implementation. The unfortunate side effect would be to update current usages of the hook to pass `document` in. 
```
  React.useEffect(() => {
    const node = runIfFn(env)

    node?.addEventListener(event, listener, options)
    return () => {
      node?.removeEventListener(event, listener, options)
    }
  }, [event, env, options, listener])

```
### Alternatives

_No response_

### Additional Information

_No response_

mlajtos:
Closes #5156 

## 📝 Description

This PR adds ability to turn off closing of the Tooltip via Esc key.

Every Tooltip registers a `keydown` handler to listen for Esc key press. If there are a lot of Tooltips on the page (>1000), the performance might suffer.

## 🚀 New behavior

User can opt-out from using Esc to close Tooltip via new optional `closeOnEsc` prop that has `true` as default value to preserve current behavior.

No, but it adds new prop to the Tooltip API. Current users are not affected.


packages/hooks/src/use-event-listener.ts
```diff
@@ -27,11 +27,15 @@ export function useEventListener<K extends keyof DocumentEventMap>(
   React.useEffect(() => {
     const node = runIfFn(env) ?? document
 
+    if (!handler) {
+      return () => {}
```
TimKolberger, line 31: We can safely remove the noop function here:

```suggestion
      return
```

TimKolberger (changes requested):
Thank you for working on this @mlajtos 💖

I added a tiny comment in my review. Could you kindly add a test for this? 
If you need support with that feel free to reach out!

mlajtos:
@TimKolberger Thank you for your suggestions. What kind of test you have in mind? For `useEventListener` hook, or `closeOnEsc` prop?

TimKolberger:
I'd like to make sure that the feature "closeOnEsc" will survive refactorings.
The implementation detail "useEventListener" is not very important for me (in this case).

TimKolberger (approved):

mlajtos:
@TimKolberger Thank you for merging. One question though – test ran okay on my machine, but they were failing on CI. What's up with that?

TimKolberger:
An unsolved issue, could not pinpoint it yet. It mostly happens with the yarn script test:ci which uses jests runinband option, so I am guessing there is somewhere a failed mock cleanup/setup.

PR #5206, opened 2021-12-08
ci: add action for snapshot releases

segunadebayo:
This PR improves our release process by adding:

- Pull request action to release a snapshot when we add the `release:canary` label. Adding the label will be controlled by maintainers.
- Main branch action to release an `rc` version when we push changes and the changeset hasn't been published yet.
- Notification to Discord and update changelog after release


.github/workflows/pr.yml
```diff
@@ -30,16 +41,85 @@ jobs:
...
+    name: Release Snapshot (if labelled)
+    runs-on: ubuntu-latest
+    if: contains( github.event.pull_request.labels.*.name, 'release:canary' )
+    steps:
+      - name: Cancel previous runs
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get Yarn cache path
```
cschroeter, line 65: I wouldn't rush to introduce caching. It has its downsides and if the speed gain is only a few seconds, I would reconsider.

.github/workflows/pr.yml
```diff
@@ -30,16 +41,85 @@ jobs:
...
 
       - name: Run tests
         run: yarn test:ci
+
+  release:
+    name: Release Snapshot (if labelled)
+    runs-on: ubuntu-latest
+    if: contains( github.event.pull_request.labels.*.name, 'release:canary' )
+    steps:
+      - name: Cancel previous runs
```
cschroeter, line 60: I would be very careful with that. This can cause quite strange side effects, depending on the time at which a run is aborted.

I would rather make sure that only one job running at any time, and the next one is queued.

.github/workflows/pr.yml
```diff
@@ -30,16 +41,85 @@ jobs:
...
+      - name: Load Yarn cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --ignore-engines
```
cschroeter, line 86: `--ignore-engines` should be fine for web dev - wonder why we need this

.github/workflows/pr.yml
```diff
@@ -30,16 +41,85 @@ jobs:
...
+
+      - name: Build packages
+        run: yarn build
+
+      - name: Build prop docs
+        run: yarn build:prop-docs
+
+      - name: Create .npmrc
```
cschroeter, line 94: You could also use `.yarnrc` since we are using yarn not npm :P

PR #5218, opened 2021-12-09
fix(styled-system): trigger release after cli fixed the theme typings

TimKolberger:
## 📝 Description

Typings for default theme went missing after the migration to preconstruct. 
This PR only consists of a changeset to trigger the release of `@chakra-ui/styled-system` with updated theme typings

No autocomplete for theme values like sizes and colors 😨 

## 🚀 New behavior

Autocomplete 🧘 

No


TimKolberger:
nvm - fixed in latest version

PR #5240, opened 2021-12-13
fix(radio): add aria-describedby to improve accessibility

SamHecquet:
## 📝 Description

> "The `aria-describedby` attribute is used to indicate the IDs of the elements that describe the object. It is used to establish a relationship between widgets or groups and text that described them." ([reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute))

- In order to improve accessibility, the radio element needs to follow the convention described by **MDN Web Docs** and to have the attribute "_aria-describedby_".

```js
<Radio aria-describedby="label_id">Hello</Radio>
```
`aria-describedby` is not in the list of the props allowed for `<input>` defined by [`getInputProps()`](https://github.com/SamHecquet/chakra-ui/blob/main/packages/radio/src/use-radio.ts#L200) so adding this prop when using `<Radio>` ends up adding it to the `<span>` following the input:

image

## 🚀 New behavior

image

No

## 📝 Additional Information
It was done for the checkbox component in this [PR](https://github.com/chakra-ui/chakra-ui/pull/3678)

### DIFF
--- a/.changeset/olive-lemons-lay.md
+++ b/.changeset/olive-lemons-lay.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/radio": patch
+---
+
+Add `aria-describedby` to the radio props to improve accessibility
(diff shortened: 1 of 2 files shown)


TimKolberger (approved):
✨ LGTM!

PR #5243, opened 2021-12-13
feat(cli): generate strict theme variant and size types

### LINKED ISSUE #5153: [TS] Strongly typed Chakra props
# Description

We would love the possibility of strongly typing Chakra props, in particular `variant` & `size`, when generating our custom theme types with `@chakra-ui/cli`.

You can **get intellisense to suggest your actual variants** (let's say `primary` and `secondary`) but the prop accepts `(string & {})`, so any string input is considered valid.

## In short
👉  **`variant="wrong-variant-name"` should be a TypeScript error**.

# Problem Statement/Justification

The purpose of the `chakra-cli tokens` [command](https://chakra-ui.com/docs/theming/advanced#usage) is to provide precise typings for your theme, so that you get auto-completion when writing props.

If you use the command, there is no reason to think these types could be bypassed, so **the possibility for any Chakra prop to be of type `(string & {})` is not justified**.

# Proposed Solution or API

## Proof of concept 

In the file `@chakra-ui/system/dist/types/system.types.d.ts`:

### Replace 

```ts
variant?: ThemeComponent extends keyof ThemeTypings["components"] ? ThemeTypings["components"][ThemeComponent]["variants"] | (string & {}) : string;
size?: ThemeComponent extends keyof ThemeTypings["components"] ? ThemeTypings["components"][ThemeComponent]["sizes"] | (string & {}) : string;
```

### By 

```ts
variant?: ThemeComponent extends keyof ThemeTypings["components"] ? ThemeTypings["components"][ThemeComponent]["variants"] : string;
size?: ThemeComponent extends keyof ThemeTypings["components"] ? ThemeTypings["components"][ThemeComponent]["sizes"] : string;
```

# Limitations
As far as I know, this naive solution would generate TS errors for users who created new variants without using the `@chakra-ui/cli` type generation.

But I'm sure we could find some cleaner way to enforce types only for themes that were typed with `@chakra-ui/cli`, maybe by differentiating the default `ThemeTypings` and the generated one.

# Alternatives

## Enforcing types for all style props

In a similar, but more strict approach, we would like to enforce using theme scales **in all Chakra style props**.

For example, for a scale where spaces are defined like `space-100`:

```ts
margin="space-100" // ok

margin="none" // error
margin={2} // error
margin="8px" // error
```

This is a bit hardcore but relevant in a strict TS codebase.
What do you think? :)

# Additional Information

Here would be the result, as a developer:

## Before

Screenshot 2021-11-25 at 17 22 34

## After

Screenshot 2021-11-25 at 17 23 02

TimKolberger:
Closes #5153

## 📝 Description

The type for variants and sizes are hardcoded as loose types:

```tsx
<Button variant="abc" /> // valid type but not available in the theme
```

```tsx
<Button variant="abc" /> // invalid
```  

## 🚀 New behavior

```bash
chakra-cli tokens --strict-component-types
```

will generate following types:

```ts
  // ...

  components: {
    Button: {
      sizes: "sm" | "md" | "lg"
      variants: "solid" | "outline" | "ghost"
    }
  }
}
```

No, because of opt in.

## Additional Information

This PR could have a follow up to expand the strict mode to the theme tokens as well.

### DIFF
--- a/.changeset/empty-tools-smash.md
+++ b/.changeset/empty-tools-smash.md
@@ -1,0 +1,21 @@
+---
+"@chakra-ui/system": minor
+"@chakra-ui/cli": minor
+---
+
+Use the feature flag `--strict-component-types` for `@chakra-ui/cli tokens` to
+generate strict component type for the theming props `variant` and `size`.
+
+```bash
+chakra-cli tokens --strict-component-types
+```
+
+```tsx
+// before
+<Button variant="abc" />
+// valid type but variant is not available in the theme
+
+// after
+<Button variant="abc" /> // invalid
+// Type '"abc"' is not assignable to type '"link" | "outline" | "ghost" | "solid" | "unstyled" | undefined'.
+```
--- a/packages/system/src/system.types.tsx
+++ b/packages/system/src/system.types.tsx
@@ -10,12 +10,12 @@ import * as React from "react"
 
 export interface ThemingProps<ThemeComponent extends string = string> {
   variant?: ThemeComponent extends keyof ThemeTypings["components"]
-    ? ThemeTypings["components"][ThemeComponent]["variants"] | (string & {})
-    : string
+    ? ThemeTypings["components"][ThemeComponent]["variants"]
+    : never
   size?: ThemeComponent extends keyof ThemeTypings["components"]
-    ? ThemeTypings["components"][ThemeComponent]["sizes"] | (string & {})
-    : string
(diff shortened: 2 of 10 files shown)


segunadebayo (approved):
Love this improvement to the CLI, Tim.

Feel free to send a follow-up PR for the `--strict-tokens` flag as well.

dbousamra:
If I don't use CLI, but I have custom variants, how do I use it? I don't want to generate types using the CLI, but would like to use custom variants.

PR #5253, opened 2021-12-16
Fix 3884 text styles not applied correctly

### LINKED ISSUE #3884: textStyles defined inside extendTheme aren’t applied correctly

primos63:
Closes #3884  

## 📝 Description

> Process text and layer styles correctly

> If a component already has a property for which a textStyle property would overwrite, the update is skipped.

## 🚀 New behavior

> Each style property is processed in precedence order. For each style the CSS for the style is generated, the CSS is then merged with an accumulator for the final computed CSS. This ensures that properties, including responsive ones, have the correct values.


packages/system/src/system.ts
```diff
@@ -38,20 +46,38 @@ interface GetStyleObject {
...
+    //   sx
+    stylePrecedence.push(__css)
+    stylePrecedence.push(runIfFn(baseStyle, props))
+    stylePrecedence.push(otherProps)
+    stylePrecedence.push(styleProps)
+    stylePrecedence.push(sx)
+
+    const computedCSS = stylePrecedence.reduce((accStyle: any, style: any) => {
+      return mergeWith(accStyle, css(style)(theme))
+    }, {})
```
TimKolberger, line 77: The `mergeWith` function causes to increase the runtime complexity to increase from O(n) to O(n^2). Sadly this is in the `styled-system` algorithm hot path, which will be called by every chakra component on every render. 

A very simple loop test gave us the following average time in ms for a run with 10000 iterations:

PR solution: 0.045435ms
Current main branch: 0.02831ms

The large chunk of the performance regression is caused by running `mergeWith` and `css` function in a loop, compared to the current approach that uses `Object.assign` and runs the `css` function once.

TimKolberger (commented):
Thank you @primos63 for this PR :sparkles:

I really like the improvements you did to the code structure and readability.

Kindly see my comments in the code to see the performance impact.
I think we may need a re-architecture of the styled-system that's focused heavily on performance. When we do that, **we will include the idea behind your solution**.

We are not able to move forward with this PR. Feel free to reach out to me to discuss this more!

Thanks for understanding

PR #5276, opened 2021-12-22
fix(grid): move `area` prop from Grid to GridItem

prashanth-cpaul:
## 📝 Description

Allows for usage of GridItem's that have named template areas to be used in conjunction with a Grid that has a defined template area.
This matches the usage pattern of a grid container and its children using grid template areas.
Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-areas

Currently `GridItem` doesn't have an `area` prop which means we cannot have a `Grid` that defines a grid template area to contain `GridItems` with a named grid area. The `area` prop is currently defined on `Grid` so the following works but is not semantically correct:
```
<Grid templateAreas='"one two three"'>
   <Grid area='one'>one</Grid>
   <Grid area='two'>two</Grid>
   <Grid area='three'>three</Grid>
</Grid>
```

## 🚀 New behavior
Have moved the `area` prop from `Grid` to `GridItem` so the following is now possible:
```
<Grid templateAreas='"one two three"'>
   <GridItem area='one'>one</Grid>
   <GridItem area='two'>two</Grid>
   <GridItem area='three'>three</Grid>
</Grid>
```

Yes. Users that have implemented `Grid` with template areas and have setup named template areas using the `area` prop on nested `Grid` elements will need to change their implementation to use `GridItem` with the `area` prop instead.

### DIFF
--- a/.changeset/itchy-singers-behave.md
+++ b/.changeset/itchy-singers-behave.md
@@ -1,0 +1,18 @@
+---
+"@chakra-ui/layout": patch
+---
+
+### Add support for `area` prop on `GridItem`
+
+Deprecated `area` prop on `Grid` and added support for `area` prop to be used
+with `GridItem` instead. This allows for usage of `GridItem`'s that have named
+template areas to be used in conjunction with a `Grid` that has a defined
+template area.
+
+```jsx live=false
+<Grid templateAreas='"one two three"'>
+   <GridItem area='one'>one</Grid>
+   <GridItem area='two'>two</Grid>
+   <GridItem area='three'>three</Grid>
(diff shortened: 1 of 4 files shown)


segunadebayo (changes requested):
Great work here @prashanth-cpaul. I totally agree with your thoughts on this.

To avoid breaking changes, I'd recommend:

- Add a `@deprecated` js doc comment in the `area` prop for the `Grid` component but **keep it there.**
- Add `area` prop to the `GridItem` component (you've done this already)

Please add a changeset as well by running `yarn changeset` and following the prompt

PR #3732, opened 2021-04-03
Feat/gh3675 change spinner position

jatin33:
Closes #3675

## 📝 Description

> This PR adds a prop to button component/package that allows to change position of loading spinner

## ⛳️ Current behavior (updates)

> Currently position of loading spinner is fixed

## 🚀 New behavior

> Now Developer will be able to change/customize the position of loading spinner.

## 💣 Is this a breaking change (Yes/No):

No

PR #5284, opened 2021-12-23
fix(button): use iconSpacing for ButtonSpinner

### LINKED ISSUE #5056: Make `iconSpacing` control `ButtonSpinner` margin
### Description

Currently `iconSpacing` property of `Button` component only controls the spacing between `leftIcon` / `rightIcon` and button text. It doesn't control the spacing between `ButtonSpinner` and button text when `isLoading={true}`.

From my point of view, it should control it. Otherwise, there is a design shift when `isLoading` goes from `false` to `true`.

Currently `iconSpacing` property of `Button` component only controls the spacing between `leftIcon` / `rightIcon` and button text. It doesn't control the spacing between `ButtonSpinner` and button text when `isLoading={true}`.

### Proposed Solution or API

Controling ButtonSpinner margin via `iconSpacing` property.

### Alternatives

_No response_

### Additional Information

_No response_

marek-sed:
Closes #5056 

## 📝 Description

> iconSpacing got removed from spinner as part of this [pull request](https://github.com/chakra-ui/chakra-ui/pull/3732)

> Spinner has hardcoded spacing of **0.5**

## 🚀 New behavior

> passing iconSpace to spinner and using it for spacing

No

### DIFF
--- a/packages/button/src/button-spinner.tsx
+++ b/packages/button/src/button-spinner.tsx
@@ -37,12 +37,12 @@ export const ButtonSpinner: React.FC<ButtonSpinnerProps> = (props) => {
       display: "flex",
       alignItems: "center",
       position: label ? "relative" : "absolute",
-      [marginProp]: label ? "0.5rem" : 0,
+      [marginProp]: label ? spacing : 0,
       fontSize: "1em",
       lineHeight: "normal",
       ...__css,
     }),
-    [__css, label, marginProp],
+    [__css, label, marginProp, spacing],
   )
 
   return (
--- a/packages/button/src/button-spinner.tsx
+++ b/packages/button/src/button-spinner.tsx
@@ -21,7 +21,7 @@ export const ButtonSpinner: React.FC<ButtonSpinnerProps> = (props) => {
   const {
     label,
     placement,
-    spacing = "0.5",
+    spacing = "0.5rem",
     children = <Spinner color="currentColor" width="1em" height="1em" />,
     className,
     __css,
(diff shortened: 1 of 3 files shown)


packages/button/src/button-spinner.tsx
```diff
@@ -37,12 +37,12 @@ export const ButtonSpinner: React.FC<ButtonSpinnerProps> = (props) => {
       display: "flex",
       alignItems: "center",
       position: label ? "relative" : "absolute",
-      [marginProp]: label ? "0.5rem" : 0,
+      [marginProp]: label ? spacing : 0,
```
TimKolberger, line 40: I think we should add a default value for `spacing`. E.g. 

```
// Line 24 
spacing = "0.5rem"
```

TimKolberger (changes requested):
Thank you for this PR @marek-sed 💖 
I added a comment, then it looks good to me 🚀

packages/button/src/button-spinner.tsx
```diff
@@ -21,7 +21,7 @@ export const ButtonSpinner: React.FC<ButtonSpinnerProps> = (props) => {
   const {
     label,
     placement,
-    spacing,
+    spacing = "0.5",
```
TimKolberger, line 24: `0.5` refers to a theme token (= `0.125rem`), let's the previous value `0.5rem` to keep the look consistent.

```suggestion
    spacing = "0.5rem",
```

marek-sed: sry was on autopilot, in our app we are only using theme-token values :)

updated.

TimKolberger (approved):

PR #5287, opened 2021-12-24
fix(tooltip): add some background props for options to style arrow bg

### LINKED ISSUE #5283: The `background` and `bg` properties of the `Tooltip` component are not interchangeable
### Description

When I use the Tooltip component, I expect both the `background` and `bg` properties to also style the arrow when `hasArrow` is true.  Instead, using `bg` styles the arrow with the correct color, while using `background` does not change the arrow color.

### Link to Reproduction

https://codesandbox.io/s/charming-thompson-4sdmf

### Steps to reproduce

1. Hover over the text on the left.  Notice the arrow is the same color as the tooltip.
Screen Shot 2021-12-22 at 2 38 19 PM

2. Hover over the arrow on the right.  Notice the arrow is black, while the tooltip is blue.
Screen Shot 2021-12-22 at 2 38 31 PM

You can see in the code that the only difference between the two elements is the property used to style the background.

### Chakra UI Version

1.7.3, also present in 1.5.1

### Browser

Chrome 96

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

This appears to be because the Tooltip component only considers the `bg` prop when styling the arrow: https://github.com/chakra-ui/chakra-ui/blob/main/packages/tooltip/src/tooltip.tsx#L81

ashalfarhan:
Closes #5283

## 📝 Description

Fix an issue where arrow tooltip background color only consider `bg` props.

Currently `Tooltip` only consider `bg` props.

## 🚀 New behavior

While, in this PR, I try to add some additional bg props to be considered while customize the arrow background color.

No.

### DIFF
--- a/packages/tooltip/src/tooltip.tsx
+++ b/packages/tooltip/src/tooltip.tsx
@@ -73,14 +73,22 @@ export const Tooltip = forwardRef<TooltipProps, "div">((props, ref) => {
     hasArrow,
     bg,
     portalProps,
+    background,
+    backgroundColor,
+    bgColor,
     ...rest
   } = ownProps
 
-  if (bg) {
-    styles.bg = bg
-    styles[popperCSSVars.arrowBg.var] = getCSSVar(theme, "colors", bg)
-  }
+  const userDefinedBg = background ?? backgroundColor ?? bg ?? bgColor
 
+  if (userDefinedBg) {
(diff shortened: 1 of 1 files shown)


TimKolberger (approved):
LGTM 🚀

PR #5297, opened 2021-12-28
fix(modal): show footer with size=full

### LINKED ISSUE #4903: Modal footer is not visible on iOS browsers
### Description

Footers of modals with ```size="full"``` and ```scrollBehavior="inside"```  are not visibile on iOS 

### Link to Reproduction

https://codesandbox.io/s/romantic-greider-xwuy6

### Steps to reproduce

1. Open the sandbox link
2. Click on Open modal
3. The footer is not visible

### Chakra UI Version

1.6.10

### Browser

Safari and Chrome 

### Operating System

iOS 14 & 15

### Additional Information

The same issue has been reported before here: https://github.com/chakra-ui/chakra-ui/issues/3684
It's been subsequently fixed but it seems it's now happening again. 

I did some digging and it seems this is caused by `ModalContent`'s `minHeight` being `100vh`:
https://github.com/chakra-ui/chakra-ui/blob/3adda59f95986abfc7b4d97a855900270cde2d3c/packages/theme/src/components/modal.ts#L86

The problem goes away if the codesandbox example is modified like this:
```
<ModalContent minHeight="-webkit-fill-available">
...
</ModalContent>
```

TimKolberger:
Closes #4903

## 📝 Description

`<Modal size="full">` is currently not completely visible on iOS.

Footer is missing

## 🚀 New behavior

Use `@supports(min-height: -webkit-fill-available)` query to apply the correct height.

No

## 📝 Additional Information

Really looking forward to [Dynamic Viewport Units](https://css-tricks.com/the-large-small-and-dynamic-viewports/)

### DIFF
--- a/.changeset/twelve-papayas-judge.md
+++ b/.changeset/twelve-papayas-judge.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/theme": patch
+---
+
+Fixed an issue where the Modal Footer was out of the viewport for `size="full"`.
--- a/packages/theme/src/components/modal.ts
+++ b/packages/theme/src/components/modal.ts
@@ -83,7 +83,14 @@ const baseStyle: PartsStyleFunction<typeof parts> = (props) => ({
 function getSize(value: string): PartsStyleObject<typeof parts> {
   if (value === "full") {
     return {
-      dialog: { maxW: "100vw", minH: "100vh", my: 0 },
+      dialog: {
+        maxW: "100vw",
+        minH: "100vh",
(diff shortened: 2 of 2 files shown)


segunadebayo:
Great work here @TimKolberger. Maybe this is something we can resolve from within `styled-system` itself by transforming height, min-height, and max-height `100vh` values to this:

```js
 minH: "100vh",
 "@supports(min-height: -webkit-fill-available)": {
   minH: "-webkit-fill-available",
},
```

PR #5335, opened 2022-01-06
fix(descendant): enable property access via index

### LINKED ISSUE #5317: @chakra-ui/descendant fails TypeScript strict type-checking, breaking CI
### Description

When using chakra-ui in a project with `noPropertyAccessFromIndexSignature` enabled (a strict type-checking option in tsconfig.json), @chakra-ui/descendant causes tsc to fail and complain due to two easily fixable errors

### Link to Reproduction

https://www.npmjs.com/package/@chakra-ui/cra-template-typescript

### Steps to reproduce

1. Clone any chakra-ui project/template (`npx create-react-app issue-chakra-types --template @chakra-ui/typescript`)
2. Add `"noPropertyAccessFromIndexSignature": true` to `tsconfig.json`->`compilerOptions`
3. Have tsc check types, f.e. with a `package.json` script (`"check": "tsc"`)
4. Watch tsc cry about two minor mistakes:
```
node_modules/@chakra-ui/descendant/src/descendant.ts:64:31 - error TS4111: Property 'index' comes from an index signature, so it must be accessed with ['index'].

64       descendant.node.dataset.index = descendant.index.toString()
                                 ~~~~~

node_modules/@chakra-ui/descendant/src/use-descendant.ts:60:50 - error TS4111: Property 'index' comes from an index signature, so it must be accessed with ['index'].

60     const dataIndex = Number(ref.current.dataset.index)
                                                    ~~~~~
```

### Chakra UI Version

1.7.3

### Browser

Node 17.3.0 / TS 4.5.4

### Operating System

- [x] macOS
- [ ] Windows
- [X] Linux

### Additional Information

Perhaps do some CI checks with a strict `tsconfig.json`: :wink: 
```
"strict": true /* Enable all strict type-checking options. */,
"noImplicitAny": true /* Enable error reporting for expressions and declarations with an implied `any` type.. */,
"strictNullChecks": true /* When type checking, take into account `null` and `undefined`. */,
"strictFunctionTypes": true /* When assigning functions, check to ensure parameters and the return values are subtype-compatible. */,
"strictBindCallApply": true /* Check that the arguments for `bind`, `call`, and `apply` methods match the original function. */,
"strictPropertyInitialization": true /* Check for class properties that are declared but not set in the constructor. */,
"noImplicitThis": true /* Enable error reporting when `this` is given the type `any`. */,
"useUnknownInCatchVariables": true /* Type catch clause variables as 'unknown' instead of 'any'. */,
"alwaysStrict": true /* Ensure 'use strict' is always emitted. */,
"noUnusedLocals": true /* Enable error reporting when a local variables aren't read. */,
"noUnusedParameters": true /* Raise an error when a function parameter isn't read */,
"noImplicitReturns": true /* Enable error reporting for codepaths that do not explicitly return in a function. */,
"noFallthroughCasesInSwitch": true /* Enable error reporting for fallthrough cases in switch statements. */,
"noUncheckedIndexedAccess": true /* Include 'undefined' in index signature results */,
"noImplicitOverride": true /* Ensure overriding members in derived classes are marked with an override modifier. */,
"noPropertyAccessFromIndexSignature": true /* Enforces using indexed accessors for keys declared using an indexed type */,
```

antwigambrah:
Closes #5317 

>  Strict type checking with "noPropertyAccessFromIndexSignature": true enables accessing a field via the “dot” (obj.key) syntax, and “indexed” (obj["key"]) .Property is accessed via the "dot"(obj.key) syntax

## 🚀 New behavior

>Since the Property 'index' comes from an index signature, it is now accessed with ['index']

No

### DIFF
--- a/packages/descendant/src/descendant.ts
+++ b/packages/descendant/src/descendant.ts
@@ -31,7 +31,7 @@ export type Descendant<T, K> = DescendantOptions<K> & {
  */
 export class DescendantsManager<
   T extends HTMLElement,
-  K extends Record<string, any> = {}
+  K extends Record<string, any> = {},
 > {
   private descendants = new Map<T, Descendant<T, K>>()
 
@@ -61,7 +61,7 @@ export class DescendantsManager<
     this.descendants.forEach((descendant) => {
       const index = descendants.indexOf(descendant.node)
       descendant.index = index
-      descendant.node.dataset.index = descendant.index.toString()
+      descendant.node.dataset["index"] = descendant.index.toString()
     })
   }
 
--- a/packages/descendant/src/use-descendant.ts
+++ b/packages/descendant/src/use-descendant.ts
@@ -26,13 +26,12 @@ export interface UseDescendantsReturn
   NB:  I recommend using `createDescendantContext` below
  * -----------------------------------------------------------------------------------------------*/
 
-const [
-  DescendantsContextProvider,
-  useDescendantsContext,
-] = createContext<UseDescendantsReturn>({
-  name: "DescendantsProvider",
-  errorMessage: "useDescendantsContext must be used within DescendantsProvider",
-})
+const [DescendantsContextProvider, useDescendantsContext] =
+  createContext<UseDescendantsReturn>({
+    name: "DescendantsProvider",
+    errorMessage:
+      "useDescendantsContext must be used within DescendantsProvider",
+  })
 
 /**
  * @internal
@@ -57,7 +56,7 @@ function useDescendant<T extends HTMLElement = HTMLElement, K = {}>(
 
   useSafeLayoutEffect(() => {
     if (!ref.current) return
-    const dataIndex = Number(ref.current.dataset.index)
+    const dataIndex = Number(ref.current.dataset["index"])
     if (index != dataIndex && !Number.isNaN(dataIndex)) {
       setIndex(dataIndex)
     }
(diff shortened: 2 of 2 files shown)


TimKolberger:
Thank you for this PR @antwigambrah ✨ 

The TS flag [`noPropertyAccessFromIndexSignature`](https://www.typescriptlang.org/tsconfig#noPropertyAccessFromIndexSignature) is an opinionated flag to signal intent in your calling syntax about how certain you are this property exists. It has no effect during runtime by my knowledge.

By only fixing the typos in this package without updating the `tsconfig.json` sets this up for failure in the future, because we wont get notified when we break this rule again.

Merging this PR is a question of code style and I'd like to invite @chakra-ui/react-team to vote on this.

If we vote for this codestyle I'd love to see the updated `tsconfig.json` and possible more fixes over the codebase.
If we vote against this codestyle I think you can workaround this by using `skipLibCheck`.

ljosberinn:
the code itself is fine although obviously "risky" since `index` may or may not be defined. this pr however doesn't change anything about that. please see https://github.com/chakra-ui/chakra-ui/issues/5317#issuecomment-1013689959

3x071c:
@TimKolberger @ljosberinn While it is true that the flag doesn't affect runtime behavior, the intention behind it isn't code style, but rather to prevent accidental access to non-existing properties on types with an index signature. That's why I'm enforcing it in my project, it forces you to explicitly differentiate between accessing a declared type property or a possibly not existing property from its Index signature, catching production bugs and typos earlier. Unfortunately, `skipLibCheck` (which I have turned on) doesn't prevent TypeScript from checking (indirectly) imported modules. If I use any chakra-ui component which uses `descendant` internally, it will be evaluated (regardless of if `node_modules is being explicitly ignored). I hope you adopt the flag, I don't know how else I can work around this.

3x071c:
Any updates on this? 😁

TimKolberger (approved):
I don't mind merging this PR, but keep the current tsconfig as is.

PR #5355, opened 2022-01-09
fix: type check repo

TimKolberger:
## 📝 Description

This PR adds TS type checking back to the CI process.

We dropped the type check somehow.

## 🚀 New behavior

Run `tsc --no-emit` on `yarn lint` and `yarn lint:types`

No


packages/layout/stories/text.stories.tsx
```diff
@@ -21,7 +21,10 @@ const theme = extendTheme({
 // see https://github.com/chakra-ui/chakra-ui/issues/2464
 export const withVariant = () => (
   <ChakraProvider theme={theme}>
-    <Text variant="customCaps">
+    <Text
+      // @ts-expect-error Text has no variants in default theme. Use @chakra-ui/cli to generate theme typings.
```
Brennvo, line 25: I might be missing something, but I'm not sure if this is necessary. It's also holding up the PR workflows.

PR #5362, opened 2022-01-10
fix: index issue with range slider

### LINKED ISSUE #5346: Range Slider issue
Please check the line363 in use-range-slider.js, I think we should replace "index = thumbsAtPosition.length - 1" with "index = index + thumbsAtPosition.length - 1". It is an index increment instead of override the index value.

_Originally posted by @weiliao715 in https://github.com/chakra-ui/chakra-ui/issues/5081#issuecomment-970310308_

sohamsshah:
Closes #5346 

## 📝 Description

As per the discussion in #5081 , the index value needs to be updated for the range-slider.

No

### DIFF
--- a/.changeset/popular-monkeys-matter.md
+++ b/.changeset/popular-monkeys-matter.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/slider": patch
+---
+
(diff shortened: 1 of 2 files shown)


burhanuday (approved):
looks good

PR #5371, opened 2022-01-11
refactoring(theme): Simplify exports

selbekk:
Remove more than half the code by using more modern syntax.

## 📝 Description

> Came across a unnecessarily bloated way of re-exporting default exports. This change simplifies it a bit.

Nope.

## 📝 Additional Information

Totally understand if this is a stylistic choice, so if you don't want it, just close my PR, no feelings will be hurt.

### DIFF
--- a/.changeset/popular-jars-know.md
+++ b/.changeset/popular-jars-know.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/theme": patch
+---
+
+refactoring(theme): Simplify exports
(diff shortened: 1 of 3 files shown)


TimKolberger:
The new export syntax is great ✨ 
Currently it fails due the removed default export. Could you update the import of the theme components?

TimKolberger (changes requested):
The new export syntax is great ✨
Currently it fails due the removed default export. Could you update the import of the theme components?

selbekk:
Of course 😅 Thanks!

selbekk:
Hm, not sure why the Release Snapshot fails, looks like there's an issue with my npm user or what not? Either way, the build passes now :)

PR #5422, opened 2022-01-20
fix: fix useAriaHidden deps

dqn:
## 📝 Description

Fix `useAriaHidden` deps.

Currently, the `useInert` of `Modal` is not working because `useAriaHidden` is not triggered correctly.

## 🚀 New behavior

`useAriaHidden` will be triggered correctly and the target elements will be given the `aria-hidden` attribute.

No

### DIFF
--- a/packages/modal/src/use-modal.ts
+++ b/packages/modal/src/use-modal.ts
@@ -1,7 +1,7 @@
 import { useIds } from "@chakra-ui/hooks"
 import { callAllHandlers } from "@chakra-ui/utils"
 import { mergeRefs, PropGetter } from "@chakra-ui/react-utils"
-import { hideOthers, Undo } from "aria-hidden"
+import { hideOthers } from "aria-hidden"
 import {
   KeyboardEvent,
   MouseEvent,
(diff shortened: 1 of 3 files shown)


packages/modal/src/use-modal.ts
```diff
@@ -217,5 +217,7 @@ export function useAriaHidden(
         undo?.()
       }
     }
-  }, [shouldHide, ref])
+    // Even if `ref.current` changes, `ref` does not change, so `ref.current` is specified.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
TimKolberger, line 221: We should be able to restructure the code to remove the necessity to disable the eslint rule.

```tsx
const currentElement = ref.current
useEffect(() => {
  // still use ref.current in the effect, because it could have changed since triggering the effect
}, [shouldHide, currentElement])
```

TimKolberger (changes requested):
Thank you for this PR @dqn ✨ 
Please see my comment above.

dqn:
@TimKolberger Thanks for reviewing! I've fixed it.

packages/modal/src/use-modal.ts
TimKolberger: Thank you for the fast response!
I think I did not express my self clear enough.
Your first implementation was logically correct, only eslint did not like the implementation.
With my code comment I meant, that we should still use `ref.current` inside of the effect.

Full example: 
```tsx
export function useAriaHidden(
  ref: RefObject<HTMLElement>,
  shouldHide: boolean,
) {
  // save current ref in a local var to trigger the effect on identity change
  const currentElement = ref.current

  useEffect(() => {
    // keep using `ref.current` inside the effect
    // it may have changed during render and the execution of the effect
    if (!ref.current) return undefined

    let undo: Undo | null = null

    if (shouldHide && ref.current) {
      undo = hideOthers(ref.current)
    }

    return () => {
      if (shouldHide) {
        undo?.()
      }
    }
  }, [shouldHide, ref, currentElement])
}
```

dqn: Oh, sorry, I didn't check carefully.
And I realized that I could write it more concisely. Could you please review it?

TimKolberger:
Thank you for updating the PR ✨ 
The code looks much cleaner now! 

I am not quite sure if/how this change affects the behavior of this hook. 
What do you think about adding a test to ensure this won't break again in the future?

dqn:
@TimKolberger Thank you for your comment. I added tests for `useAriaHidden`. Could you please review it?

packages/modal/src/use-modal.ts
TimKolberger: 💖

PR #5472, opened 2022-01-27
Increase scan depth for Chakra CLI on non-color tokens

### LINKED ISSUE #5457: Increase scan depth for Chakra CLI on non-color tokens
### Description

I would be happy if the scan depth for non-color tokens in Chakras CLI script for generating typings could be increased from one.

Currently, the chakra CLI tool scans theme variables and generates typings for them, by scanning for a depth of up to 3 for color variables, but only up to one for everything else. This means that a color with the token name "general.brand.default" would be found, but "general.brand.default.subprop" not. However, in our project we also use nested font sizes and font weights, e.g. "body.small" is one of the font size values. If the CLI tool would scan that as well, it would help us by providing better typings.

### Proposed Solution or API

* In https://github.com/chakra-ui/chakra-ui/blob/2bc045a671b23b5d3436d9038dfce4c7f75bbea0/tooling/cli/src/command/tokens/config.ts, add `maxScanDepth: 3` to `fontSizes` and `fontWeights`

or 

* Change the default scan depth to 3 here: https://github.com/chakra-ui/chakra-ui/blob/ae6fd7a25c543d089d500e328596b399d85afe8e/tooling/cli/src/command/tokens/extract-property-paths.ts#L39

Of course I can open a PR with this if this proposed change is considered.

### Alternatives

Alternatively the default scan depth could maybe be customizable via CLI argument, which also should not be that much effort.

### Additional Information

_No response_

lukasbach:
Closes #5457

## 📝 Description

I would be happy if the scan depth for non-color tokens in Chakras CLI script for generating typings could be increased from one.

Currently, the chakra CLI tool scans theme variables and generates typings for them, by scanning for a depth of up to 3 for color variables, but only up to one for everything else. This means that a color with the token name "general.brand.default" would be found, but "general.brand.default.subprop" not. However, in our project we also use nested font sizes and font weights, e.g. "body.small" is one of the font size values. If the CLI tool would scan that as well, it would help us by providing better typings.

## 🚀 New behavior

Default scan depth increased to 3.

No

### DIFF
--- a/.changeset/six-bags-bake.md
+++ b/.changeset/six-bags-bake.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/cli": minor
+---
+
+Increased scan depth for tokens in cli tooling
--- a/tooling/cli/src/command/tokens/extract-property-paths.ts
+++ b/tooling/cli/src/command/tokens/extract-property-paths.ts
@@ -36,7 +36,7 @@ export function printUnionMap(
 /**
  * Extract recursively all property paths with a max depth
  */
-export function extractPropertyPaths(target: unknown, maxDepth = 1) {
+export function extractPropertyPaths(target: unknown, maxDepth = 3) {
   if ((!isObject(target) && !Array.isArray(target)) || !maxDepth) {
     return []
   }


TimKolberger (approved):
Thank you for this PR @lukasbach 💖

TimKolberger:
@all-contributors please add @lukasbach for code

Issue #5467, opened 2022-01-27
useMediaQuery not supported on iOS 13<


Edjevw12:
### Description

useMediaQuery hook not supported on safari browsers v13 <=

### Link to Reproduction

https://codesandbox.io/s/nostalgic-https-mwzhs?file=/pages/index.tsx

### Steps to reproduce

open the sandbox on safari browsers v13 <=

### Chakra UI Version

1.6.3

### Browser

Safari 13 <=

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

useMediaQuery hook not supported on safari browsers v13 <=

`"Support for matchMedia is only partial in Safari (Desktop and iOS) including Safari 13 (Technology Preview), because the MediaQueryList object that is returned only supports the legacy .addListener() method, but not .addEventListener()."` 

Similair issue: https://github.com/mdn/sprints/issues/858

![image](https://user-images.githubusercontent.com/37038043/151396831-02282153-ed86-4a6e-af33-53c91a1be223.png)

noobinthisgame:
See https://github.com/chakra-ui/chakra-ui/issues/5197

Edjevw12:
fix: #5474

pushing the fix with patch-package to prod, and issue solved

TimKolberger:
Closed by #5474 and @chakra-ui/react@1.8.3

Issue #5197, opened 2021-12-06
When I use the safari browser, an error is reported, Firefox and Google browsers are normal


shenhaoguangDayDayUp:
### Description

When I use the safari browser, an error is reported .

### Link to Reproduction

https://test.zhanheyibiao.com/#/cube

### Steps to reproduce

![image](https://user-images.githubusercontent.com/17617821/144769663-b35b22ec-5b28-4bcd-82cc-ab5e56c7059d.png)

### Chakra UI Version

1.7.2

### Browser

safari 13.1.3

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Looking forward to your reply

primos63:
Please provide the version of Safari this occurred on.

shenhaoguangDayDayUp:
> Please provide the version of Safari this occurred on.

Ok.

primos63:
[Safari support](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList) for `MediaQueryList inherits EventTarget` began in version 14. Version 13.1.3 was released on October 15, 2019. Version 14 was released on September 16, 2020 and version 15 was released on September 20, 2021.
 
 The `addListener` and `removeListener` API methods were deprecated.
 > Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.
 
 I would advise updating your browser to version 14 or greater; otherwise, you will need to use an earlier version of Chakra.

shenhaoguangDayDayUp:
> [Safari support](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList) for `MediaQueryList inherits EventTarget` began in version 14. Version 13.1.3 was released on October 15, 2019. Version 14 was released on September 16, 2020 and version 15 was released on September 20, 2021.
> 
> The `addListener` and `removeListener` API methods were deprecated.
> 
> > Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.
> 
[...]

PR #5474, opened 2022-01-28
fix(use-media-query): conditional check if addEventListener is supported

Edjevw12:
Closes # 

#5467 and #5197

## 📝 Description

> Add a brief description

Support useMediaQuery for older browsers. Conditionally check if the
MediaQueryList object supports the addEventListener() method, else fallback to
the legacy .addListener() method.

Website crashes on Safari browsers < v13

## 🚀 New behavior

the method `addListener()` is used instead of `addEventListener()`

No

### DIFF
--- a/packages/media-query/src/use-media-query.ts
+++ b/packages/media-query/src/use-media-query.ts
@@ -54,10 +54,10 @@ export function useMediaQuery(query: string | string[]): boolean[] {
       // media query is matched. Using addEventListener on the window object
       // to listen for the resize event will call the callback on every
       // viewport resize.
-      if(typeof mediaQueryList[index].addEventListener === 'function') {
-        mediaQueryList[index].addEventListener("change", listener);
+      if (typeof mediaQueryList[index].addEventListener === "function") {
+        mediaQueryList[index].addEventListener("change", listener)
       } else {
-        mediaQueryList[index].addListener(listener);
+        mediaQueryList[index].addListener(listener)
       }
 
       return listener
@@ -65,10 +65,13 @@ export function useMediaQuery(query: string | string[]): boolean[] {
 
     return () => {
       mediaQueryList.forEach((_, index) => {
-        if(typeof mediaQueryList[index].removeEventListener === 'function') {
-          mediaQueryList[index].removeEventListener("change", listenerList[index]);
+        if (typeof mediaQueryList[index].removeEventListener === "function") {
+          mediaQueryList[index].removeEventListener(
+            "change",
+            listenerList[index],
+          )
         } else {
-          mediaQueryList[index].removeListener(listenerList[index]);
+          mediaQueryList[index].removeListener(listenerList[index])
         }
       })
     }
--- a/packages/media-query/src/use-media-query.ts
+++ b/packages/media-query/src/use-media-query.ts
@@ -54,14 +54,22 @@ export function useMediaQuery(query: string | string[]): boolean[] {
       // media query is matched. Using addEventListener on the window object
       // to listen for the resize event will call the callback on every
       // viewport resize.
-      mediaQueryList[index].addEventListener("change", listener)
+      if(mediaQueryList[index].addEventListener) {
+        mediaQueryList[index].addEventListener("change", listener);
+      } else {
+        mediaQueryList[index].addListener(listener);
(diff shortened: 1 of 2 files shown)


cschroeter:
```typescript
typeof mediaQueryList[index].addEventListener === "function"
```

packages/media-query/src/use-media-query.ts
```diff
@@ -54,14 +54,22 @@ export function useMediaQuery(query: string | string[]): boolean[] {
       // media query is matched. Using addEventListener on the window object
       // to listen for the resize event will call the callback on every
       // viewport resize.
-      mediaQueryList[index].addEventListener("change", listener)
+      if(typeof mediaQueryList[index].addEventListener === 'function') {
+        mediaQueryList[index].addEventListener("change", listener);
+      } else {
+        mediaQueryList[index].addListener(listener);
+      }
```
TimKolberger, line 61: This is a viable solution to define a polyfill for `addEventListener`, but I'd love to see this implementation somewhere else.
We maintain a `utils` package where the helper function `addDomEvent` is located.

https://github.com/chakra-ui/chakra-ui/blob/2133a4122672f23a5816d7010d8441067cba13a9/packages/utils/src/dom.ts#L63-L73

Kindly add the polyfill there and use the `addDomEvent` function here to attach the change listener.

TimKolberger (changes requested):
Thank you for this PR @Edjevw12 ✨ 
I added my thoughts as comment above.

packages/media-query/src/use-media-query.ts
Edjevw12: @TimKolberger 

the `addEventListener` on a DOM event is different from the MediaQueryList event

The util would look like:

```ts
export function addMediaQueryListEvent(
  type: "change",
  target: MediaQueryList,
  handler: (mqlEvent: MediaQueryListEvent) => void,
  options?: AddEventListenerOptions,
) {
  if(typeof target.addEventListener === 'function') {
    target.addEventListener(type, handler, options);
    return () => target.removeEventListener(type, handler, options);
  } 
  target.addListener(handler);
  return () => target.removeListener(handler)
}
```

but in the forEach, we can't use the return value of the `addMediaQueryListEvent`, since we are returning the listener.

So we have to define another util?

```ts
export function removeMediaQueryListEvent(
  type: "change",
  target: MediaQueryList,
  handler: (mqlEvent: MediaQueryListEvent) => void,
  options?: AddEventListenerOptions,
) {
  if(typeof target.removeEventListener === 'function') {
    target.removeEventListener(type, handler, options);
  } 
  target.removeListener(handler)
}
```

So it would be:

```ts
    addMediaQueryListEvent("change", mediaQueryList[index], listener)
    return listener
    })

   return () => {
     mediaQueryList.forEach((_, index) => {
      removeMediaQueryListEvent("change", mediaQueryList[index], listenerList[index])
    })
```

TimKolberger: Thank you for the detailled description!

TIL that this is only the case for `mediaQueryList` for Safari 14, not for all event listeners.

> Before Safari 14, MediaQueryList is based on EventTarget, so you must use addListener() and removeListener() to observe media query lists. - [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener)

This invalidates my suggestion and I think we can proceed with the current solution.

One last thing - kindly run prettier on this file to add a space between `if` and `(`.
I think then its good to go 🚀

Edjevw12: > Thank you for the detailled description!
> 
> TIL that this is only the case for `mediaQueryList` for Safari 14, not for all event listeners.
> 
> > Before Safari 14, MediaQueryList is based on EventTarget, so you must use addListener() and removeListener() to observe media query lists. - [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener)
> 
> This invalidates my suggestion and I think we can proceed with the current solution.
> 
> One last thing - kindly run prettier on this file to add a space between `if` and `(`. I think then its good to go 🚀

ahh right, that makes sense. 

I've run prettier 🚀

TimKolberger (approved):
LGTM 🚀

PR #5486, opened 2022-01-29
fix(cli): use latest pkg version of /utils

TimKolberger:
## 📝 Description

cli package has a dependency to /utils@^1.10.0 which is not a `latest` release. See [npmjs.org](https://www.npmjs.com/package/@chakra-ui/utils)

You can end up with multiple installed chakra versions.

## 🚀 New behavior

Replaced with `^1.9.1`

No

### DIFF
--- a/.changeset/dirty-ravens-drum.md
+++ b/.changeset/dirty-ravens-drum.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/styled-system": patch
+---
+
+Escape CSS custom property names
--- a/.changeset/dirty-ravens-drum.md
+++ b/.changeset/dirty-ravens-drum.md
@@ -1,5 +1,0 @@
----
-"@chakra-ui/styled-system": patch
----
-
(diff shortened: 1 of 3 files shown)


ishowta:
It looks like `@chakra-ui/theme-tools:1.3.2` is also not `latest`. Is there a problem with the release?

.changeset/dirty-ravens-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+"@chakra-ui/styled-system": patch
+---
+
+Escape CSS custom property names
```
TimKolberger, line 5: Oops - this changeset needs to be removed.

TimKolberger:
@ishowta Yes, the releases 1.7.4, 1.7.5 and 1.8.0 had issues with using the `latest` tag.
If you find more packages with invalid latest tags, kindly share them in a new issue - we can address them all at once.

PR #5492, opened 2022-01-31
fix: EditablePreview becomes unclickable when empty

### LINKED ISSUE #5423: Editable without value isn't clickable
### Description

In a "editable" input, when it is empty, is not possible to click on it again, cause the width of the editable seems to be zero.

### Link to Reproduction

https://chakra-ui.com/docs/form/editable

### Steps to reproduce

1. Go to 'https://chakra-ui.com/docs/form/editable'
2. Click on 'editable'
3. Clean the 'editable'
4. Lost the focus
4. Try to click on 'editable' again

### Chakra UI Version

1.7.5

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

shamoilarsi:
Closes #5423

## 📝 Description

adds a minWIdth to EditablePreview

No

### DIFF
--- a/packages/theme/src/components/editable.ts
+++ b/packages/theme/src/components/editable.ts
@@ -9,6 +9,7 @@ const baseStylePreview: SystemStyleObject = {
   py: "3px",
   transitionProperty: "common",
   transitionDuration: "normal",
+  minWidth: "2px",
 }
 
 const baseStyleInput: SystemStyleObject = {
--- a/.changeset/funny-steaks-punch.md
+++ b/.changeset/funny-steaks-punch.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/theme": patch
+---
+
+added minWidth to EditablePreview
+
+fixes #5423


burhanuday (approved):
Looks good

noobinthisgame:
Looks good to me :)

packages/theme/src/components/editable.ts
```diff
@@ -9,6 +9,7 @@ const baseStylePreview: SystemStyleObject = {
   py: "3px",
   transitionProperty: "common",
   transitionDuration: "normal",
+  minWidth: "2px",
```
segunadebayo, line 12: What's the reason behind using "2px"? It seems arbitrary

shamoilarsi: "2px" was selected so that the pointer has a small width to click on. 

On second thought, 
since the width of `EditableInput` is set to `"full"`, we could also set the width of `EditablePreview` the same?
`minWidth` can be removed in this case.

segunadebayo:
Thanks for working on this @shamoilarsi. Please see my comment in the issue.

PR #5502, opened 2022-02-01
fix(useCheckboxGroup): fixing the issue when using the useCheckboxGroup hook with numbers

### LINKED ISSUE #5501: useCheckboxGroup not working with numbers
### Description

When we use `useCheckboxGroup` with number values the checkboxes don't reflect the changes

### Link to Reproduction

https://codesandbox.io/s/vibrant-saha-56fr6?file=/src/index.tsx

### Steps to reproduce

1. Go to https://codesandbox.io/s/vibrant-saha-56fr6?file=/src/index.tsx
2. Click on the checkboxes
3. The checkboxes aren't reflecting the changes

### Chakra UI Version

1.8.1

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

nikolovlazar:
Closes #5501

## 📝 Description

When using the `useCheckboxGroup` hook with numbers instead of strings the checkboxes aren't reflecting the changes. I noticed the issue in two places:
- `use-checkbox-group.ts` on line 89: the `values` array may contain both string and number values. The comparison doesn't take that into account. Instead of `includes`, which relies on the type as well, I replaced it with `value.some((val) => String(props.value) === String(val))`, so that both strings and numbers can be taken into an account.
- `use-checkbox-group.ts` on line 70: I replaced the `remove` method with a local `filter` that compares the string values for the same reason.

The checkboxes don't reflect the changes.

## 🚀 New behavior

The checkboxes do reflect the changes.

No (hopefully)

## 📝 Additional Information
I've also added an additional example in the storybook to test this fix.

### DIFF
--- a/.changeset/soft-countries-sin.md
+++ b/.changeset/soft-countries-sin.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/checkbox": patch
+"@chakra-ui/utils": patch
+---
+
+Fixing a bug that happens when using the useCheckboxGroup hook with number
+values instead of string values
--- a/packages/checkbox/src/use-checkbox-group.ts
+++ b/packages/checkbox/src/use-checkbox-group.ts
@@ -85,8 +85,7 @@ export function useCheckboxGroup(props: UseCheckboxGroupProps = {}) {
       const checkedKey = isNative ? "checked" : "isChecked"
       return {
         ...props,
-        [checkedKey]:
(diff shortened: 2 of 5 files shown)


TimKolberger (approved):
Thanks @nikolovlazar ✨ 
LGMT 🚀

nikolovlazar:
@allcontributors please add @nikolovlazar for bugs.

PR #5504, opened 2022-02-01
Remove redundant `useMemo` from `FormControl`

Andarist:
This is a very simple change that just drops a line of code that is basically a "noop" here.


TimKolberger (approved):
Thank you @Andarist! Good catch ✨

Note to self: add changeset before merging.

PR #5507, opened 2022-02-01
fix: resolve native input size prop conflict with input theme props

### LINKED ISSUE #5489: Theming prop `size` conflicts with input prop `size`
### Description

When I want to set the DOM attribute `size={3}` of an `<input>` element, I am unable to do it with `<Input>` since `size` is interpreted as a theme prop, and not a native attribute.

### Link to Reproduction

https://codesandbox.io/s/jovial-mendeleev-upylk?file=/src/index.js

### Steps to reproduce

See above CodeSandbox

### Chakra UI Version

1.6.3

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

noobinthisgame:
Closes #5489 

## 📝 Description

Add htmlSize prop to enable the usage of the native size prop and to resolve the conflict with the input theme prop size.

The native input size prop can't be used because it gets overriden by the input theme prop size.

## 🚀 New behavior

There is a new htmlSize prop to pass the native size prop to an input.

No

## 📝 Additional Information

Added Storybook story to test the behaviour with the native size prop

### DIFF
--- a/packages/input/src/input.tsx
+++ b/packages/input/src/input.tsx
@@ -58,15 +58,11 @@ export const Input = forwardRef<InputProps, "input">((props, ref) => {
   const input = useFormControl<HTMLInputElement>(ownProps)
   const _className = cx("chakra-input", props.className)
 
-  // This is used to remove the default width=100% when size is set, which otherwise conflicts
-  const { width, ...fieldStyles } = styles.field
-  const cssStyles = htmlSize ? { ...fieldStyles } : { ...styles.field }
-
   return (
     <chakra.input
       size={htmlSize}
       {...input}
-      __css={cssStyles}
+      __css={styles.field}
       ref={ref}
       className={_className}
     />
--- a/packages/input/src/input.tsx
+++ b/packages/input/src/input.tsx
@@ -6,11 +6,28 @@ import {
   ThemingProps,
   useMultiStyleConfig,
   HTMLChakraProps,
+  PropsOf,
 } from "@chakra-ui/system"
 import { cx, __DEV__ } from "@chakra-ui/utils"
 import * as React from "react"
 
-interface InputOptions {
+interface NativeInputOptions {
+  /**
+   * The native HTML `size` attribute to be passed to the `input`
+   */
+  htmlSize?: number
+}
+
+interface NativeInputProps extends PropsOf<"input">, NativeInputOptions {}
+
+const NativeInput = React.forwardRef(
+  (props: NativeInputProps, ref: React.Ref<any>) => {
+    const { htmlSize, ...rest } = props
+    return <input size={htmlSize} ref={ref} {...rest} />
+  },
+)
+
+interface InputOptions extends NativeInputOptions {
   /**
    * The border color when the input is focused. Use color keys in `theme.colors`
    * @example
@@ -47,15 +64,23 @@ export interface InputProps
  * Element that allows users enter single valued data.
  */
 export const Input = forwardRef<InputProps, "input">((props, ref) => {
-  const styles = useMultiStyleConfig("Input", props)
-  const ownProps = omitThemingProps(props)
+  const { htmlSize, ...rest } = props
+
+  const styles = useMultiStyleConfig("Input", rest)
+  const ownProps = omitThemingProps(rest)
   const input = useFormControl<HTMLInputElement>(ownProps)
   const _className = cx("chakra-input", props.className)
 
+  // This is used to remove the default width=100% when size is set, which otherwise conflicts
+  const { width, ...fieldStyles } = styles.field
+  const cssStyles = htmlSize ? { ...fieldStyles } : { ...styles.field }
+
   return (
     <chakra.input
+      as={NativeInput}
+      htmlSize={htmlSize}
       {...input}
-      __css={styles.field}
+      __css={{ ...cssStyles }}
       ref={ref}
       className={_className}
     />
--- a/packages/input/src/input.tsx
+++ b/packages/input/src/input.tsx
@@ -6,28 +6,11 @@ import {
   ThemingProps,
   useMultiStyleConfig,
   HTMLChakraProps,
-  PropsOf,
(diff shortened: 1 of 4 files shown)


packages/input/src/input.tsx
```diff
@@ -6,11 +6,28 @@ import {
...
+
+interface NativeInputProps extends PropsOf<"input">, NativeInputOptions {}
+
+const NativeInput = React.forwardRef(
+  (props: NativeInputProps, ref: React.Ref<any>) => {
+    const { htmlSize, ...rest } = props
+    return <input size={htmlSize} ref={ref} {...rest} />
+  },
+)
+
```
TimKolberger, line 29: I think we can solve this without introducing an additional component. 
The intended behavior can be achieved by using the chakra factory function `<chakra.input />`

packages/input/src/input.tsx
```diff
@@ -47,15 +64,23 @@ export interface InputProps
...
+  const { width, ...fieldStyles } = styles.field
+  const cssStyles = htmlSize ? { ...fieldStyles } : { ...styles.field }
+
   return (
     <chakra.input
+      as={NativeInput}
+      htmlSize={htmlSize}
       {...input}
-      __css={styles.field}
+      __css={{ ...cssStyles }}
```
TimKolberger, line 83: We can safely omit the spread operator here: 

```suggestion
      __css={cssStyles}
```

packages/input/src/input.tsx
```diff
@@ -47,15 +64,23 @@ export interface InputProps
...
   const input = useFormControl<HTMLInputElement>(ownProps)
   const _className = cx("chakra-input", props.className)
 
+  // This is used to remove the default width=100% when size is set, which otherwise conflicts
+  const { width, ...fieldStyles } = styles.field
+  const cssStyles = htmlSize ? { ...fieldStyles } : { ...styles.field }
+
   return (
     <chakra.input
+      as={NativeInput}
```
TimKolberger, line 80: I think we can omit this, because

```tsx
<chakra.input size="2" />
// renders
<input size="2" />
```

```suggestion
```

TimKolberger (changes requested):
Thank you for this PR @noobinthisgame 💖 
I added my thoughts as comments. 
Feel free to reach out if you have further questions!

packages/input/src/input.tsx
noobinthisgame: Indeed that works. Definitely cleaner.

packages/input/src/input.tsx
noobinthisgame: Right that's a relict from testing. Remove it.

packages/input/src/input.tsx
```diff
@@ -47,15 +51,22 @@ export interface InputProps
...
+  const { htmlSize, ...rest } = props
+
+  const styles = useMultiStyleConfig("Input", rest)
+  const ownProps = omitThemingProps(rest)
   const input = useFormControl<HTMLInputElement>(ownProps)
   const _className = cx("chakra-input", props.className)
 
+  // This is used to remove the default width=100% when size is set, which otherwise conflicts
+  const { width, ...fieldStyles } = styles.field
+  const cssStyles = htmlSize ? { ...fieldStyles } : { ...styles.field }
```
TimKolberger, line 63: I'm not quite sure we want to predict the user's goal here.

From the perspective of the using side I find it much easier to understand whats happening when the width is explicitly set like `<Input htmlSize="10" width="auto" />`.

Otherwise we would need to filter the props `width` and `w` respectively.

What do you think about that?

packages/input/tests/input.test.tsx
```diff
@@ -45,3 +45,9 @@ test("Readonly input renders correctly", () => {
 
   expect(screen.getByRole("textbox")).toHaveAttribute("aria-readonly", "true")
 })
+
+test("Input with native size renders correctly", () => {
+  render(<Input htmlSize={4} />)
+
+  expect(screen.getByRole("textbox")).toHaveAttribute("size", "4")
+})
```
TimKolberger, line 53: 🦄 💖

packages/input/src/input.tsx
noobinthisgame: I think that building around the default theme width is overcomplicating the component. I'd be happy to remove it to make the component more transparent in usage and less funky. Just didn't wanted to make too many changes everywhere.
Removing the default theme width might result in a breaking change tho, as we don't know how many people rely on this, as it's a default. 😕

TimKolberger: > I'd be happy to remove it to make the component more transparent in usage and less funky.

Sounds great! 

> Removing the default theme width might result in a breaking change 

Yes, this would be a breaking change IMO and not necessary to introduce a new html prop.

noobinthisgame: Actually I don't get why we wouldn't need the new htmlProp ? We still have the issue that theme size and native size conflict here ?

TimKolberger: I think I did not express my self clearly 😅 
It is great to add support for the html `size` attribute! But I think this does not mean to introduce a breaking change to the theme. 
If a user wants to use `htmlSize` I think it is much cleaner to let them use `<Input htmlSize="10" width="auto" />` instead of internally omitting the styleProp `width`.

noobinthisgame: Ahhh I right I got it. Docs thing then for the usage of the htmlSize and to override default width with auto. Gotcha. Will make those changes and open a PR in docs aswell 😄

TimKolberger (approved):
lgtm 🚀

noobinthisgame:
@TimKolberger Here is the according docs https://github.com/chakra-ui/chakra-ui-docs/pull/277 
Let me know if thats enough 😊

noobinthisgame:
@nikolovlazar Would it make sense to have labels for PRs like this one that should only be merged when a feature is actually released in the main repo ?

PR #5508, opened 2022-02-01
feat: allow all dom elements for the chakra factory

TimKolberger:
## 📝 Description

Allow all `JSX.IntrinsicElements` for the chakra factory. This allows to use [every DOM element](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/30a2f70db2f9ac223fd923ff1f8bcc175c082fd0/types/react/index.d.ts#L3111-L3288) with the shorthand version:

```jsx live=false
<chakra.header>Header</chakra.header>
<chakra.main>Main</chakra.main>
<chakra.footer>Many more</chakra.footer>
```

It's a hardcoded list

## 🚀 New behavior

It derives the types from `JSX.IntrinsicElements` and resolves the components lazy with a `Proxy` and a `Map` as cache.

No

## 📝 Additional Information

[MDN Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)


packages/system/src/system.ts
```diff
@@ -63,7 +65,7 @@ interface StyledOptions {
 
 export function styled<T extends As, P = {}>(
   component: T,
-  options?: StyledOptions,
+  options?: StyledOptionsChakra,
```
segunadebayo, line 68:
```suggestion
  options?: ChakraStyledOptions,
```

packages/system/src/system.utils.ts
```diff
@@ -1,74 +1,12 @@
...
-  "textarea",
-  "tfoot",
-  "th",
-  "thead",
-  "tr",
-  "ul",
-] as const
-
-export type DOMElements = UnionStringArray<typeof domElements>
+export type DOMElements = keyof JSX.IntrinsicElements
```
segunadebayo, line 9: Such a breath of fresh air! Thanks for working on this

packages/system/src/factory.ts
```diff
@@ -0,0 +1,28 @@
...
+  ): ChakraComponent<T, P>
+}
+
+function factory() {
+  const cache = new Map<DOMElements, ChakraComponent<DOMElements>>()
+
+  return new Proxy(styled, {
+    apply(target, thisArg, argArray: [DOMElements, StyledOptionsChakra]) {
+      return styled(...argArray)
+    },
```
segunadebayo, line 18: Any reason this is needed?

TimKolberger: To allow `chakra("div")` or `chakra(AnotherComponent)` besides `chakra.div` :)

segunadebayo: I see what you mean, but because the Proxy was created from `styled`, my guess is that we may not need this `apply`. Did you try removing the apply and testing it already?

PR #5527, opened 2022-02-04
fix(cli): add failure exit code

### LINKED ISSUE #5523: Chakra CLI returns exit code 0 when failing to parse theme
### Description

When I have some code that the chakra CLI can't parse, the command exits with code 0, suggesting success.

### Link to Reproduction

https://codesandbox.io/s/nameless-moon-9ufmh

### Steps to reproduce

1. Download the sandbox (top left menu button, File, Export to zip) and extract
2. Install dependencies (I used yarn)
3. Run `generate:theme` script
```
$ chakra-cli tokens ./src/theme.ts

 Chakra UI CLI  v1.7.0 by Chakra UI
Generate theme typings for autocomplete


✖ An error occurred
SyntaxError: Invalid or unexpected token
✨  Done in 10.78s.
```

If you then check the exit code (`echo $?` in bash), you'll find that it exited with a zero (suggesting success) when it failed.

This is an issue because e.g. in CI, we want to fail the build if this fails.

### Chakra UI Version

1.8.1

### Browser

Node 16

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I suspect this is related to the code around reading the theme in https://github.com/chakra-ui/chakra-ui/blob/main/tooling/cli/src/command/tokens/index.ts#L15 but not dug into it too much.

with-heart:
Closes #5523

## 📝 Description

This PR adds a `process.exit(1)` call to the `try`/`catch` of `generateThemeTypings`.

Process exits with a success (`0`) status code in all cases.

## 🚀 New behavior

Process exits with an "Uncaught Fatal Exception" (`1`) when `generateThemeTypings` fails.

No

### DIFF
--- a/tooling/cli/src/command/tokens/index.ts
+++ b/tooling/cli/src/command/tokens/index.ts
@@ -69,6 +69,8 @@ export async function generateThemeTypings({
     if (e instanceof Error) {
       console.error(e.message)
     }
+    spinner.stop()
+    process.exit(1)
   } finally {
     spinner.stop()
   }
--- a/.changeset/shy-cobras-wonder.md
+++ b/.changeset/shy-cobras-wonder.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/cli": patch
+---
+
+When the [Chakra CLI](https://chakra-ui.com/docs/theming/advanced#theme-typings)
+fails to generate theme typings, it now exits with a status code of `1`. This
+resolves an issue where failures exited with a success status code.


tooling/cli/src/command/tokens/index.ts
```diff
@@ -69,6 +69,8 @@ export async function generateThemeTypings({
     if (e instanceof Error) {
       console.error(e.message)
     }
+    spinner.stop()
+    process.exit(1)
   } finally {
     spinner.stop()
```
with-heart, line 75: Just wanted to note that we have to duplicate `spinner.stop()` calls here because `process.exit` blows away the call stack, so we never reach the `finally` block.

TimKolberger:
Thank you @with-heart 💖

PR #4643, opened 2021-08-30
fix(checkbox): add form-control support to checkbox

takethefake:
Closes #4629 

## 📝 Description

Currently, Checkbox/Checkbox Group aren't compatible with FormControl (see 4629)

## 🚀 New behavior

Added FormControl Support for checkbox

## 💣 Is this a breaking change (Yes/No):

No

## 📝 Additional Information

Current Drawbacks there are some Props (`isDisabled` `isReadOnly`...) rendered to the which results in React-Errors. have to check out where this happens and fix it accordingly.

PR #5564, opened 2022-02-12
fix(use-checkbox): add form-control support to use-checkbox

### LINKED ISSUE #4629: FormControl does not provide context to CheckboxGroup
### Description

When I tried to use a FormControl to disable a CheckboxGroup, it did not work as expected. I tested with invalid and the same issue existed. I assume the context is not being provided.

### Link to Reproduction

https://codesandbox.io/s/busy-mclean-oygwx?file=/src/index.js

### Steps to reproduce

1. See code in codesandbox.io link above.
2. Notice how CheckboxGroup is not being passed the isDisabled or isInvalid context from FormGroup.

### Chakra UI Version

"@chakra-ui/react": "^1.3.3"

### Browser

Google Chrome Version 94.0.4606.20 (Official Build) dev (64-bit)

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

_No response_

santialbo:
Closes #4629

## 📝 Description

This is an alternative to #4643. Instead of using `useFormControlProps` inside the checkbox component. I use it inside the `useCheckbox` hook. This way all implementations (Checkbox, Switch) benefit from it.

## 🚀 New behavior

Checkbox and Switch now support FormControl

No

### DIFF
--- a/.changeset/fair-clocks-thank.md
+++ b/.changeset/fair-clocks-thank.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/checkbox": patch
+---
+
+Add Form-Control support for useCheckbox
--- a/packages/checkbox/package.json
+++ b/packages/checkbox/package.json
@@ -38,6 +38,7 @@
     "url": "https://github.com/chakra-ui/chakra-ui/issues"
   },
   "dependencies": {
+    "@chakra-ui/form-control": "1.5.6",
     "@chakra-ui/hooks": "1.8.2",
     "@chakra-ui/react-utils": "1.2.2",
     "@chakra-ui/utils": "1.10.2",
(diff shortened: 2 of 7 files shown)


takethefake:
Hej @santialbo thanks for your contribution! 

I've changed the initial PR according to your suggestion. So we could close this?

santialbo:
> I've changed the initial PR according to your suggestion. So we could close this?

Don't get me wrong, you are a member of the Chakra team, do as you wish. But given that I spent the time to come up with a better implementation I think it would good etiquette to merge this instead.

takethefake:
Closed my PR in favour of yours. Good work thanks for coming up with a better idea! :)

TimKolberger:
Thank you for working on this @santialbo and @takethefake 💖

This looks great! I love the the extensive tests 🔥

🚀

PR #4067, opened 2021-05-21
✨ Support generating token type definitions from theme packages

jrolfs:
## 📝 Description

I'm distributing a Chakra theme within my organization as a package on NPM and there's currently no good way to distribute the associated token type definitions.

## ⛳️ Current behavior (updates)

The Chakra CLI (**@chakra-ui/cli**) expects a theme to exist in a file within a project (e.g: `src/path/to/theme.ts`) with which to generate definitions from.

## 🚀 New behavior

Add support to **chakra-ui/cli** to consume themes distributed as NPM packages, e.g: `chakra-cli tokens @my-org/chakra-theme`.

## 💣 Is this a breaking change (Yes/No):

No.

## 📝 Additional Information

I'm curious if this sort of use case has come up before. I did try to search around a bit in the issues and discussions on GitHub and didn't find anything. Ideally, it would be possible to define custom theme type definitions via declaration merging (similar to how [**styled-components** does it](https://styled-components.com/docs/api#create-a-declarations-file)), that way they could be distributed with the theme. That said, I understand the choice in distributing Chakra with the default theme's definitions as otherwise, it would take work for a user to get completion working with the default theme out of the box. With this change it's at least possible to add a simple step to the package's readme to add the `postinstall` hook.

### Implementation Notes

- I'm happy to add tests for this, but I didn't see any existing coverage for the functionality I updated. Perhaps an integration test makes sense for the CLI?
- I'm also happy to add additional documentation, but I wanted to get some guidance with regards to my question above first.

Thank you!

PR #5290, opened 2021-12-26
fix(cli): replace runtime require of prettier with import

TimKolberger:
Follow up PR to #5225
Closes #5219

## 📝 Description

Prettier is used to format the TS interface created by the `chakra-cli tokens` command.

## ⛳️ Current behavior (updates)

Prettier was required during runtime, and if it was not available it should be skipped. 
This breaks the node worker script setup.

## 🚀 New behavior

Added prettier as dependency (previous PR) and import prettier the "normal" way.

## 💣 Is this a breaking change (Yes/No):

No

PR #5579, opened 2022-02-14
Support for custom theme types via TypeScript module augmentation

jrolfs:
## 📝 Description

Here I'm following up on a previous idea I had mentioned [in this PR](https://github.com/chakra-ui/chakra-ui/pull/4067#pullrequestreview-666149028). The TypeScript experience when distributing a component library based on Chakra still isn't _great_ on account of the Chakra CLI step being required for consumers to the library to get the proper language server/editor experience. Not only is it an extra step, but there are multiple ways in which it can break preventing consumers of our library from getting accurate type definitions.

In this PR, I'm proposing a small change to the way that these custom types are defined that both allow the existing CLI functionality to continue working exactly as it had previously, but also providing an optional advanced feature that would allow custom type definitions to be shipped _with_ component libraries that are based on Chakra that would not require end users to install and run the CLI. It also makes it possible to manually customize the type definitions, if desired.

Defining custom theme typings relies on the **@chakra-ui/cli** package _always_ running and overwriting the `ThemeTypings` interface in the end users' `./node_modules/` directory as documented [here](https://chakra-ui.com/docs/theming/advanced#theme-typings).

## 🚀 New behavior

This PR makes it possible to also customize the `ThemeTypings` interface using TypeScript's [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) and declaration merging features, much like custom theme types are implemented in [Styled Components](https://styled-components.com/docs/api#create-a-declarations-file) and [Emotion](https://emotion.sh/docs/typescript#define-a-theme).

In order to take advantage of this approach, the [`CustomThemeTypings`](https://github.com/jrolfs/chakra-ui/blob/feature/theme-typings-module-augmentation/packages/styled-system/src/theme.types.ts#L72) interface needs to be "augmented" and at least implement the [`DefaultThemeTypings`](https://github.com/jrolfs/chakra-ui/blob/feature/theme-typings-module-augmentation/packages/styled-system/src/theme.types.ts#L3-L26) interface, otherwise it will fall back to the existing [`ThemeTypings`](https://github.com/jrolfs/chakra-ui/blob/feature/theme-typings-module-augmentation/packages/styled-system/src/theming.types.ts#L6) interface that the CLI overwrites.

Example `types/chakra.d.ts`:

```ts
import "@chakra-ui/styled-system";

type DefaultSizes = 'small' | 'medium' | 'large';

declare module "@chakra-ui/styled-system" {
  export interface CustomThemeTypings {
    // Example custom `borders` tokens
    borders: 'none' | 'thin' | 'thick';
	// ...
	// Other custom tokens (everything in `DefaultThemeTypings` must be defined here)
	// ...
    components: {
      Button: {
        // Example custom component sizes and variants
        sizes: DefaultSizes;
        variants: 'solid' | 'outline' | 'wacky' | 'chill';
      };
      // ...
     }
  }
}
```

No, see above.

## 📝 Additional Information

In it's current form, this PR simply makes this possible and adds some [rudimentary inline documentation](https://github.com/jrolfs/chakra-ui/blob/feature/theme-typings-module-augmentation/packages/styled-system/src/theme.types.ts#L28-L72). I think it would be reasonable (barring maybe some feedback on how I named things) to merge this PR as is. With that said, I am very much ready to add documentation explaining the feature and I was even thinking we could combine this approach with the CLI with a new option to ouput a version of the definitions that leverages the above approach as opposed to modifying the definitions directly in `./node_modules`.

In addition to making it possible to _include_ custom type definitions with a component library that's based on Chakra, I think avoiding messing around in `./node_modules/` has some pretty clear benefits. Let me know what y'all think!

### DIFF
--- a/.changeset/grumpy-moons-look.md
+++ b/.changeset/grumpy-moons-look.md
@@ -1,0 +1,43 @@
+---
+"@chakra-ui/styled-system": minor
+---
+
+Modify theme types to make it possible to customize token types via TypeScript
+module augmentation and declaration merging in addition to allowing
+customization via the Chakra CLI.
+
+This makes it possible to do the following:
+
+- Distribute custom types with a component library based on Chakra
+- Customize theme types by hand
+- Version control your theme types
+
+To customize themes using the new mechanism, augment the `CustomThemeTypings`
+type in a definitions file such as `types/chakra.d.ts`:
+
+> ⚠️ NOTE: your `CustomThemeTypings` _must_ implement/extend `BaseThemeTypings`,
+> otherwise the types will fall back to the default Chakra types (or custom
+> output from **@chakra-ui/cli**)
+
+```ts
+import { BaseThemeTypings } from "@chakra-ui/styled-system";
+
+type DefaultSizes = 'small' | 'medium' | 'large';
+
+declare module "@chakra-ui/styled-system" {
+  export interface CustomThemeTypings extends BaseThemeTypings {
+    // Example custom `borders` tokens
+    borders: 'none' | 'thin' | 'thick';
+    // ...
+    // Other custom tokens
+    // ...
+    components: {
+      Button: {
+        // Example custom component sizes and variants
+        sizes: DefaultSizes;
+        variants: 'solid' | 'outline' | 'wacky' | 'chill';
+      };
+      // ...
+     }
+  }
+```
--- a/packages/styled-system/src/index.ts
+++ b/packages/styled-system/src/index.ts
@@ -1,7 +1,6 @@
 export * from "./config"
 export * from "./css"
 export * from "./system.types"
-export * from "./theming.types"
 export * from "./system"
 export * from "./create-theme-vars"
 export * from "./pseudos"
@@ -13,3 +12,4 @@ export type {
 } from "./utils"
 export { tokenToCSSVar } from "./utils/create-transform"
 export type OmitSpaceXY<T> = Omit<T, "spaceX" | "spaceY">
+export type { ThemeTypings, BaseThemeTypings, CustomThemeTypings } from "./theme.types"
--- a/packages/styled-system/src/theme.types.ts
+++ b/packages/styled-system/src/theme.types.ts
@@ -1,0 +1,63 @@
+import type { ThemeTypings as GeneratedThemeTypings } from "./theming.types"
+
+export interface BaseThemeTypings {
+  borders: string
+  colors: string
+  breakpoints: string
+  colorSchemes: string
+  fonts: string
+  fontSizes: string
+  fontWeights: string
+  layerStyles: string
+  letterSpacings: string
+  lineHeights: string
+  radii: string
+  shadows: string
+  sizes: string
+  space: string
+  textStyles: string
+  zIndices: string
+  components: {
+    [componentName: string]: {
+      sizes: string
+      variants: string
+    }
+  }
+}
+
+/**
+ * This is a placeholder meant to be implemented via TypeScript's Module
+ * Augmentation feature and is an alternative to running `npx @chakra-ui/cli
+ * tokens`
+ *
+ * @example
+ * ```ts
+ * import { BaseThemeTypings } from "@chakra-ui/styled-system";
+ * 
+ * type DefaultSizes = 'small' | 'medium' | 'large';
+ * 
+ * declare module "@chakra-ui/styled-system" {
+ *   export interface CustomThemeTypings extends BaseThemeTypings {
+ *     // Example custom `borders` tokens
+ *     borders: 'none' | 'thin' | 'thick';
+ *     // ...
+ *     // Other custom tokens
+ *     // ...
+ *     components: {
+ *       Button: {
+ *         // Example custom component sizes and variants
+ *         sizes: DefaultSizes;
+ *         variants: 'solid' | 'outline' | 'wacky' | 'chill';
+ *       };
+ *       // ...
+ *      }
+ *   }
+ * ```
+ *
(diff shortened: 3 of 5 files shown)


jrolfs:
Oops, I forgot a changeset. I added one and updated the inline documentation to extend `BaseThemeTypings`. Do y'all have any feedback for me here? I'm happy to open a corresponding PR against the documentation site as well.

Thank you!

TimKolberger:
Love this PR! 💖 

I'll review it over the weekend, because I have currently only access to my phone..

I am excited! 🎉

TimKolberger:
Hi @jrolfs, 
I tried to test this feature in a new CRA project, with the built and linked chakra packages from this PR, but I failed.
The example app is available here: https://github.com/TimKolberger/chakra-5579 where I described my setup in the README and I appended it here.

I just got autocomplete from the default theme and not the augmented types (IDEA nor VSCode).

image

Can you kindly give me hint how to test this?

```
cd <chakra-repo>
# checkout PR branch
cd <chakra-repo>/packages/react
yarn link
cd <chakra-repo>/packages/styled-system
yarn link

cd <this-repo>
yarn link "@chakra-ui/react"
yarn link "@chakra-ui/styled-system"

# try autocomplete in /src/App.tsx
```

jrolfs:
I just pulled down your repo and linked in my existing Chakra branch and it seems to be working in Code.

cra

I'll dig in a bit more tomorrow and make sure I didn't forget something in the PR.

I was hoping to be able to just set up a CodeSandbox with the PR releases to test and demo this, but now that I look I don't see those happening on PRs anymore? I thought every PR got released on the PR dist-tag?

jrolfs:
Aha, I forgot to export `CustomThemeTypings` from the index of **@chakra-ui/styled-system**! I amended my commit and pushed that up. You should be able to just pull down this PR again and rebuild with that project linked.

TimKolberger:
Awesome! I'll try it this evening.

> Here's an example of the PR releases I was talking about: #5290 (comment)

The PR releases were only available for core members until I disabled them yesterday completely, because they had a bug where we released some packages unwillingly without an explicit release and without a pr dist tag 😬

Releasing unreviewed code in the chakra nom namespace raises some security concerns.

TimKolberger:
Thanks, that worked great! ✨ 

I tinkered a bit with the module augmentation and came up with this question:

The following module augmentation works [in my test repository ](https://github.com/TimKolberger/chakra-5579/commit/9e9ffd9a670060edaff16e2dd91dc110424aa84e) with the **current** version of chakra ui:

```ts
type DefaultSizes = 'small' | 'medium' | 'large';

declare module "@chakra-ui/styled-system" {
    borders: 'none' | 'thin' | 'thick';
    components: {
      Button: {
        sizes: DefaultSizes;
        variants: 'wacky' | 'chill';
      };
    }
  }
}
```

Why do we need to introduce a new interface `CustomThemeTypings`?
If we augment `ThemeTypings` in our application codebase (or shipped via a "on-top-chakra component lib") those typings will have precedence over the typings generated by the CLI (in `node_modules`). 
If the user wants to use the chakra cli because they are using `extendTheme` in their codebase, the changes would not be picked up, because still the augmented types have precedence.

In my current understanding the only way to override the augmented types from the "on-top-chakra component lib" for the user is to use augmented types again.

```mermaid
graph TD
    Chakra[chakra default ThemeTypings] -->|via CLI| CompLib[on-top-chakra component lib]
    CompLib -->|via TS Module Augmentation| App
    App -->|ONLY possible via TS Module Augmentation| App
```

You mentioned that we could add a feature to the CLI to generate the augmented type definitions:

> I was even thinking we could combine this approach with the CLI with a new option to ouput a version of the definitions that leverages the above approach as opposed to modifying the definitions directly in ./node_modules.

And I think this is a great idea! And I think this would make this PR obsolete 🤔 

Kindly let me know if I am missing something!

jrolfs:
Sorry, I guess I should taken notes back when I started trying to _only_ leverage module augmentation to accomplish this. Going from your latest commit, it looks like component/theme props (`size`, `variant`) etc. do work, but I cannot get the types to work for token/regular style props:

When we don't import `@chakra-ui/styled-system`, these props (like `margin`) don't show up at all:
space--no-import

when we do import it, the default types take precedence:
space--with-import

Maybe we can modify the [`Token`](https://github.com/jrolfs/chakra-ui/blob/main/packages/styled-system/src/utils/types.ts#L16-L21) type? Although, I suspect it might require something similar to what I did with `ThemeTypings`.

TimKolberger:
> ...but I cannot get the types to work for token/regular style props [... without the changes in this PR]

True that! Thank you for laying our your thought process, I did learn a lot 💖 

IMO this is good to go as is 🚀 

This feature implicates a few follow up steps:

- update the docs as you mentioned
    -  kindly add a warning that the CLI generated types might not work if module augmentation is used 
- add a feature to the CLI to generate the augmented TS file

✨ Awesome work @jrolfs!

TimKolberger (approved):

jrolfs:
Hey, @TimKolberger thank you so much for testing this and for the feedback! I meant to at least acknowledge your concerns illustrated with that pretty diagram ;) — I think for this particular use case that further type customization at the application level won't really be a thing, but I do think it's important that we make that limitation clear...
> kindly add a warning that the CLI generated types might not work if module augmentation is used

as, yes, once `CustomThemeTypings` is implemented, the default CLI output will no longer work. I'm hoping that explaining each use case in the documentation will make this clear, but I will make sure to call that out specifically. I hope to have PRs up for this stuff later this week :)

PR #5595, opened 2022-02-16
fix(accordion):  introduce `styles.root` for `Accordion`

### LINKED ISSUE #5593: Accordion items not stylable as a theme part (or at least not how you'd expect it to)
### Description

When I specify styles for the accordion's `item` parts key.

### Link to Reproduction

https://codesandbox.io/s/dazzling-lichterman-95c36v?file=/src/index.js

### Steps to reproduce

1. Open the code sandbox
2. Notice how the theme.components.Accordion.baseStyle.item styles are ignored
3. Notice how the theme.components.Accordion.baseStyle.container styles are placed on the AccordionItem

I'm implementing my own design for accordions through specifying my own theme.

When I specify changes for the item key, they're just ignored.

When I look at the code, it looks like the `styles.container` styles are spread onto the AccordionItem, not the Accordion.

This should probably be addressed, and doing so would be pretty straight forward, but it would break existing implementations. 

Should we change it so that `container` styles are added to the `Accordion` component, and `item` styles to the `AccordionItem` component, or should we save this change for a future major release?

### Chakra UI Version

1.6.4

### Browser

_No response_

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

_No response_

takethefake:
closes #5593 

## 📝 Description

Updated style config for `Accordion`

This change introduces `styles.root` for `Accordion`

No

### DIFF
--- a/packages/accordion/src/accordion.tsx
+++ b/packages/accordion/src/accordion.tsx
@@ -66,6 +66,7 @@ export const Accordion = forwardRef<AccordionProps, "div">(
               ref={ref}
               {...htmlProps}
               className={cx("chakra-accordion", props.className)}
+              __css={styles.root}
             >
               {children}
             </chakra.div>
@@ -86,14 +87,12 @@ if (__DEV__) {
 
 type AccordionItemContext = Omit<UseAccordionItemReturn, "htmlProps">
 
-const [
-  AccordionItemProvider,
-  useAccordionItemContext,
-] = createContext<AccordionItemContext>({
-  name: "AccordionItemContext",
-  errorMessage:
-    "useAccordionItemContext: `context` is undefined. Seems you forgot to wrap the accordion item parts in `<AccordionItem />` ",
-})
+const [AccordionItemProvider, useAccordionItemContext] =
+  createContext<AccordionItemContext>({
+    name: "AccordionItemContext",
+    errorMessage:
+      "useAccordionItemContext: `context` is undefined. Seems you forgot to wrap the accordion item parts in `<AccordionItem />` ",
+  })
 
 export interface AccordionItemProps
   extends Omit<HTMLChakraProps<"div">, keyof UseAccordionItemProps>,
--- a/packages/anatomy/src/index.ts
+++ b/packages/anatomy/src/index.ts
@@ -2,13 +2,14 @@ import { anatomy } from "@chakra-ui/theme-tools"
 
 /**
  * **Accordion anatomy**
- * - Item: the accordion item contains the button and panel
+ * - Root: the root container of the accordion
+ * - Container: the accordion item contains the button and panel
  * - Button: the button is the trigger for the panel
  * - Panel: the panel is the content of the accordion item
  * - Icon: the expanded/collapsed icon
  */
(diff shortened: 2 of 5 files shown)


TimKolberger:
Thank you @takethefake for your time creating this PR!

I think there might be a way to fix this issue without a breaking change.

What do you think about adding a new part `root` to the [accordionAnatomy](https://github.com/chakra-ui/chakra-ui/blob/main/packages/anatomy/src/index.ts) and using it instead of moving `container` from the AccordionItem to the Accordion?

With v2 in mind we could reuse `root` for every MultipartComponent to make it a pattern. Why `root` and not `container`? Some components might have a container, like Tab or Accordion, but when thinking about a button it is hard to define where the container is. `root` is IMO easy to identify as the style root, which is always the top most DOM element of a component.

Kindly let me know your thoughts!

takethefake:
Hej @TimKolberger 

this sounds like a good approach to provide new functionality. We definitely need to make sure in v2 that we unify the usage of `root` / `container` to be clearly.

Adjusted the pr accordingly.

packages/anatomy/src/index.ts
```diff
@@ -2,13 +2,14 @@ import { anatomy } from "@chakra-ui/theme-tools"
...
- * - Item: the accordion item contains the button and panel
+ * - Root: the root container of the accordion
+ * - Container: the accordion item contains the button and panel
  * - Button: the button is the trigger for the panel
  * - Panel: the panel is the content of the accordion item
  * - Icon: the expanded/collapsed icon
  */
 export const accordionAnatomy = anatomy("accordion")
-  .parts("container", "item", "button", "panel")
+  .parts("root", "container", "button", "panel")
```
TimKolberger, line 12: I think the `item` is missing in the list.

packages/accordion/src/accordion.tsx
```diff
@@ -58,6 +58,10 @@ export const Accordion = forwardRef<AccordionProps, "div">(
       [context, reduceMotion],
     )
 
+    const rootStyles: SystemStyleObject = {
```
TimKolberger, line 61: Could this be inlined to the __css prop?

TimKolberger (changes requested):
Thank you for updating the PR so fast, I added a few comments.
Then it looks good to go 🚀

packages/accordion/src/accordion.tsx
takethefake: sure!

packages/anatomy/src/index.ts
takethefake: we don't use `item` in accordion. I think this is why those styles are kind of mixed up.

TimKolberger: Ah got it. Thanks 👍

TimKolberger (approved):

packages/accordion/src/accordion.tsx
TimKolberger: Awesome 😊

takethefake: @TimKolberger adjusted it

PR #5599, opened 2022-02-16
fix: not overwritable aria label prop on TagCloseButton

### LINKED ISSUE #5587: Allow for properties of a component to be overriden via props
### Description

If I pass in a prop to a chakra component, I expect my props to have a higher priority than the default props. 

As applications develop, there are likely different uses that lead to the need to have finer control over props on a component. Or else we'd need to compose our own components.

Recently, I wanted to add an aria label to `TagCloseButton` because `close` wasn't as descriptive as I'd like. However, I was unable to because the component's defaults have a higher priority.

I'm not sure how many other props in other components have this issue. Maybe this should be scoped down to aria labels and different issues for other properties should be brought up accordingly(?)

### Proposed Solution or API

In the case of `TagCloseButton` at least, the props could be reordered here:
https://github.com/chakra-ui/chakra-ui/blob/ead2bd65978fdaa84c14fd3d60d15c3ec4285d87/packages/tag/src/tag.tsx#L111-L119

to be:
```js
    <chakra.button
      type="button"
      aria-label="close"
      disabled={isDisabled}
      __css={btnStyles}
      {...rest}
    >
      {children || <TagCloseIcon />}
    </chakra.button>
```

### Alternatives

_No response_

### Additional Information

_No response_

m4x3d:
Closes #5587 

## 📝 Description

Change order of aria-label prop on TagCloseButton to be overwritable

aria-label of TagCloseButton can not be overwritten

## 🚀 New behavior

aria-label of TagCloseButton can be overwritten

No

### DIFF
--- a/.changeset/yellow-baboons-hang.md
+++ b/.changeset/yellow-baboons-hang.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/tag": patch
+---
+
+Change order of aria-label prop on TagCloseButton to be overwritable
--- a/packages/tag/src/tag.tsx
+++ b/packages/tag/src/tag.tsx
@@ -109,9 +109,9 @@ export const TagCloseButton: React.FC<TagCloseButtonProps> = (props) => {
 
   return (
     <chakra.button
+      aria-label="close"
       {...rest}
       type="button"
-      aria-label="close"
       disabled={isDisabled}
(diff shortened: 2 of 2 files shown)


TimKolberger:
Thank you @m4x3d 💖🚀

PR #5601, opened 2022-02-17
feat(theme): migrate from mode to _dark and _light selectors

takethefake:
## 📝 Description

Currently working on a codemod to migrate from `mode` in our components to `_dark` and `_light` pseudo selectors

This represents the current result.

See codemod implementation: https://github.com/chakra-ui/chakra-codemod/pull/22

### DIFF
--- a/.changeset/rude-rats-sit.md
+++ b/.changeset/rude-rats-sit.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/theme": minor
+---
+
+Using `_dark`&`_light` pseudo selectors in favor of mode for all components when
+ever possible
--- a/packages/theme-tools/src/color.ts
+++ b/packages/theme-tools/src/color.ts
@@ -48,12 +48,11 @@ export const isLight = (color: string) => (theme: Dict) =>
  * @param color - the color in hex, rgb, or hsl
  * @param opacity - the amount of opacity the color should have (0-1)
  */
-export const transparentize = (color: string, opacity: number) => (
-  theme: Dict,
-) => {
-  const raw = getColor(theme, color)
-  return new TinyColor(raw).setAlpha(opacity).toRgbString()
-}
+export const transparentize =
+  (color: string, opacity: number) => (theme: Dict) => {
+    const raw = getColor(theme, color)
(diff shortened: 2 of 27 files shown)


TimKolberger:
Thank you for working on this @takethefake!

Creating a codemod for this will be a great help for large codebases ✨

For the migration from 0.8 to 1.0 we created a codemod package: https://github.com/chakra-ui/chakra-codemod

I think it would be a great addition to add the mode-to-_dark mod there.

To merge this PR we should get the codemod out there first to provide our users with a tool to migrate their codebase.

I am not quite sure if we should label this as a breaking change or not. I remember a discussion if changes in the styling should be breaking changes. In this case I am leaning towards yes:

If the userland codebase uses the mode function to implement dark mode, it will be overridden by the current changes, because the pseudo selector _dark has a higher CSS precedence. This would result in broken dark mode styles for all codebases which use the extendTheme function in conjunction with the mode function.

What do you think @chakra-ui/react-team?

takethefake:
Current progress can be seen here: https://github.com/chakra-ui/chakra-codemod/pull/22

segunadebayo:
Why don't we do this manually within the codebase so we can get it released? Then we can provide a codemod for our consumers?

takethefake:
Hej @segunadebayo,@TimKolberger!

I've adjusted our components to reflect `_dark`&`_light`-selectors wherever possible. 

- `SystemStyleFunction` became `SystemStyleObject` if `props` are no longer needed
-`PartStyleFunction` became `PartStyleObject` if `props` are no longer needed.

The codemod get's most cases right  - regarding the chakra-ui codebase this was about 80%/90%.

segunadebayo:
After much thought, I think this is going to be a much bigger breaking change than most users would like as Tim described above.

I honestly appreciate your work on this @takethefake. I'd recommend that we take a design-centric approach and rely on semantic tokens instead. That way, the breaking changes are minimal and properties can be sanely overridden.

Let's discuss this at the meeting today but permit me to close this for now. Let's encourage consumers to use the codemod within their codebases instead.

Thanks for understanding.

PR #5602, opened 2022-02-17
feat(theme-utils): fallback to semanticToken-value when calling `getColor`

takethefake:
## 📝 Description

When you are defining custom components-themes with semantic tokens and need to resolve your color value on your own (.e.g. `box-shadow`) you can't resolve the token as usual with `getColor` since `getColor` only supports a lookup into `colors.${tokenValue}` 

## 🚀 New behavior

`getColor` now also tries to resolve the color from  `semanticTokens` before returning the fallback

### DIFF
--- a/packages/theme-tools/test/color.test.ts
+++ b/packages/theme-tools/test/color.test.ts
@@ -1,0 +1,30 @@
+import { getColor } from "../src"
+
+test("gets color from theme", () => {
+  expect(
+    getColor({ colors: { red: { 200: "#ff0000" } } }, "red.200", "red.200"),
+  ).toBe("#ff0000")
+})
+
+test("can get semantic token from theme", () => {
+  expect(
+    getColor(
+      { semanticTokens: { colors: { error: { 200: "#ff0000" } } } },
+      "error.200",
+      "error.200",
+    ),
+  ).toBe("#ff0000")
+})
+
+test("takes color over semanticToken", () => {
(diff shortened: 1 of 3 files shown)


packages/theme-tools/test/color.test.ts
```diff
@@ -0,0 +1,30 @@
...
+test("gets color from theme", () => {
+  expect(
+    getColor({ colors: { red: { 200: "#ff0000" } } }, "red.200", "red.200"),
+  ).toBe("#ff0000")
+})
+
+test("can get semantic token from theme", () => {
+  expect(
+    getColor(
+      { semanticTokens: { colors: { error: { 200: "#ff0000" } } } },
```
TimKolberger, line 12: The semanticTokens property of the theme only supports one level of nesting. E.g.

```js
{ semanticTokens: { colors: { "error.200": "#ff0000" } } }
```

And I think the lodash get function has does not quite work with this.

TimKolberger (commented):
Thanks for this PR!

The semantic tokens mostly consist of references to base tokens or even a conditional configuration object. 

E.g. 

```js
"error": "red.200",
"formBg": { default: "gray.50", _dark: "gray.700" }
```

Because the nature of semantic tokens are context sensitive I cannot think of a way to simply "get the color of a semantic token".

IMO it is fine that getColor only resolves base colors.

What do you think?

PR #5618, opened 2022-02-18
feat(form-control): add optional indicator for `FormLabel`

takethefake:
## 📝 Description

Added `optionalIndicator` to `FormLabel` to provide a way to signalize if a field is optional

Currently, we only show `<RequiredIndicator/>` if a field is `required` to do the same for the inverse side the `Indicator has to be passed to the `children`

## 🚀 New behavior

made `optionalIndicator` a prop which can be passed to `<FormLabel/>` with default value `null` which renders if the field is not required.

### DIFF
--- a/.changeset/sharp-crews-wave.md
+++ b/.changeset/sharp-crews-wave.md
@@ -1,0 +1,8 @@
+---
+"@chakra-ui/form-control": minor
+---
+
+Introduced `optionalIndicator` for `FormLabel`
+
+Similar to the `RequiredIndicator` the `OptionalIndicator` signalizes when a
+field is optional.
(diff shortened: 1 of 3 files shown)


.changeset/sharp-crews-wave.md
```diff
@@ -0,0 +1,8 @@
+---
+"@chakra-ui/form-control": patch
+---
+
+Introduced `optionalIndicator` for `FormLabel`
```
segunadebayo, line 5: This is a feature, so it should be a `minor`

segunadebayo: We should also create a PR to the docs repo to document this feature

takethefake: updated to minor

PR #5623, opened 2022-02-19
fix(popover): isLazy mounting for Popover

### LINKED ISSUE #5557: Popover isLazy not working
### Description

When I use the isLazy prop on a popover I expect the popover content to be lazily loaded

### Link to Reproduction

https://codesandbox.io/s/strange-lichterman-3o1lb?file=/src/index.tsx

### Steps to reproduce

Go to console and see the "im rendering" log.

It shows before clicking the popover trigger.

### Chakra UI Version

latest

### Browser

all

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

Brennvo:
Closes #5557

## 📝 Description

Updates the `useAnimationState` hook in order to determine if the animation should be present in the UI or not.

The `Popover` component's lazy behavior is initially mounting the popover content, which removes the benefit of its [intended lazy behavior](https://chakra-ui.com/docs/overlay/popover#lazily-mounting-popover).

## 🚀 New behavior

As a result of updating the `useAnimationState`'s `hidden` property, lazy content inside of the `Popover` will *not* mount until the `PopoverTrigger` is clicked (assuming the `Popover` is rendered with `unMount` as its `lazyBehavior` prop). 

No

## 📝 Additional Information

Unit tests for the `Popover` have been updated to better test the lazy behavior of popover content. This bug likely got through due to an initial mixup between the testing-library's [`toBeInTheDocument`](https://github.com/testing-library/jest-dom#tobeinthedocument) assertion versus the [`toBeVisible`](https://github.com/testing-library/jest-dom#tobevisible) assertion.

The other component using `useAnimationState` is the `Menu` component, which [doesn't seem to be using any logic associated with the `useAnimationState`'s `present` property](https://github.com/chakra-ui/chakra-ui/blob/main/packages/menu/src/menu.tsx#L182) (which is what is updated in this PR).


TimKolberger:
Thank you @Brennvo for this excellent work 💖

PR #5571, opened 2022-02-13
chore: remove deep imports

matt-tingen:
No issue

## 📝 Description

Removes deep imports that prevent storybook from building

## ⛳️ Current behavior (updates)

Storybook prints errors like:

```
Module not found: Error: Can't resolve '@chakra-ui/hooks/use-animation-state' in '<snip>/chakra-ui/packages/menu/src'
```

## 🚀 New behavior

Storybook builds successfully

## 💣 Is this a breaking change (Yes/No):

No

PR #5624, opened 2022-02-19
chore: update storybook instructions

matt-tingen:
No issue; prompted by https://github.com/chakra-ui/chakra-ui/pull/5571

## 📝 Description

Recent changes to entry points have caused the storybook build to be
inadequate resulting in errors like
`Can't resolve '@chakra-ui/hooks/use-animation-state'`.

Following the instructions for running storybook results in a failed build.

## 🚀 New behavior

The instructions reflect the current requirements for running storybook

No

## 📝 Additional Information

Is this accurate? Or does `yarn build` only need to be run once when changing branches (and then you just need `yarn preconstruct` + `yarn storybook` to get SB running)?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,9 @@ end with `.stories.tsx`.
 
 ### Storybook
 
-Build components in isolation with Storybook using `yarn storybook`
+Build components in isolation with Storybook. When using Storybook, the build
+process needs to be run separately. Run `yarn build && yarn preconstruct watch`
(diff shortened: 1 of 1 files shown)


TimKolberger:
> Is this accurate? Or does yarn build only need to be run once when changing branches (and then you just need yarn preconstruct + yarn storybook to get SB running)?

The recipe `yarn build && yarn preconstruct watch` will work in all circumstances.
You could omit the build script if there are no changes in the `babel` or `cli` packages, but let's not overcomplicate the onboarding docs.

Thank you for this PR @matt-tingen! ✨

PR #5631, opened 2022-02-20
Fix/menu item option icon

### LINKED ISSUE #5629: Allow disabling Check Icon in `MenuItemOption`
### Description

In the `Menu` component, we can use `MenuOptionGroup` to have a `radio` or `checkbox` functionality, The chosen item usually has a check Icon:
CleanShot 2022-02-20 at 19 26 47@2x

The `MenuItemOption` has an `icon` prop, that allows us to dictate what we we might want to use in place of this `Check` Icon. Now I'm using this functionality, but I don't want any icon at all. I tried setting `icon` to `null`. But that doesn't help.

### Proposed Solution or API

By normal React convention, a null component is not displayed, so I feel we should not display the Check Icon when null is passed, If not, we can accept a prop perhaps on the `MenuOptionGroup` or the Item itself to disable this icon.

### Alternatives

My current approach is to pass `<></>` to the `icon` prop

anubra266:
Closes #5629

## Explanations ca be found [in issue page](https://github.com/chakra-ui/chakra-ui/issues/5629)

This PR prevents rendering the `MenuItemOption` Icon when `null` is passed as it's value

## 💣 Is this a breaking change (Yes/No): No

### DIFF
--- a/.changeset/twelve-singers-prove.md
+++ b/.changeset/twelve-singers-prove.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/menu": patch
+---
+
+Fixed bug where passing `null` as value of `icon` prop in `MenuOptionItem` still
+rendered the icon.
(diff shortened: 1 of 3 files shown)


TimKolberger (approved):
Awesome @anubra266 💖

PR #5576, opened 2022-02-14
fix: use sorted breakpoints in useBreakpoint

TimKolberger:
Closes #5555 (not yet. The CSS responsive object uses media min queries, where as the useBreakpoint uses minmax queries)

## 📝 Description

Fixed an issue where the hook `useBreakpoint` did not work as expected with custom breakpoints

## ⛳️ Current behavior (updates)

useBreakpoint used user defined values from the theme. It is operating on an unsorted object.

## 🚀 New behavior

It uses the processed and sorted breakpoints from `toCSSVar(theme)`.

## 💣 Is this a breaking change (Yes/No):

No

PR #5638, opened 2022-02-21
Fix/media query default breakpoint unused

ishowta:
## 📝 Description

Since `matchMedia` always returns `false` during SSR, the code to check `defaultBreakpoint` was not reached by  [this refactor](https://github.com/chakra-ui/chakra-ui/pull/5576)
Also, the test was not able to set up the same situation as SSR, so it was not failed.

https://github.com/chakra-ui/chakra-ui/blob/60fc221f362c85a80d1534525dc45c499116433f/packages/media-query/src/use-breakpoint.ts#L27-L45

Always return undefined during SSR

## 🚀 New behavior

return `defaultValue` when SSR

No

### DIFF
--- a/.changeset/thick-swans-attend.md
+++ b/.changeset/thick-swans-attend.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/media-query": patch
+---
+
+Fixed an issue that undefined is returned when calling the hook `useBreakpoint`
+with `defaultValue` specified in SSR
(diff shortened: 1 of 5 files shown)


TimKolberger (approved):
Thank you for this PR @ishowta 💖
Especially for the updated test suite!

TimKolberger:
Thank you again for you work @ishowta 

I used the changes to fix another SSR / CSR issue with the useBreakpointValue hook.

Your commits + history are still intact and all credits are going to you for fixing this bug!
I'll close this PR in favor of #5651.

@all-contributors please add @ishowta for code.

PR #5639, opened 2022-02-21
fix: add color mode script runtime safeguard

sQVe:
## 📝 Description

This PR adds a runtime safeguard to `ColorModeScript` since there's no need for the values to be dynamic and it's creating a `script` with `dangerouslySetInnerHTML`.

The `initialColorMode` passed to `ColorModeScript` is currently only guarded via TypeScript typing. Setting `initialColorMode` to evil values like `(() => { console.log('hello'); return 'light'; })()` is possible.

This security issue is obviously far fetched but it is nevertheless there, and made worst since `ColorModeScript` obscures the fact that it's using `dangerouslySetInnerHTML`. 

## 🚀 New behavior

Adds hardening to `ColorModeScript` by runtime checking for three specific color mode values: `dark`, `light` and `system`.

No

### DIFF
--- a/.changeset/poor-llamas-obey.md
+++ b/.changeset/poor-llamas-obey.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/color-mode": patch
+---
+
+Adds a runtime safeguard for `ColorModeScript`.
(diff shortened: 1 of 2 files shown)


sQVe:
I'm unsure why the `test` are failing. It's passing fine locally on my machine and I'm unable to rerun the action.

TimKolberger (approved):

PR #5640, opened 2022-02-21
Add option to specify initially focused menu item

### LINKED ISSUE #5598: Add option to specify initially focused menu item
### Description

Add an option in the menu's props to define which item is focused initially when it is opened.

We use the menu as a custom select component, which works in functionality for most parts, except that a select component requires that the previously selected item is the one that is focused initially when it is opened. Chakra always automatically focuses the first menu item when it is opened. I understand that this might be a stretch for an use case for the menu component, since it is not a select component, but the solution I believe is really easy to impelement and might be helpful to others that use the component in a similar way (e.g. #3411).

### Proposed Solution or API

* Add a `initialFocusIndex?: number;` entry in the `UseMenuProps` here: https://github.com/chakra-ui/chakra-ui/blob/main/packages/menu/src/use-menu.ts#L69
* Extract it with `initialFocusIndex = -1,` from here: https://github.com/chakra-ui/chakra-ui/blob/main/packages/menu/src/use-menu.ts#L144
* Use it as initial focus index value for initialization by using `const [focusedIndex, setFocusedIndex] = React.useState(initialFocusIndex)` here: https://github.com/chakra-ui/chakra-ui/blob/main/packages/menu/src/use-menu.ts#L217

### Alternatives

* Use a regular HTML select input, however that does not yield the customization possibilities as does Chakras menu component

### Additional Information

Of course I'm happy to take a PR on this if this feature is desired.

lukasbach:
Closes #5598

## 📝 Description

Adds the option to specify which menu item is focused initially when the menu is opened.

Always focuses the first item when `autoSelect` is true.

## 🚀 New behavior

Behavior remains the same if `initialFocusIndex` is undefined, and focuses the item at index `initialFocusIndex` if provided and `autoSelect` is true.

No. The new behavior is controlled by a new prop.

## 📝 Additional Information

-

### DIFF
--- a/packages/menu/src/use-menu.ts
+++ b/packages/menu/src/use-menu.ts
@@ -119,6 +119,12 @@ export interface UseMenuProps
    * as it might affect scrolling performance.
    */
   computePositionOnMount?: boolean
+  /**
+   * If set, the initially focused item will be the item with the given index.
+   * Keyboard navigation will start out at this item. If not set, the first item
+   * will be focused.
+   */
+  initialFocusIndex?: number;
 }
 
 /**
@@ -142,6 +148,7 @@ export function useMenu(props: UseMenuProps = {}) {
     lazyBehavior = "unmount",
     direction,
     computePositionOnMount = false,
+    initialFocusIndex = -1,
     ...popperProps
   } = props
   /**
@@ -214,7 +221,7 @@ export function useMenu(props: UseMenuProps = {}) {
     direction,
   })
 
-  const [focusedIndex, setFocusedIndex] = React.useState(-1)
+  const [focusedIndex, setFocusedIndex] = React.useState(initialFocusIndex)
 
   /**
    * Focus the button when we close the menu
--- a/packages/menu/src/use-menu.ts
+++ b/packages/menu/src/use-menu.ts
@@ -124,7 +124,7 @@ export interface UseMenuProps
    * Keyboard navigation will start out at this item. If not set, the first item
    * will be focused.
    */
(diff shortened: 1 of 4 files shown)


lukasbach:
Demo: https://chakra-ui-storybook-git-fork-lukasbach-main-chakra-ui.vercel.app/?path=/story/components-overlay-menu--with-initial-focus

packages/menu/src/use-menu.ts
```diff
@@ -162,13 +169,13 @@ export function useMenu(props: UseMenuProps = {}) {
     })
   }, [])
 
-  const focusFirstItem = React.useCallback(() => {
+  const focusInitialItem = React.useCallback(() => {
     const id = setTimeout(() => {
-      const first = descendants.firstEnabled()
-      if (first) setFocusedIndex(first.index)
+      const itemIndex = initialFocusIndex ?? descendants.firstEnabled()?.index
```
burhanuday, line 174: the type has changed from `descendants.firstEnabled()` to `descendants.firstEnabled()?.index`

segunadebayo:
Thanks for working on this. I think we should favor `initialFocusRef` over `initialFocusIndex` since it aligns with other APIs and relies less on the DOM order of menu items (which isn't SSR friendly).

lukasbach:
Hi @segunadebayo, thanks for your feedback. I'm busy this week, but will look into updating this next week.

segunadebayo:
Closing in favor of https://github.com/chakra-ui/chakra-ui/pull/7032

PR #5234, opened 2021-12-10
fix: #4251 layout flashing from useBreakpoint

nxtman123:
If `env.window.matchMedia` is available when the hook mounts, use it to find the initial state. This prevents the value changing the next "tick" after the hook mounts, often causing layout changes that look like flashes on the screen.

I am having trouble with the tests, the SSR test suite doesn't mock matchMedia, but it seems to be "working" and matching the "sm" breakpoint. In SSR it should return the default breakpoint as I have tested for `env.window.matchMedia`.

Addresses #4251

## 📝 Description

Even though `defaultBreakpoint` was added, the essential problem of an incorrect first value is still present.

## ⛳️ Current behavior (updates)

When useBreakpoint mounts, its first returned value is `defaultBreakpoint` or `undefined`, even when `window.matchMedia` is available.
Screen Shot 2021-12-10 at 2 36 17 PM

## 🚀 New behavior

When `window.matchMedia` is available, the first returned value is correct.
Screen Shot 2021-12-10 at 2 33 34 PM

## 💣 Is this a breaking change (Yes/No):

No

PR #5651, opened 2022-02-23
fix: first render consistency of useBreakpointValue in SSR+CSR envs

### LINKED ISSUE #5506: Chakra-UI + NextJS - useBreakpointValue not working with NextJS!
### Description

 When I use the useBreakpointValue hook inside a component using NextJS, the application breaks and I get the error: Warning: Did not expect server HTML to contain a div in div.

### Link to Reproduction

https://codesandbox.io/s/keen-chaplygin-n7zcw

### Steps to reproduce

1. Go to https://codesandbox.io/s/keen-chaplygin-n7zcw
2. Wait for the application to load and see the console

### Chakra UI Version

@chakra-ui/core = 0.8.0 / @chakra-ui/react = 1.8.1 

### Browser

Google Chrome 97.0.4692.99

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

@chakra-ui/react: latest (1.8.1)
@emotion/react: latest (11.7.1)
@emotion/styled: latest (11.6.0)
@types/node: 17.0.12
@types/react: 17.0.38
framer-motion: latest (6.2.3)
next: 12.0.9
react: 17.0.2
react-dom: 17.0.2
typescript: 4.5.5

image

TimKolberger:
Closes #5506
Based on PR #5638 💖  thanks to @ishowta

## 📝 Description

Ensure first render consistency in SSR+CSR environments.

The PR #5234 introduced the side effect of mismatching SSR+CSR renders.
This was not fixed by my previous PR #5576.

## 🚀 New behavior

To ensure the render consistency on the server and client I reordered the priority of the parameter `defaultBreakpoint` for `useBreakpoint`.
If a `defaultValue` is set, it has priority over `window.matchMedia`. 

The hook `useBreakpointValue` uses now a default value for `defaultBreakpoint` (`base`) to make use of this priority change.

No

### DIFF
--- a/packages/media-query/src/use-breakpoint-value.ts
+++ b/packages/media-query/src/use-breakpoint-value.ts
@@ -8,7 +8,7 @@ import { useBreakpoint } from "./use-breakpoint"
  * provided responsive values object.
  *
  * @param values
- * @param [defaultBreakpoint="base"] default breakpoint name
+ * @param [defaultBreakpoint] default breakpoint name
  * (in non-window environments like SSR)
  *
  * For SSR, you can use a package like [is-mobile](https://github.com/kaimallea/isMobile)
@@ -19,7 +19,7 @@ import { useBreakpoint } from "./use-breakpoint"
  */
 export function useBreakpointValue<T = any>(
   values: Partial<Record<string, T>> | T[],
-  defaultBreakpoint = "base", // default value ensures SSR+CSR consistency
+  defaultBreakpoint?: string,
 ): T | undefined {
   const breakpoint = useBreakpoint(defaultBreakpoint)
   const theme = useTheme()
--- a/packages/media-query/src/use-breakpoint-value.ts
+++ b/packages/media-query/src/use-breakpoint-value.ts
@@ -8,7 +8,7 @@ import { useBreakpoint } from "./use-breakpoint"
  * provided responsive values object.
  *
  * @param values
- * @param defaultBreakpoint default breakpoint name
+ * @param [defaultBreakpoint="base"] default breakpoint name
  * (in non-window environments like SSR)
  *
  * For SSR, you can use a package like [is-mobile](https://github.com/kaimallea/isMobile)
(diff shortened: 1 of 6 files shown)


packages/media-query/src/use-breakpoint-value.ts
```diff
@@ -19,7 +19,7 @@ import { useBreakpoint } from "./use-breakpoint"
  */
 export function useBreakpointValue<T = any>(
   values: Partial<Record<string, T>> | T[],
-  defaultBreakpoint?: string,
+  defaultBreakpoint = "base", // default value ensures SSR+CSR consistency
```
ishowta, line 22: If `defaultBreakpoint` is set by default, layout flashing will probably occur even in a Vanilla environment.

Although rendering mismatch may be more serious than layout flashing...

ishowta: I think it would be better to have a global `defaultBreakPoint` setting for users who use SSR.

TimKolberger: Exactly! We need to support SSR. And this is by my knowledge only possible with a rerender on the client.

It's a great idea to configure this via the theme. I think it's best to do this iterative. 

1. fix the bug
2. add the theme config in another PR

cschroeter:
LGTM.

Fixes issues with Next.js

cschroeter (approved):

packages/media-query/src/use-breakpoint.ts
```diff
@@ -25,20 +25,25 @@ export function useBreakpoint(defaultBreakpoint?: string) {
   )
```
ishowta, line 26: Don't we need to set the default value to `useBreakpoint` as well as `useBreakpointValue`?

TimKolberger: Your right, I think we should move the default value from useBreakpointValue to useBreakpoint to ensure a safe SSR+CSR for both hooks.

Issue #5664, opened 2022-02-28
TabIndicator background color can briefly flicker over an adjacent tab when tabs have differing widths and the selected tab changes.


josulliv101:
### Description

When tab widths differ and the TabIndicator is present, there are times when the TabIndicator bg flickers over an adjacent tab when the transition to the newly selected tabs starts.

### Link to Reproduction

https://codesandbox.io/s/pedantic-dawn-ds3hfu?file=/src/index.tsx

### Steps to reproduce

1. Go to 'https://codesandbox.io/s/pedantic-dawn-ds3hfu?file=/src/index.tsx'
2. Click on the second tab
3. Click back to the first tab
4. See error appear -- the green background of the TabIndicator flickers over the left side of the 3rd tab at the start of the transition.

### Chakra UI Version

1.8.5

### Browser

Google Chrome 98.0.4758.109

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

The code is doing what is suppose to do, it is just that when a newly selected tab happens to be wider than the others, the extra width can make the transition less than ideal visually because items start to overlap. One solution is to add "width" to the list of transitionProperties of the TabIndicator to prevent the overlap -- the transition looks cleaner in those circumstances.

A work-around exists by including the following property `<TabIndicator transitionProperty="top, left, bottom, right, width !important" ...` -- this seems like it may come up somewhat frequently for users of TabIndicator so I was going to put in a PR (https://github.com/chakra-ui/chakra-ui/pull/5665) to fix via including "width".

See below for an example of the bg flicker behavior on the tab
![tabindicator-bg-flash-when-long-text](https://user-images.githubusercontent.com/1703430/156041376-70b40fd6-bf19-41a5-8539-dba8d10ccf66.gif)

PR #5665, opened 2022-02-28
fix(TabIndicator): avoid bg flash when long tab text 

josulliv101:
Closes (https://github.com/chakra-ui/chakra-ui/issues/5664)

## 📝 Description

Add "width" to the list of transitionProperties of the TabIndicator to prevent the background from overlapping adjacent tabs. This issue occurs when one tab's width is larger than the other tabs.

The transition behavior when the selected tab changes.

## 🚀 New behavior

Including "width" as a transition property of the TabIndicator prevents any overlapping when the selected tab changes (assuming the tabs differ in width so the issue shows itself)

No

## 📝 Additional Information
Example of the issue:
https://codesandbox.io/s/pedantic-dawn-ds3hfu?file=/src/index.tsx

I did not include a test since the nature of the issue is more visual/timing related -- it seemed hard to create a test to capture that.

### DIFF
--- a/packages/tabs/src/use-tabs.ts
+++ b/packages/tabs/src/use-tabs.ts
@@ -276,7 +276,7 @@ export interface UseTabProps
  */
 export function useTab<P extends UseTabProps>(props: P) {
   const { isDisabled, isFocusable, ...htmlProps } = props
-
+  // test
   const { setSelectedIndex, isManual, id, setFocusedIndex, selectedIndex } =
     useTabsContext()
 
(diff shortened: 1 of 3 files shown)


TimKolberger (approved):

packages/tabs/src/use-tabs.ts
```diff
@@ -449,7 +444,7 @@ export function useTabIndicator(): React.CSSProperties {
 
   return {
     position: "absolute",
-    transitionProperty: "left, right, top, bottom",
+    transitionProperty: "left, right, top, bottom, width",
```
segunadebayo, line 447: I guess `height` should be included here as well, in the case of vertical tabs. Right?

josulliv101: Makes sense. I've included `height` now as well.

PR #5666, opened 2022-02-28
docs: Adds a badge to show off usage and adoption on live sites

davidbielik:
## 📝 Description

Adds a badge to show a live website count and examples across the web.

image


segunadebayo:
Hi @davidbielik,

Thanks for creating this PR. Although I see the value that this brings, I think the numbers showing currently aren't truly reflective of the "actual" number of sites using Chakra UI.

It's pretty low and can somehow discourage potential users of Chakra UI. We have over 1.1M downloads per month and it's hard to believe that only 568 websites use Chakra UI.

We won't be adding this to the README for now.

Thank you for understanding.

PR #5632, opened 2022-02-20
fix: conflicting zindex of linkoverlay with opacity

m4x3d:
Closes #5630 

## 📝 Description

Fixed an issue regarding the zIndex of the LinkOverlay in LinkBox.

## ⛳️ Current behavior (updates)

Content outside the LinkOverlay can not be clicked when having an opacity below 1.\
Codesandbox from the issue: https://codesandbox.io/s/laughing-sun-q1qjmw?file=/src/index.tsx

## 🚀 New behavior

Content outside the LinkOverlay can be clicked when having an opacity below 1.
Codesandbox with an override of the changed zIndex of this PR: https://codesandbox.io/s/elegant-varahamihira-9zuwk6?file=/src/index.tsx

## 💣 Is this a breaking change (Yes/No):

No

PR #5679, opened 2022-03-02
fix: buttons zIndex in linkbox with usage of link overlay

### LINKED ISSUE #5673: LinkOverlay breaks IconButton/Button within LinkBox
### Description

The latest release (1.8.6) introduced a new bug.

When using a LinkBox + LinkOverlay, any IconButton within the LinkBox becomes unresponsive.

### Link to Reproduction

https://codesandbox.io/s/bold-leaf-rhjbxj?file=/src/index.tsx

### Steps to reproduce

Hover over the + icon, you can see that there is no animation, and clicking it doesn't log anything to the console.

### Chakra UI Version

1.8.6

### Browser

Safari 

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Removing the LinkOverlay component seems to fix it, but obviously that makes the whole link box feature not very useful.

m4x3d:
Closes #5673 

## 📝 Description

This fixes a problem that buttons are no longer clickable within a `LinkBox` with a `LinkOverlay`  because of the zIndex. This is a result of the changes of https://github.com/chakra-ui/chakra-ui/pull/5632.

Buttons are not clickable if they are within a `LinkBox` with a `LinkOverlay`, see e.g. the CSB from the issue: https://codesandbox.io/s/bold-leaf-rhjbxj?file=/src/index.tsx

## 🚀 New behavior

Buttons are clickable again, this should represent the fix: https://codesandbox.io/s/stoic-glitter-9em593

No

## 📝 Additional Information

It worked also with `zIndex: 1` on Firefox 97.0 and Chromium 99.0.4844.51 but it feels wrong so that I set it to 2 to avoid unexpected behavior on other browsers/versions.

### DIFF
--- a/packages/layout/src/link-box.tsx
+++ b/packages/layout/src/link-box.tsx
@@ -54,8 +54,8 @@ export const LinkBox = forwardRef<LinkBoxProps, "div">((props, ref) => {
       {...rest}
       className={cx("chakra-linkbox", className)}
       __css={{
-        /* Elevate the links and abbreviations up */
-        "a[href]:not(.chakra-linkbox__overlay), abbr[title]": {
+        /* Elevate the links, abbreviations and buttons up */
+        "a[href]:not(.chakra-linkbox__overlay), abbr[title], button": {
           position: "relative",
           zIndex: 1,
         },
--- a/.changeset/happy-ads-confess.md
+++ b/.changeset/happy-ads-confess.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/layout": patch
+---
+
+Fix buttons zIndex within LinkBox and usage of LinkOverlay


packages/layout/src/link-box.tsx
```diff
@@ -54,10 +54,10 @@ export const LinkBox = forwardRef<LinkBoxProps, "div">((props, ref) => {
       {...rest}
       className={cx("chakra-linkbox", className)}
       __css={{
-        /* Elevate the links and abbreviations up */
-        "a[href]:not(.chakra-linkbox__overlay), abbr[title]": {
+        /* Elevate the links, abbreviations and buttons up */
+        "a[href]:not(.chakra-linkbox__overlay), abbr[title], button": {
           position: "relative",
-          zIndex: 1,
+          zIndex: 2,
```
TimKolberger, line 60: Did you encounter any issues with `zIndex: 1`?
I think `zIndex: 1` should suffice, because the `::before` element will always be _before_ any other child DOM element. Trailing DOM elements with `zIndex: 1` should be on top.

m4x3d: Aaah, that makes sense. I did not figured out why `zIndex: 1` worked as expected so I set it to 2 to avoid any other unexpected behavior with other browsers/browser version.

I force-pushed it.

Issue #5288, opened 2021-12-25
Image component loads fallbackSrc even without error


mythosil:
### Description

When I pass `src` and `fallbackSrc` props to Image component, I expect that `fallbackSrc` get loaded only when the browser fails to load `src` but the browser loads both simultaneously.

### Link to Reproduction

https://codesandbox.io/s/t51j5

### Steps to reproduce

1. Open a new tab in Chrome
2. Open developer tools' network tab
3. Go to https://codesandbox.io/s/t51j5
4. You can see fallbackSrc https://via.placeholder.com/240 is loaded

### Chakra UI Version

1.7.3

### Browser

Chrome 96, Firefox 95

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Image component loads fallbackSrc when status is not "loaded". The initial value of status is "loading", so fallbackSrc get loaded.
https://github.com/chakra-ui/chakra-ui/blob/96c4aabb/packages/image/src/image.tsx#L120-L135

It might be an intended design but I think fallbackSrc should be loaded only when error occurs to suppress network traffic.

primos63:
Changed status check from `status !== "loaded"` to `status === "failed"`. The fallback image is then only loaded when the image fails to load.
The [example](https://codesandbox.io/s/sandpack-project-forked-dqh98?file=/src/components/image/src/image.tsx) is from #5285 which this change will also fix. For this issue please comment out `fallback` and remove the `//` from `fallbackSrc`. If you add a 1 to the end of the `src` value the load will fail and the fallback will load.

mythosil:
I found this in image.stories.tsx:
> Chakra has support for fallback images so in event the image falls to load, or while the image is loading, you can show a fallback.
https://github.com/chakra-ui/chakra-ui/blob/96c4aabb/packages/image/stories/image.stories.tsx#L17-L19

But the docs mention only "error":
> You can provide a fallback image for when there is an error loading the src of the image.
https://chakra-ui.com/docs/media-and-icons/image#fallback-support

Maybe issue #1027 is related to this behavior.

segunadebayo:
Great thought @primos63 and thanks for opening this @mythosil.

For now, we can update the docs to reflect the true implementation.

An approach could be to include a `fallbackStrategy = "onError"  | "beforeLoadOrError"` which will determine how the fallback loading logic would be. Just an idea

trybick:
[...]

PR #5680, opened 2022-03-02
feat(image): add imageFallbackStrategies

### LINKED ISSUE #5581: Image fallback strategy
### Description

Add a `fallbackStrategy` prop to the Image component that defines the fallback image loading behavior. The prop should accept either `onError` or `beforeLoadOrError` (current behavior).

The current implementation of the Image component loads both the main and fallback image (if provided). The fallback image is displayed if the main one is currently loading or if an error occurred while loading it. I think developers should be given the option to define whether to display the fallback image or not while the main one is loading, as there are some use cases that loading the fallback only in case of error would be a better choice (saving resources for example).

related issues: #5288

### Proposed Solution or API

Implement a `fallbackStrategy` prop that would accept either 'onError' or 'beforeLoadOrError'

This idea was suggested by @segunadebayo on issue #5288 

### Alternatives

_No response_

### Additional Information

I'm happy to volunteer and work on the implementation!

takethefake:
Closes #5581 #5288

## 📝 Description

added fallback strategies to the image component

Currently unsure in a few points how to write a useful test for this since I'm not sure how to verify resource loads in jest

Also, not sure how to handle this specific case:

`fallbackStrategy==="onError" && Boolean(fallbackSrc) && status==="loading"`

In this case, I currently render `null` as `fallbackImage` which could lead to unexpected behaviour in the layout.

**The main problem is:** when `fallbackStrategy==='onError'` I want to load the `fallbackSrc` only if `status==='error'` therefor we don't have a resource to display when `status==='loading'`  

Possible Alternatives:
- render an `img` without the `src`-Attribute, but this doesn't comply to html5 standards since `src` is a required attribute`
- remove unnecessary `img` attributes and render a `Box` as a placeholder with the same props as the image component to just take the same space.

This showcases the current behavior: https://codesandbox.io/s/create-react-app-ts-forked-9zv672?file=/src/App.tsx to see the `loading`-case set throttling to `fast 3G`

What do you think @TimKolberger?

No

### DIFF
--- a/.changeset/chatty-hornets-shake.md
+++ b/.changeset/chatty-hornets-shake.md
@@ -1,0 +1,11 @@
+---
+"@chakra-ui/image": patch
+---
+
+Added `fallbackStrategies` for `Image` to provide different behaviors when to
+display/load the `fallback`
+
+- `beforeLoadOrError` is the default strategy and previous behaviour of `Image`
+  which displays/loads the placeholder when `loading` the `src`/`srcSet` and on
+  `error`
+- `onError` strategy displays/loads the fallback image only on `error`
--- a/packages/image/src/image.tsx
+++ b/packages/image/src/image.tsx
@@ -7,7 +7,12 @@ import {
 } from "@chakra-ui/system"
 import { omit, __DEV__ } from "@chakra-ui/utils"
 import * as React from "react"
-import { useImage, UseImageProps } from "./use-image"
+import {
+  FallbackStrategy,
+  shouldShowFallbackImage,
+  useImage,
+  UseImageProps,
+} from "./use-image"
 
 interface NativeImageOptions {
   /**
@@ -69,6 +74,16 @@ interface ImageOptions extends NativeImageOptions {
    * If `true`, opt out of the `fallbackSrc` logic and use as `img`
    */
   ignoreFallback?: boolean
+
+  /**
+   * @see https://github.com/chakra-ui/chakra-ui/issues/5581
+   *
+   *
+   * - beforeLoadOrError(default): loads the fallbackImage while loading the src
+   * - onError: loads the fallbackImage only if there is an error fetching the src
+   *
+   */
+  fallbackStrategy?: FallbackStrategy
 }
 
 export interface ImageProps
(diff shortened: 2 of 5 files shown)


TimKolberger:
Interesting issue! Thanks for reaching out - I hope my braindump will help 😅 
I think the proposed fallback strategies are fundamentally different:

## beforeLoadOrError

The strategy `beforeLoadOrError` is our current solution on the main branch.

The `useImage` hook loads the target image while the `Image` component is rendering the fallback image or element as DOM element. This takes place simultaneously (in parallel).

1. `<Fallback />` while loading `` in `useImage`
1. ✅  
1. ``

If an error occurs while loading the target image it keeps the fallback in place.

1. `<Fallback />` while loading `` in `useImage`
1. 💥 
1. `<Fallback />`  


## onError

The strategy `onError` has serial approach.
It renders the target img DOM element from the start...

1. ``
1. ✅ 
1. ``

... and listens for an error, to replace the image with the fallback image or component.

1. `<chakra.img className="chakra-image" onError={() => setStatus("failed")} />`
1. 💥 
1. `<Fallback />`

This strategy does not need the `useImage` hook, so we should opt out by passing `ignoreFallback: true` for this branch and register the error handler on the actual img DOM element.
We might need to introduce an additional state to keep track of the error state only in this fallback strategy.

```tsx
const [hasStrategyOnErrorError, { on: onStrategyOnErrorError }] = useBoolean(false)

if (status !== "loaded" || hasStrategyOnErrorError) {
  // ... render fallback
}

return (
  <chakra.img
    as={NativeImage}
    src={src}
    srcSet={srcSet}
    crossOrigin={crossOrigin}
    loading={loading}
    className="chakra-image"
    {...shared}
    onError={callAllHandlers(props.onError, () => onStrategyOnErrorError())}
  />
)
```

Does this sound right? 🤔

takethefake:
Hej @TimKolberger thanks for taking your time on this!

Yes sounds valid. will adjust this PR tomorrow! :)

takethefake:
hej @TimKolberger !

I found a way to incorporate both strategies together!

```
  const providedFallback = fallbackSrc !== undefined || fallback !== undefined
  /**
   * Defer to native `img` tag if `loading` prop is passed
   * @see https://github.com/chakra-ui/chakra-ui/issues/1027
   *
   * shouldIgnoreFallbackImage determines if we have the possibility to render a fallback image
   */
  const shouldIgnoreFallbackImage =
    loading != null ||
    // use can opt out of fallback image
    ignoreFallback ||
    // if the user doesn't provide any kind of fallback we should ignore it
    !providedFallback
    ```

This describes if we have a possibility to render a `fallbackImage` at all. if this is `true` the status from `useImage` will be `loaded` and we just render the "main-image".

If it is `false` the `status` from `useImage` representates the loading status of our "main-image"

we can use this `status` to apply our strategies

```
  /**
   * shouldShowFallbackImage
   *
   * `shouldIgnoreFallbackImage===false` implies that we have the data to display a fallback image
   *
   * also, it implies that `status` from `useImage` can have another value than `loaded`
   *
   * therefor we can apply different strategies when we want to render the fallback image based on the status
   *
   * `shouldShowFallbackImage` brings the strategies together with `status`
   */
  const shouldShowFallbackImage =
    (status !== "loaded" && fallbackStrategy === "beforeLoadOrError") ||
    (status === "failed" && fallbackStrategy === "onError")
    ```

if `shouldShowFallbackImage` is true we render `fallback` or `fallbackSrc` instead of the "main-image" 


Feeling quite confident that this works, but would be more confident with tests which I will write now

takethefake:
Added a test to verify the strategies

lcswillems:
Hi @takethefake ,

Great PR!

For the naming of the strategies, I think the following are clearer:

- "onError" and "beforeLoadedOrOnError"
- or even: "error" and "pendingLoadingError"
- or even: ["error"], ["pending", "loading", "error"]

In my case, I would also like to have a fallback for "pending" / "loading", and another fallback for "error". It might not be in the scope of this PR but it is related so might be worth taking into consideration in the design.

TimKolberger (approved):

TimKolberger:
Great work @takethefake! Kindly update the docs accordingly.

PR #5684, opened 2022-03-03
fix(editable): add prevValue as controllable prop

### LINKED ISSUE #5670: Editable: setting controlled value does not propagate into Editable's internal value
### Description

Using React useState to control the value of Editable, calling setState does not seem to change the internal value of Editable.

### Link to Reproduction

https://codesandbox.io/s/naughty-glade-ufqr78?file=/pages/index.tsx

### Steps to reproduce

1. The name state is set to "John" in useEffect once initially
2. Click on the Editable text: "John"
3. You can change, or not to change the Editable text.
4. Press Escape key to abort.
5. Editable shows placeholder text and the name state becomes the default empty string specified in useState("")

### Chakra UI Version

1.8.6

### Browser

Google Chrome 99

### Operating System

- [ ] macOS
- [ ] Windows
- [X] Linux

### Additional Information

_No response_

takethefake:
Closes #5670 

## 📝 Description

make `prevValue` controllable so it can be updated, when the value outside of editable changes

No


TimKolberger:
The value of Editable does not change when I try to type something in this story: https://chakra-ui-storybook-6w7yss5li-chakra-ui.vercel.app/?path=/story/components-forms-editable--editable-event-handler

I think this is the case because the onChange handler does not update the state in the story.

takethefake:
yep, my bad - fixed it accordingly! :)

PR #5696, opened 2022-03-07
feat(modal): introduce useImperativeModal

takethefake:
## 📝 Description

Added `useImperativeModal` to display a Modal and await its result inside a function call.

this makes the following possible:

```tsx
const Component: React.FC = ()=>{
  const [modalNode, openModal] = useImperativeModal()

  const onClick = async () => {
    // render a modal inside a function and get the result back from it
    const modalResult = await openModal<boolean>(({ onClose }) => (
      <ExampleModal
        isOpen
        onClose={() => onClose(false)}
        onCancel={() => onClose(false)}
        onConfirm={() => onClose(true)}
      />
    ))
    if (modalResult) {
      alert("user confirmed alert")
    } else {
      alert("user canceled alert")
    }
  }
  return (
    <>
      {modalNode}
      <Button onClick={onClick}>Click me to open a modal imperatively</Button>
    </>
  )
}
```

take a look at this CSB to checkout the feature: https://codesandbox.io/s/useimperativemodal-oxgwym?file=/src/App.tsx

### DIFF
--- a/.changeset/beige-bikes-wait.md
+++ b/.changeset/beige-bikes-wait.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/modal": patch
+---
+
+Add a possibility to open a modal imperatively inside a function with
+`useImperativeModal`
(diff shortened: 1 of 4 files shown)


segunadebayo:
Thinking about it deeply. I don't think we should add imperative APIs like this to the core of Chakra. It's probably not a good idea even though I'm the one that came up with it.

Please make this a recipe in the docs repo so we can share with others, and possibly add it to the docs.

Thanks for working on this, Daniel.

PR #5701, opened 2022-03-08
feat(cli): add strict-token-types option to CLI

### LINKED ISSUE #5520: Strict type tokens for other theme variables
### Description

https://github.com/chakra-ui/chakra-ui/pull/5243 added support for generating strict types for variants and sizes

I'd like to be able to extend that to other properties (e.g. `colors`, `space` etc.)

We'd like to ensure that only known colours/sizes etc are used in the app to ensure consistency.

For example, `colors` supports `| (string & {})`, so any colour can be provided. This includes HTML named colours, which make it hard to tell from looking at some code whether it is using known theme values or just arbitrary values.
```tsx
<Text color="red">Hello world</Text>
```
This would display as red, but if I don't have a red colour defined in my theme it is doing it based on the HTML colour, which undermines using the theme for consistency.

### Proposed Solution or API

A new flag for Chakra CLI, maybe `--strict-theme-tokens`.

When provided, the unions in the output wouldn't include `| (string & {})`.

At runtime, custom strings would still be allowed, but the typechecker would enforce correctness.

### Alternatives

_No response_

### Additional Information

_No response_

### LINKED ISSUE #6843: Props are not type-safe
### Description

When I wrote an invalid property such as `bg="thiscolordoesnotexist"` I expected TSC to throw an error, but nothing happened instead

### Link to Reproduction

https://stackblitz.com/edit/vitejs-vite-wzx6kp?file=src%2FApp.tsx&terminal=dev

### Steps to reproduce

1. Go to https://stackblitz.com/edit/vitejs-vite-wzx6kp?file=src%2FApp.tsx&terminal=dev
2. Open `App.tsx`
3. See that TypeScript doesn't complain about invalid props

### Chakra UI Version

^2.3.6

### Browser

Version 106.0.5249.119 (Official Build) (64-bit)

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

_No response_

lexanth:
Allows generating tokens without `| (string & {})` in the union, to allow stricter typechecking.

Closes #5520
Closes #6843

## 📝 Description
Based on https://github.com/chakra-ui/chakra-ui/pull/5243, which added support for strict component props (variants and sizes).

This does the same for other theme tokens (colours, spacing etc.). It also tweaks `Token` to exclude plain string from the union.

```tsx
<Box padding="abc" />
// valid type, but "abc" is not available in the theme
```

## 🚀 New behavior
```bash
chakra-cli tokens --strict-token-types
```
will generate types like:
```ts
  blur: "none" | "sm" | "base" | "md" | "lg" | "xl" | "2xl" | "3xl"
  borders: "none" | "1px" | "2px" | "4px" | "8px"
  borderStyles: never
  borderWidths: never
  breakpoints: "base" | "sm" | "md" | "lg" | "xl" | "2xl"
  // ...
}
```

```tsx
<Box padding="abc" /> // invalid
// Type '"abc"' is not assignable to type '"1" | "2" | ... | undefined'.
```

No
* The CLI change is an extra option
* The `Token` change is not breaking - if users don't use the new CLI option, the theme types will still have `| (string & {})` anyway, so the `Union<>` in `Token` is redundant.

## 📝 Additional Information
`Token` still includes `CSSType` even if we do have token fields for it.

For example:
```
padding?: Token<CSS.Property.Padding | number, "space">
```
would still allow any number to be passed as a padding value

Additionally, `CSS.Property.Padding` and `CSS.Color` both include `| (string & {})` anyway, which diminishes the value of this change.

We could change `Token` to 
```
export type Token<
  CSSType,
  ThemeKey = unknown,
> = ThemeKey extends keyof ThemeTypings
  ? ResponsiveValue<ThemeTypings[ThemeKey]>
  : ResponsiveValue<CSSType>
```
which would be the strictness that we're looking for. However, this may be seen as a breaking change (not sure what Chakra's policy on type changes as breaking changes is - I know some projects don't deem breaking type changes to be breaking changes).


lexanth:
@TimKolberger I couldn't reproduce the test failure locally, so I rebased this - could you trigger the unit tests again?

packages/stat/tests/stat.test.tsx
```diff
@@ -32,7 +32,7 @@ describe("<StatGroup />", () => {
         </Stat>
       </StatGroup>,
     )
-    expect(findByTestId("group")).toBeTruthy()
```
lexanth, line 35: `findByTestId` is async - this seemed to be leaving async handlers open, which caused stuff to fail after the test had already passed (and the assertion would always pass regardless of the result, because the promise is always truthy).

For some reason my PR seems to have tweaked the order/performance of the tests slightly such that it would trigger reasonably reliably in CI (but only very intermittently locally)

TimKolberger (approved):
Thank you @lexanth for this perfect PR 💖 

As soon as the React 18 stuff is out of the way I am happy to bring this forward!

Awesome work!

lexanth:
@TimKolberger any chance that someone could have a further look at this? Shall I resolve the conflicts?

TimKolberger:
@lexanth yes, this is still a very valuable PR I'd love to merge. Kindly rebase it onto main.
I'd like to test the change of the type `Token` in a larger project, to see the implication of this.

luketanner:
Is there anything blocking this PR from moving forward? If so, is there anything I can help out with?

lexanth:
I've rebased this again - any chance we could get this merged?

PR #5724, opened 2022-03-12
feat(prose): Created Prose component

nikolovlazar:
Closes https://github.com/chakra-ui/chakra-ui/discussions/5566

## 📝 Description

Initial version of the `Prose` component. The `Prose` component adds ready-made styles for rendering remote HTML content. Inspired by Tailwind's typography plugin.

There's an entry in Storybook under the `Typography` category where you can check out the `Prose` styling.

No


segunadebayo:
Quick thoughts:

I believe the intention is to make Prose an opt-in feature. In that case, we can't add prose to the theme object, it'll make it quite bloated (it's already bloated as is). Let's export the theme for Prose within the package itself `import { proseTheme }  from "pkg"` and let the user add it to their theme object manually.

segunadebayo:
Quick thoughts:

I believe the intention is to make Prose an opt-in feature. In that case, we can't add prose to the theme object, it'll make it quite bloated (it's already bloated as is). Let's export the theme for Prose within the package itself `import { proseTheme }  from "pkg"` and let the user add it to their theme object manually.

packages/theme/src/components/prose.ts
```diff
@@ -0,0 +1,173 @@
+import { mode, SystemStyleFunction } from "@chakra-ui/theme-tools"
+
+import { getComponentBaseStyle } from ".."
+
+const baseStyle: SystemStyleFunction = (props) => {
+  const { theme } = props
+  const headingBase = getComponentBaseStyle(theme, "Heading")
+  const linkBase = getComponentBaseStyle(theme, "Link")
```
segunadebayo, line 8: I think Markdown headings and links could be styled differently from the components. Feel free to create their own standalone styles. This code assumes that the user hasn't deleted, omitted or changed the styles for Heading and/or Link.

packages/theme/src/components/prose.ts
```diff
@@ -0,0 +1,173 @@
...
+      lineHeight: 6,
+      my: 6,
+    },
+    a: {
+      ...linkBase,
+      fontWeight: 500,
+    },
+    hr: {
+      my: { base: 12, md: 14 },
+      borderColor: mode("gray.200", "gray.600")(props),
```
segunadebayo, line 50: Prefer to use `_dark` instead of `mode(...)` function

packages/theme/src/components/prose.ts
```diff
@@ -0,0 +1,173 @@
...
+    "h3 + *": {
+      mt: 0,
+    },
+    "h4 + *": {
+      mt: 0,
+    },
+    "hr + *": {
+      mt: 0,
+    },
+  }
```
segunadebayo, line 168: Can we inline all the selectors into a single one?

nikolovlazar:
Thanks for you input @segunadebayo. Here's what I did:
- Moved the theme file from the `theme` package into the `prose` package
- Built a theme extension `withProse` that injects the Prose theme into the theme
- Refactored the styling to use `_dark` instead of `mode`
- Edited `.storybook/preview.tsx` to use the `withProse` theme extension
   - I thought to implement a Context that would allow us to modify the theme from within each story, but since Prose's story is the only story that needs that I decided to just use the `withProse` in the main file.
   
This PR is ready to be reviewed again.

nikolovlazar:
Closing this because we decided to roll out the Prose component as a separate package: https://github.com/nikolovlazar/chakra-ui-prose. We decided to roll it out as a separate package is to avoid bloating the core library while we figure out how much people actually need it. I've created a discussion thread to help us understand if people need it, and if it should be moved to the core library: https://github.com/chakra-ui/chakra-ui-docs/discussions/469.

Issue #5380, opened 2022-01-13
[Chakra Storybook Addon] Color mode switch breaks visual testing tools 


guilleironhack:
### Description

When using visual testing tools that take screenshots of stories, the color mode switch appears in all of them, triggering a false positives.

### Link to Reproduction

N/A

### Chakra UI Version

@chakra-ui/storybook-addon@1.0.0

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Some example:
![image](https://user-images.githubusercontent.com/44772724/149331083-1ed0e996-f2d8-40a0-9006-ed8601658558.png)

Ideally, the color mode selector should at the toolbar, or at least would be great to have the option to disable the color mode switcher.

ngxCoder:
If nobody else solves this, I will try to add an option to remove the color mode switcher this week.
I thought about this feature when I released it, but I am having some problems with my bills, haha

PR #5732, opened 2022-03-14
feat: add configurable parameters to the ColorModeSwitch of the storybook-addon package

### LINKED ISSUE #5544: Storybook Addon: configuration to remove the color mode switcher
### Description

I recently began migrating my CSS to Chakra from CSS modules and already have Storybook stories publishing visual regression tests to Chromatic. I would like to be able to disable the color mode switch button injected to my stories by the Chakra add-on because it causes a visual diff in my existing visual regression tests.

As someone who has existing visual regression tests created with a different CSS framework, it's important to me that my migration to Chakra UI results in no visual diffs to my existing UI. In an ideal world, I could add Chakra, re-run my visual regression tests, and have them all pass because I have refactored my styles without changing my UI.

The automatically-injected color mode switch button causes visual diffs, creating roadblocks as I try to use the Chakra storybook add-on.

### Proposed Solution or API

If the color mode switcher could be disabled via configuration parameters, I could use the storybook add-on as-is without impediment.

_preview.js_
```
export const parameters = {
  chakra: {
    theme,
    enableColorModeControl: false, // default could be true
  },
}
```

### Alternatives

I'm currently working around this by importing the storybook add-on's `webpackFinal` function. This allows me to get the benefit of the add-on's web pack handling without without the color mode button.

_main.js_
```
const {
  webpackFinal,
} = require('@chakra-ui/storybook-addon/preset/dist/chakra-ui-storybook-addon-preset.cjs.dev')

module.exports = {
  // [...]
  webpackFinal: async (config) => {
    return webpackFinal(config)
  },
}

```

### Additional Information

_No response_

ngxCoder:
Closes #5544 #5380 

## 📝 Description

This PR enhances the configuration of the [Storybook Addon](https://storybook.js.org/docs/react/addons/writing-addons/) for Chakra-UI. This will allow users to configure the Color Mode Switch with following parameters:

| Parameter | Type             | Default       | Allowed values                                           | Description                                                                                                                                            |
| --------- | ---------------- | ------------- | -------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
| enable    | boolean          | `true`        | `true\|false`                                            | Show or hide                                                                                                                                           |
| position  | string           | `'top-right'` | `'top-left'\|'top-right'\|'bottom-left'\|'bottom-right'` | Change the position to a different corner                                                                                                              |
| zIndex    | number \| string | `9999`        | any valid value to Chakra `zIndex` Style System          | This property goes directly to the `zIndex` of the [Style Props](https://chakra-ui.com/docs/styled-system/features/style-props#position)               |
| margin    | number \| string | `1rem`        | any valid value to Chakra `position` Style System        | This property goes directly to `top\|right\|bottom\|left` of the [Style Props](https://chakra-ui.com/docs/styled-system/features/style-props#position) |

All the parameters are **optional**.

### Example Usage
```js
//.storybook/preview.js
export const parameters = {
  chakra: {
    ...
    colorModeSwitch: { //default values
      enable: true,
      position: 'top-right',
      margin: '1rem',
      zIndex: 9999
    }
  }
}
```

The Color Mode Switch can't be modified

## 🚀 New behavior

This PR will include:
- New parameters to configure the Color Mode Switch and their properties
- Docs explaining the new parameters, example usage and allowed values

No

### DIFF
--- a/.changeset/wicked-deers-hug.md
+++ b/.changeset/wicked-deers-hug.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/storybook-addon": minor
+---
+
+Add enable, position, zIndex and margin as parameters to modify the Color Mode
+Switch properties
--- a/tooling/storybook-addon/README.md
+++ b/tooling/storybook-addon/README.md
@@ -19,6 +19,8 @@ yarn add -D @chakra-ui/storybook-addon
 npm i -D @chakra-ui/storybook-addon
 ```
 
+## Usage
+
 Add the addon to your configuration in `.storybook/main.js`:
 
 ```js
@@ -27,7 +29,7 @@ module.exports = {
 }
 ```
 
-## Configuring `ChakraProvider`
+### Configuring `ChakraProvider`
 
 If you need to customize the props passed to `ChakraProvider` for your stories
 (to use a custom `theme`, for example), you'll need to create custom Storybook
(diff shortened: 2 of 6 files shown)


ngxCoder:
ok ok, I will think about how to place it in the toolbar

PR #5739, opened 2022-03-15
fix: memoize switch object in useBoolean

ambar:
This allows us to use it in hook deps:

```js
const [flag, setFlag] = useBoolean();

useEffect(() => {}, [setFlag]);
```

### DIFF
--- a/packages/hooks/src/use-boolean.ts
+++ b/packages/hooks/src/use-boolean.ts
@@ -1,4 +1,4 @@
-import { useState } from "react"
+import { useMemo, useState } from "react"
 
 type InitialState = boolean | (() => boolean)
 
@@ -9,11 +9,13 @@ type InitialState = boolean | (() => boolean)
  */
 export function useBoolean(initialState: InitialState = false) {
   const [value, setValue] = useState(initialState)
-  const [switches] = useState(() => ({
-    on: () => setValue(true),
-    off: () => setValue(false),
-    toggle: () => setValue(prev => !prev),
-  }))
-
-  return [value, switches] as const
+  const callbacks = useMemo(
+    () => ({
+      on: () => setValue(true),
+      off: () => setValue(false),
+      toggle: () => setValue((prev) => !prev),
+    }),
+    [],
+  )
+  return [value, callbacks] as const
 }
--- a/packages/hooks/src/use-boolean.ts
+++ b/packages/hooks/src/use-boolean.ts
@@ -1,4 +1,4 @@
-import { useCallback, useState } from "react"
+import { useState } from "react"
 
 type InitialState = boolean | (() => boolean)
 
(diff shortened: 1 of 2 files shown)


packages/hooks/src/use-boolean.ts
```diff
@@ -22,5 +22,5 @@ export function useBoolean(initialState: InitialState = false) {
     setValue((prev) => !prev)
   }, [])
 
-  return [value, { on, off, toggle }] as const
+  return [value, useRef({ on, off, toggle }).current] as const
```
TimKolberger, line 25: Kindly replace this useRef with a useMemo call, e.g:

```suggestion
  const helpers = useMemo(() => ({ 
    on: ( ) => setValue(true), 
    off: ( ) => setValue(false),
    toggle: () => setValue(prev => !prev)  
  }), [on, off, toggle])
  return [value, helpers] as const
```

TimKolberger (changes requested):
Thanks for creating this PR 💖 
I left a thought as comment

packages/hooks/src/use-boolean.ts
ambar: https://reactjs.org/docs/hooks-reference.html#usememo
> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render

ambar: Replaced with `useState`

TimKolberger: Why do we need a guarantee that the helpers object won't ever change their identity?

I think this change is a performance optimization - the hook will work as intended even without any useMemo, useRef or useState usage.

ambar: To avoid running effects multiple times, and avoid using `exhaustive-deps` from eslint plugin:

```js
const [flag, setFlag] = useBoolean()

useEffect(() => {
  if (foo && flag) {
    // dot sth
  }
}, [foo, flag, setFlag])
```

TimKolberger: Instead of depending on a stable wrapper object we can extract the needed function and pass them in the dependency array:

```ts
const [flag, { off }] = useBoolean();

React.useEffect(() => {
  if (foo && flag) {
    off();
  }
}, [foo, flag, off]);
```

This way we wouldn't even need to memoize the helper object.
I don't think it is a good idea enforce a guaranteed stable reference for the wrapper object.

---

If your application logic has a dependency on this effect, consider moving this code into an event handler or state machine.

The workaround I currently often use is `useLatestRef` - it's available in the @chakra-ui/hooks package or as [thirdparty package here](https://www.npmjs.com/package/use-latest-ref).

```ts
const [flag, helpers] = useBoolean();
const latestOnFnRef = useLatestRef(helpers.on);

React.useEffect(() => {
  if (foo && flag) {
    latestOnFnRef.current();
  }
}, [foo, flag, latestOnFnRef]);
```

packages/hooks/src/use-boolean.ts
```diff
@@ -9,18 +9,11 @@ type InitialState = boolean | (() => boolean)
  */
 export function useBoolean(initialState: InitialState = false) {
   const [value, setValue] = useState(initialState)
+  const [switches] = useState(() => ({
+    on: () => setValue(true),
+    off: () => setValue(false),
+    toggle: () => setValue(prev => !prev),
+  }))
```
burhanuday, line 16: question - what value does this provide over using useCallback

segunadebayo: Sorry, I've updated the hook to a slightly better version that uses `useMemo`.

Don't see much value in a double `useState`

PR #5754, opened 2022-03-21
fix(ButtonGroup): collapse borders with isAttached enabled

### LINKED ISSUE #4531: [ButtonGroup] Use negative spacing to hide extra border
### Description

Use ButtonGroup's `spacing` instead of Button's `mr` to set the negative value once.

Currently, you will have to repeat the code on every button (except for the last one) and on every button group. Moving this logic to the parent element would seem to provide easier customisation and lesser code.

### Proposed Solution or API

So instead of doing this:
```
<ButtonGroup isAttached>
    <Button mr="-px">Day</Button>
    <Button mr="-px">Week</Button>
    <Button>Month</Button>
  </ButtonGroup>
```

do this:
```
<ButtonGroup isAttached spacing="-px">
    <Button>Day</Button>
    <Button>Week</Button>
    <Button>Month</Button>
  </ButtonGroup>
```

### Alternatives

Or why not just remove the extra borders automatically when `isAttached` is passed?

### Additional Information

_No response_

novascreen:
Closes #4531

## 📝 Description

Collapse borders in outlined buttons inside `ButtonGroup` with `isAttached`.

Reinstates the behavior of https://github.com/chakra-ui/chakra-ui/pull/230

Currently the `ButtonGroup` requires manually adding `mr="-px"` to each outlined button when using `isAttached` to get rid of the double border.

## 🚀 New behavior

`ButtonGroup` applies the negative margin automatically when using `isAttached` with `variant="outline"`

## 📝 Additional Information

Before:
image

After:
image

### DIFF
--- a/packages/button/src/button-group.tsx
+++ b/packages/button/src/button-group.tsx
@@ -73,6 +73,9 @@ export const ButtonGroup = forwardRef<ButtonGroupProps, "div">((props, ref) => {
       "> *:not(:first-of-type):not(:last-of-type)": { borderRadius: 0 },
       "> *:not(:first-of-type):last-of-type": { borderStartRadius: 0 },
     }
+    if (variant === "outline") {
+      groupStyles["> *:not(:last-of-type)"] = { marginEnd: "-1px" }
+    }
   } else {
     groupStyles = {
       ...groupStyles,
--- a/packages/button/src/button-group.tsx
+++ b/packages/button/src/button-group.tsx
@@ -87,7 +87,7 @@ export const ButtonGroup = forwardRef<ButtonGroupProps, "div">((props, ref) => {
         role="group"
         __css={groupStyles}
         className={_className}
-        data-attached={isAttached ? true : undefined}
+        data-attached={isAttached ? "" : undefined}
         {...rest}
       />
     </ButtonGroupProvider>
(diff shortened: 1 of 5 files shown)


packages/button/src/button-group.tsx
```diff
@@ -73,6 +73,9 @@ export const ButtonGroup = forwardRef<ButtonGroupProps, "div">((props, ref) => {
       "> *:not(:first-of-type):not(:last-of-type)": { borderRadius: 0 },
       "> *:not(:first-of-type):last-of-type": { borderStartRadius: 0 },
     }
+    if (variant === "outline") {
+      groupStyles["> *:not(:last-of-type)"] = { marginEnd: "-1px" }
+    }
```
segunadebayo, line 78: The issue I see with this is that we're assuming that all Chakra UI users have a variant named "outline" in their theme object.

The button component should have no knowledge of the variant named consumers to use in their design system or app. I think we should document this instead of creating a fix for it.

segunadebayo: Here's another way to resolve this:

- Add a `data-flushed`  attribute to the root `chakra.div` element when `isAttached` is passed to the `ButtonGroup`
- In the button theme, you can use this attribute to change the styles for the outline variant when it is within the context of that data attribute, something like.   `.chakra-button__group[data-flushed] & : {...}`

packages/button/src/button-group.tsx
```diff
@@ -87,6 +87,7 @@ export const ButtonGroup = forwardRef<ButtonGroupProps, "div">((props, ref) => {
         role="group"
         __css={groupStyles}
         className={_className}
+        data-attached={isAttached ? true : undefined}
```
segunadebayo, line 90:
```suggestion
        data-attached={isAttached ? "" : undefined}
```

PR #5756, opened 2022-03-21
feat(storybook-addon): Replace button on the corner with toggle tool 

### LINKED ISSUE #5544: Storybook Addon: configuration to remove the color mode switcher

ngxCoder:
Closes #5544 #5380 

## 📝 Description

This PR replaces the ColorModeSwitch with a Tool.
This tool allows the users to toggle the color mode of their stories (the same purpose).

The addon creates a button on the corner of every storybook and breaks some visual tests.

## 🚀 New behavior

- Remove the ColorModeSwitch
- Create a tool to toggle the color mode

No

## 📝 Additional Information
The tool "tries" to get the persisted state of the theme getting the local storage key.
If the local storage key is unavailable, click it again.

¿Anyone knows a better way to do this?


tooling/storybook-addon/src/components/Tool.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useCallback } from "react"
+import { useAddonState } from "@storybook/api"
+import { Icons, IconButton } from "@storybook/components"
+import { addons } from "@storybook/addons"
+import { ADDON_ID, TOOL_ID, EVENTS } from "../constants"
+
+export const Tool = () => {
+  const persistedColorMode =
+    localStorage.getItem("chakra-ui-color-mode") === "dark"
```
burhanuday, line 9:
```suggestion
  const isDarkMode =
    localStorage.getItem("chakra-ui-color-mode") === "dark"
```

burhanuday: I think its better to move local storage access to useEffect. Otherwise this key will be fetched from localstorage on every render of this component

tooling/storybook-addon/src/components/Tool.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const [darkMode, setDarkMode] = useAddonState(ADDON_ID, persistedColorMode)
+  const channel = addons.getChannel()
+
+  const toggleMyTool = useCallback(
+    () => {
+      channel.emit(EVENTS.TOGGLE_COLOR_MODE, !darkMode ? "dark" : "light")
+      setDarkMode(!darkMode)
```
burhanuday, line 16:
```suggestion
      setDarkMode(darkMode => !darkMode)
```

tooling/storybook-addon/src/components/Tool.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { addons } from "@storybook/addons"
+import { ADDON_ID, TOOL_ID, EVENTS } from "../constants"
+
+export const Tool = () => {
+  const persistedColorMode =
+    localStorage.getItem("chakra-ui-color-mode") === "dark"
+  const [darkMode, setDarkMode] = useAddonState(ADDON_ID, persistedColorMode)
+  const channel = addons.getChannel()
+
+  const toggleMyTool = useCallback(
```
burhanuday, line 13:
```suggestion
  const toggleDarkMode = useCallback(
```

tooling/storybook-addon/src/components/Tool.tsx
```diff
@@ -0,0 +1,33 @@
...
+    () => {
+      channel.emit(EVENTS.TOGGLE_COLOR_MODE, !darkMode ? "dark" : "light")
+      setDarkMode(!darkMode)
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
burhanuday, line 18: if `channel` variable is the one causing an issue, can you initialise channel inside this useCallback itself?

tooling/storybook-addon/src/components/Tool.tsx
ngxCoder: Yes, it could be a useEffect with an empty deps array.

tooling/storybook-addon/src/components/Tool.tsx
ngxCoder: What do you mean by an issue? 
The channel is not the problem, it works. 

The part that I want to improve here is how I retrieve the color mode's persisted state.
Imagine when a dev needs to change the storageManager to cookies or in a future release, Chakra UI decides to remove the localstorage key or change it.

ngxCoder:
Did I miss something? Please tell me if I need to include something

segunadebayo:
Thanks for all your work on this @ngxCoder.

I've gone ahead to refactor a few items as needed. Here's a screenshot of the toolbar now. In our storybook add-on, we need to have both a toggle for color mode and writing direction (rtl or ltr).

In the future, we can provide a way to hide some of them.

CleanShot 2022-04-22 at 14 17 55@2x

TimKolberger:
Seems like 3 persons worked on the same issue 😆 
@segunadebayo I think we need to consolidate your latest changes to this PR and #5922.

segunadebayo:
@TimKolberger. Yeah it seems so 🤣

No worries though. Let's use your PR as the source of truth for this work.

segunadebayo:
Closing in favor of #5922

PR #5772, opened 2022-03-26
feat(wrap): separate x and y margins

mix3d:
## 📝 Description

* Adds the option to specify vertical and horizontal spacing options for the `<Wrap>` component, should the user choose. 
* Falls back to the `spacing` prop for defaults or to set both h and v at the same time.
* Also fixes the code example of Wrap in the layout package.

The Wrap component adds the `spacing` prop to both horizontal and vertical margins, with no opportunity for the user to specify anything else.

## 🚀 New behavior

Users can optionally specify `hSpacing` and/or `vSpacing` to choose a specific margin for that axis.

Example:

```jsx
<Wrap vSpacing={0} hSpacing={['8px', '12px', '24px']}>
  <Box/>
  <Box/>
  <Box/>
  <Box/>
</Wrap>
```

No, intentionally made backwards compatible.

## 📝 Additional Information
Happy to rename if this prop naming is not preferred.
I recognize that there could be room for improvements on `spacing`-only implementations, but wasn't sure what the best way to conditionally use one or two css values in the styles memo without mostly duplicating the return object.

### DIFF
--- a/packages/layout/README.md
+++ b/packages/layout/README.md
@@ -61,9 +61,10 @@ Wrap is used to manage the distribution of child elements that are liable to
 wrap. It is mostly used for button groups, tag group, badge group, and chips.
 
 ```jsx
-<AspectRatio ratio={16 / 9}>
-  <img src="./some-ig-story" alt="Instagram story" />
-</AspectRatio>
+<Wrap spacing={3}>
+  <Box>Box 1</Box>
+  <Box>Box 2</Box>
+</Wrap>
 ```
 
 Badge is used to render a badge. It can comes in different variants and color
--- a/packages/layout/src/wrap.tsx
+++ b/packages/layout/src/wrap.tsx
@@ -76,9 +76,9 @@ export const Wrap = forwardRef<WrapProps, "div">((props, ref) => {
     }
     return {
       "--chakra-wrap-x-spacing": (theme: Dict) =>
-        mapResponsive(y, (value) => tokenToCSSVar("space", value)(theme)),
-      "--chakra-wrap-y-spacing": (theme: Dict) =>
         mapResponsive(x, (value) => tokenToCSSVar("space", value)(theme)),
+      "--chakra-wrap-y-spacing": (theme: Dict) =>
+        mapResponsive(y, (value) => tokenToCSSVar("space", value)(theme)),
       "--wrap-x-spacing": "calc(var(--chakra-wrap-x-spacing) / 2)",
       "--wrap-y-spacing": "calc(var(--chakra-wrap-y-spacing) / 2)",
       display: "flex",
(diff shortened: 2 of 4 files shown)


packages/layout/src/wrap.tsx
```diff
@@ -10,10 +10,20 @@ import * as React from "react"
...
+  /**
+   * The horizontal space between the each child (even if it wraps). Defaults to `spacing` if not defined.
+   * @type SystemProps["margin"]
+   */
+  hSpacing?: SystemProps["margin"]
+  /**
+   * The horizontal and vertical space between the each child (even if it wraps). Defaults to `spacing` if not defined.
+   * @type SystemProps["margin"]
+   */
+  vSpacing?: SystemProps["margin"]
```
segunadebayo, line 26: Let's change this to `spacingX` and `spacingY` to match the API in `SimpleGrid`

packages/layout/README.md
```diff
@@ -61,9 +61,10 @@ Wrap is used to manage the distribution of child elements that are liable to
 wrap. It is mostly used for button groups, tag group, badge group, and chips.
 
 ```jsx
-<AspectRatio ratio={16 / 9}>
-  <img src="./some-ig-story" alt="Instagram story" />
-</AspectRatio>
+<Wrap spacing={3}>
+  <Box>Box 1</Box>
+  <Box>Box 2</Box>
+</Wrap>
```
segunadebayo, line 67: Good catch!

segunadebayo (changes requested):

packages/layout/src/wrap.tsx
```diff
@@ -57,25 +69,32 @@ export const Wrap = forwardRef<WrapProps, "div">((props, ref) => {
...
+  const styles = React.useMemo(() => {
+    const { spacingX: x = spacing, spacingY: y = spacing } = {
+      spacingX,
+      spacingY,
+    }
+    return {
+      "--chakra-wrap-x-spacing": (theme: Dict) =>
+        mapResponsive(y, (value) => tokenToCSSVar("space", value)(theme)),
+      "--chakra-wrap-y-spacing": (theme: Dict) =>
+        mapResponsive(x, (value) => tokenToCSSVar("space", value)(theme)),
```
segunadebayo, line 81:
```suggestion
      "--chakra-wrap-x-spacing": (theme: Dict) =>
        mapResponsive(x, (value) => tokenToCSSVar("space", value)(theme)),
      "--chakra-wrap-y-spacing": (theme: Dict) =>
        mapResponsive(y, (value) => tokenToCSSVar("space", value)(theme)),
```

segunadebayo: It seems the x and y got mixed up here.

mix3d: sure did. Sorry!

segunadebayo (approved):
Thanks for working through all the changes here @mix3d.

Great work 💖.

PR #5814, opened 2022-04-04
feat(react-18): umbrella PR for React 18 support

TimKolberger:
## 📝 Description

React 18 has arrived 🎉 
Chakra UI will support React 18 with a major update.
This PR should be the target of all React 18 related changes.

All packages use React 17 as dependency.

## 🚀 New behavior

All packages use React 18 as dependency.

Yes


TimKolberger:
We are removing this PR in favor of the `next` branch

PR #5815, opened 2022-04-04
test(react-18-test-suite): update test suite for React 18

TimKolberger:
## 📝 Description

React 18 has arrived :tada:

The test suite uses React 17 and the corresponding versions of `@testing-library/react` etc.

## 🚀 New behavior

The test suite uses React 18 and latest versions of the testing libs.

No

## 📝 Additional Information

Depends on #5797


nicksrandall:
The `@types/react` package has a some breaking changes as well. Not sure if it is scope for this PR but I thought you should know incase you wanted to upgrade the types to v18 as well. 

Also, there is a codemod that can help with some of the breaking changes: https://github.com/eps1lon/types-react-codemod

matamatanot:
Error in `PopoverTrigger` regarding `children`.

```
error TS2559: Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes'.
660 <PopoverTrigger>.
```

PR #5841, opened 2022-04-08
docs: add story and test for nested menu

lexanth:
Closes # 

## 📝 Description
It's currently possible to create a nested menu, but it isn't documented.

Before adding an example in https://github.com/chakra-ui/chakra-ui-docs/blob/main/pages/docs/components/overlay/menu.mdx I thought it would be a good idea to add a story and test to make sure it doesn't get broken

It's currently possible to create a nested menu, but it isn't documented.

## 🚀 New behavior
Adds a story and unit test to make sure it keeps working

No

### DIFF
--- a/.changeset/gentle-rivers-bow.md
+++ b/.changeset/gentle-rivers-bow.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/menu": patch
+---
+
+Add story and tests for nested menu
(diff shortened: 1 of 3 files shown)


segunadebayo:
Hi @lexanth,

Thanks for working on this. There are a few issues with the nested menu example you shared:

- Keyboard navigation doesn't work for the nested menus.
- Pointer over interaction for the menu item doesn't work. It works when you click on it.

This isn't the ideal experience we want for the nested menu. Permit me to close this.

We plan to design proper support for the nested menu after v2 is released.

Thanks for understanding.

PR #5850, opened 2022-04-10
chore(toast): remove dependency to @reach-ui/alert

TimKolberger:
## 📝 Description

This PR removes the dependency to @reach-ui/alert, because it is not compatible with React 18. 
It uses `ReactDOM.render` to announce the toast message.
According to this reach-ui PR https://github.com/reach/reach-ui/pull/906 it may not be necessary to reattach the message for a11y reasons.

The toast package wraps the toast message in Reach UIs Alert component.
This leads to a ReactDOM error:
> Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot

## 🚀 New behavior

Replaced the Reach Alert with a div element with following properties:

```ts
<chakra.div 
  role="status"
  aria-live="polite" // do not interrupt current announcements
  aria-atomic="true" // announce whole alert
  />
```

The ReactDOM error is not showing anymore.
The alert is still announced by VoiceOver, on macOS. 

No

### DIFF
--- a/.changeset/forty-plums-attend.md
+++ b/.changeset/forty-plums-attend.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/toast": patch
+---
+
+Removed dependency to @reach-ui/alert
--- a/packages/toast/package.json
+++ b/packages/toast/package.json
@@ -8,9 +8,7 @@
     "ui mode",
     "ui"
   ],
-  "sideEffects": [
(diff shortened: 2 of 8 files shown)


nicksrandall:
Thank you! I was just about to submit a PR to fix this.

Issue #3198, opened 2021-01-28
Editable submitOnBlur submits on button click


vogler:
# 🐛 Bug report

Following the example from https://chakra-ui.com/docs/form/editable#with-custom-controls
I would expect it to behave the same with `submitOnBlur`, i.e., <kbd>esc</kbd> should undo any change instead of calling `onSubmit`. Also, the cancel button should call `onCancel` instead of `onSubmit`.

## 💥 Steps to reproduce

The example linked below first calls `onCancel` and afterwards `onSubmit` but does not revert changes.
If you click the cancel button, it only calls `onSubmit`. I understand that clicking the button does blur `EditableInput`, but maybe the blur should be scoped to `Editable` instead. How can I get the button to call `onCancel` otherwise?

## 💻 Link to reproduction

 [TypeScript CodeSandbox template](https://codesandbox.io/s/happy-hypatia-okuw3?file=/src/App.tsx)

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Chakra UI        |      1.1.5      |
| Browser          |     Chrome 87.0.4280.88       |
| Operating System |      macOS 11.1 (20C69)      |

with-heart:
Please fill out the whole report template when filing an issue. We ask for that information for a reason 😞  What browser and OS are you on?

vogler:
Sorry, added above. Your setup behaves differently?

with-heart:
Interesting! We have very similar setups but I'm only experiencing the problem with clicking the cancel button. `esc` is canceling as expected.

vogler:
Huh, didn't expect a browser issue. Just tried it in Safari 14.0.2, there `esc` works.

vogler:
Ok, if I disable the [Surfingkeys](https://github.com/brookhong/Surfingkeys) extension, `esc` works as expected.
The extension removes focus from the current input on `esc`. I assume it only passes through `esc` afterwards.
I'll open an issue there.

So, just the button issue remains.

segunadebayo:
Looking into this!

segunadebayo:
Pushed a fix for this! Should be available in the next release

vogler:
Thanks! Just had to change from `useEditableState` to `useEditableControls` in my `EditableControls`.

I noticed that `getCancelButtonProps` sets `aria-label` but it's not reflected in the type (`string | undefined` instead of `string`).
[...]

Issue #424, opened 2020-02-19
[Editable]: Call onCancel when submitOnBlur is false


Saifadin:
**Is your feature request related to a problem? Please describe.**
When the `Editable` Component is used with Custom Handling and the field is blurred, the `isEditing` prop never changes to false. I am currently using it like this:

```
<Editable defaultValue={email} mb="2" fontFamily="heading" position="relative" submitOnBlur={false}>
  {({ isEditing, onSubmit, onCancel }) => {
    return (
      <>
        <EditablePreview />
        <EditableInput />
        {isEditing ? (
          <Box position="absolute" right="0" bottom="-32px">
            <ButtonGroup justifyContent="center" size="xs">
              <IconButton icon="check" aria-label="Check" onClick={onSubmit} />
              <IconButton icon="close" aria-label="cancel" onClick={onCancel} />
            </ButtonGroup>
          </Box>
        ) : null}
      </>
    );
  }}
</Editable>
```

**Describe the solution you'd like**
The solution I would suggest is to return to the initialValue on Blur when `submitOnBlur={false}` is set. Basically call `onCancel` instead of `onSubmit` here: 

https://github.com/chakra-ui/chakra-ui/blob/06b65dd4117d8c1207d3ed80f4932d8c2b0d7dee/packages/chakra-ui/src/Editable/index.js#L218

This could be changed to `submitOnBlur ? onSubmit() : onCancel();`

**Additional context**
I can create an MR if this behavior is desired. I think this would be a better UX in this specific use case.

segunadebayo:
Hi @Saifadin,

Thanks for this suggestion. It sounds like a good idea. I'll look into it for the next release.

Saifadin:
@segunadebayo I will create a MR and you can look into it :)

stefanbugge:
It seems to me that this newly added `onCancel()` call in the `onBlur` handler basically blocks the `onSubmit` renderProp from ever being called when using custom controls. I am guessing the `onBlur` is triggered immediately and the custom "check" button never receives the mouse event before it has been unmounted (due to the `onCancel` call).
As a workaround I can add a custom noop `onBlur` handler to `<EditableInput />` to bypass this.

Staz:
Hey, I'm experiencing this same issue on v1.6.4. Can we please get this "Call onCancel when submitOnBlur is false
" behaviour back again?

PR #5859, opened 2022-04-12
fix(editable): cancel on blur

s97712:
Related #424 #3198

Call onCancel when submitOnBlur is false
## 📝 Description

> Add a brief description

## 🚀 New behavior

### DIFF
--- a/packages/editable/src/use-editable.ts
+++ b/packages/editable/src/use-editable.ts
@@ -235,7 +235,11 @@ export function useEditable(props: UseEditableProps = {}) {
       const isValidBlur = !targetIsCancel && !targetIsSubmit
 
       if (isValidBlur) {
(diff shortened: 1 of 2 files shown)


packages/editable/src/use-editable.ts
```diff
@@ -234,8 +234,8 @@ export function useEditable(props: UseEditableProps = {}) {
       const targetIsSubmit = contains(submitButtonRef.current, relatedTarget)
       const isValidBlur = !targetIsCancel && !targetIsSubmit
 
-      if (isValidBlur && submitOnBlur) {
-        onSubmit()
+      if (isValidBlur) {
+        submitOnBlur ? onSubmit(): onCancel()
```
segunadebayo, line 238:
```suggestion
        if (submitOnBlur) onSubmit()
        else onCancel()
```

PR #5864, opened 2022-04-13
added `selected hidden disabled` attrs

### LINKED ISSUE #5863: Make `placeholder` of `Select` unselectable
### Description

When I fill the `placeholder` prop of the `Select` I get extra unwanted element.
image

Better is to have it only as a hint  not option, and when the user clicks it fades:
image
image

According to MDN, a placeholder is a "string that provides a brief hint to the user". However, when used with select it provides an extra unwanted `option` element. It must not be included in the user's options. 

### Proposed Solution or API

Just add `disabled hidden` attribute to the placeholder option tag. 

### Alternatives

_No response_

### Additional Information

_No response_

ahmedelq:
Closes #5863

According to MDN, a placeholder is a "string that provides a brief hint to the user". However, when used with select it provides an extra unwanted `option` element. It must not be included in the user's options. 

`placeholder` places an extra `option` element
image

## 🚀 New behavior

Placeholder provides only a hint and doesn't add an extra element:
image
image

No

### DIFF
--- a/packages/select/src/select.tsx
+++ b/packages/select/src/select.tsx
@@ -30,7 +30,7 @@ export const SelectField = forwardRef<SelectFieldProps, "select">(
         ref={ref}
         className={cx("chakra-select", className)}
       >
-        {placeholder && <option value="">{placeholder}</option>}
+        {placeholder && <option value="" selected hidden disabled>{placeholder}</option>}
         {children}
       </chakra.select>
     )


segunadebayo:
Based on my comment here: https://github.com/chakra-ui/chakra-ui/issues/5863#issuecomment-1118485412

PR #5872, opened 2022-04-15
tooling: use `swc` for `ts-node`

bwlt:
## 📝 Description

I have updated the ts-node dependency in order to use a version which supports the [swc: true option](https://typestrong.org/ts-node/docs/swc/).
This improvement will speed up the ts-node, in particular it affects the `@chakra-ui/cli` package: in a personal project I was able to reduce the `chakra-cli tokens <path/to/your/theme.(js|ts)>` script execution time from~15s to ~3s.
In my PR I have also run a `yarn-deduplicate` script (see [yarn-deduplicate](https://www.npmjs.com/package/yarn-deduplicate)) in order to clean up the lockfile.

slow `chakra-cli tokens <path/to/your/theme.(js|ts)>`

## 🚀 New behavior

faster `chakra-cli tokens <path/to/your/theme.(js|ts)>` ⚡ 

No

## 📝 Additional Information

-

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,7 +158,7 @@
     "rimraf": "3.0.2",
     "shelljs": "0.8.5",
     "storybook-addon-performance": "0.16.1",
-    "ts-node": "10.4.0",
+    "ts-node": "^10.7.0",
     "typescript": "^4.6.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "@storybook/react": "~6.5.0-alpha.64",
     "@storybook/theming": "~6.5.0-alpha.64",
     "@swc-node/jest": "^1.3.5",
+    "@swc/core": "^1.2.174",
     "@types/edit-json-file": "^1.6.1",
     "@types/jest": "^27.4.1",
     "@types/mkdirp": "1.0.2",
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "@storybook/react": "~6.5.0-alpha.64",
     "@storybook/theming": "~6.5.0-alpha.64",
     "@swc-node/jest": "^1.3.5",
-    "@swc/core": "^1.2.174",
+    "@swc/core": "^1.2.177",
     "@types/edit-json-file": "^1.6.1",
     "@types/jest": "^27.4.1",
     "@types/mkdirp": "1.0.2",
--- a/tooling/cli/bin/tsconfig.json
+++ b/tooling/cli/bin/tsconfig.json
@@ -8,6 +8,7 @@
     "esModuleInterop": true
   },
   "ts-node": {
+    "swc": true,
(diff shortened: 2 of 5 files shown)


morellodev (changes requested):
Please add the `@swc/core` dependency

TimKolberger:
Awesome idea to use swc in the Chakra CLI! 🚀 ⚡ 
I think it should be possible to eliminate the dependency to ts-node by using [@swc-node/register](https://github.com/swc-project/swc-node/tree/master/packages/register) directly. This way we would replace a dependency in the CLI package instead of adding one on top. 
What do you think?

bwlt:
It's a good idea to replace `ts-node` with `@swc-node/register` but it is not so easy to perform on the entire codebase:
- in the `package.json` scripts section I can easily replace the `ts-node` occurences with `node -r @swc-node/register`. For example ["csb:install"](https://github.com/bwlt/chakra-ui/blob/67340f98c0247ea06adf72d8387d5ae8925580ad/package.json#L38) `"csb:install": "yarn && node -r @swc-node/register scripts/csb.ts"`
- those scripts runs with different compiler options: for example the `compilerOptions.module` [is set to "CommonJS"](https://github.com/bwlt/chakra-ui/blob/67340f98c0247ea06adf72d8387d5ae8925580ad/tsconfig.json#L30) instead of ["esnext"](https://github.com/bwlt/chakra-ui/blob/67340f98c0247ea06adf72d8387d5ae8925580ad/tsconfig.json#L4). As far as I know the @swc-node/register does not read from the `ts-node` key of the tsconfig.json file, so we have to create another file like `tsconfig.node.json` and reference it with [SWC_NODE_PROJECT or TS_NODE_PROJECT env](https://github.com/swc-project/swc-node/tree/master/packages/register#read-tsconfigjson)
- the `tooling/cli/bin/tsconfig.json` is not affected by the above issues, so it's easy to replace with `@swc-node/register` here
- the ts-node invocation in `tooling/props-docs/package.json` [build script](https://github.com/bwlt/chakra-ui/blob/67340f98c0247ea06adf72d8387d5ae8925580ad/tooling/props-docs/package.json#L31) is ambiguous with the `-T` command line flag. I think it could be easily omitted.

What do you think of the above considerations?

TimKolberger:
Thank you for the awesome task breakdown @bwlt ! 

This PR is simply too good not to get merged asap. So little change, such impact! ✨ 
Kindly resolve the merge conflicts, then this is good to go IMO.
 
[You mentioned a few perfect PR topics](https://github.com/chakra-ui/chakra-ui/pull/5872#issuecomment-1114903647), feel free to followup with them!

bwlt:
Thank you for the review! the conflicts are now solved. I will eventualy open other PRs for replacing ts-node

TimKolberger (approved):

PR #5946, opened 2022-04-29
refactor: color mode and tests

segunadebayo:
## 📝 Description

As we prepare for v2 release, I think it's important to ensure we get color mode right this time or at least improve it. The goals of this refactoring are:

- Ensure color mode evaluation is deterministic
- Make cookie and local storage managers uniform (reduce size as well)
- Change color mode instantly, disabling element CSS transitions when switching
- Simplify tests to match user interactions, not implementation details


packages/color-mode/src/color-mode-provider.tsx
```diff
@@ -57,109 +53,73 @@ export function ColorModeProvider(props: ColorModeProviderProps) {
...
-      if (initialColorMode === "system") {
-        return rawSetColorMode(systemColorWithFallback)
-      }
-      return rawSetColorMode(defaultColorMode)
+    if (!isBrowser) return
+    const utils = getColorModeUtils({ doc: document })
+
+    const managerValue = colorModeManager.get()
+    if (managerValue) {
+      return rawSetColorMode(managerValue)
```
TimKolberger, line 74: Because `useEffect` treats a returned function as cleanup fn, this line gave me the impression `rawSetColorMode` returned a function, but is simply a setState call.

```suggestion
     rawSetColorMode(managerValue)
     return
```

packages/color-mode/src/color-mode-provider.tsx
```diff
@@ -57,109 +53,73 @@ export function ColorModeProvider(props: ColorModeProviderProps) {
...
-  }, [colorMode, document])
+    const systemValue = utils.getColorScheme(defaultColorMode)
 
-  const setColorMode = React.useCallback(
-    (value: ColorMode, isListenerEvent = false) => {
-      if (!isListenerEvent) {
-        colorModeManager.set(value)
-      } else if (colorModeManager.get() && !useSystemColorMode) return
+    if (initialColorMode === "system") {
+      return rawSetColorMode(systemValue)
```
TimKolberger, line 80: Same as above

```suggestion
     rawSetColorMode(systemValue)
     return
```

packages/color-mode/src/color-mode-script.tsx
```diff
@@ -1,63 +1,35 @@
...
+  const k = isCookie ? storageKey : `'${storageKey}'`
 
-  // Runtime safeguard against invalid color mode values.
-  const validColorModeValues = ["dark", "light", "system"] as const
-  if (!validColorModeValues.includes(initialColorMode)) {
-    initialColorMode = "light"
-  }
+  const cookieScript = `!function(){try{var t="(prefers-color-scheme: dark)", e=window.matchMedia(t).matches?"dark":"light", d=document.documentElement;var ck=document.cookie.match(new RegExp(\`(^| )${k}=([^;]+)\`));var m=ck && ck[2]; if(!m) return document.cookie=\`${k}=${init}; max-age=31536000; path=/\`,d.dataset.theme="system"==='${init}'?e:'${init}';var yy="system"===m?e:m; d.dataset.theme=yy;d.style.colorScheme=yy}catch(t){}}()`
+
+  const localStorageScript = `!function(){try{var t="(prefers-color-scheme: dark)",e=window.matchMedia(t).matches?"dark":"light",d=document.documentElement,m=localStorage.getItem(${k});if(!m)return localStorage.setItem(${k},${init}),d.dataset.theme="system"===${init}?e:${init};var yy="system"===m?e:m; d.dataset.theme=yy;d.style.colorScheme=yy}catch(t){}}()`
```
TimKolberger, line 23: Looks interesting, but seems better than calling `toString()` on a function to get the source.

TimKolberger (changes requested):
Love it:
image

I added a few tiny comments :)

packages/color-mode/src/color-mode-script.tsx
segunadebayo: My aim with this function is to really slim down the script given that it won't change that often compared to the rest of the codebase. Using the shebang `!function(){}()` also ensures that browser executes the script immediately.

TimKolberger (approved):

Issue #4975, opened 2021-11-02
Reduce theme bundle size


hiroppy:
### Description

Hi, I'm @hiroppy from webpack team. First, thank you for creating a wonderful library and I am satisfied with it other than one thing which is the huge bundle size. I want to discuss you have a plan to re-architecture the code structure or not. The bundle size is so important to improve performance and chakra-ui has many unnecessary variables even if bundled.

### Problem Statement/Justification

The code of chakra-ui isn't friendly with dead code elimination so the bundle size will be huge. I created [a loader of webpack](https://github.com/hiroppy/chakra-ui-optimization-loader) to improve chakra-ui's bundle size and it can reduce bundle size, but I know it's a bad practice because I won't realize breaking changes of charka-ui.

e.g. an example of just importing the chakra-ui

You can see all themes exist in a chunk **even if you don't use a component**, this is because of importing all themes as an object field. See https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/index.ts#L17

<img width="1304" alt="w-o-loader" src="https://user-images.githubusercontent.com/3367801/139788628-45ea300a-04bc-4a75-9abb-dde3a77ebb3b.png">

Also, even if you don't use css variables of Twitter, Facebook, and so on, the chunk has them. See https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/foundations/colors.ts#L9-L248

I think we don't need many tokens on production...

When ignore 'Alert', 'Table', 'Tabs', 'Slider'

<img width="1345" alt="w-loader" src="https://user-images.githubusercontent.com/3367801/139789266-8b900585-ecb9-431b-8269-20a6549a3741.png">

the bundle size is down over 10kb

From now on, chakra will introduce many components and the bundle will be bigger than now so I want to discuss preventing this situation.

### Proposed Solution or API

I think we have many ideas.

- provide a config file like tailwind to tell information to bundlers, not runtime config(extendTheme)
- each theme should be in each component, don't add it to a global object, should be as on-demand 
- do you have others...?

### Additional Information

Our company uses the chakra-ui on production but I have this issue. The article is [here](https://engineering.mercari.com/blog/entry/20210823-a57631d32e/) written in Japanese.

hiroppy:
oh, sorry I didn't read [this issue](https://github.com/chakra-ui/chakra-ui/issues/4972), maybe "Core (Styled System)" is similar to this issue🙂

segunadebayo:
[...]

PR #5949, opened 2022-04-29
refactor(theme-tools): replace tinycolor functionality

noobinthisgame:
Relates to #4975 

## 📝 Description

The goal of this PR is to remove the dependency to `@ctrl/tinycolor`. Which would shrink the bundle size by quite a bit.

Most of our color utility functions in `theme-tools` are not used internally in the codebase of Chakra UI. The tradeoff here for the higher bundle size is not worth the little usage. Also removing those functions simplifies the code to maintain.

Users that want to work with advanced color functions should pick their library of choice and overwrite the theme.

Let me know what you think.

Steps after discussion:
- [ ] write tests
- [ ] cleanup
- [ ] changesets

Currently we have `@ctrl/tinycolor` as dependency and use some of its functionality for our color utility functions.

## 🚀 New behavior

All the functionality of `@ctrl/tinycolor` is either replaced by custom functions or removed / deprecated completely.

Yes if we remove everything. No if we deprecated in a first step.

## 📝 Additional Information

I wasn't sure if removing everything directly is a good call, as we shipped those functions. So I chose to mark all the functions as deprecated to inform potential users. Let me know what's best here.

Also the getColor function could possibly be removed. It is not used consistently in the theme. This would lead to even less code to be maintained and no new code needed to replace the functionality from `@ctrl/tinycolor`

### DIFF
--- a/packages/theme/src/components/avatar.ts
+++ b/packages/theme/src/components/avatar.ts
@@ -1,5 +1,5 @@
 import { avatarAnatomy as parts } from "@chakra-ui/anatomy"
-import { isDark, mode, randomColor } from "@chakra-ui/theme-tools"
+import { mode } from "@chakra-ui/theme-tools"
 import type {
   PartsStyleFunction,
   PartsStyleObject,
@@ -23,18 +23,27 @@ const baseStyleExcessLabel: SystemStyleFunction = (props) => {
 }
 
 const baseStyleContainer: SystemStyleFunction = (props) => {
-  const { name, theme } = props
-  const bg = name ? randomColor({ string: name }) : "gray.400"
-  const isBgDark = isDark(bg)(theme)
+  const { theme } = props
 
-  let color = "white"
-  if (!isBgDark) color = "gray.800"
+  const colors = Object.keys(theme.colors).filter(
+    (c) =>
+      ![
+        "whiteAlpha",
+        "blackAlpha",
+        "white",
+        "black",
+        "transparent",
+        "current",
+      ].includes(c),
+  )
+
+  const bgColor = colors[Math.floor(Math.random() * colors.length)]
 
   const borderColor = mode("white", "gray.800")(props)
 
   return {
-    bg,
-    color,
+    bg: mode(`${bgColor}.300`, `${bgColor}.700`)(props),
+    color: mode("gray.800", "white")(props),
     borderColor,
     verticalAlign: "top",
   }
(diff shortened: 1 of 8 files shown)


packages/theme/src/components/avatar.ts
```diff
@@ -23,18 +23,27 @@ const baseStyleExcessLabel: SystemStyleFunction = (props) => {
...
+        "whiteAlpha",
+        "blackAlpha",
+        "white",
+        "black",
+        "transparent",
+        "current",
+      ].includes(c),
+  )
+
+  const bgColor = colors[Math.floor(Math.random() * colors.length)]
```
TimKolberger, line 40: We cannot use `Math.random` in code which is run multiple times, because this will return _random_  colors on each run.
This code will be run on each render of the Avatar component, so we need a stable algorithm.

I think the implementation of `randomColor` does only use tinycolor to create a fallback value, which we could either remove or replace.

TimKolberger (changes requested):
Thanks for working on this @noobinthisgame 💖 

I think deprecating the functions is a good call - we could remove them with Chakra v3.
I added a comment regarding the replacement for `randomColor`.

It would be very valuable to know the amount of bundle size reduction - how many bytes are we saving with the removal of tinycolor?

packages/theme/src/components/avatar.ts
noobinthisgame: Good point. I'll check again on that :)

noobinthisgame:
@TimKolberger I'll provide a comparison of package size :) 
What do you think about the `getColor()` function overall ? It is not really used everywhere consistently. 
In my opinion it provides only little value, checking if a color is really valid. Isn't that a users responsibility ? Am I missing something ?

packages/theme/src/components/avatar.ts
noobinthisgame: I've added a custom fallback to replace the tinycolor implementation. The fallback is not used in our own code. https://github.com/chakra-ui/chakra-ui/blob/next/packages/theme/src/components/avatar.ts#L27
Not using the function without input, thus no case for the avatar.

noobinthisgame:
Here is a comparison of bundle size of before my changes and after my changes in a plain vitejs react and a plain chakra installation:
Before :
image
After:
image

TimKolberger:
Awesome work @noobinthisgame! ✨ 
Kindly remove all the code changes and just deprecate all helper functions related to tinycolor.

noobinthisgame:
I'll open a new PR for this :) We can close this one as reference.

PR #5951, opened 2022-05-01
feature: support for adding custom icons

### LINKED ISSUE #5930: Customize AlertIcon for Toast
### Description

The icon that is being shown in the toast is the default Chakra icon and cannot be modified in the options.

Specifying your own icon makes the Toast component more adjustable.

### Proposed Solution or API

Currently toast takes a `status` and will render an `<AlertIcon />` based on that, which will be a Chakra default icon. If you want to display your own specific icon, but keep all the other functionality, you'd have to use `render` in the options and write out the whole component to show, like:
```
<Alert> 
  <AlertIcon as={youOwnIcon} /> 
</Alert>
```
It would be nice to be able to specify your own icon when calling toast. Something like 
```
toast({
  ...toastOptions,
  alertIcon: youOwnIcon
})
```

### Alternatives

_No response_

### Additional Information

_No response_

haneenmahd:
Closes #5930

## 📝 Description

> Adds support for passing in custom icons into toasts. 

> we do not have a possibility to render custom icons on the toast element.

## 🚀 New behavior

> The toasts elements will now render custom icons, if they are passed into their props. If there are no icons passed in, we will render the Default `<AlertIcon />` component. The icon is a React Node that will be passed into the component using the prop `icon`, which is `optional`.

No, this does not break anything instead adds a new feature.

## 📝 Additional Information
This would be a great feature for devs who want to customise their toast's UI. So, I just need to know that if this feature will fit match UI perfectly.

### DIFF
--- a/packages/toast/src/toast.tsx
+++ b/packages/toast/src/toast.tsx
@@ -19,7 +19,7 @@ export interface ToastProps
 }
 
 export const Toast: React.FC<ToastProps> = (props) => {
-  const { status, variant, id, title, isClosable, onClose, description } = props
+  const { status, variant, id, title, isClosable, onClose, description, icon } = props
 
   const alertTitleId =
     typeof id !== "undefined" ? `toast-${id}-title` : undefined
@@ -37,7 +37,7 @@ export const Toast: React.FC<ToastProps> = (props) => {
       width="auto"
       aria-labelledby={alertTitleId}
     >
-      <AlertIcon />
+      {icon || <AlertIcon />}
       <chakra.div flex="1" maxWidth="100%">
         {title && <AlertTitle id={alertTitleId}>{title}</AlertTitle>}
         {description && (
--- a/packages/toast/src/toast.tsx
+++ b/packages/toast/src/toast.tsx
@@ -37,7 +37,7 @@ export const Toast: React.FC<ToastProps> = (props) => {
       width="auto"
       aria-labelledby={alertTitleId}
     >
-      {icon || <AlertIcon />}
+      <AlertIcon>{icon}</AlertIcon>
       <chakra.div flex="1" maxWidth="100%">
         {title && <AlertTitle id={alertTitleId}>{title}</AlertTitle>}
         {description && (
--- a/packages/toast/stories/toast.stories.tsx
+++ b/packages/toast/stories/toast.stories.tsx
@@ -381,7 +381,7 @@ export const AsyncToast = () => {
   )
 }
 
-export const ToastWidthCustomIcon = () => {
+export const ToastWithCustomIcon = () => {
   const toast = useToast()
   const id = "toast-with-custom-icon"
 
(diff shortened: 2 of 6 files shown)


packages/toast/stories/toast.stories.tsx
```diff
@@ -380,3 +380,44 @@ export const AsyncToast = () => {
     </ButtonGroup>
   )
 }
+
+export const ToastWidthCustomIcon = () => {
```
haneenmahd, line 384: Replace 'ToastWidth...' with 'ToastWith...'

haneenmahd (commented):
Everything seems good 👍🏻 !

packages/toast/src/toast.tsx
```diff
@@ -37,7 +37,7 @@ export const Toast: React.FC<ToastProps> = (props) => {
       width="auto"
       aria-labelledby={alertTitleId}
     >
-      <AlertIcon />
+      {icon || <AlertIcon />}
```
TimKolberger, line 40: By replacing the <AlertIcon /> with custom icon we remove the possibility to style it via the theme, because the theme styles are applied within the AlertIcon component.

https://github.com/chakra-ui/chakra-ui/blob/a26f7f8c02561f2a20cc6fac416fa5e758175eb1/packages/alert/src/alert.tsx#L122-L138

When we allow the AlertIcon to receive `children` we have the best out of both worlds.

```tsx
// packages/alert/src/alert.tsx
return (
    <chakra.span
      display="inherit"
      {...props}
      className={cx("chakra-alert__icon", props.className)}
      __css={css}
    >
      {props.children || <BaseIcon h="100%" w="100%" />}
    </chakra.span>
  )
```

and we can update the implementation here to

```diff
-{icon || <AlertIcon />}
+<AlertIcon>{icon}</AlertIcon>
```

What do you think?

TimKolberger (changes requested):
Awesome work @haneenmahd ✨ 
I left a comment regarding the AlertIcon component.

packages/toast/src/toast.tsx
haneenmahd: Yes, I think this kinda way the using custom icon is much more maintainable with the theme and stuff. Great suggestion 👍 .
Thanks 🤟🏻

haneenmahd: Hey, I have changed the code according to what you said.

TimKolberger: Awesome 🚀 
I think we still need the change in `alert.tsx`:

```diff
// packages/alert/src/alert.tsx
return (
    <chakra.span
      display="inherit"
      {...props}
      className={cx("chakra-alert__icon", props.className)}
      __css={css}
    >
-      <BaseIcon h="100%" w="100%" />
+      {props.children || <BaseIcon h="100%" w="100%" />}
    </chakra.span>
  )
```

haneenmahd: I'm sorry, Just forgot that  😅

haneenmahd: I have made the change, take a look 👀

TimKolberger (approved):
⚡ ⚡ ⚡

anubra266 (approved):

PR #5963, opened 2022-05-03
Make `leastDestructiveRef` property non-nullable

RunDevelopment:
## 📝 Description

There was likely an oversight when creating the type definition for `AlertDialog`. `AlertDialogProps` requires the `leastDestructiveRef` property to be present, even though the property can be trivially set to `undefined`. This makes the property effectively optional, but the types don't (currently) allow this.

When creating a `AlertDialog` in a TS project, the elements requires the `leastDestructiveRef` property. However, the property can trivially be set to `undefined`.

```tsx
      <AlertDialog
        isOpen
        onClose={() => { ... }}
        leastDestructiveRef={undefined}
      >
```

## 🚀 New behavior

When creating a `AlertDialog` in a TS project, the `leastDestructiveRef` property is now optional and can be omitted.

```tsx
      <AlertDialog
        isOpen
        onClose={() => { ... }}
      >
```

The behavior is identical to setting `leastDestructiveRef` to undefined.

No.

### DIFF
--- a/.changeset/nervous-keys-look.md
+++ b/.changeset/nervous-keys-look.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/modal": patch
+---
+
+Make `leastDestructiveRef` non-nullable
--- a/packages/modal/src/alert-dialog.tsx
+++ b/packages/modal/src/alert-dialog.tsx
@@ -3,7 +3,7 @@ import * as React from "react"
 import { Modal, ModalContent, ModalContentProps, ModalProps } from "./modal"
 
 export interface AlertDialogProps extends Omit<ModalProps, "initialFocusRef"> {
-  leastDestructiveRef?: ModalProps["initialFocusRef"]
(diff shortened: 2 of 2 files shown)


segunadebayo:
Hi @RunDevelopment,

This was indeed an oversight when we created the type definitions. We did intend to make `leastDestructiveRef ` required though given the semantic meaning of an alert dialog (a dialog that requires an action).

This actually means we should update the type to not allow `undefined`.

RunDevelopment:
In that case, `AlertDialog` should also throw a runtime error if `leastDestructiveRef` is undefined. This way, the static types match the runtime behavior.

RunDevelopment:
Anyway, I made `leastDestructiveRef` non-nullable for now. A runtime check can still be added later if so desired.

cheshireoctopus:
👋 @segunadebayo 

Apologies for commenting on merged PR's.

Are you able to help me understand the importance of `leastDestructiveRef`?

Interestingly named property!

Appreciate you help - thank you.

PR #6003, opened 2022-05-11
feat: fix size issue in drawer

### LINKED ISSUE #5921: [Drawer] "size" props isn't used when placement="top|bottom"
### Description

When I pass a size prop to the drawer I expect it to be sized accordingly to the prop regardless of what the placement prop is

### Link to Reproduction

https://codesandbox.io/s/priceless-benz-rviclr?file=/src/index.tsx

### Steps to reproduce

_No response_

### Chakra UI Version

1.8.8

### Browser

_No response_

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

image

hardikhere:
Closes #5921 

## 📝 Description

Drawer was not handling size prop when placement was top or bottom 

Size prop doesn't work if the drawer has placement = "top" or "bottom"

## 🚀 New behavior

Size prop works for top and bottom too.

### DIFF
--- a/packages/modal/src/drawer.tsx
+++ b/packages/modal/src/drawer.tsx
@@ -75,6 +75,7 @@ export function Drawer(props: DrawerProps) {
         isOpen={isOpen}
         onClose={onClose}
         styleConfig={drawerStyleConfig}
+        _drawerPlacement={placement}
         {...rest}
       >
         {children}
--- a/packages/modal/src/modal.tsx
+++ b/packages/modal/src/modal.tsx
@@ -119,6 +119,11 @@ export interface ModalProps
    * Fires when all exiting nodes have completed animating out
    */
   onCloseComplete?: () => void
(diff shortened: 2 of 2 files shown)


hardikhere:
Hello @segunadebayo Can you please review if this solution is correct 
Thanks.

segunadebayo:
Hi @hardikhere,

Thanks for working on this. This is considered a breaking change and I'll prefer we find an alternative approach to this.

Thanks for understanding.

hardikhere:
> Hi @hardikhere,
> 
> Thanks for working on this. This is considered a breaking change and I'll prefer we find an alternative approach to this.
> 
> Thanks for understanding.

Yeah, I also think That we need an alternative approach, previously I tried this with hook but it was giving a bunch of errors. Let me know if you come up with another approach I will try to contribute.

Issue #6054, opened 2022-05-19
BorderRadius on Tag.container Theme not working


johannespein:
### Description

When changing the `borderRadius` on the `baseStyle.container` of the Tag component, I expect it to change on the UI.

### Link to Reproduction

https://codesandbox.io/s/chakra-ui-borderradius-on-tag-not-working-hlxcwp?file=/src/index.js

### Steps to reproduce

Change the `borderRadius` prop on the Theme of the Tag container

### Chakra UI Version

1.7.4

johannespein:
I made it work by doing this:

```js
const Tag = {
  variants: {
    none: {
      container: {
        borderRadius: '20px'
      }
    }
  },
  defaultProps: {
    variant: 'none'
  }
}
```

but now I need to define the borderRadius for every variant individually

PR #6056, opened 2022-05-19
fix: baseStyle borderRadius on Tag component

johannespein:
Closes [6054](https://github.com/chakra-ui/chakra-ui/issues/6054)

## 📝 Description

Tag component variants borderRadius was overwriting baseStyle eventhough
borderRadius was same for all variants. borderRadius is now part of the
baseStyle, so it can be overwritten by a custom theme on the baseStyle instead of variants. 

No

### DIFF
--- a/.changeset/nasty-cows-begin.md
+++ b/.changeset/nasty-cows-begin.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/theme": patch
+---
+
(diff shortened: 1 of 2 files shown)


johannespein:
https://codesandbox.io/s/create-react-app-ts-forked-k918qt?file=/src/App.tsx

PR #6061, opened 2022-05-19
fix(slider): Updated the styling for the _disabled states.

### LINKED ISSUE #6060: Fix `_disabled` state of slider component
### Description

When I hover the disabled slider component, I expect to see a `not-allowed` cursor type but I don't see any cursor.

### Link to Reproduction

https://codesandbox.io/s/fix-slider-add-disabled-state-cd8q66

### Steps to reproduce

1. Render a `<Slider/>` component, add the `isDisabled` prop.
2. Hover over the slider.
3. Observe that there is no `not-allowed` cursor.
> **Side Note:** Please see the disabled state on other components such as `Button`, `Checkbox`, or `Input` as examples. These components will provide a `not-allowed` cursor when disabled.

### Chakra UI Version

2.0.2

### Browser

Google Chrome 101.0.4951.64

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Here is what it looks like after the fix applied, sorry for the picture of screen. I couldn't get the hover state to say while taking a screenshot, it would appear even with a snapshot delay focus is given to the screenshot program before the actual capture.
slider_disabled

SpencerDuball:
Closes #6060 

## 📝 Description

Updating the styling of the `slider` component to add the `not-allowed` cursor on hover when `isDisabled`.

Currently, there is no cursor displayed when hovering the `slider` component when it is disabled.
See live example here: https://codesandbox.io/s/fix-slider-add-disabled-state-cd8q66

## 🚀 New behavior

When hover, the cursor will display as `not-allowed`.

No, not a breaking change.

## 📝 Additional Information

(1) Updated baseStyleContainer to show the "not-allowed" cursor when disabled.
(2) Updated thumbOrientation _active -> _disabled state to translateY when active and disabled.
    This update was necessary, because the pointerEvents on baseStyleContainer needed to be turned
    to "auto" from "none" or else the "not-allowed" cursor would not be shown.

### DIFF
--- a/packages/theme/src/components/slider.ts
+++ b/packages/theme/src/components/slider.ts
@@ -15,6 +15,9 @@ function thumbOrientation(props: StyleFunctionProps): SystemStyleObject {
       transform: `translateX(-50%)`,
       _active: {
         transform: `translateX(-50%) scale(1.15)`,
+        _disabled: {
+          transform: `translateX(-50%)`,
+        },
       },
     },
     horizontal: {
@@ -22,6 +25,9 @@ function thumbOrientation(props: StyleFunctionProps): SystemStyleObject {
       transform: `translateY(-50%)`,
       _active: {
         transform: `translateY(-50%) scale(1.15)`,
+        _disabled: {
+          transform: `translateY(-50%)`,
+        },
       },
     },
   })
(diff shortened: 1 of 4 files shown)


packages/theme/src/components/slider.ts
```diff
@@ -15,13 +15,19 @@ function thumbOrientation(props: StyleFunctionProps): SystemStyleObject {
       transform: `translateX(-50%)`,
       _active: {
         transform: `translateX(-50%) scale(1.15)`,
+        _disabled: {
+          transform: `translateX(-50%)`,
+        },
```
segunadebayo, line 20: Do we really need these styles?

SpencerDuball: Please see below comment, same discussion applies but onto the x-axis attributes.

PR #6063, opened 2022-05-20
fix(tabs): Adding disabled state for all variants.

### LINKED ISSUE #6062: Fix `_disabled` state of tabs component
### Description

When I hover a disabled tab I expect to see a `not-allowed` cursor and `0.4` opacity, but I see the `pointer` cursor and `1.0` opacity.

### Link to Reproduction

https://codesandbox.io/s/fix-tabs-add-disabled-state-buczy5

### Steps to reproduce

**Line Variant**
1. Visit the reproduction link, above.
2. Hover over the "Line Variant" tab "Three".
3. You should see the `not-allowed` cursor. This is good, this variant already has `_disabled` state added.
4. You will also see the `0.4` opacity is applied to the `_disabled` tab, this is also good.
5. Now, click the "Line Variant" tab "Three", and you will see the background color changes indicating a successful click. This is the only issue with the "Line Variant". Please see the Button, Checkbox, and Input components to compare. These components do not give click feedback via a background color change when disabled.

**Enclosed Variant**
1. Visit the reproduction link, above.
2. Hover over the "Enclosed Variant" tab "Three".
3. You should not see the `not-allowed` cursor, we see the standard `pointer` cursor which may frustrate users if they think the tab is clickable but cannot get the state to change.
4. You will also not see the `0.4` opacity at any time when `_disabled`, this is also an issue.

**Enclosed Colored Variant**
_Repeat the same steps as Enclosed Variant, see the same issues with the Enclosed Colored Variant_

**Soft Rounded Variant**
_Repeat the same steps as Enclosed Variant, see the same issues with the Enclosed Colored Variant_

**Solid Rounded Variant**
_Repeat the same steps as Enclosed Variant, see the same issues with the Enclosed Colored Variant_

### Chakra UI Version

2.0.2

### Browser

Google Chrome 101.0.4951.64

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I did not add any test, or change the stories. I noticed the tabs stories already has a `_disabled` state component, so I did not change/add any stories. I also noticed that test for styling does not seem to be present, I also did not add tests for this reason.

Please let me know if you would like me to add any tests/stories that demonstrate or test this issue.

SpencerDuball:
Closes #6062 

## 📝 Description

1. Adds disabled state to variants [`enclosed`, `enclosed-colored`, `soft-rounded`, `solid-rounded`] of the Tabs component. 
2. Also added a fix that prevents visual feedback of a successful click on the `_disabled` tabs of variant `line`.

1. Previously there was no visual feedback if a tab was disabled to the end user.
2. Previously, when clicking the tab would show a background color that gave indication of a successful click, other components in the library such as Button, Checkbox, and Input do not display this.

## 🚀 New behavior

1. When a tab is `_disabled` the `cursor: 'not-allowed'` will be shown, also the `opacity: 0.4` will be displayed.
2. When clicking the `_disabled` tab, the background will be `background: none`.

No, this is not a breaking change.

## 📝 Additional Information
(1) Added the disabled state to variantEnclosed, variantEnclosedColored, variantSoftRounded, variantSolidRounded
(2) Set bg to "none" in variantLine when _disabled. This new styling mimics that of other components such as the
    button, checkbox, and input which do not have styling changes on clicks when disabled. AKA it shouldn't look
    like a click when the tab button is disabled.

### DIFF
--- a/packages/theme/src/components/tabs.ts
+++ b/packages/theme/src/components/tabs.ts
@@ -26,6 +26,10 @@ const baseStyleTab: SystemStyleFunction = (props) => {
       zIndex: 1,
       boxShadow: "outline",
     },
+    _disabled: {
+      cursor: "not-allowed",
+      opacity: 0.4,
+    },
   }
 }
 
@@ -102,8 +106,6 @@ const variantLine: PartsStyleFunction<typeof parts> = (props) => {
         bg: mode("gray.200", "whiteAlpha.300")(props),
       },
       _disabled: {
-        opacity: 0.4,
-        cursor: "not-allowed",
         _active: { bg: "none" },
       },
     },
@@ -122,10 +124,6 @@ const variantEnclosed: PartsStyleFunction<typeof parts> = (props) => {
         color: mode(`${c}.600`, `${c}.300`)(props),
         borderColor: "inherit",
         borderBottomColor: mode(`white`, `gray.800`)(props),
-      },
-      _disabled: {
-        cursor: "not-allowed",
-        opacity: 0.4,
       },
     },
     tablist: {
@@ -154,10 +152,6 @@ const variantEnclosedColored: PartsStyleFunction<typeof parts> = (props) => {
         borderTopColor: "currentColor",
         borderBottomColor: "transparent",
       },
-      _disabled: {
-        cursor: "not-allowed",
-        opacity: 0.4,
-      },
     },
     tablist: {
       mb: "-1px",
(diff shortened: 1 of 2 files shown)


packages/theme/src/components/tabs.ts
```diff
@@ -122,6 +123,10 @@ const variantEnclosed: PartsStyleFunction<typeof parts> = (props) => {
         borderColor: "inherit",
         borderBottomColor: mode(`white`, `gray.800`)(props),
       },
+      _disabled: {
+        cursor: "not-allowed",
+        opacity: 0.4,
+      },
```
segunadebayo, line 129: Great work here. Thank you!

How about we add this to the base style of the tab instead of duplicating it in each variant? I think we should update the `baseStyleTab` styles instead.

SpencerDuball: Good point, I will get this updated today. Thanks!

PR #6069, opened 2022-05-21
fix: use-media-query ssr

segunadebayo:
## 📝 Description

Fix hydration mismatch issue when using `useBreakpoint` or `useBreakpointValue` in an SSR environment.

Throws a mismatch error

## 🚀 New behavior

Works

No


velddev:
This PR makes me very happy!

PR #6075, opened 2022-05-22
Update version and fix ts error

fredckl:
- Update @types version
- Fix error ts when get Element by ID


tooling/cra-template-typescript/template/src/index.tsx
```diff
@@ -5,22 +5,29 @@ import { App } from "./App"
 import reportWebVitals from "./reportWebVitals"
 import * as serviceWorker from "./serviceWorker"
 
-const container = document.getElementById("root")
-const root = ReactDOM.createRoot(container)
+(() => {
+  try {
+    const container = document.getElementById("root")
+    if (!container) throw new Error('Failed to find the root element');
+    const root = ReactDOM.createRoot(container)
```
segunadebayo, line 13: Could we do without this since the `root` is definitely added to the `index.html`? No need for a runtime check

PR #6090, opened 2022-05-23
docs: Add typescript details to `AlertDialog`

KenzoBenzo:
## 📝 Description

> The `AlertDialog` component currently has an example that Typescript throws an error for. This PR provides a reference point for people hitting the same issue.


packages/modal/README.md
```diff
@@ -189,6 +189,16 @@ function AlertDialogExample() {
 }
 ```
 
+## Usage with Typescript
+
+Using the above example will throw a typescript error. You'll want to modify the type of the `useRef` hook to the HTML element you're placing the focus on and default the value to null.
+
+
+```jsx
```
segunadebayo, line 197:
```suggestion
```tsx
```

PR #6106, opened 2022-05-25
feat(modal): add border radius of zero on full size

### LINKED ISSUE #6066: Modal fullscreen should not have a border radius by default
### Description

The modal component by default has a border radius, when using `size="full"`. When using its fullscreen size this is not necessary.

The fullscreen 'feeling' is less due to the fact that the fullscreen size is not really _full_

See: [https://codesandbox.io/s/chakra-ui-fullscreen-modal-xqgf2f?file=/src/index.js](https://codesandbox.io/s/chakra-ui-fullscreen-modal-xqgf2f?file=/src/index.js)

### Proposed Solution or API

Removing the border radius if the size is `full` 

### Alternatives

_No response_

### Additional Information

_No response_

RubenSmn:
Closes #6066 

## 📝 Description
Add `borderRadius` of `none` to the default style of `full` size modal.

Current full size modal has a border radius by default.

## 🚀 New behavior
Now the full size modal has no border radius by default.

No.

### DIFF
--- a/packages/theme/src/components/modal.ts
+++ b/packages/theme/src/components/modal.ts
@@ -90,6 +90,7 @@ function getSize(value: string): PartsStyleObject<typeof parts> {
           minH: "-webkit-fill-available",
         },
         my: 0,
+        borderRadius: "none",
       },
     }
   }
--- a/packages/theme/src/components/modal.ts
+++ b/packages/theme/src/components/modal.ts
@@ -90,7 +90,7 @@ function getSize(value: string): PartsStyleObject<typeof parts> {
           minH: "-webkit-fill-available",
         },
         my: 0,
-        borderRadius: "none",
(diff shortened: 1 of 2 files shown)


packages/theme/src/components/modal.ts
```diff
@@ -90,6 +90,7 @@ function getSize(value: string): PartsStyleObject<typeof parts> {
           minH: "-webkit-fill-available",
         },
         my: 0,
+        borderRadius: "none",
```
trybick, line 93: @RubenSmn I think it would be more proper to go with `0` instead of `none`. I tried `none` in the browser and it looks like it's not accepted.
Screen Shot 2022-06-09 at 8 53 14 AM

trybick (approved):

PR #6115, opened 2022-05-27
fix(tooltip): don't show tooltips after opening a modal

### LINKED ISSUE #2503: Tooltip delay + Drawer = bug
# Bug report

## Describe the bug

Tooltips with a delay applied to buttons that open drawer causes the following problem:

- If the user clicks on the button **before** the tooltip appears, 
    - The drawer appear, so far so good.
- Then if the user moves the cursor out of where the button was, still **before** the tooltip appears
    - Then the tooltip **still** appears after the drawer, on top of it, creating a weird graphical bug. The tooltip is then impossible to dismiss without leaving the drawer.

This might sound like a rare case, but it happens all the time if you put a normal delay of 500-1000ms to the tooltip, and use the tooltip on a button that opens a drawer. 

Some variants of the bug happen depending on the exact configuration. 

## To reproduce

https://codesandbox.io/s/youthful-rgb-iit3e?file=/src/App.tsx

## Minimal reproduction

https://codesandbox.io/s/youthful-rgb-iit3e?file=/src/App.tsx

## Expected behavior

Tooltip should only appear if its target is visible on screen for example. Here it happens even though there is a drawer in front of it.

## Screenshots

Screen Recording 2020-11-15 at 01 38 14 mov

## System information

- OS: macOS but I suspect all
- Browser: Chrome but I suspect all
- Version of @chakra-ui/react: 1.0.0

itkrt2y:
Closes #2503

## 📝 Description

Avoid rendering multiple tooltips so that they are not appeared after opening a modal

If the user clicks the modal trigger button before opening the tooltip, `openWithDelay` will be called twice by `onMouseEnter` and `onFocus`. `openWithDelay` using `setTimeout` to open a tooltip.
#2503 is caused by this double `setTimeout` .

When the user click the trigger button, `onClick` or `onMouseDown` is called, and they call `closeWithDelay` to close a tooltip and clear an enter timer.
However, `enterTimeout` ref stores only the latest timer id, so `closeWithDelay` clears only the latest timer and the previous timer displays a tooltip after opening the modal.

## 🚀 New behavior

Don't show tooltips after opening a modal

No

## 📝 Additional Information

Testable sandbox https://codesandbox.io/s/serene-worker-i9mpqg

### DIFF
--- a/packages/tooltip/src/use-tooltip.ts
+++ b/packages/tooltip/src/use-tooltip.ts
@@ -117,7 +117,7 @@ export function useTooltip(props: UseTooltipProps = {}) {
   const exitTimeout = React.useRef<number>()
 
   const openWithDelay = React.useCallback(() => {
-    if (!isDisabled) {
+    if (!isDisabled && !enterTimeout.current) {
       enterTimeout.current = window.setTimeout(onOpen, openDelay)
     }
   }, [isDisabled, onOpen, openDelay])
@@ -125,6 +125,7 @@ export function useTooltip(props: UseTooltipProps = {}) {
   const closeWithDelay = React.useCallback(() => {
     if (enterTimeout.current) {
       clearTimeout(enterTimeout.current)
+      enterTimeout.current = undefined
     }
     exitTimeout.current = window.setTimeout(onClose, closeDelay)
   }, [closeDelay, onClose])
--- a/packages/tooltip/src/use-tooltip.ts
+++ b/packages/tooltip/src/use-tooltip.ts
@@ -131,16 +131,16 @@ export function useTooltip(props: UseTooltipProps = {}) {
   }, [closeDelay, onClose])
 
   const onClick = React.useCallback(() => {
-    if (closeOnClick) {
+    if (isOpen && closeOnClick) {
       closeWithDelay()
     }
-  }, [closeOnClick, closeWithDelay])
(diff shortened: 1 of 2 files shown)


packages/tooltip/src/use-tooltip.ts
```diff
@@ -117,29 +117,30 @@ export function useTooltip(props: UseTooltipProps = {}) {
   const exitTimeout = React.useRef<number>()
 
   const openWithDelay = React.useCallback(() => {
-    if (!isDisabled) {
+    if (!isDisabled && !enterTimeout.current) {
```
itkrt2y, line 120: Do not set the enter timer if it still exists to avoid rendering multiple tooltips

packages/tooltip/src/use-tooltip.ts
```diff
@@ -117,29 +117,30 @@ export function useTooltip(props: UseTooltipProps = {}) {
...
     if (enterTimeout.current) {
       clearTimeout(enterTimeout.current)
+      enterTimeout.current = undefined
     }
     exitTimeout.current = window.setTimeout(onClose, closeDelay)
   }, [closeDelay, onClose])
 
   const onClick = React.useCallback(() => {
-    if (closeOnClick) {
+    if (isOpen && closeOnClick) {
```
itkrt2y, line 134: Avoid clearing an enter timer before opening the tooltip when clicking the trigger button

PR #6119, opened 2022-05-28
fix(modal): fix modal not to change position when the radio button is focused

### LINKED ISSUE #5926: Modal position is affected when a radio button is clicked inside a modal
### Description

When a radio button or checkbox is clicked inside a modal which has some long content, the modal position is affected.

### Link to Reproduction

https://chakra-modal-click-inside-issue.vercel.app/

### Steps to reproduce

1. Go to https://chakra-modal-click-inside-issue.vercel.app/
3. Click on any of the modal buttons to open the modal
4. Scroll down to the bottom of the opened modal and click on the radio input
5. You'll see that the modal moves up! It seems the degree to which the modal moves up is related to how much of content is before that input 😅

You can go through the code of the deployed version on the public repo: https://github.com/jaisharx/chakra-modal-click-inside-issue. I've also tried adding a very simple radio group to one of the modals in the chakra docs itself (it happens with scrollBehaviour="inside"), and the issue was reproducible!

Screenshot 2022-04-25 at 6 15 12 PM

Screenshot 2022-04-25 at 6 15 33 PM

Screenshot 2022-04-25 at 6 17 05 PM

### Chakra UI Version

^1.8.8

### Browser

Edge 100

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

itkrt2y:
Closes #5926

## 📝 Description

Fix a bug in chromium browsers where the modal position would change when focusing on the radio button in the modal.

When a radio button is focused inside a modal which has some long content, the modal position is affected in chromium browsers.
This is caused by `position: absolute` of the [`visuallyHiddenStyle`](https://github.com/chakra-ui/chakra-ui/blob/d26c7d7aa2ca8951794cdfdc5c340bf8ff0866ec/packages/visually-hidden/src/visually-hidden.tsx#L18) used by [`<input type="radio">`](https://github.com/chakra-ui/chakra-ui/blob/d26c7d7aa2ca8951794cdfdc5c340bf8ff0866ec/packages/radio/src/radio.tsx#L117).

## 🚀 New behavior

Wrap the `<input type="radio">` with `position: relative` and fix it so that the modal doesn't move when focusing on the radio button.

No

## 📝 Additional Information

- Checkbox also uses `visuallyHiddenStyle`, but this issue doesn't occur because [it's already wrapped with `position: relative`](https://github.com/chakra-ui/chakra-ui/blob/d26c7d7aa2ca8951794cdfdc5c340bf8ff0866ec/packages/checkbox/src/checkbox.tsx#L35).
- Testable sandbox https://codesandbox.io/s/jolly-margulis-uftvky

### DIFF
--- a/.changeset/quick-fans-act.md
+++ b/.changeset/quick-fans-act.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/modal": patch
+---
+
+Fix a bug in chromium browsers where the modal position would change when
+focusing on the radio button in the modal which has some long content.
--- a/.changeset/quick-fans-act.md
+++ b/.changeset/quick-fans-act.md
@@ -1,5 +1,5 @@
 ---
-"@chakra-ui/modal": patch
+"@chakra-ui/radio": patch
 ---
 
 Fix a bug in chromium browsers where the modal position would change when
--- a/packages/radio/src/radio.tsx
+++ b/packages/radio/src/radio.tsx
@@ -114,7 +114,9 @@ export const Radio = forwardRef<RadioProps, "input">((props, ref) => {
 
   return (
     <chakra.label className="chakra-radio" {...rootProps} __css={rootStyles}>
-      <input className="chakra-radio__input" {...inputProps} />
+      <chakra.div position="relative">
+        <input className="chakra-radio__input" {...inputProps} />
+      </chakra.div>
       <chakra.span
         className="chakra-radio__control"
         {...checkboxProps}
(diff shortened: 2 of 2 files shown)


segunadebayo:
Thanks for working on this @itkrt2y.

Why not add the `position: relative` to the `chakra.label` as we do in the checkbox? I mean adding the style here:
https://github.com/chakra-ui/chakra-ui/blob/220626650f2e53938e19ccb779a0ed80b3e6fdcc/packages/radio/src/radio.tsx#L93

itkrt2y:
> Why not add the `position: relative` to the `chakra.label` as we do in the checkbox?

Because I'm not sure it would be a breaking change. That's why I chose a safe way.

segunadebayo:
I think the change is a welcome, non-breaking change. It addresses the original intention for the Radio component

itkrt2y:
OK. I fixed it 3e300547ad93893f20b7c1b419b106b9a2565734

PR #6137, opened 2022-06-02
feat(StyleProvider): style provider factory w/ helpful error message

### LINKED ISSUE #5119: Confusing error "useStyles: `styles` is undefined" from a simple mistake

Patrick-Ullrich:
Closes #5119

## 📝 Description

As outlined in the issue, because all components share a pre created StyleContext, the error message was not very descriptive. This PR replaces the current implementation with a helper to create a consistent StyleProvider in every component. This will ensure that the error message is:
1. Consistent
2. Helpful

Currently the error message when missing wrapping a component in the correct parent component with the correct StyleProvider reads:
```
useStyles: `styles` is undefined. Seems you forgot to wrap the components in `<StylesProvider />` 
```

## 🚀 New behavior

```useStyles: "styles" is undefined. Seems you forgot to wrap the components in "<__COMPONENT_NAME__" />"```

Screen Shot 2022-06-02 at 9 33 13 AM

The context name now also includes the component name.
>> <StyleProvider to <__COMPONENT_NAME__StyleProvider>

No

### DIFF
--- a/.changeset/nervous-bulldogs-cheer.md
+++ b/.changeset/nervous-bulldogs-cheer.md
@@ -1,0 +1,23 @@
+---
+"@chakra-ui/accordion": patch
+"@chakra-ui/alert": patch
+"@chakra-ui/avatar": patch
+"@chakra-ui/breadcrumb": patch
+"@chakra-ui/editable": patch
+"@chakra-ui/form-control": patch
+"@chakra-ui/input": patch
+"@chakra-ui/layout": patch
+"@chakra-ui/menu": patch
+"@chakra-ui/modal": patch
+"@chakra-ui/number-input": patch
+"@chakra-ui/popover": patch
+"@chakra-ui/progress": patch
+"@chakra-ui/slider": patch
+"@chakra-ui/stat": patch
+"@chakra-ui/system": patch
+"@chakra-ui/table": patch
+"@chakra-ui/tabs": patch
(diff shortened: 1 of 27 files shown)


asfand987:
Is this also fixed for table? I am getting exact same error for Table component after upgrading from v1 to v2.

PR #6153, opened 2022-06-05
refactor: remove annoying focus outline

segunadebayo:
## 📝 Description

Removes the annoying focus outline by leveraging the `:focus-visible` pseudo-class and a `trackFocusVisible` helpers from Zag.js

A focus outline is always shown because we use the `:focus` selector

## 🚀 New behavior

Focus outline shows only when the keyboard is used to interact with the component

No

### DIFF
--- a/.changeset/breezy-boats-relax.md
+++ b/.changeset/breezy-boats-relax.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/checkbox": minor
+"@chakra-ui/styled-system": minor
+"@chakra-ui/theme": minor
+---
+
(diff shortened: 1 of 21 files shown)


AshMW2724:
This would be so nice so I can remove this crappy fix

Screen Shot 2022-06-06 at 12 41 15 AM

Ploppy3:
@segunadebayo any chance for this to be implemented in v1 ? React 18 compatibility issues make it hard to switch to v2.

simontaisne:
Nice job! Focusing on the `Radio` component, however, no longer shows any outline. I'll open an issue 👍

PR #6157, opened 2022-06-06
fix(media-query): get correct first render value for csr apps

### LINKED ISSUE #6130: useMediaQuery value is incorrect on first render
### Description

When using useMediaQuery, the value it returns is incorrect on first render.

### Link to Reproduction

https://codesandbox.io/s/usemediaquery-testing-xkv2dh?file=/src/App.tsx

### Steps to reproduce

1. Go to the codesandbox link
2. Open the console
3. Notice the initial value is `false` while subsequent values are `true`

### Chakra UI Version

2.1.2

### Browser

Google Chrome 102

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Why is it important that the initial value be correct? Some components might take a prop like `initialView` which gets passed the first value from useMediaQuery and updating this prop doesn't help because it only looks at the initial value. Ideally, the initial value of useMediaQuery would be accurate.

segunadebayo:
Closes #6130

## 📝 Description

Add support for client-side rendered (CSR) apps to get the correct value on the first render.

> Affected hooks: `useMediaQuery`, `useBreakpoint`, `useBreakpointValue`.

These hooks are built to work in server-side rendering (SSR) applications by default. You might notice a quick flash of incorrect media query values when you use them.

If you're creating a CSR-only app, you can now leverage the `ssr` argument to get the correct value on the first render.

```jsx live=false
const [isMobile] = useMediaQuery("(max-width: 768px)", {
  // you can now pass `ssr: false`
  ssr: false,
})
const buttonSize = useBreakpointValue(
  { base: "sm", lg: "md" },
  // you can now pass `ssr: false`
  { ssr: false },
)
// you can now pass `ssr: false`
const breakpoint = useBreakpoint({ ssr: false })
```

CSR only apps still face the Incorrect value on first render

## 🚀 New behavior

With the new option, this doesn't happen anymore

No

### DIFF
--- a/packages/media-query/src/use-media-query.ts
+++ b/packages/media-query/src/use-media-query.ts
@@ -4,38 +4,49 @@ import { useEffect, useLayoutEffect, useState } from "react"
 
 const useSafeLayoutEffect = isBrowser ? useLayoutEffect : useEffect
 
+export type UseMediaQueryOptions = {
+  fallback?: boolean | boolean[]
+  ssr?: boolean
+}
+
 /**
  * React hook that tracks state of a CSS media query
  *
  * @param query the media query to match
- * @param defaultValues the default values to match
+ * @param options the mediq query options { fallback, ssr }
  */
 export function useMediaQuery(
   query: string | string[],
-  defaultValues?: boolean | boolean[],
+  options: UseMediaQueryOptions = {},
 ): boolean[] {
+  const { ssr = true, fallback } = options
+
   const env = useEnvironment()
 
   const queries = Array.isArray(query) ? query : [query]
 
-  let defaults = Array.isArray(defaultValues) ? defaultValues : [defaultValues]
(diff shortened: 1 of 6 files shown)


packages/media-query/src/use-media-query.ts
```diff
@@ -4,38 +4,49 @@ import { useEffect, useLayoutEffect, useState } from "react"
...
+  fallback?: boolean | boolean[]
+  ssr?: boolean
+}
+
 /**
  * React hook that tracks state of a CSS media query
  *
  * @param query the media query to match
- * @param defaultValues the default values to match
+ * @param options the mediq query options { fallback, ssr }
```
trybick, line 16: small typo on "media".  @segunadebayo Instead of passing `ssr` option on every hook, would it be worth considering passing `ssr` to `ChakraProvider`? Either way, thanks for your hard work!

PR #6170, opened 2022-06-08
fix(modal): add size to modal with reference to theme.sizes

Ryuta1346:
no issues.

## 📝 Description

- add size to modal with reference to theme.sizes.

modal is able to set size until ‘6xl’. 

## 🚀 New behavior

in theme.sizes, there are the size until ‘8xl’.

No.

### DIFF
--- a/.changeset/serious-panthers-brake.md
+++ b/.changeset/serious-panthers-brake.md
@@ -1,6 +1,6 @@
 ---
-"@chakra-ui/modal": major
-"@chakra-ui/theme": major
+"@chakra-ui/modal": minor
+"@chakra-ui/theme": minor
 ---
 
 add size to modal with reference to theme.sizes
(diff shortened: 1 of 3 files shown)


.changeset/serious-panthers-brake.md
```diff
@@ -0,0 +1,6 @@
+---
+"@chakra-ui/modal": major
+"@chakra-ui/theme": major
```
segunadebayo, line 3:
```suggestion
"@chakra-ui/modal": minor
"@chakra-ui/theme": minor
```

segunadebayo:
Does this PR solve any inherent problem that you have?

Ryuta1346:
@segunadebayo 
I think so.

segunadebayo:
I'm not convinced we should add more options just because we can.

I'd much rather wait till it's a pressing need before adding it. I'll revive this PR when the time is right.

I appreciate the work here, though.

PR #6182, opened 2022-06-11
Apply the same constraint to this type parameter as the place it's used

RyanCavanaugh:
Mitigates an upcoming correctness improvement in TypeScript 4.8

## 📝 Description

TypeScript 4.8 has improved checking around unconstrained type parameters and now correctly identifies potentially-unsafe instantiations of generic types with `null` and `undefined`

See https://github.com/microsoft/TypeScript/issues/49489

This code will not build in TypeScript 4.8

## 🚀 New behavior

Now it will

Sort of? Code that invoked this with `null` or `undefined` will now correctly see an error, but this code presumably already crashed

### DIFF
--- a/packages/descendant/src/use-descendant.ts
+++ b/packages/descendant/src/use-descendant.ts
@@ -7,7 +7,7 @@ import { useSafeLayoutEffect, cast } from "./utils"
  * @internal
  * React hook that initializes the DescendantsManager
  */
-function useDescendants<T extends HTMLElement = HTMLElement, K = {}>() {
+function useDescendants<T extends HTMLElement = HTMLElement, K extends Record<string, any> = {}>() {
   const descendants = useRef(new DescendantsManager<T, K>())
   useSafeLayoutEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/descendant/src/use-descendant.ts
```diff
@@ -7,7 +7,7 @@ import { useSafeLayoutEffect, cast } from "./utils"
  * @internal
  * React hook that initializes the DescendantsManager
  */
-function useDescendants<T extends HTMLElement = HTMLElement, K = {}>() {
+function useDescendants<T extends HTMLElement = HTMLElement, K extends Record<string, any> = {}>() {
   const descendants = useRef(new DescendantsManager<T, K>())
```
RyanCavanaugh, line 11: Context: the type `DescendantsManager<T, K>` has a constraint of `Record<string, any>` on `K`, so cannot be instantiated with `null` or `undefined`, but `useDescendants` *can* be instantiated with those types in that position, hence there's a potential bug here without this PR in place.

PR #6190, opened 2022-06-14
fix(avatar): __css prop does not work on Avatar #6188

### LINKED ISSUE #6188: __css prop does not work on Avatar
### Description

When I use the `__css` prop, I expected to be able to change the style of the component but nothing happens

### Link to Reproduction

https://codesandbox.io/s/avatar-css-vcx7tp?file=/src/app.tsx

### Steps to reproduce

1. Add `__css` prop to an `Avatar` component to change the style
2. See that style is not applied
3. Use the same CSS object in the `sx` prop
4. See that style is applied

### Chakra UI Version

2.1.0

### Browser

any

### Operating System

any

### Additional Information

`sx` prop works fine

adrivelasco:
Closes #6188 

## 📝 Description

`__css` prop doesn't work on Avatar since it's not being properly passed down to component. This PR gets `__css` working by spreading its value into **avatarStyles** variable.

`__css` prop doesn't work on **Avatar** component since its being ignored

## 🚀 New behavior

`__css` prop works as expected on **Avatar** component just as `sx` prop does.

No

### DIFF
--- a/.changeset/lucky-tigers-juggle.md
+++ b/.changeset/lucky-tigers-juggle.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/avatar": patch
+---
+
+\_\_css prop doesn't work on Avatar since it's not being properly passed down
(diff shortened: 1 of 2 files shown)


segunadebayo:
Hi @adrivelasco. This was an intentional decision to only consume `__css` within the components and not expose it to the consumers.

Kindly use `sx` prop or spread props for overrides.

Thanks for understanding

PR #6214, opened 2022-06-19
feat: replace global with local css reset

TimKolberger:
## 📝 Description

Early DRAFT!
This PR replaces the global CSS reset with a local, element specific one.

Ideas and feedback welcome :)

The CSS reset uses by Chakra UI clashes with other CSS frameworks on the site and thus makes it hard to migrate older projects onto Chakra UI.

You can pass `resetCSS={false}` to the ChakraProvider to disable the global CSS reset, but it won't look good with the default theme.

## 🚀 New behavior

The chakra factory looks up the element to render, and applies the CSS reset for this element.

No, but we can remove the global CSS reset definitions in the next major.

## 📝 Additional Information

- [ ] needs manual testing in different browsers - maybe a storybook story with all elements?


segunadebayo:
Thanks for working on this @TimKolberger. This approach duplicates the reset style objects across all components which negatively impacts the performance (slightly).

Let's consider using CSS `:where` or `:is` selectors combined with an attached class to all chakra components. I'll have a PR ready for this shortly.

PR #6241, opened 2022-06-24
WIP: Adding the Chakrathon winnig select to Chakra-UI

### LINKED ISSUE #4311: Custom Dropdown/Select
### Description

Create a custom, [keyboard compliant](https://www.w3.org/WAI/perspective-videos/keyboard/) dropdown/select component with search capabilities.

- Most designs and apps need a more dynamic `select` component, such as `react-select`.
- I needed to create a custom `Select` component, while using Chakra-UI's infra and system utils, hooks. Referencing the `Menu` component, I added optional search to filter out the listed options, the `DescendantsContext` for keyboard focus management, along with custom `SelectOptions` for rending labels and optional icons, etc.

### Proposed Solution or API

I found the `react-select` format much more versatile in terms of controlling and rendering data.

`Select` will render and array of `OptionTypes`.
```js
export type OptionType = {
  icon?: string;
  label: string;
  value: string; 
}
```

Example usage:
```js
function SelectUsage() {
  const [state, setState] = React.useState(null)
  const options = [
    { label: 'Green', value: 'green' },
    { label: 'Green-Yellow', value: 'greenyellow' },
    { label: 'Red', value: 'red' },
    { label: 'Violet', value: 'violet' },
    { label: 'Forest', value: 'forest' },
    { label: 'Tangerine', value: 'tangerine' },
    { label: 'Blush', value: 'blush' },
    { label: 'Purple', value: 'purple' },
  ]
  return (
    <>
      <Select
        isSearchable
        defaultValue={options[0].value}
        value={state}
        options={options}
        onChange={(value) => setState(value)}
      />
    </>
  )
}
```
- container `SelectDescendantsProvider`, `SelectProvider` & `StylesProvider`
- `Select` consumes four internal components, each with its relative context.
```
<SelectContainer />
  <SelectInputValue />
<SelectList />
   <SelectOption />
```

### Alternatives

- Structure: maybe moving the components out of the context, leaving the developers more control of the general structure of the elements? I found this be more difficult in terms of controlling data.
```
<Select>
  <SelectContainer>
    <SelectInputValue /> // handles current state, and search state
  </SelectContainer>
  <SelectList>
    <SelectOption label="Green" value="green" />
    <SelectOption label="Red" value="red" />
    <SelectOption label="Yellow" value="yellow" />
    <SelectOption label="Blue" value="blue" />
    <SelectOption label="Black" value="black" />
  </SelectList>
</Select>
```

### Additional Information
- current component development

https://user-images.githubusercontent.com/19299064/124046312-e5fd3880-d9ce-11eb-8938-25b946a9cfd9.mov

MasterGordon:
Closes #4311

## 📝 Description

This PR adds the Chakrathon winning select of team Dragonball to Chakra-UI
https://github.com/rpehlivan1/chakrathon/

I have change the code quite a bit to make it more chakra-ui like:

- stripped some logic to use prebuild chakra-ui hooks
- removed to custom contexts to use the ChakraUI ones
- Improved to styling to support darkmode
- Refactored stories
- Fixed linting issues
- Trimmed down some component code

## 🚀 New behavior

- added new Select Component
- Renamed old Select component to SimpleSelect

## 💣 Is this a breaking change Yes:

The Old Select has been renamed to `SimpleSelect`

## 📝 Additional Information / TODOs

- Add Tests
- Added Prop documentation
- Improve file structure
- A lot more testing


packages/theme/src/components/select.ts
```diff
@@ -1,70 +1,135 @@
...
+      borderWidth: "1px",
+      borderStyle: "solid",
+      height: "100%",
+      _disabled: {
+        opacity: 0.4,
+        cursor: "not-allowed",
+      },
+      ...(invalid && {
+        borderColor: mode("red.500", "red.300")(props),
+        boxShadow: `0 0 0 1px ${getColor(
```
MasterGordon, line 42: Is there an alternative not to use deprecated getColor?

rome2o:
Do we have any ETA on this?

dbousamra:
> Do we have any ETA on this?

I feel like there are hundreds of us waiting on a custom Select component. Super keen to see this, and a Multiselect & Datepicker in Chakra.

jangroeneveld:
Is there any update on this? I would really like to incorporate this in a project

packages/theme/src/components/select.ts
leomelzer: @MasterGordon maybe you looked at this already, but you can leave out the color value and it'll default to `currentColor`, hence you can set the `color` property directly on the selector as a theme token (`color: "red.500"`).

segunadebayo:
A quick update on the Select component. I've started some work to model the logic of the select component here:
https://github.com/chakra-ui/zag/blob/main/packages/machines/select/src/select.machine.ts

This will land in Chakra shortly after testing and documentation are done.

> Once this is shipped, we'll release the multi-select version next.

BrickheadJohnny:
Hello @segunadebayo, just bumping this up. 😉  I wonder if there's any ETA on when could we see the Zag based (?) combobox/select in Chakra UI? 👀 We're using chakra-react-select for now, but it certainly isn't perfect and a simpler solution would be perfect for us.

csandman:
> We're using chakra-react-select for now, but it certainly isn't perfect and a simpler solution would be perfect for us.

As the creator of `chakra-react-select`... ouch 😂

I completely agree though, I made it originally to fill a need that I had because it was missing from the original package. I did the best I could to make it as customizable as possible, but at the end of the day, it's a big package and the architecture for implementing it really doesn't follow the Chakra pattern. I'll be happy to (mostly) retire is when Chakra finally has a single _and_ multi select in place.

eshwaruppala:
@segunadebayo any update on this?

PR #6290, opened 2022-07-09
fix(border.ts): create new tag br for borderRadius

LucasLuisBorges:
I use chakra a lot and I realized that with the br tag for borderRadius, it would be much simpler.

Just updated so it can be passed just br instead of borderRadius.


segunadebayo:
Hey @LucasLuisBorges,

Thanks for creating this. We already have an alias for border radius called "radius". I think this is sufficient

We don't want to create more ways to do the same thing.

Thanks for understanding

PR #6301, opened 2022-07-11
fix(utils): breakpoint calculation

### LINKED ISSUE #6269: useBreakpointValue returns unexpected result between breakpoints
### Description

When I have a screen size somewhere between lg and xl, I expect useBreakpointValue to express my current breakpoint as "lg" or "xl"  but instead it returns "base".

### Link to Reproduction

https://codesandbox.io/s/laughing-phoebe-i3wybo?file=/src/index.tsx

### Steps to reproduce

1. Open the CodeSandbox example in a new full window
2. Open developer tools and set your screen width to 1279px.
3. Instead of "large" being printed to the screen, "base" is printed instead. If you reduce/increase the screen size by even just one pixel the correct value is printed to the screen.

### Chakra UI Version

2.2.1

### Browser

Google Chrome 103.0.5060.66

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

It's as if the default breakpoints for `useBreakpointValue` are exclusive and not inclusive to the actual pixel width for a breakpoint. If `1279px` is the start of the screen size for "xl" then I'd expect the breakpoint value to be "xl" at `1279px`.

I should also mention this is happening with a default set of breakpoint values. I have not extended the breakpoint values.

effectivetom:
Closes #6269

## 📝 Description

Fixes breakpoint calculation which was causing certain viewport widths to be excluded from media query coverage.

Media queries which are generated from theme breakpoints do not cover all possible pixel widths.

## 🚀 New behavior

Media queries cover all possible viewport widths.

No

### DIFF
--- a/.changeset/selfish-bees-kneel.md
+++ b/.changeset/selfish-bees-kneel.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/styled-system": patch
+"@chakra-ui/utils": patch
+---
+
+Fix incorrect breakpoint calculation which excluded some viewport widths from
+media query coverage
--- a/packages/styled-system/tests/style-config.test.ts
+++ b/packages/styled-system/tests/style-config.test.ts
@@ -4,7 +4,7 @@ test("should handle single variant", () => {
   expect(recipe({ theme, variant: "solid", size: ["sm", "md"] }))
     .toMatchInlineSnapshot(`
     Object {
-      "@media screen and (min-width: 0px) and (max-width: 39.9365em)": Object {
+      "@media screen and (min-width: 0px) and (max-width: 39.9375em)": Object {
         "fontSize": "sm",
         "h": 8,
         "minW": 8,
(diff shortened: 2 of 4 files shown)


effectivetom:
I just tested this by installing the 2.0.4 release and the issue is still occurring. After looking at my `yarn.lock` file, I think this might be because `@chakra-ui/react-utils` still depends on version 2.0.2 of `@chakra-ui/utils`.

Is this something I should have addressed in my PR, or is there another process for ensuring that other packages also get versioned?

ShawnFumo:
@effectivetom Should this be working now? I upgraded all the chakra stuff and see that react-utils is depending on utils 2.0.11. If I search the built code in devtools sources tab, I only see -0.0625 coming back. But still am seeing strange stuff happen at 1535px and 1279px widths. At those exact widths, some components are still switching to mobile mode. Any ideas?

effectivetom:
@ShawnFumo see #6743 and #6744 which are far better researched and implemented than this fix.

Issue #5255, opened 2021-12-17
RangeSlider's RangeSliderThumb is rendered incorrectly when within Accordion


dufia:
### Description

When you place RangeSlider in the Accordion the RangeSliderThumb is rendered too much to the right

### Link to Reproduction

https://codesandbox.io/s/homepage-forked-3v00t?file=/src/App.tsx:507-523

### Steps to reproduce

1. Go to [link](https://codesandbox.io/s/homepage-forked-3v00t?file=/src/App.tsx:507-523)
2. If you look **very** closely you can see that on the second RangeSlider, RangeSliderThumb are offset to the right a little bit

<img width="716" alt="Screenshot 2021-12-17 at 18 05 37" src="https://user-images.githubusercontent.com/5569649/146582262-7de1e4f1-f965-46ef-9010-37de7bc9a735.png">

### Chakra UI Version

1.7.3

### Browser

Chrome 96.0.4664.93

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

primos63:
If you align the standalone `RangeSlider` with the one in the `AccordionPanel` you'll see that the left and right thumb are to the right of where those in the standalone slider are. This is because of the calculation that is done to position the thumbs relies on the `offsetWidth` of the slider thumb. Before the panel is expanded the `offsetWidth` is 0 and this causes them to be placed farther to the right.

This only works when the `Accordion`'s internal state prop, `isExpanded`, is used. The `RangeSlider` needs to be visible to be able to get the offsetWidth.
```jsx
  <Accordion allowToggle={true} borderX="1px" borderTop="1px">
    <AccordionItem>
      {({ isExpanded }) => (
        <>
          <AccordionButton>
            <Box flex="1" textAlign="left">
              Click to expand
            </Box>
            <AccordionIcon />
          </AccordionButton>
          <AccordionPanel bg="cyan.100">
            {isExpanded ? <CustRangeSlider /> : null}
          </AccordionPanel>
        </>
      )}
    </AccordionItem>
  </Accordion>
```

My [CodeSandbox example](https://codesandbox.io/s/homepage-forked-uy0ci?file=/src/App.tsx) shows the RangeSlider with my proposed change.

dufia:
Thanks @primos63

dufia:
@segunadebayo was it fixed? Cheers

dufia:
@segunadebayo?

jdddog:
[...]

PR #6360, opened 2022-07-22
Update thumbRects based on latest thumbs width and height

jdddog:
Closes https://github.com/chakra-ui/chakra-ui/issues/5255

## 📝 Description

Uses @primos63's solution to fix the RangerSliderThumb rendering too far to the right when the RangeSlider is placed in an Accordion that is not expanded by default.

When the Accordion is not expanded by default, the offsetWidth and offsetHeight of the range slider thumbs are 0 and they are not updated when the AccordionPanel opens. This means that the RangeSliderThumb is shifted to the right as the calculation for positioning the RangeSliderThumb requires the width and height each thumb.

## 🚀 New behavior

Update thumbRects until the width and height matches the width and height of the thumbs.

No

## 📝 Additional Information
I'm not sure if there is a better way subscribe to changes in the DOM when the slider thumbs' offsetWidth and offsetHeight change?

### DIFF
--- a/.changeset/popular-ants-impress.md
+++ b/.changeset/popular-ants-impress.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/slider": minor
+---
+
+Fix RangerSliderThumb rendering too far to the right when the RangeSlider is
+placed in an Accordion that is not expanded by default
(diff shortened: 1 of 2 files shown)


segunadebayo:
Closing in favor of #6472

Thanks for taking a stab at this @jdddog

jdddog:
> Closing in favor of #6472
> 
> Thanks for taking a stab at this @jdddog

No worries, thanks for fixing it @segunadebayo.

PR #6366, opened 2022-07-23
fix(modal): Add wrapper to make centered modal scrollable

akira02:
## 📝 Description
[According to docs](https://chakra-ui.com/docs/components/modal/usage#make-modal-vertically-centered), if the content within the modal overflows beyond the viewport, using this prop will make it unscrollable.

But actually can add a wrapper using `min-height: 100vh` under fixed div. This will make the content centered, but also scrollable when viewport smaller than content.

Unscrollable when having `isCentered` prop.

Jul-23-2022 16-49-33

## 🚀 New behavior
Scrollable when having `isCentered` prop.

Jul-23-2022 16-44-58

No.

### DIFF
--- a/.changeset/two-suits-learn.md
+++ b/.changeset/two-suits-learn.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/modal": minor
+---
+
+Add wrapper to make centered modal scrollable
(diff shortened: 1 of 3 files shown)


mohammadizanloo55:
see an example of this bug : 
https://codesandbox.io/s/exciting-vaughan-ehy0c3?file=/src/App.tsx

when you click on the "normal Modal" button everything is ok but in "centered Modal" has a bug in the scroll

segunadebayo:
Hi @akira02,

This is considered a breaking change to the structure of the modal's HTML. we'll consider adding it to the next major release.

Thanks for working on this.

Issue #3696, opened 2021-03-30
Is it possible to turn off motion/animation for all components?


coodoo:
Currently there's `motionPreset` for `Modal`, it would be great if there's a global switch to turn off all motion/animation effects for every component, any thought?

vickOnRails:
Hey there @coodoo 👋🏼

First off, I didn't find a perfect way to turn off animations on all components. 

But I took a little time to look further. I've found there are 2 ways you might want to go about this
- `usePrefersReducedMotion` custom hook ([here](https://chakra-ui.com/docs/hooks/use-prefers-reduced-motion))
- CSS [Reduced Motion](https://css-tricks.com/introduction-reduced-motion-media-query/) Media Query

The first approach would involve you obtaining and storing the user's motion preference at some top-level place in your app.

```jsx
import {  usePrefersReducedMotion } from "@chakra-ui/react";

export default App() {
  // You can then store this value in some state or localstorage
  const prefersMotion = usePrefersReducedMotion();
}
```

You'll then have to use it in specific components that allow you to turn on/off motion (Like in your modal example)

```jsx

import React from "react";
import ReactDOM from "react-dom";
import {
  Button,
  Modal,
  ChakraProvider,
  usePrefersReducedMotion
} from "@chakra-ui/react";

export function App() {
  const prefersReducedMotion= usePrefersReducedMotion();

  const { isOpen, onOpen, onClose } = useDisclosure();
  return (
    <ChakraProvider>
      <Button onClick={onOpen}>Trigger modal</Button>

      <Modal
        onClose={onClose}
        isOpen={isOpen}
        isCentered
        motionPreset={prefersReducedMotion ? "none" : "slideInBottom"}
      >
        {...ModalContent}
      </Modal>
    </ChakraProvider>
  );
}
```
I've put together a working demo [here](https://6913w.csb.app/)

> You'll have to turn off/on motion on your OS to test

The only downside to this approach is that you can't turn it off globally (Not that I know of)
--- 

- The second approach I thought of didn't work. It would involve using the native CSS [prefers motion](https://css-tricks.com/introduction-reduced-motion-media-query/) query and turning off transitions on every element on the page. 

```css
@media screen and (prefers-reduced-motion) {
  * {
    /* 🤷🏼‍♀️ */
    transition: none !important;
  }
}
```

I'll update this issue as I learn more. Or @segunadebayo, is there a way around this already?

segunadebayo:
[...]

PR #6368, opened 2022-07-24
feature: introduce Feature Composing System for reducing bundle size

yukukotani:
Related to #4975, #3696

## 📝 Description

This PR is a proposal for the Feature Composing System for reducing bundle size. This is just a rough PoC implementation to get reviews about the approach.

The concept is like dependency injection. Features directly depend on external large libraries such as framer-motion. Child components refer to it via Context so that they don't need to have large dependencies directly.

Current `ChakraProvider` is still fully featured, so users can start with it,  then optimize by replacing it with `ChakraComposingProvider` on they need it.

### Case 1:

With the code below, there is no animation but `framer-motion` will be included into bundle, because `ChakraProvider` contains `ToastProvider` that depends on `framer-motion`.

```tsx
import {
  Box,
  ChakraProvider,
} from "@chakra-ui/react";
import  { FC, useEffect } from "react";
import ReactDOM from "react-dom/client";

const App: FC = () => {
  return (
    <Box>foo</Box>
  );
};

ReactDOM.createRoot(document.getElementById("root")!).render(
  <ChakraProvider>
    <App />
  </ChakraProvider>
);
```

### Case 2:

With the code below, `framer-motion` will be included into bundle, because `CheckBox` depends on `framer-motion`.

```tsx
import {
  Box,
  Checkbox,
  ChakraProvider,
} from "@chakra-ui/react";
import  { FC, useEffect } from "react";
import ReactDOM from "react-dom/client";

const App: FC = () => {
  return (
    <Box>
      <Checkbox>foo</Checkbox>
    </Box>
  );
};

ReactDOM.createRoot(document.getElementById("root")!).render(
  <ChakraProvider>
    <App />
  </ChakraProvider>
);
```

## 🚀 New behavior

### Case 1

With the code below, `framer-motion` will be included into bundle, because `createToastFeature` contains `ToastProvider` that depends on `framer-motion`.

Once you removed `toastFeature`, `framer-motion` will be removed by tree-shaking and `useToast` will stop working.

```tsx
import {
  Box,
  ChakraComposableProvider,
  createToastFeature,
  useToast,
} from "@chakra-ui/react";
import  { FC, useEffect } from "react";
import ReactDOM from "react-dom/client";

const App: FC = () => {
  const toast = useToast();

  useEffect(() => {
    toast();
  }, []);

  return (
    <Box>foo</Box>
  );
};

const toastFeature = createToastFeature();

ReactDOM.createRoot(document.getElementById("root")!).render(
  <ChakraComposableProvider features={[toastFeature]}>
    <App />
  </ChakraComposableProvider>
);
```

### Case 2

With the code below, `framer-motion` will be included into bundle, because `createAnimationFeature` depends on `framer-motion`.

Once you removed `animationFeature`, `framer-motion` will be removed by tree-shaking but `Checkbox` will stop working.

```tsx
import {
  Box,
  Checkbox,
  ChakraComposableProvider,
  createAnimationFeature,
} from "@chakra-ui/react";
import  { FC, useEffect } from "react";
import ReactDOM from "react-dom/client";

const App: FC = () => {
  return (
    <Box>
      <Checkbox>foo</Checkbox>
    </Box>
  );
};

const animationFeature = createAnimationFeature();

ReactDOM.createRoot(document.getElementById("root")!).render(
  <ChakraComposableProvider features={[animationFeature]}>
    <App />
  </ChakraComposableProvider>
);
```

No

### Remaining points

- Naming of Feature Composing System (since i'm not English native)
- Package structures

### DIFF
--- a/packages/checkbox/package.json
+++ b/packages/checkbox/package.json
@@ -50,6 +50,7 @@
     "react": "^18.0.0"
   },
   "peerDependencies": {
+    "@chakra-ui/react": ">=2.0.0-next.0",
     "@chakra-ui/system": ">=2.0.0-next.0",
     "framer-motion": ">=4.0.0",
     "react": ">=18"
--- a/packages/checkbox/package.json
+++ b/packages/checkbox/package.json
@@ -45,6 +45,7 @@
     "@zag-js/focus-visible": "0.1.0"
   },
   "devDependencies": {
+    "@chakra-ui/react": "2.2.4",
     "@chakra-ui/system": "2.2.2",
     "framer-motion": "^6.2.9",
     "react": "^18.0.0"
--- a/packages/checkbox/src/checkbox-icon.tsx
+++ b/packages/checkbox/src/checkbox-icon.tsx
@@ -1,19 +1,8 @@
-import { chakra, PropsOf, ChakraComponent } from "@chakra-ui/system"
-import { AnimatePresence, CustomDomComponent, motion } from "framer-motion"
+import { chakra, PropsOf } from "@chakra-ui/system"
+import { AnimatePresence, motion } from "framer-motion"
 import * as React from "react"
 
-function __motion<T extends ChakraComponent<any, any>>(
-  el: T,
-): CustomDomComponent<PropsOf<T>> {
-  const m = motion as any
-  if ("custom" in m && typeof m.custom === "function") {
-    return m.custom(el)
-  }
-  return m(el)
-}
-
-// @future: only call `motion(chakra.svg)` when we drop framer-motion v3 support
-const MotionSvg = __motion(chakra.svg)
+const MotionSvg = motion(chakra.svg)
 
 const CheckIcon = (props: PropsOf<typeof MotionSvg>) => (
   <MotionSvg
--- a/packages/checkbox/src/checkbox-icon.tsx
+++ b/packages/checkbox/src/checkbox-icon.tsx
@@ -1,86 +1,126 @@
+import { useFramerMotion } from "@chakra-ui/react"
 import { chakra, PropsOf } from "@chakra-ui/system"
-import { AnimatePresence, motion } from "framer-motion"
+import { motion } from "framer-motion"
 import * as React from "react"
 
-const MotionSvg = motion(chakra.svg)
+function getMotionSvg() {
+  return motion(chakra.svg)
+}
 
-const CheckIcon = (props: PropsOf<typeof MotionSvg>) => (
-  <MotionSvg
-    width="1.2em"
-    viewBox="0 0 12 10"
-    variants={{
-      unchecked: {
-        opacity: 0,
-        strokeDashoffset: 16,
-      },
-      checked: {
-        opacity: 1,
-        strokeDashoffset: 0,
-        transition: { duration: 0.2 },
-      },
-    }}
-    style={{
-      fill: "none",
-      strokeWidth: 2,
-      stroke: "currentColor",
-      strokeDasharray: 16,
-    }}
-    {...props}
-  >
-    <polyline points="1.5 6 4.5 9 10.5 1" />
-  </MotionSvg>
-)
+function useMotionSvg(): ReturnType<typeof getMotionSvg> {
+  const framerMotion = useFramerMotion()
 
-const IndeterminateIcon = (props: PropsOf<typeof MotionSvg>) => (
-  <MotionSvg
-    width="1.2em"
-    viewBox="0 0 24 24"
-    variants={{
-      unchecked: {
-        scaleX: 0.65,
-        opacity: 0,
-      },
-      checked: {
-        scaleX: 1,
-        opacity: 1,
-        transition: {
-          scaleX: { duration: 0 },
-          opacity: { duration: 0.02 },
+  if (!framerMotion) {
+    throw new Error(
+      "You need to add animationFeature to use Checkbox component",
+    )
+  }
+
+  return framerMotion.motion(chakra.svg)
+}
+
+function useAnimatePresence() {
+  const framerMotion = useFramerMotion()
+
+  if (!framerMotion) {
+    throw new Error(
+      "You need to add animationFeature to use Checkbox component",
+    )
+  }
+
+  return framerMotion.AnimatePresence
+}
+
+const CheckIcon = (props: PropsOf<ReturnType<typeof useMotionSvg>>) => {
+  const MotionSvg = useMotionSvg()
+
+  return (
+    <MotionSvg
+      width="1.2em"
+      viewBox="0 0 12 10"
+      variants={{
+        unchecked: {
+          opacity: 0,
+          strokeDashoffset: 16,
         },
-      },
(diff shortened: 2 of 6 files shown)


packages/checkbox/package.json
```diff
@@ -45,11 +45,13 @@
     "@zag-js/focus-visible": "0.1.0"
   },
   "devDependencies": {
+    "@chakra-ui/react": "2.2.2",
```
yukukotani, line 48: I know this is not good. Will be fixed after deciding package structure.

packages/checkbox/src/checkbox-icon.tsx
```diff
@@ -1,97 +1,121 @@
...
+      "You need to add animationFeature to use Checkbox component",
+    )
   }
-  return m(el)
+
+  return framerMotion.motion(chakra.svg)
 }
 
-// @future: only call `motion(chakra.svg)` when we drop framer-motion v3 support
-const MotionSvg = __motion(chakra.svg)
```
yukukotani, line 16: framer-motion v3 support is already dropped

packages/checkbox/src/checkbox-icon.tsx
```diff
@@ -1,97 +1,121 @@
...
+function useAnimatePresence() {
+  const framerMotion = useFramerMotion()
+
+  if (!framerMotion) {
+    throw new Error(
+      "You need to add animationFeature to use Checkbox component",
+    )
+  }
```
yukukotani, line 24: I think we can fallback to the component without animation

packages/checkbox/package.json
```diff
@@ -45,11 +45,13 @@
     "@zag-js/focus-visible": "0.1.0"
   },
   "devDependencies": {
+    "@chakra-ui/react": "2.2.4",
```
yukukotani, line 48: I know this is not good. Will be fixed after deciding package structure.

packages/react/src/chakra-composable-provider.tsx
```diff
@@ -0,0 +1,56 @@
...
+export function createToastFeature(
+  toastOptions?: ToastProviderProps,
+): ChakraFeature {
+  return {
+    id: "toast",
+    Provider(props) {
+      return <ToastProvider {...props} {...toastOptions} />
+    },
+  }
+}
```
yukukotani, line 25: This is an independent function so that the reference to `ToastProvider` can be dropped by tree-shaking.

yukukotani:
@segunadebayo  Hi, what do you mean by draft? Should I do something additional to get reviewed or just wait?

segunadebayo:
Hi @yukukotani. Based on your comment, "This is just a rough PoC".

There are a lot of breaking changes this will introduce and I want to give you the chance to prove your ideas before reviewing them.

yukukotani:
@segunadebayo Thanks, understood.

I think we can introduce this without any breaking changes but anyway the PR is ready for getting feedback about my idea!

TheThirdRace:
Just a general feedback, I think it's a step in the good direction, but it won't solve the root problem.

Framer-Motion is a big 30 KB gzipped monolith. Given what Chakra-UI uses it for, it could be replaced by a `div` and 2 css classes, aka, 100 bytes gzipped. Chakra should never have adopted a full-on animation framework, they're a library and as such they should not dictate the rest of our architecture...

Having your component breaks without implementing animation in the provider might not be the best option. Anything that is not required on load should be lazy-loaded. I do not need an animation on checkboxes when my page load. I might need one on hover, focus or click, but that's what dynamic imports are for.

The goal is to avoid loading a gigantic amount of JS on first load, which Chakra is crippled by at the moment. You can't tree-shake part of Framer-Motion, which means they would dynamically import 30 KB gzipped of code when lazy-loaded. There are only 3 solutions to this problem:

1. Framer-Motion magically release a tree-shakeable version that would allow less than 5 KB gzipped if you keep to simple animations
2. Change the animation framework to something more palatable and tree-shakeable
3. Replace the animations with 1 `div` and 2 css classes for less than 100 bytes gzipped

The solution `#1` won't happen in at least the next 2 years... let's stop dreaming about it...

Both solutions `#2` and `#3` have the same thing in common: the complete removal of Framer-Motion. There is not really any way around it.

Your solution could be a intermediary step for the time being, but as I said, it won't solve the root cause of the problem. A 30KB gzipped monolith is simply too big of a liability and it will always affect performance no matter how or when you load it.

yukukotani:
@TheThirdRace Thank you for the feedback!

First, I totally agree with your opinion that Framer-Motion is the root problem and we should remove that. 

But we will use react-transition-group or something like that to replace Framer-Motion and it's needless for users who don't need animations. They don't want to load animation-related codes even if it's lazy-loaded. So Feature Composing System will still be effective for reducing bundle size after Framer-Motion is wholly removed.

Also, global components such as Toast are bundled regardless of whether it's used or not. Feature Composing System can resolve that.

> Having your component breaks without implementing animation in the provider might not be the best option. 

Yes. We can fall back to the non-animated component if animation is not implemented in the provider.

So the summary of my opinion is:

1. Framer-Motion should be removed.
2. Use Feature Composing System to avoid loading animation if the user doesn't need it.
3. Lazy-load animation if the user needs it.

segunadebayo:
Yes, I agree with the conclusions above. Removing framer-motion is a priority for me in the next major release. I believe we can design the components to fit into any animation library or native CSS animations, and I have ideas around this.

Thanks for taking the time to work on this @yukukotani.

mattgperry:
@TheThirdRace Framer Motion does allow lazy loading animations https://www.framer.com/docs/guide-reduce-bundle-size/#async-loading

So I can better understand the problem (as reducing bundle size is something I'm actively thinking about) - what is it about the above approach that falls short?

TheThirdRace:
@mattgperry 

It's not as much the lazy-loading per se than the actual bundle size.

The way Chakra-UI uses Framer-Motion, the package adds `31.97 KB` gzipped.

By itself, this is already way too big. For comparison sake, in a typical app, `react-dom` would be the only package bigger than Framer-Motion at a whooping `38 KB` gzipped. Nothing else come close to that size.

There have been attempts to use lazy-loading in Chakra-UI in the past. I know of at least 1 PR I followed back in the days that did everything possible to reduce Framer-Motion bundle size and came up at `25 KB` gzipped. I'm paraphrasing, but the conclusion was that everything eventually links to the `LazyMotion` component, meaning there was no further tree-shaking possible... While the saving were better than nothing, the resulting bundle size was still at least `10 KB` gzipped bigger than pretty much any other common package in a typical app.

It's very hard to justify a 25 to 32 KB gzipped dependency when you know you can replace most animations with 2 css classes and a prop.

I want to point out I'm not denying the hard work made on Framer-Motion. It's actually an awesome library and if I needed an animation framework , it would definitely be at the top of the list. It's just that in the context of Chakra-UI, it's not a good design decision for them to decide for us what's gonna be our animation framework for our whole app. Their use of animations is very simple and minimal, there's no point in using a nuclear strike to squash a fly 😅

PR #6380, opened 2022-07-26
Potentially hacky fix for `thumbRects[i] is undefined`

### LINKED ISSUE #6375: `TypeError: thumbRects[i] is undefined` when dynamically adding points to `RangeSlider`
### Description

I am receiving an error (`TypeError: thumbRects[i] is undefined`) while trying to dynamically add thumbs, when using the `RangeSlider` component.

### Link to Reproduction

https://codesandbox.io/s/zealous-fire-jhhugv

### Steps to reproduce

Here is a minimal component that demonstrates the issue. It starts with 2 thumbs and then a 3rd one gets added after one second:

```typescript
export function HomePicker() {
    const [points, setPoints] = useState<number[]>([30, 70]);
    useEffect(() => {
        setTimeout(() => {
            setPoints([ 30, 50, 70 ]);
        }, 1000)
    }, []);
    return (
        <RangeSlider value={points}>
            <RangeSliderTrack>
                <RangeSliderFilledTrack />
            </RangeSliderTrack>
            {points.map((_p, index) => (
                <RangeSliderThumb key={index} index={index} />
            ))}
        </RangeSlider>
    );
}
```

The component renders properly at first, but after adding the 3rd point in the setTimeout, I get this error:

Screen Shot 2022-07-26 at 1 44 12 AM
 

### Chakra UI Version

2.2.4

### Browser

Firefox 102.0.1

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

TooTallNate:
Closes #6375.

## 📝 Description

This fixes the linked issue by falling back to using the size of the
first "thumb" when calculating the thumbs styles in the case where the
desired index is not present.

This is probably not a fully "correct" fix, in the case where thumbs are
different sizes, but this does fix my issue even if it is hacky. If
there is a more _correct_ solution then that would be even better.

When trying to dynamically add new "thumbs" to a `RangeSlider` after
intial instantiation, the error `thumbRects[i] is undefined` is thrown.

## 🚀 New behavior

Error is no longer thrown.

No

## 📝 Additional Information

### DIFF
--- a/packages/slider/src/slider-utils.ts
+++ b/packages/slider/src/slider-utils.ts
@@ -45,10 +45,14 @@ export function getStyles(options: {
     ...orient({
       orientation,
       vertical: {
-        bottom: `calc(${thumbPercents[i]}% - ${thumbRects[i].height / 2}px)`,
+        bottom: `calc(${thumbPercents[i]}% - ${
+          (thumbRects[i] || thumbRects[0]).height / 2
+        }px)`,
       },
       horizontal: {
-        left: `calc(${thumbPercents[i]}% - ${thumbRects[i].width / 2}px)`,
+        left: `calc(${thumbPercents[i]}% - ${
+          (thumbRects[i] || thumbRects[0]).width / 2
+        }px)`,
       },
     }),
   })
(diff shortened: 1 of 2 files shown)


TooTallNate:
@segunadebayo Did you have any thoughts on this one?

segunadebayo:
Hey @TooTallNate,

I'll get on this today. I think we need a better fix for the size measurements instead of assuming the size of the first thumb.

segunadebayo:
Closing in favor of #6472

Thanks for taking a stab at this @TooTallNate 💖

PR #6402, opened 2022-08-01
fix(media-query): avoid useMediaQuery re-renders

### LINKED ISSUE #6401: useBreakpointValue causes re-renders
### Description

When using useBreakpointValue you get multiple re-renders on first render, even when the value does not change.

### Link to Reproduction

https://codesandbox.io/s/usebreakpointvalue-renders-vg2khe

This basic example sees 3 renders on first render, and any resizing between breakpoints causes 2 renders. I've removed the code for running in strictmode since that would be misleading in counting the number of renders. I've also used React devtools to verify that it renders again because "Hook 1 changed", which is the `useState()` inside `useMediaQuery()`. It also renders a third time because "context changed" which I'm not entirely sure why?

### Steps to reproduce

1. Use `useBreakpointValue()` in a component
2. This component will re-render at least once on initial load

### Chakra UI Version

@chakra-ui/react@2.2.4

### Browser

Chrome Version 103.0.5060.134 (Official Build) (64-bit)

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

I'd say the main problem is that any component using the `useBreakpointValue()` hook will render at least twice. This is somewhat made worse by the fact that the first of these two renders always uses the fallback value, even when running purely in browser.

This is essentially a re-open of https://github.com/chakra-ui/chakra-ui/issues/4267

### Workaround

I think the workaround currently is to 

* Probably always use [`memo()`](https://reactjs.org/docs/react-api.html#reactmemo) for anything relying on breakpoint-values. Remember memo uses referential equality by default, either define other methods for checking the diff, or only use primitive breakpoint values to make sure memorization works as expected. 

* Make sure you define what breakpoint to use as a fallback in the second argument to reduce the number of times your memoized component renders. Either make sure your fallback to match your expected most common platform, or populate it with information from use-agent or similiar.
  ```tsx
  const value = useBreakpointValue({ base: 'green', lg: 'blue' }, { fallback: 'lg' })
  ```

* If you're not using Server-Side Rendering, you should also pass `ssr: false` to ensure the first value is actual mediaqueried and not just the fallback.
    ```tsx
    const value = useBreakpointValue({ base: 'green', lg: 'blue' }, { ssr: false })
    ```

oBusk:
This is a simple solution that I could think of with my limited understanding of the code.

Closes #6401

## 📝 Description

An attempt at avoiding unecessary re-renders from the `useState()` inside the `useMediaQuery` hook, which `useBreakpoint()` and `useBreakpointValue()` depend on.

On first render, any component using `useBreakpointValue()` will render at least twice, as a `useEffect()` function will call `setState()`. 

`useMediaQuery()` will also first give the fallback value, even in environments where the mediaquery could be performed. If `ssr` is true, which is the default value, the intial value will be the fallback value, even if the mediaquer could have been performed. This means that even if you memoize your components relying on `useBreakpointValue()`, it would render twice.

## 🚀 New behavior

- `useMediaQuery` now tries to use `matchMedia()` to check if mediaqueries matches already when the state is initialised, and only uses fallback values if this fails _and_ `ssr == true`.
- `useMediaQuery` now avoids the re-render in the `useEffect()` by checking if the new value is different from the init value when setting the state.
- `useBreakpoint` now accepts configuration which is a list of the breakpoints that we are listening for. This list is used to only ask `useMediaQuery` to track these breakpoints, rather than all breakpoints in the theme. 
- `useBreakpointValue` uses the previously mentioned configuration of `useBreakpoint` to only track the breakpoints which the user provided values for. 

So all in all, if you use `useBreakpointValue({ base: '_base_', lg: '_lg_ ' }, 'base')`

- If the screen that the component is being rendered at matches `lg`, it will no longer first be rendered with the value `_base_` and should render only once.
- Moving the screen between `sm` and `md` no longer triggers a re-render

The change to `useMediaQuery` should only reduce unexpected re-renders. If you are rendering in an environment where `matchMedia` is available, you are rendering it with `ssr: true` but you expect it to first render with the fallback value, you will see a change in behaviour.
The change to `useBreakpoint` is only affected when using the new configuration to limit which breakpoints to track.
The change to `useBreakpointValue` should only mean fewer unnecessary re-renders.

## 📝 Additional Information

This is all based on my limited understanding of the chakra media query code, and my studies into the documentation of `useState()`.


oBusk:
I added a changeset, but not entirely sure how large/small the change should be. Also rebased the branch to stay up to date.

oBusk:
Got into react development the other day and ran into the annoyances of `useMediaQuery()` again so trying to revive this PR. Not sure if it's wanted. Maybe there's some improvements to these issues with the v3 coming up. 

Tests are failing now because I'm getting a double render, will look into later.

natBizitza:
> Got into react development the other day and ran into the annoyances of `useMediaQuery()` again so trying to revive this PR. Not sure if it's wanted. Maybe there's some improvements to these issues with the v3 coming up.

I think this might solve the issue with CLS that I have. I use useMediaQuery with min-height value and this type of conditions <Flex fontSize={isHeightSmallerThan1000 ? '2xs: 'xs'}>. It causes an awful CLS.

segunadebayo:
Thanks for working on this @oBusk.

I've added the changes in this PR to our upcoming v3 PR.

I appreciate your work on this.

Issue #5934, opened 2022-04-27
PopoverArrow always has box shadow blur, even when border is solid.


Tenkir:
### Description

When I have a PopoverArrow, I expect the arrow to have the same border as the Popover. Instead it has a 1px blur.

### Link to Reproduction

https://codesandbox.io/s/chakra-popover-bug-v2-4rr1i1

### Steps to reproduce

1. Go to reproduction
2. Click on button labeled `Click`
4. Observe PopoverArrow has a blurred box shadow.

### Chakra UI Version

2.2.4

### Browser

Firefox, Chrome, Safari

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

The style which adds this box shadow is an inline style, so it's not able to be overridden with a `boxShadow` property on the `PopoverArrow` component.

### Screenshot
![Screen Shot 2022-04-27 at 10 34 41 AM](https://user-images.githubusercontent.com/3649828/165543153-3cf8f3c1-d232-4383-840f-65d1534c6151.png)

YosefBlandin:
I can't reproduce the issue
![1](https://user-images.githubusercontent.com/65220054/173479670-eab6c52f-98f9-43f0-a250-54071fe055c0.png)

Tenkir:
> I can't reproduce the issue ![1](https://user-images.githubusercontent.com/65220054/173479670-eab6c52f-98f9-43f0-a250-54071fe055c0.png)

Looks like an update to codesandbox had broken the reproduction. Updated one is attached, and I have updated the original issue link:

https://codesandbox.io/s/chakra-popover-bug-v2-4rr1i1

robcrock:
If this is resolved can the issue be closed?

segunadebayo:
This has already been fixed in one of the recent commits. We'll release an update shortly.

If the issue persists after the release, I'll re-open this issue.

Tenkir:
@segunadebayo This is fixed in so far as it is no longer blurry, but it is still not correct. The borders protrude out the bottom of the border. My fix in #6412 addressed this.

![image](https://github.com/chakra-ui/chakra-ui/assets/3649828/7395917f-1dd1-4ae9-baf5-c1fcfaa70129)

PR #6411, opened 2022-08-02
fix: Resolving issue where PopoverArrow would have different border than contents

Tenkir:
Closes [#5934](https://github.com/chakra-ui/chakra-ui/issues/5934)

## 📝 Description

PopoverArrow used a box-shadow with blur to generate the borders on it's edges. This resulted in a blurry border, when the border of the rest of the popover was solid. This was particularly noticeable on dark backgrounds.

Old borders over dark background

PopoverArrow always had blurred border.

## 🚀 New behavior
New borders over dark background

PopoverArrow will have the same border as the rest of the PopOver or Tooltip.

This PR uses variations of border to generate the borders, rather than a box-shadow implementation. 
This results in accurate borders which match the rest of the container.

This PR also removes the now unused `getBoxShadow` function from `packages/popper/src/utils.ts`

No

## Additional Information

Not sure if this small visual change requires a test, but I can add one if necessary.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2100,6 +2100,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "tenkir",
+      "name": "Jonathan Blair",
+      "avatar_url": "https://avatars.githubusercontent.com/u/3649828?v=4",
+      "profile": "https://github.com/tenkir",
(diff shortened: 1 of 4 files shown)


Tenkir:
Discovered bug when border is not set, closing to resolve.

Bug when border is not set

PR #6437, opened 2022-08-07
fix(env): render span for env detection in portal

### LINKED ISSUE #6374: EnvironmentProvider appends a span to DOM which breaks some unit tests
### Description

I've noticed that EnvironmentProvider appends an empty span to the document which in my case breaks some unit tests that check if specified components do NOT render any elements. I'm using react-testing-library.

### Link to Reproduction

https://codesandbox.io/s/chakraprovider-empty-span-issue-qbmbdw

### Steps to reproduce

Visible in any case when ChakraProvider is being used.

### Chakra UI Version

2.2.4

### Browser

N/A

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

It was already mentioned https://github.com/chakra-ui/chakra-ui/issues/5855#issuecomment-1095400843.

In my case, when using RTL to unit test a component, with following test:
```
    const { container } = customRender(<ComponentThatShouldNotRenderAnything />);

    expect(container).toBeEmptyDOMElement();
```

and setup like:
```
const TestProviders = ({ children }: { children: React.ReactNode }) => {
  return (
      <ChakraProvider theme={theme}>{children}</ChakraProvider>
  );
};

const customRender = (
  ui: React.ReactElement,
  options?: RenderOptions,
): RenderResult => render(ui, { wrapper: TestProviders, ...options });
```

The above test is failing when using `ChakraProvider`.

I believe it could be solved by adjusting a `mounted` conditional in `EnvironmentProvider` and making it like: `mounted && !node`. This is how I solved my issue for now, I made a patch using `patch-package`. However, I'm not aware of any potential issues it may cause. Maybe there is a better way of solving this issue.

yukukotani:
Closes #6374

`span` is rendered next to the children of `ChakraProvider`. This breaks some tests to check DOM structure.

## 🚀 New behavior

`span` is rendered in the portal, so DOM structure of the component isn't affected.

No

## 📝 Additional Information

It seems that we can skip rendering `span` if `environmentProp` is given. I'll work on that in separate PR because it's not related to the issue.

### DIFF
--- a/.changeset/fair-readers-remember.md
+++ b/.changeset/fair-readers-remember.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react-env": minor
+---
+
+render span for env detection in portal
--- a/packages/env/package.json
+++ b/packages/env/package.json
@@ -32,7 +32,8 @@
     "url": "https://github.com/chakra-ui/chakra-ui/issues"
   },
   "dependencies": {
-    "@chakra-ui/utils": "workspace:*"
+    "@chakra-ui/utils": "workspace:*",
+    "@chakra-ui/portal": "workspace:*"
   },
   "peerDependencies": {
     "react": ">=18"
--- a/packages/env/src/env.tsx
+++ b/packages/env/src/env.tsx
@@ -1,4 +1,5 @@
 import { isBrowser, __DEV__ } from "@chakra-ui/utils"
+import { Portal } from "@chakra-ui/portal"
 import {
   createContext,
   startTransition,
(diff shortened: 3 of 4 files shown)


segunadebayo:
@yukukotani Kindly resolve the conflicts in this PR. I'd like to get this merged.

yukukotani:
@segunadebayo Resolved. But lock file has large diff because of pnpm's version update. Is it ok?

segunadebayo:
Thinking about this again. It seems to me that using a portal defeats the purpose of the env in the first place.

Its purpose is to get the correct reference to `window` or `document` regardless of the environment used. Portal will append to the `document.body` by default. The `document` might be incorrect in most cases.

The best way to test this is to render any component that uses the env within an iframe (using react-frame-component) and see if it still works as intended.

yukukotani:
@segunadebayo I think this works because Portal also detects env by `tempNode`.

Anyway, we need tests for the env but it's not easy since [jsdom doesn't support iframe's srcdoc](https://github.com/jsdom/jsdom/issues/1892). `react-frame-component` doesn't work with React Testing Library.

e2e testing is one of the options.

ksocha:
That might be a dumb proposition, but what if we just could prevent the `span` element being rendered after the node is set, like so:
```
{mounted && !node && (
        <span
          ref={(el) => {
            startTransition(() => {
              if (el) setNode(el)
            })
          }}
        />
      )}
```

segunadebayo:
So far, setting the span to be `hidden` has worked, and I would leave it as is for now.

I'll use a Portal as a fallback implementation if this becomes an issue.

Thanks for your efforts on this one @yukukotani

PR #6438, opened 2022-08-07
chore: enforce code format with prettier

yukukotani:

Prettier has been set up, but the format is not checked in CI.

Some codes are not formatted and it will cause irrelevant changes in PRs.

## 🚀 New behavior

CI ensures that the code is formatted.

No


packages/env/src/mock-document.ts
```diff
@@ -38,4 +38,4 @@ const doc = {
   },
 }
 
-export const ssrDocument = (doc as unknown) as Document
```
yukukotani, line 41: `pnpm format:write` generated changes except `quality.yml` and `package.json`.

PR #6448, opened 2022-08-09
chore(cli): add styled system to deps

### LINKED ISSUE #6427: Missing dependency for `@chakra-ui/cli`
### Description

When I use pnpm as my package manager, and I run the `chakra-cli tokens` command, I get an error about styled-system can't be found in node_modules

### Link to Reproduction

https://github.com/Belco90/personal-website/pull/38

### Steps to reproduce

The provided repro link is actually fixed. You can reproduce it by

1. Uninstalling `@chakra-ui/styled-system` from dev dependencies
2. Running `pnpm gen:theme-typings`

You can run into the same issue in any repo using Chakra UI by:

1. Switching the package manager to pnpm
2. Setting up Chakra CLI to generate custom tokens
3. Running the command to generate custom tokens

### Chakra UI Version

2.2.6

### Browser

n/a

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

This is the error in the terminal:
chakra-pnpm

This seems a phantom dependency to me. Since pnpm manages the dependencies in a non-flat node_modules, code has no access to arbitrary packages. The default output for `@chakra-ui/cli` token command is `@chakra-ui/styled-system`, which is not explicitly set as dependency of `@chakra-ui/cli`. This works fine with npm since all packages are hoisted in the node_modules, so `@chakra-ui/styled-system` is found implicitly installed by other package (not sure which one, but probably another chakra package).

I think just adding `@chakra-ui/styled-system` as a dependency of `@chakra-ui/cli` would fix this issue. As a workaround, I had to explicitly install `@chakra-ui/styled-system` in the dev dependencies of my repo.

anubra266:
Closes #6427

Fixes issue where Styled system is not resolved in pnpm cli dependencies

### DIFF
--- a/.changeset/happy-knives-remember.md
+++ b/.changeset/happy-knives-remember.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/cli": patch
+---
+
+Fixed issue where Styled system is not resolved in pnpm cli dependencies
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: 5.3
+lockfileVersion: 5.4
 
 overrides:
   react: ^18.2.0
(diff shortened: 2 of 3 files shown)


itkrt2y:
This PR does not fix this issue.
reproduction => `pnpm --filter @chakra-ui/cli build && pnpm chakra-cli tokens packages/theme/src`

This issue is caused by `resolveThemingDefinitionPath` in `tooling/cli/src/command/tokens/resolve-output-path.ts` .

`resolveThemingDefinitionPath` looks for `theming.types.d.ts` under `node_modules` and expects `node_modules` is flat. 
But pnpm creates non-flat `node_modules`. Therefore chakra-cli cannot find `theming.types.d.ts`. (ref: https://pnpm.io/blog/2020/05/27/flat-node-modules-is-not-the-only-way)

If the app installs `@chakra-ui/styeld-system` as a direct dependency, `@chakra-ui/styled-system` will be placed flat in `node_modules`. So [this workaround](https://github.com/chakra-ui/chakra-ui/issues/6427#issuecomment-1206808711) works.

PR #6451, opened 2022-08-09
fix(toast): fix toast provider ignoring defaultOptions

### LINKED ISSUE #6370: ChakraProvider toast defaultOptions not applied
### Description

When I set a position in defaultOptions in toastOptions in ChakraProvider, it doesn't get applied

### Link to Reproduction

https://codesandbox.io/s/busy-rui-szzc85?file=/src/index.tsx

### Steps to reproduce

1. Click on "Click me"
2. See toast is in standard default position (bottom) rather than the specified default position (top)

### Chakra UI Version

2.2.3

### Browser

Chrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I can't see anything being done with `defaultOptions` in https://github.com/chakra-ui/chakra-ui/blob/main/packages/toast/src/toast.provider.tsx

sofiyevsr:
Closes #6370 

## 📝 Description

Receives toastOptions from ChakraProvider and passes it down to useToast hook where data is merged with options passed to hook

Passing toastOptions to ChakraProvider does nothing

## 🚀 New behavior

ToastOptions passed to ChakraProvider is applied to all toasts when toast function or useToast hook called with no options 

No

## 📝 Additional Information
This is my first contribution, so i am not very familiar with internals. I think changes in react/src/chakra-provider.tsx can break some stuff


packages/toast/test/use-toast.test.tsx
```diff
@@ -121,4 +121,98 @@ describe("useToast", () => {
...
+
+    const button = await screen.findByText("Toast")
+    await user.click(button)
+
+    const allByTitle = await screen.findAllByText(title)
+    const allByDescription = await screen.findAllByText(description)
+
+    expect(allByTitle).toHaveLength(1)
+    expect(allByDescription).toHaveLength(1)
+  })
```
sztadii, line 217: Nice and readable tests, but one test case is missing.
You should check the case where we define default options in ToastProvider and useToast and we will be going to override it in the toast function.

sztadii (changes requested):

codebutler:
Hi, I rebased @sofiyevsr's fixes and added the additional test requested by @sztadii: https://github.com/codebutler/chakra-ui/commit/33c1343b7c014d2a617e0f879ce083b8e7b76af6

can we please get this in? thanks

packages/react/src/chakra-provider.tsx
```diff
@@ -19,8 +19,7 @@ export function ChakraProvider({
 }: ChakraProviderProps) {
   return (
     <BaseChakraProvider {...restProps}>
-      {children}
-      <ToastProvider {...toastOptions} />
+      <ToastProvider {...toastOptions}>{children}</ToastProvider>
```
segunadebayo, line 22: The toast provider holds an internal state that'll cause the entire "app" to re-render when it changes. I'd prefer we create a separate context for the toast options and wrap the children within that.

packages/toast/src/toast.provider.tsx
```diff
@@ -90,6 +91,18 @@ export type ToastProviderProps = React.PropsWithChildren<{
   portalProps?: Pick<PortalProps, "appendToParentPortal" | "containerRef">
 }>
 
+export interface ToastDataContext {
```
segunadebayo, line 94:
```suggestion
export interface ToastOptionContext {
```

packages/toast/src/toast.provider.tsx
```diff
@@ -90,6 +91,18 @@ export type ToastProviderProps = React.PropsWithChildren<{
...
+  topLevelDefaultOptions: UseToastOptions | undefined
```
segunadebayo, line 95: We could use "defaultOptions" or "globalOptions"

```suggestion
defaultOptions: UseToastOptions | undefined
```

packages/toast/src/toast.provider.tsx
```diff
@@ -90,6 +91,18 @@ export type ToastProviderProps = React.PropsWithChildren<{
...
+}
+
+/**
+ * Passes default options down to be used by toast creator function
+ */
+export const [ToastDataContext, useToastDataContext] =
+  createContext<ToastDataContext>({
+    name: `ToastDefaultOptionsContext`,
+  })
```
segunadebayo, line 104:
```suggestion
export const [ToastOptionProvider, useToastOptionContext] =
  createContext<ToastOptionContext>({
    name: `ToastOptionsContext`,
  })
```

segunadebayo (changes requested):
Thanks for working on this @sofiyevsr.

Left a few comments.

packages/react/src/chakra-provider.tsx
sofiyevsr: Done. I was unsure about moving provider to a new file, so i kept the context in toast.provider and imported it inside chakra-provider.

PR #6464, opened 2022-08-11
Add explicit type declaration to `ChakraProvider.defaultProps`

### LINKED ISSUE #6463: `ChakraProvider.defaultProps` is implicitly typed any in generated declarations for `@chakra-ui/react`
### Description

`ChakraProvider.defaultProps` is currently declared as `var defaultProps`, which can cause TypeScript builds to fail in projects that depend on `@chakra-ui/react` and TypeScript strict mode.

### Link to Reproduction

https://stackblitz.com/edit/node-uqdb5a?file=tsconfig.json

### Steps to reproduce

1. Run `npm run build:tsc`
2. Observe the following output:
    ```
    $ tsc -p .
    node_modules/@chakra-ui/react/dist/index.d.ts:60:9 - error TS7005: Variable 'defaultProps' implicitly has an 'any' type.
     
    60     var defaultProps;
               ~~~~~~~~~~~~


    Found 1 error in node_modules/@chakra-ui/react/dist/index.d.ts:60
    ```

### Chakra UI Version

@chakra-ui/react 2.2.6

### Browser

N/A

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

This error appears when using TypeScript strict mode (see reproduction).

rossng:
Closes #6463

## 📝 Description

This PR adds an explicit declaration to `chakra-provider.tsx` that will forcefully change the declaration output from this:

```
declare function ChakraProvider({ children, toastOptions, ...restProps }: ChakraProviderProps): JSX.Element;
declare namespace ChakraProvider {
    var defaultProps;
}
```

to this:

```
declare function ChakraProvider({ children, toastOptions, ...restProps }: ChakraProviderProps): JSX.Element;
declare namespace ChakraProvider {
    let defaultProps: {
        theme: Theme;
    };
}
```

I was

The current implementation leaves TypeScript to infer the type that must be emitted for `ChakraProvider.defaultProps`, which it fails to do. This causes TypeScript builds to fail in strict mode for consumers of `@chakra-ui/react`.

## 🚀 New behavior

The new implementation explicitly declares the type that needs to be emitted.

## 💣 Is this a breaking change (Yes/**No**):

I don't believe this is a real breaking change, though I guess it could _potentially_ break builds in some edge cases.

## 📝 Additional Information

I did find alternative ways to do this without the `declare namespace`, but they require depending on the `@chakra-ui/theme-tools` package and inline the gigantic theme type into the emitted declaration file.

### DIFF
--- a/.changeset/three-jokes-swim.md
+++ b/.changeset/three-jokes-swim.md
@@ -2,4 +2,4 @@
 "@chakra-ui/react": minor
 ---
 
-Add explicit type declaration for `ChakraProvider.defaultProps`.
+Remove `ChakraProvider.defaultProps` and move logic inside the component.
--- a/.changeset/three-jokes-swim.md
+++ b/.changeset/three-jokes-swim.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react": minor
+---
+
+Add explicit type declaration for `ChakraProvider.defaultProps`.
--- a/packages/react/src/chakra-provider.tsx
+++ b/packages/react/src/chakra-provider.tsx
@@ -2,7 +2,7 @@ import {
   ChakraProvider as BaseChakraProvider,
   ChakraProviderProps as BaseChakraProviderProps,
 } from "@chakra-ui/provider"
-import { theme } from "@chakra-ui/theme"
+import { theme, Theme } from "@chakra-ui/theme"
 import { ToastProvider, ToastProviderProps } from "@chakra-ui/toast"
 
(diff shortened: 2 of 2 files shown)


segunadebayo:
We really want to move away from default props for the most part and just add the default from within the component.  If you can refactor the code to do that, that'll be great.

rossng:
Thanks for the review. I've revised the PR - does this match what you were suggesting?

PR #6465, opened 2022-08-11
fix regression where theme typing could not be used

### LINKED ISSUE #6440: VSCode autocomplete for theme props not working
### Description

When filling in theme related props like bg color I expected intellisense suggestions with theme tokens but none are showing up

### Workaround

Use exact version of chakra `2.2.0`. Thanks [system32uwu](https://github.com/system32uwu)

### Link to Reproduction

https://codesandbox.io/s/purple-water-b30s3k?file=/src/index.tsx

### Steps to reproduce

1. Create a simple `<Box />` component
2. Add a prop that uses theme tokens, for example`bg`
3. CTRL+Space to have the intellisense suggestions pop up
4. Note that no suggestions are shown with theme tokens like `red.100`, `red.200`, etc

image

### Chakra UI Version

2.2.6

### Browser

_No response_

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

First tested in next,js using pnpm, then CRA using npm and the chakra template, then codesandbox using the chakra template. Same result everytime

I'm not 100% sure this is a supported feature, most I could find in the [docs is the following which makes me believe autocomplete suggestions are supported](https://chakra-ui.com/docs/styled-system/advanced-theming#theme-typings):
image

itkrt2y:
Closes #6440

## 📝 Description

Before migrating to tsup, `.d.ts` files were created for each source file. However, with the migration to tsup, it is now bundled in single file.
`chakra-cli tokens` creates `styled-system/dist/theming.types.d.ts`, but with this change, `styled-system/dist/index.d.ts` no longer import this type file.

I fixed so that `styled-system/dist/theming.types.ts` is created and imported from `styled-system/dist/index.d.ts` as before tsup was introduced.

Cannot use theme typing created by chakra-cli

## 🚀 New behavior

Works as expected

No


packages/styled-system/package.json
```diff
@@ -45,7 +45,8 @@
   },
   "sideEffects": false,
   "scripts": {
-    "build": "tsup src/index.ts --dts",
+    "build": "tsup src/index.ts && pnpm build:dts",
+    "build:dts": "tsup --entry src/index.ts --entry src/theming.types.ts --entry src/shared.types.ts --dts-only",
```
segunadebayo, line 49: How about we update the the `build` command to this:

```sh
tsup src/ --dts
```

That way, it'll generate individual chunks for the code and types and everything works as needed.

itkrt2y: I tested `tsup src/ --dts` and this command generated some duplicated code.
So I think `tsup src/index.ts` is more efficient and better.

packages/styled-system/package.json
```diff
@@ -45,7 +45,8 @@
...
     "dev": "pnpm build -- --watch",
```
segunadebayo, line 50: In that case, let's restructure this a bit so that the `dev` command and `build:fast` can work.

itkrt2y: Fixed 4a2d4531bdd2d42a4c1f13fbff42dc120a53e8c1

Issue #5252, opened 2021-12-16
Allow Input Group width to be configurable


hpersson:
### Description

We have overridden the default `width: 100%` property for Input and Select in our project using our custom Chakra theme. 

However, we are unable to configure the default width of InputGroups via the theme object - https://github.com/chakra-ui/chakra-ui/blob/main/packages/input/src/input-group.tsx#L72

We wrap the Chakra InputGroup component and pass styles to provide a uniform experience when building forms. But it would be nice if this could be configured like the other properties.

### Problem Statement/Justification

Without being able to configure InputGroup's width from 100%, it is hard to use Chakra directly with inputs of a fixed width.

Other components allow you to configure "container" components - e.g. FormControl can be configured to not have `width: 100%`

### Proposed Solution or API

Add a "container" or "group" part to the theme for inputs

primos63:
> However, we are unable to configure the default width of InputGroups via the theme object 

Yes, because there is no theme for `InputGroup`. 

> We wrap the Chakra InputGroup component and pass styles to provide a uniform experience when building forms.

I assume you have a simple wrapper that passes a `width` prop to the `InputGroup` which is valid.

I cannot see the use case for this. Every form you have with `InputGroup`s are always the same width? I would expect that the children of `InputGroup` would always take up 100% of the parent's width. If your form's container had a width of 300px, I would expect your inputs to also have that width.

If you could please provide a CodeSandbox example of what you are trying to do, it would be appreciated.

hpersson:
> I cannot see the use case for this. Every form you have with InputGroups are always the same width?  If your form's container had a width of 300px, I would expect your inputs to also have that width.

All of our Inputs have the same width. Technically, though, the override I would like to make to the InputGroup width is to set it to "fit-content".

While I can understand the choice to have Inputs and InputGroups take up 100%, I don't think this is necessarily the only obvious case. In fact, I don't think Chakra itself has always assumed this, given that Input has [deprecated props](https://chakra-ui.com/docs/form/input) for `isFullWidth` which makes me think that these components behaved differently, or were able to be configured to behave differently, at some point.

I filed this as a feature request and not a bug because I fully acknowledge that I can have a wrapper that passed width to the InputGroup, so I'm not blocked/missing any functionality. But, it is not clear to me why I need a wrapper component for an InputGroup when most other customization of this variety I have been able to configure through theme.

Here's an example:
[...]

PR #6490, opened 2022-08-15
fix: restore zindex of input element

### LINKED ISSUE #5742: Conflicting zIndex of InputRightElement and MenuList
### Description

When I have an Input with an InputRightElement it overlaps with the MenuList of a Menu because of the constellation of zIndex in both components.

### Link to Reproduction

https://codesandbox.io/s/objective-almeida-i2rcvl

### Steps to reproduce

1. Go to the CodeSandbox
2. Click on the buttons to see the bug and possible solutions

### Chakra UI Version

1.8.6

### Browser

Firefox 98.0 and Chromium 99.0.4844.51

### Operating System

- [ ] macOS
- [ ] Windows
- [X] Linux

### Additional Information

Use case for this is in my case a search bar with a lens icon in the top right corner of the page, directly under the mobile header.

I figured out that the change of the zIndex of InputLeftElement and InputRightElement was increased from 1 to 2 because of [issue 1900](https://github.com/chakra-ui/chakra-ui/issues/1900). I can not reproduce this bug when setting it to 1 again, so I guess something else has changed in the code that it is not required anymore.

In the codesandbox are 3 solutions from me how we can fix this:
1. set a high zIndex for MenuList in the theme, fits probably the best [between banner and overlay](https://chakra-ui.com/docs/styled-system/theming/theme#z-index-values) at ~1250 -> this should be a breaking change but in my opinion the way to go
2. set zIndex of the InputElements to 1 again and set the zIndex of MenuList to 2 OR wrap the MenuList in a Portal (to keep it on top if both have a zIndex of 1)
3. set zIndex of MenuList to 3 -> probably the worst solution

Feel free to assign me when we found out which solution is the best :smiley:

m4x3d:
Closes #5742 

## 📝 Description

Restore of `InputElement`s zIndex from 2 to 1 which leads to problems in overlapping with e.g. `MenuList` and `Popover`. This is basically a revert of https://github.com/chakra-ui/chakra-ui/pull/1906

Overlapping `InputElements` on e.g. open `MenuList`, see first example in: 
https://codesandbox.io/s/objective-almeida-i2rcvl?file=/src/index.tsx

## 🚀 New behavior

Restore to 'old' zIndex 1 since the bug from back then seems not to exist anymore with the current Chakra UI version (copy&pasted example):
https://codesandbox.io/s/serene-platform-3wuu74?file=/src/index.tsx

Possibly not

### DIFF
--- a/.changeset/clever-planets-fetch.md
+++ b/.changeset/clever-planets-fetch.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/input": minor
+---
+
+Restore zIndex of InputElement to 1
--- a/packages/input/src/input-element.tsx
+++ b/packages/input/src/input-element.tsx
@@ -18,7 +18,7 @@ const StyledInputElement = chakra("div", {
     justifyContent: "center",
     position: "absolute",
     top: "0",
-    zIndex: 2,
+    zIndex: 1,
   },
 })
 


noobinthisgame:
Thanks for working on this :) 

This looks good to me 🚀

Please resolve the conflicts and we are ready to go.

lexanth:
The bug that the original fix was to address was caused by https://github.com/chakra-ui/chakra-ui/pull/1583/files
* In the outline variant, when focused the input's z-index would increase to 1
* The input used to have a white background
* The input element was before the input in the DOM, so got hidden

The situation now:
* In the outline variant, the background is now inherit by default
* The z-index still gets increased, but the background is now transparent so you can still see the element

However, with this fix, you can still trigger that bug if you try to set a background on the input.


A cleaner fix would be to keep z-index as 2 and set `isolation: isolate` on the `InputGroup`
https://codesandbox.io/s/nostalgic-newton-pwjm0z?file=/src/index.tsx - added a "Solution 4" with this

this scopes the z-index on the input element to just the input group anyway (by creating a new stacking context), so it isn't above the popover/menu etc.

Unfortunately, InputGroup cannot be styled using the theme - https://github.com/chakra-ui/chakra-ui/issues/5252.

I've raised https://github.com/chakra-ui/chakra-ui/pull/7104 to apply my suggested fix.

segunadebayo:
Closing in favor of #7104

PR #6516, opened 2022-08-19
Refactor image component

anubra266:
refactor image


packages/image/package.json
```diff
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "@chakra-ui/hooks": "workspace:*",
```
segunadebayo, line 36: Can we remove this dependency?

segunadebayo (changes requested):

PR #6517, opened 2022-08-19
Refactor Input

anubra266:
Refactor Input

### DIFF
--- a/utilities/shared-utils/src/index.ts
+++ b/utilities/shared-utils/src/index.ts
@@ -79,24 +79,3 @@ export function callAll<T extends AnyFunction>(...fns: (T | undefined)[]) {
     })
   }
 }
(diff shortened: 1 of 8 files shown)


utilities/shared-utils/src/index.ts
```diff
@@ -79,3 +79,24 @@ export function callAll<T extends AnyFunction>(...fns: (T | undefined)[]) {
...
+    if (shouldPass) {
+      result[key] = value
+    }
+  })
+
+  return result
+}
+
+export const filterUndefined = (object: Record<string, any>) =>
+  objectFilter(object, (val) => val !== null && val !== undefined)
```
segunadebayo, line 102: Can we use compact from object-utils for this?

PR #6522, opened 2022-08-20
fix(tooltip): show at most one tooltip at a time

### LINKED ISSUE #4685: [tooltip] Show at most one tooltip at a time
### Description

The strategy to display at most one tooltip at a time tends to improve UX in certain scenarios.

When using `closeDelay` (which I find preferable from the UX point of view) it seems highly undesirable to have 2 tooltips rendered at once. This can easily happen when triggering tooltips with keyboard navigation and when "jumping" through elements. The issue gets worse with long longer labels as tooltips might overlap with each other as each one gets positioned without taking the other one into consideration.

Take a look at this sandbox:
https://codesandbox.io/s/trusting-darkness-2yqz6?file=/src/App.js

### Proposed Solution or API

Use module-level (global~) variables to track open tooltip and if there is any tooltip shown when displaying a new one - dismiss the older one (immediately, without taking `closeDelay` into account for this action).

### Alternatives

_No response_

### Additional Information

_No response_

itkrt2y:
Closes #4685

## 📝 Description

When opening a tooltip, close other tooltips even if `closeDelay` has not passed

Multiple tooltips can be displayed if `closeDelay` is specified

## 🚀 New behavior

Show at most one tooltip at a time

No


packages/tooltip/src/use-tooltip.ts
```diff
@@ -246,3 +257,13 @@ export function useTooltip(props: UseTooltipProps = {}) {
...
+
+const closeEventName = "chakra-ui:close-tooltip"
+
+function useCloseEvent(close: () => void) {
+  useEffect(() => {
+    document.addEventListener(closeEventName, close)
+    return () => document.removeEventListener(closeEventName, close)
+  }, [close])
+  return () => document.dispatchEvent(new CustomEvent(closeEventName))
+}
```
segunadebayo, line 269: Great work! Can we ensure we use the correct reference to `window` and `document` for the tooltip?

doc => `ref.current.ownerDocument` 
win => `doc.defaultView`

itkrt2y: Fixed 19ba261b1d3ed1d50ed794583d90f7dd4b7d20bd

segunadebayo: I don't think that's the correct fix. Here's what you can consider

1. Allow useCloseEvent to accept a ref (same ref that the tooltip creates)
2. Within the useEffect of the useCloseEvent and useTooltip, read the `doc` and `win` from there

This way, we can avoid the useForceUpdate completely

itkrt2y: Thanks for your review! I fixed not to use `useForceUpdate`. af89344346a1318480710c6df1ab5a6d9eedda61

PR #6523, opened 2022-08-20
feat: LoadingOverlay

Pagebakers:
## 📝 Description

This PR implements the LoadingOverlay RFC (https://github.com/chakra-ui/rfcs/pull/3)

## 📝 Additional Information

A couple of questions arose while developing the new component.

1. Should the container element have padding, and how much? It's 4 now.
2. Should we use AnimatePresence / MotionBox for the animation?
3. The overlay / fill variants have a translucent background, should it be translucent and how much? right now it's whiteAlpha.300 / blackAlpha.300
4. We set aria-live="polite" and aria-busy="true" on the containerRef. Are these the appropriate attributes for all situations? Eg if the main content of a page is loading, maybe it should be assertive? as compared to a small less important card.
5. Which variant should be the default, it's `fill` right now.


segunadebayo:
Thanks for working on this @Pagebakers.

> Should the container element have padding, and how much? It's 4 now.
Yes, it should have a padding. It's much easier to give feedback on this when there's a working Storybook or Figma I can see.

> Should we use AnimatePresence / MotionBox for the animation?
I'd like to avoid using framer motion within the library going forward. If we can achieve those with CSS animation, that'll be great.


> The overlay / fill variants have a translucent background, should it be translucent and how much? right now it's whiteAlpha.300 / blackAlpha.300
Again, Storybook will help me get a better idea of this one.

> We set aria-live="polite" and aria-busy="true" on the containerRef. Are these the appropriate attributes for all situations? Eg if the main content of a page is loading, maybe it should be assertive? as compared to a small less important card.
Let's skip these for now and get the component shipped. We can resolve this later on.

Pagebakers:
Thanks! I see the storybook didn't deploy, I'll get that fixed.

The containerRef can be removed completely for now?

Pagebakers:
@segunadebayo The storybook is live: https://chakra-ui-storybook-git-fork-pagebakers-feat-l-b87635-chakra-ui.vercel.app/?path=/story/components-feedback-loadingoverlay--basic

Made a few minor changes.
- Removed the hook and containerRef functionality.
- Looked into using keyframes, but sticking with framer for now. I think it'd be better if we create some animation hooks to replace framer first, but out of scope for this PR.

segunadebayo:
Hi @Pagebakers,

Thanks for your patience here.

A couple of things are left before merging this:
- I'd like to reduce the use framer-motion for new components, let's see how we can model this without framer-motion since it's not a complex component.
- I'd prefer to use CSS variables for the theme instead of the `mode()` function. Please refer to the updated theme for most components to see how it's done.

Sincere apologies for keeping this on for so long. Happy to get this in ASAP

Pagebakers:
@segunadebayo Thanks for the update.

Yeah this definitely took too long :(, but I understand there have been a lot of architectural changes in the last few months.

I'm working on v2 of Saas UI, so for me it's a good time to move the component to Chakra UI.
I'll try to introduce the improvements in the coming weeks.

PR #6537, opened 2022-08-23
chore(styled-system): refactor

anubra266:
chore(styled-system): refactor

### DIFF
--- a/.changeset/big-elephants-deny.md
+++ b/.changeset/big-elephants-deny.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/modal": patch
+---
+
+Refactor modal
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -41,16 +41,18 @@
   "dependencies": {
     "@chakra-ui/close-button": "workspace:*",
     "@chakra-ui/focus-lock": "workspace:*",
-    "@chakra-ui/hooks": "workspace:*",
     "@chakra-ui/portal": "workspace:*",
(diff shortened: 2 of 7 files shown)


packages/styled-system/src/style-config.ts
```diff
@@ -66,15 +61,16 @@ function createResolver(theme: Theme) {
...
+            ...result,
             [part]: isSingle ? styles[part] : { [query]: styles[part] },
           })
         })
         continue
       }
 
       if (!isMultipart) {
-        if (isSingle) mergeWith(result, styles)
+        if (isSingle) compact({ ...result, ...styles })
```
segunadebayo, line 73: This is wrong because mergeWith is a deep merge and what you have here is a shallow merge. Please revert to mergeWith.

packages/styled-system/src/style-config.ts
```diff
@@ -96,11 +92,10 @@ export function resolveStyleConfig(config: Config) {
...
-      {},
-      runIfFn(config.baseStyle ?? {}, props),
-      recipe(config, "sizes", size, props),
-      recipe(config, "variants", variant, props),
-    )
+    return compact({
+      ...runIfFn(config.baseStyle ?? {}, props),
+      ...recipe(config, "sizes", size, props),
+      ...recipe(config, "variants", variant, props),
+    })
```
segunadebayo, line 99: Please use mergeWith here as well. Try to use the `merge-utils` package instead

PR #6553, opened 2022-08-24
chore: reorganize packages

anubra266:
chore: reorganize packages


oceandrama:
Hi! I found that It breaks `@chakra-ui/pro-theme`
```
import { extendTheme, theme as baseTheme } from "@chakra-ui/react";
                      ^^^^^
SyntaxError: Named export 'theme' not found.
```
Please, release new version of it!

anubra266:
@oceandrama, this PR did not update any packages, so it can't break any one. You can also see above that there's no changeset.

anubra266:
@oceandrama, this PR did not update any packages, so it can't break any one. You can also see above that there's no changeset.

PR #6562, opened 2022-08-25
Remove redundant ARIA roles

### LINKED ISSUE #6556: Table, table cell, table row have redundant aria roles
### Description

Chakra-UI's `Table`, `Tr`, and `Td` elements render `<table>`, `<tr>`, `<td>` elements respectively. These HTML elements have implicit ARIA roles, but Chakra-UI adds redundant roles to the output.

### Link to Reproduction

https://codesandbox.io/s/gifted-cannon-u0u8eq

### Steps to reproduce

1. Open codesandbox
2. Open devtools inspector
3. Observe that `<table>`, `<tr>`, and `<td>` elements have `role="..."` attributes defined

### Chakra UI Version

2.2.8

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

A workaround is to suppress the addition of the redundant role attribute by passing `role={undefined}` to Table, Tr, and Td.

Note that `<table>` and `<tr>` have implicit roles `table` and `row` respectively, so those are purely duplicative, but `<td>` has implicit role `cell`. Chakra's `Td` currently gives it `role="gridcell"`. These are slightly different, but elements with role `gridcell` can only be descendants of a `table role="grid"` (which Chakra's `Table` is not using), so the current usage is not valid.

iansjk:
Closes #6556.

## 📝 Description

The following elements already have intrinsic ARIA roles, so specifying them again with role="..." is unnecessary.

- `<table>` has intrinsic role "table"
- `<tr>` has intrinsic role "row"
- `<td>` has intrinsic role "cell"
- `<hr>` has intrinsic role "separator"

See [this W3C article](https://www.w3.org/TR/using-aria/#aria-does-nothing) for redundant ARIA usage examples and recommendations. In particular:
> In the majority of cases setting an ARIA [role](http://www.w3.org/TR/html5/infrastructure.html#attr-aria-role) and/or aria-* attribute that matches the [default implicit ARIA semantics](http://www.w3.org/TR/html51/infrastructure.html#default-implicit-aria-semantics) is unnecessary and not recommended as these properties are already set by the browser.

See [this WAI article](https://www.w3.org/WAI/ARIA/apg/practices/structural-roles/) for a mapping of ARIA roles to native HTML elements.

The above elements have `role="..."` specified.

## 🚀 New behavior
The above elements will no longer have `role="..."` specified.

I don't think this is a breaking change, since by default, the `<table<tr><td><hr>` elements already have these roles, so removing the explicit `role` should not affect their semantics. 

## 📝 Additional Information
I was unsure at first about the effect of someone who has a component structure like
```jsx
<Table role="grid">
  <Tr>
    <Td>foo</Td>
  </Tr>
</Table>
```
Which currently produces
```html
<table role="grid">
  <tr role="row">
    <td role="gridcell">foo</td>
  </tr>
</table>
```
But after this PR would produce
```html
<table role="grid">
  <tr>
    <td>foo</td>
  </tr>
</table>
```
But it appears that using `<td>` with no `role` specified as children of a `<table role="grid">` is valid. See [this MDN example on role="grid"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/grid_role), for example.


klausbadelt:
Coming a bit late, but this is a breaking change for us who use Cypress or/and React Testing Library and [`*ByRole`](https://testing-library.com/docs/queries/byrole). `<td>`s no longer have the [implicit role](https://www.w3.org/TR/html-aria/#docconformance) of `gridcell`, failing all our `findByRole('gridcell',...)` tests. :(

iansjk:
> Coming a bit late, but this is a breaking change for us who use Cypress or/and React Testing Library and [`*ByRole`](https://testing-library.com/docs/queries/byrole). `<td>`s no longer have the [implicit role](https://www.w3.org/TR/html-aria/#docconformance) of `gridcell`, failing all our `findByRole('gridcell',...)` tests. :(

Can you give a bit more context of the tests that are now failing? Did the `<td>`s you were trying to match previously with `findByRole` have a parent with `table row="grid"`?

PR #6605, opened 2022-08-31
Fix tooltip theme

yjkwon07:
## 📝 Description

'--tooltip-fg' is not applied

image


cereallarceny:
We're also having this problem. I can confirm it for 2.2.6 @segunadebayo

rm-hull:
would be good to get this fix into a point release

PR #6614, opened 2022-09-01
chore(button): refactor theme

anubra266:
chore(button): refactor theme


noobinthisgame:
Will this be released in a new major ? 

While refactoring this you could try to get rid of all functions that use @crtl/tinycolor from the theme tools. I did already do some research about that. It would result in a reduction of bundle size. https://github.com/chakra-ui/chakra-ui/pull/5949

segunadebayo:
Going to close this for now and attempt it in v3 using semantic tokens

PR #6656, opened 2022-09-08
fix(types): disallow `checked` prop where it shouldn't be accepted

Andarist:

`checked` prop is allowed on `Switch`, `Checkbox`, `Radio`, and text-like inputs. It's especially confusing when it comes to the first mentioned group because in reality only `isChecked` prop is accepted and handled correctly.

## 🚀 New behavior

`checked` prop is no longer accepted where it shouldn't be

No

## 📝 Additional Information

:warning: this is mostly PoC and groundwork for the work that has to be done. I probably won't be able to finish this PR and a core Chakra member should take over this. I've merely located the places where this prop shouldn't be accepted.

Similar PR: https://github.com/chakra-ui/chakra-ui/pull/5296/files

### DIFF
--- a/packages/components/checkbox/src/checkbox.tsx
+++ b/packages/components/checkbox/src/checkbox.tsx
@@ -38,11 +38,7 @@ const CheckboxRoot = chakra("label", {
 
 type CheckboxControlProps = Omit<HTMLChakraProps<"div">, keyof UseCheckboxProps>
 
-type BaseInputProps = Pick<
-  PropsOf<"input">,
-  "onBlur" | "checked" | "defaultChecked"
->
-
+type BaseInputProps = Pick<PropsOf<"input">, "onBlur">
 export interface CheckboxProps
   extends CheckboxControlProps,
     BaseInputProps,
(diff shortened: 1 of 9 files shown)


packages/components/checkbox/src/checkbox.tsx
```diff
@@ -38,11 +38,7 @@ const CheckboxRoot = chakra("label", {
 
 type CheckboxControlProps = Omit<HTMLChakraProps<"div">, keyof UseCheckboxProps>
 
-type BaseInputProps = Pick<
-  PropsOf<"input">,
-  "onBlur" | "checked" | "defaultChecked"
```
Andarist, line 43: `defaultChecked` is already a part of `UseCheckboxProps` so it doesn't have to be picked here

packages/components/checkbox/src/checkbox.tsx
```diff
@@ -58,103 +54,108 @@ export interface CheckboxProps
...
-    >
-      <input
-        className="chakra-checkbox__input"
-        {...getInputProps(inputProps, ref)}
-      />
-      <CheckboxControl
-        __css={styles.control}
-        className="chakra-checkbox__control"
-        {...getCheckboxProps()}
+export const Checkbox = forwardRef<CheckboxProps, "input", "checked">(
```
Andarist, line 57: the only change is in this line, the rest comes from Prettier

packages/components/editable/src/editable-input.tsx
```diff
@@ -10,7 +10,7 @@ export interface EditableInputProps extends HTMLChakraProps<"input"> {}
  * The input used in the `edit` mode
  */
 
-export const EditableInput = forwardRef<EditableInputProps, "input">(
+export const EditableInput = forwardRef<EditableInputProps, "input", "checked">(
```
Andarist, line 13: Arguably all text inputs should there the omitted props list somehow.

packages/components/system/src/system.types.tsx
```diff
@@ -55,11 +55,15 @@ export type MergeWithAs<
...
+  Component extends As,
+  Props extends object = {},
+  Omitted extends string = never,
+> = {
   <AsComponent extends As = Component>(
     props: MergeWithAs<
-      React.ComponentProps<Component>,
-      React.ComponentProps<AsComponent>,
+      Omit<React.ComponentProps<Component>, Omitted>,
+      Omit<React.ComponentProps<AsComponent>, Omitted>,
```
Andarist, line 66: from the performance PoV using `Omit` twice here is not that great - but with the current shape of this type it has to be used on both of type arguments

jucian0:
Why not make it work as expected in a input element.

segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version where the `disabled` prop is now native.

PR #6666, opened 2022-09-10
Add Missing JSDocs and Docsite Links

TylerAPfledderer:
## 📝 Description

Ensures that all Components, functions, and hooks that have detailed pages on the Chakra site have corresponding JSDocs and links back to the docsite via IDE intellisense.

Includes add/fixing links to related WAI-ARIA patterns.

Some components, hooks, and functions did not have JSDocs and/or links back to their corresponding pages of the Chakra docsite.

## 🚀 New behavior
All components, hooks, and functions with corresponding docsite pages will now have JSDocs and site links for IDE intellisense viewing.

No, as this does not change any functionality of the package.

## 📝 Additional Information

This is to help provide more accessibility of the package within the user's IDE. Only some of the components had JSDocs with site links.

### DIFF
--- a/packages/components/button/src/icon-button.tsx
+++ b/packages/components/button/src/icon-button.tsx
@@ -28,7 +28,7 @@ export interface IconButtonProps extends BaseButtonProps {
 }
 
 /**
- * Icon button renders an icon within in a button.
+ * Icon button renders an icon within a button.
  *
  * @see Docs https://chakra-ui.com/docs/components/icon-button
  */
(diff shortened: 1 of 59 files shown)


packages/components/button/src/icon-button.tsx
```diff
@@ -27,6 +27,11 @@ export interface IconButtonProps extends BaseButtonProps {
   "aria-label": string
 }
 
+/**
+ * Icon button renders an icon within in a button.
```
la55u, line 31: typo

PR #6678, opened 2022-09-12
fix(number-input): Fix `e` character logic.

### LINKED ISSUE #6676: Number Input accepts letter e
### Description

Number Input is accepting letter e and when i un foucus from the input it gives me this number -9007199254740991 at the input which is not editable 

### Link to Reproduction

https://chakra-ui.com/docs/components/number-input/usage

### Steps to reproduce

1. Go to chakra Ui number input  example
2. Click on the input and type letter e
3. then un focus on the input by clicking somewhere else 
4. See  error

### Chakra UI Version

2.2.9

### Browser

Google Chrome

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

_No response_

anubra266:
Fixes #6676

**Info**
`E/e` is only valid syntax when in the middle of numbers. E.g. 1e2 resolves to 100, 2e4 resolves to 20000
This PR updates the number input logic to prevent entering stuff like e12, 12e, or just `e` cause they're invalid syntax.

Previously entering `e` resolved to this long negative number. That number is actually the `min` number the input can accept, but because the input understood that just `e` is invalid, it automatically clamped to that `minimum` number.


Patrick-Ullrich:
See my comment: https://github.com/chakra-ui/chakra-ui/issues/6676#issuecomment-1245336139

This is a breaking change imho.

anubra266:
> Patrick-Ullrich

@Patrick-Ullrich I updated the PR description, it can't be a breaking change because the previous behaviour was invalid syntax.

PR #6679, opened 2022-09-12
Support motion overrides

### LINKED ISSUE #5843: Update Popover to support styling (or removing) animations
### Description

The current Popover system has a baked in animation / transition for show/hide. This makes is very difficult for a developer to modify it. Having an API with which to provide animations to would be extremely helpful, and having such a system would also allow for turning _off_ the transition as well. 

For a given "copy this existing site" implementation, I need to remove the transition animations from Popover.

I would expect someone who is trying to match a specific transition / animation style would also run into this problem. 

### Proposed Solution or API

unsure, I can't even find in the ChakraUI code where the transition animation is set. 

### Alternatives

Asking on the Discord channel suggested adding hardcoded CSS overrides,  `opacity="1 !important"` and `transform="none !important"` as the only way to override the existing behavior. 

### Additional Information

_No response_

### LINKED ISSUE #5883: DrawerContent does not properly forward needed props for controlling the animation
### Description

When I add a `transition` prop to the `<DrawerContent>`, I expect the `DrawerContent` to pass that through to the Slider / motion.div, but it seems like it gets swallowed at the `StyledComponent` level, possibly due to naming conflicts.

### Link to Reproduction

https://codesandbox.io/s/chakra-ui-javascript-lzzg9?file=/src/index.js

### Steps to reproduce

1. Recreate the sandbox link code in a place where you can use React Devtools
2. Open the Drawer by clicking the button
3. Inspect the DrawerContent and subsequent Slide component.
4. Observe that the `transitionEnd` prop is passed through to the MotionComponent (motion.div), but the `transition` prop is swallowed at the parent StyledSlide factory component.

### Chakra UI Version

1.8.7

### Browser

Chrome v100

### Operating System

- [X] macOS
- [X] Windows
- [ ] Linux

### Additional Information

I would assume that the styled component is swallowing the `transition` prop, expecting it to be a CSS value, and not an object prop expecting to be passed down to children. The way the DrawerContent component is written, we cannot directly pass in a `variants` prop like the `popoverContent` and `ModalContent` can.

The Slide component's motion.div will never allow for overwriting the transition animation if it cannot receive it. I can, of course, pass through a whole new `variants` prop, but all of this is undocumented functionality that I discovered through hours of digging through the source code, just trying to change the top height of where the drawer renders.

I think if this actually worked, and there was adequate documentation to show devs how to use it, the way you've setup the animation passthrough to the variant via props is a neat way to go. But the naming could use some help AND the fact that it doesn't seem to work at all is unfortunate.

Parent Slide styled component:
image

Child Slide root component:
image

### LINKED ISSUE #6189: Allow control of Framer transitions
### Description

At present Framer motion transitions are effectively hardcoded. We have no access to the underlying component props.

Setting `transitionDuration` on `<Modals />` and `<Drawers />` seems to impact the enter transition, but not the exit. It’s also quite janky when applying either of these.

Aside from the obvious cosmetic reasoning and brand integration here, for a11y we currently cannot support `prefers-reduced-motion`.

### Proposed Solution or API

Pass `motionProps` to the underlying Framer motion transition component.

### Alternatives

An alternative approach that’s heavier handed might be to remove Framer entirely and make use of native CSS transitions, which developers can then control via the theme system.

### Additional Information

_No response_

### LINKED ISSUE #6449: Transition duration props to Accordion
### Description

I know that the `Collapse` component is able to receive `transition` as a prop and change the enter and exit animation duration, but I am not able to do that on the `AccordionPanel` component, which is using `Collapse` under the hood.

This will allow us to change the transition speed with which an `Accordion` is opened and closed

### Proposed Solution or API

I would like to be able to pass `transition` as a prop to the at least one of the `AccordionPanel/AccordionItem/Accordion` components and be able to change the speed of showing the Accordion.

### Alternatives

_No response_

### Additional Information

_No response_

segunadebayo:
Closes #6449
Closes #5843
Closes #6189
Closes #5883

## 📝 Description

This PR adds support for overriding framer motion props across components.

It's almost impossible to override the framer motion config for menu, popover, tooltip, etc.

## 🚀 New behavior

Motion-related elements now have `motionProps` prop that allows users to have full control over the styling.

No

### DIFF
--- a/.changeset/lazy-ants-attack.md
+++ b/.changeset/lazy-ants-attack.md
@@ -7,4 +7,4 @@
 "@chakra-ui/tooltip": minor
 ---
 
-Allow control of framer-motion elements
+Allow control of `framer-motion` elements via the `motionProps` prop.
--- a/.changeset/lazy-ants-attack.md
+++ b/.changeset/lazy-ants-attack.md
@@ -1,0 +1,10 @@
+---
+"@chakra-ui/accordion": minor
+"@chakra-ui/checkbox": minor
+"@chakra-ui/menu": minor
+"@chakra-ui/modal": minor
+"@chakra-ui/popover": minor
+"@chakra-ui/tooltip": minor
+---
+
+Allow control of framer-motion elements
--- a/.changeset/ninety-pugs-suffer.md
+++ b/.changeset/ninety-pugs-suffer.md
@@ -2,4 +2,4 @@
 "@chakra-ui/popover": patch
 ---
 
-Refactor popover
+Refactor popover to reduce bundle size
--- a/.changeset/perfect-cycles-argue.md
+++ b/.changeset/perfect-cycles-argue.md
@@ -2,4 +2,6 @@
 "@chakra-ui/tooltip": patch
 ---
 
-Fixed the wrapping children display + changed mouse events with pointer events
+Fixed issue where disabled tooltip triggers require an extra wrapper (via
+`shouldWrapChildren). This was fixed by switching from mouse events to pointer
+events
--- a/.changeset/strange-guests-deliver.md
+++ b/.changeset/strange-guests-deliver.md
@@ -1,5 +1,0 @@
----
-"@chakra-ui/modal": minor
----
-
-Add support for `motionProps` in components that use `framer-motion`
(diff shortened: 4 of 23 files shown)


mix3d:
I think this might close https://github.com/chakra-ui/chakra-ui/issues/5883 as well

Issue #6697, opened 2022-09-15
Use `import { m } from "framer-motion"` 


jigsawpieces:
### Description

There have previous discussions in https://github.com/chakra-ui/chakra-ui/issues/4975 about ways to reduce the bundle size of Chakra UI.

One of the issues related to this is that `framer-motion` is quite large. According to their docs, one way of trimming their bundle is by:
> Instead of importing motion, import the slimmer `m` component.
> `import { m } from "framer-motion"`
> `m` is used in the exact same way as motion, but unlike `motion`, the `m` component doesn't come preloaded with features like animations, layout animations, or the drag gesture.

See: https://www.framer.com/docs/guide-reduce-bundle-size/#how-to-reduce-the-size-of-the-motion-component

However it looks like this strategy is not being used here, and most instances I'm seeing the larger `{ motion }` being imported, e.g. here: https://github.com/chakra-ui/chakra-ui/blob/3a87f68474221ed402fe5d169f7067763c8436d5/packages/components/modal/src/modal-overlay.tsx#L9

Is there any reason/context for this, and if not, shall I raise a PR to improve how `framer-motion` is being imported in Chakra?

### Problem Statement/Justification

This is the approach suggested by `framer-motion` in order to reduce bundle size of their library.

### Proposed Solution or API

Replace instances where we're doing `import { motion } from "framer-motion"` with `import { m } from "framer-motion"`. Draft PR opened: https://github.com/chakra-ui/chakra-ui/pull/6698

jigsawpieces:
Can see that this has already been attempted https://github.com/chakra-ui/chakra-ui/pull/3488

PR #6698, opened 2022-09-15
refactor(motion-import): lazy load framer-motion 

jigsawpieces:
Closes https://github.com/chakra-ui/chakra-ui/issues/6697

## 📝 Description
Lazy load a subset of `framer-motion` feature in order to reduce overall bundle size of Chakra UI. See: https://www.framer.com/docs/lazy-motion/ and https://www.framer.com/docs/guide-reduce-bundle-size/

Currently we are using `import { motion } from "framer-motion"` which is supposedly a larger component than `m`. This PR replaces this with `import { m }`, and wraps all usage of `m` with a `<LazyMotion>` component, as per recommendations in https://www.framer.com/docs/lazy-motion/

## 🚀 New behavior
Use `import { m } from "framer-motion"` instead, and wrap all usage with `<LazyMotion features={domAnimation} />`

No

### DIFF
--- a/.changeset/serious-dodos-sniff.md
+++ b/.changeset/serious-dodos-sniff.md
@@ -1,0 +1,12 @@
+---
+"@chakra-ui/checkbox": minor
+"@chakra-ui/menu": minor
+"@chakra-ui/modal": minor
+"@chakra-ui/popover": minor
+"@chakra-ui/popper": minor
+"@chakra-ui/toast": minor
+"@chakra-ui/tooltip": minor
+"@chakra-ui/transition": minor
+---
+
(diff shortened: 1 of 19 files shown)


davidkhierl:
@jigsawpieces just wondering why this got closed. I think this is a good idea to swap out the `motion` to `m` and lazy load as there is no need to import all features from `framer-motion`

1997roylee:
Love to see the change, why did it closed?

jigsawpieces:
@davidkhierl @1997roylee it turned out this has already been attempted and the savings were very minimal: https://github.com/chakra-ui/chakra-ui/pull/3488#issuecomment-1068363607

PR #6721, opened 2022-09-19
fix(tooltip): add closeOnScroll prop

### LINKED ISSUE #61: UI elements look completely broken / don't work
Hey so I followed the getting started guide but when I try to use some components they just don't work. This might just be me being newish to reactjs, but for example when I create a Button it is really thin and the text flows out. When I try to do an input I get an error saying "blue is not defined" or something like that. I can provide more info if needed but does anyone know what I might be doing wrong?

### LINKED ISSUE #5755: Optional prop for disabling/shutting down a tooltip when scrolling
### Description

If an user has the mouse over a tooltiped element, and then starts scrolling, the tooltip scrolls through the page (see the video)

https://user-images.githubusercontent.com/88422056/159212705-7af7cda0-6fd6-476f-bb49-b5c22e1aa5e2.mp4

It would save some time on the developer side and help us avoid bigger edge cases like when the tooltip scrolling causes an overflow on the screen (see the video)

https://user-images.githubusercontent.com/88422056/159213399-22765349-622b-491d-b692-d851971a8ff5.mp4

### Proposed Solution or API

`<Tooltip label={yourLabel} hideOnScroll={true}>`

### Alternatives

I could manually check scrolling and then using the `isDisabled` prop to hide the tooltip

### Additional Information

I opened this as a feature request instead of a bug cause I'm pretty sure this is on the browser side and how it handles mouseover state changes

anubra266:
Closes #61
Closes #5755

Added `closeOnScroll` prop to allow closing tooltip when the page scrolls.


.changeset/selfish-poems-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@chakra-ui/tooltip": patch
+---
+
+Add `closeOnScroll` prop.
```
anubra266, line 5:
```suggestion
Added closeOnScroll prop to allow closing tooltip when the page scrolls.
```

PR #6722, opened 2022-09-19
fix(avatar): wrong avatar styling

### LINKED ISSUE #4304: Why are there some very tiny borders around the Avatar?
### Description

When I am using the Avatar component and refresh the page several times, there is a high probability that a very small border color will appear around the avatar. However, when I add a little style to the Avatar component, the small border above Disappeared!

### Link to Reproduction

https://codesandbox.io/s/condescending-bogdan-tuse5?file=/src/App.js

### Steps to reproduce

1. Enter the page to carefully observe whether there is a small frame, if it does not appear, refresh the page several times, if it still does not appear, refresh the codesendbox, so that it can be reproduced! It seems that this component of the official website also has this problem! You have to be very careful to see this subtle difference!

### Chakra UI Version

1.6.4

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

image
Take a closer look at the border in the picture above. Are there any very subtle colors?

But if I refresh the page many times, this very subtle border may disappear. Look at the picture below. Does anyone know what the problem is? ?

image

anubra266:
Fixes #4304
Fix issue where background remained after image is loaded

### DIFF
--- a/packages/components/avatar/src/avatar.tsx
+++ b/packages/components/avatar/src/avatar.tsx
@@ -7,12 +7,14 @@ import {
   ThemingProps,
   useMultiStyleConfig,
 } from "@chakra-ui/system"
-import { cx } from "@chakra-ui/shared-utils"
+import { cx, dataAttr, runIfFn } from "@chakra-ui/shared-utils"
 import { AvatarStylesProvider } from "./avatar-context"
 import { AvatarImage } from "./avatar-image"
 import { GenericAvatarIcon } from "./generic-avatar-icon"
 import { initials } from "./avatar-name"
 import { AvatarOptions } from "./avatar-types"
+import { UseImageProps } from "@chakra-ui/image"
+import { useState } from "react"
 
 export const baseStyle: SystemStyleObject = {
   display: "inline-flex",
@@ -43,6 +45,7 @@ export interface AvatarProps
  */
 export const Avatar = forwardRef<AvatarProps, "span">((props, ref) => {
   const styles = useMultiStyleConfig("Avatar", props)
+  const [isLoaded, setIsLoaded] = useState(false)
 
   const {
     src,
@@ -51,6 +54,7 @@ export const Avatar = forwardRef<AvatarProps, "span">((props, ref) => {
     showBorder,
     borderRadius = "full",
     onError,
+    onLoad: onLoadProp,
     getInitials = initials,
     icon = <GenericAvatarIcon />,
     iconLabel = " avatar",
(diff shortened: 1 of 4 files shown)


packages/components/avatar/src/avatar.tsx
```diff
@@ -61,6 +65,11 @@ export const Avatar = forwardRef<AvatarProps, "span">((props, ref) => {
     ...rest
   } = omitThemingProps(props)
 
+  const onLoad: UseImageProps["onLoad"] = (e) => {
+    runIfFn(onLoad, e)
+    setIsLoaded(true)
+  }
+
```
segunadebayo, line 72: You can make this more concise by passing it directly to the onLoad prop

```suggestion
  const onLoad = callAllHandlers(onLoad, () => {
    setIsLoaded(true)
  })
```

packages/components/theme/src/components/avatar.ts
```diff
@@ -35,7 +35,9 @@ const baseStyleContainer = defineStyle((props) => {
   const borderColor = mode("white", "gray.800")(props)
 
   return {
-    bg,
+    "&:not([data-loaded])": {
+      bg,
+    },
```
segunadebayo, line 40: Please use a css-var here. In case a user is already overriding the bg from the theme.

Assign the value to a CSS var and use `bg: $avatarBg.reference` at the root.

packages/components/avatar/src/avatar.tsx
```diff
@@ -61,6 +65,9 @@ export const Avatar = forwardRef<AvatarProps, "span">((props, ref) => {
     ...rest
   } = omitThemingProps(props)
 
+  const onLoad = callAllHandlers(onLoad, () => {
```
segunadebayo, line 68: Can you put this inline (directly in the onLoad prop below?

packages/components/theme/src/components/avatar.ts
```diff
@@ -35,7 +37,9 @@ const baseStyleContainer = defineStyle((props) => {
   const borderColor = mode("white", "gray.800")(props)
 
   return {
-    bg,
+    "&:not([data-loaded])": {
+      bg: $avatarBg.reference,
+    },
```
segunadebayo, line 42:
```suggestion
    bg: $avatarBg.reference,
    "&:not([data-loaded])": {
      [$avatarBg.variable]: bg,
    },
```

packages/components/theme/src/components/avatar.ts
```diff
@@ -24,6 +24,8 @@ const baseStyleExcessLabel = defineStyle((props) => {
   }
 })
 
+const $avatarBg = cssVar("avatar-background")
+
 const baseStyleContainer = defineStyle((props) => {
   const { name, theme } = props
   const bg = name ? randomColor({ string: name }) : "gray.400"
```
segunadebayo, line 31:
```suggestion
  const bg = name ? randomColor({ string: name }) : "colors.gray.400"
```

PR #6763, opened 2022-09-28
feat: expose `setValue` from `useClipboard`

### LINKED ISSUE #6759: [Request] Expose `setValue` method from `useClipboard` hook
### Description

Please expose a new `setValue` method from the hook so users can update the value the `onCopy` method will copy to the user's clipboard

There are frequently times where a page has multiple components that can all be copied to a clipboard. The `useClipboard` hook doesn't expose a way to change that value and, as a result, the dev is forced to rely on a re-render to execute for a `useState` value to update before the clipboard will work OR the dev will need to abstract the copy-able value into a new component so multiple hooks can exist on one page.

### Proposed Solution or API

``` typescript
const defaultValue = 'some string';
const { onCopy, setValue } = useClipboard(defaultValue);

// ...

const updatedValue = 'some new string';
setValue(updatedValue);
onCopy();
// updatedValue copied to clipboard
```

### Alternatives

Moving all relevant copy-able elements to their own components (ideal from an abstraction standpoint, but often unnecessary if the original file is trivially small)

### Additional Information

_No response_

noobinthisgame:
Closes #6759 

## 📝 Description

Exposing a new function from `useClipboard` to enable the control over the value even after first initialization

With the current behaviour `useClipboard` doesn't offer the any way to set the value copied after the first initialization of the hook.

## 🚀 New behavior

The `useClipboard` hook now exposes a `setValue` function that enables the control over the copied value even after the first initialisation

```
const {setValue} = useClipboard('initial value')

onCopy() // copies 'initial value'

setValue('new value')

onCopy() // copies 'new value'

```

No

### DIFF
--- a/packages/legacy/hooks/src/use-clipboard.ts
+++ b/packages/legacy/hooks/src/use-clipboard.ts
@@ -15,7 +15,7 @@ export interface UseClipboardOptions {
 /**
  * React hook to copy content to clipboard
  *
- * @param text the text or value to copy
+ * @param initialValue the text or value to copy
  * @param {Number} [optionsOrTimeout=1500] optionsOrTimeout - delay (in ms) to switch back to initial state once copied.
  * @param {Object} optionsOrTimeout
  * @param {string} optionsOrTimeout.format - set the desired MIME type
@@ -24,10 +24,12 @@ export interface UseClipboardOptions {
  * @see Docs https://chakra-ui.com/docs/hooks/use-clipboard
  */
 export function useClipboard(
-  text: string,
+  initialValue: string,
   optionsOrTimeout: number | UseClipboardOptions = {},
 ) {
   const [hasCopied, setHasCopied] = useState(false)
+
+  const [value, setValue] = useState(initialValue)
 
   const { timeout = 1500, ...copyOptions } =
     typeof optionsOrTimeout === "number"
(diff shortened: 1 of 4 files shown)


packages/legacy/hooks/src/use-clipboard.ts
```diff
@@ -24,20 +24,22 @@ export interface UseClipboardOptions {
  * @see Docs https://chakra-ui.com/docs/hooks/use-clipboard
  */
 export function useClipboard(
-  text: string,
+  initialValue: string,
   optionsOrTimeout: number | UseClipboardOptions = {},
 ) {
   const [hasCopied, setHasCopied] = useState(false)
 
+  const [value, setValue] = useState(initialValue)
```
segunadebayo, line 32: Can we use a `ref` here instead of state?

```js
  const valueRef = useRef(initialValue)
```

and then write to the ref in the setValue method

```js
const setValue = ( newValue: string ) => {
   valueRef.current = newValue
}
```

noobinthisgame: Yes absolutely :)

noobinthisgame: Actually after I tried to use `useRef` instead of `useState` all ways I found ended up with the value not beeing updated properly. So I'd stick with `useState`, even though we got the rerender. @segunadebayo

yourtallness:
With `setValue` being async, there's no way to copy the new value in `onCopy` because `value` still contains the old value at this point.

It'd be nice if `setValue` could have a callback or if `onCopy` could optionally take a value as an argument.

This would allow e.g. fetching the value from an API call, setting it & also copy it to the clipboard.

PR #6786, opened 2022-10-03
Expose toast options in `render`

### LINKED ISSUE #6613: pass options to a custom component to be passed in render of `UseToastOption`
### Description

Currently only id and onClose can be passed to render props.
ref: https://chakra-ui.com/docs/components/toast/props#:~:text=%22bottom%22-,render,-Description
I want to pass options to a custom component.
https://github.com/chakra-ui/chakra-ui/blob/7d7e04d53d871e324debe0a2cb3ff44d7dbf3bca/packages/components/toast/src/toast.tsx?plain=1#L85

I am attempting to customize the Toast rendered by useToast.
Options can be passed when calling toast function, like this `toast({ title: 'Account created.', status: 'success' })` ,
but options are not passed when using custom components with render, so options cannot be used to display dynamic messages.

### Proposed Solution or API

It would be good if it could be passed as a second argument as follows.
`render?(props: RenderProps, options: UseToastOptions): React.ReactNode`

### Alternatives

_No response_

### Additional Information

_No response_

anubra266:
Closes #6613

Expose options to toast `render`

### DIFF
--- a/.changeset/.changeset/pink-rockets-pull.md
+++ b/.changeset/.changeset/pink-rockets-pull.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/toast": patch
+---
+
+Expose original toast options to toast `render` pptions
--- a/.changeset/.changeset/pink-rockets-pull.md
+++ b/.changeset/.changeset/pink-rockets-pull.md
@@ -2,4 +2,4 @@
 "@chakra-ui/toast": patch
 ---
 
-Expose original toast options to toast `render` pptions
+Expose original toast options to toast `render` options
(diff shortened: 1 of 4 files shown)


packages/components/toast/src/toast.tsx
```diff
@@ -87,7 +87,7 @@ export function createRenderToast(
   const { render, toastComponent: ToastComponent = Toast } = options
   const renderToast: React.FC<RenderProps> = (props) => {
     if (typeof render === "function") {
-      return render(props) as JSX.Element
+      return render({ ...props, options }) as JSX.Element
```
segunadebayo, line 90: What we want here is for the options to be inlined, not as a dedicated key


```suggestion
      return render({ ...props, ...options }) as JSX.Element
```

.changeset/.changeset/pink-rockets-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@chakra-ui/toast": patch
+---
+
+Expose original toast options to toast `render` pptions
```
segunadebayo, line 5:
```suggestion
Expose original toast options to toast `render` options
```

PR #6808, opened 2022-10-10
fix: allow SystemStyleFunction to accept a user provided type for props

### LINKED ISSUE #6796: Allow extending the type of SystemStyleFunction props in order to support custom component props
### Description

Today, if I we have to write styles for custom components into the theme we can do so by using `defineStyleConfig`. Here's an example.
```ts
const StatsCardStyles = defineStyleConfig({
  baseStyle: {
    backgroundColor: 'gray.50',
  },
  variants: {
    highlighted(props: StyleFunctionProps) {
      return {
        backgroundColor: 'yellow.50'
      }
    },
  },
});
```

Now, over the components I have created, it's pretty common use case to require a custom prop from the component inside the variant function to decide on styles.

Example : 
```ts
variants: {
  highlighted(props: StyleFunctionProps) {
    const { intent } = props;
    let backgroundColor;
    if (intent === 'positive') {
      backgroundColor = 'green.50';
    } else if (intent === 'negative') {
      backgroundColor = 'red.50';
    }
    return {
      backgroundColor,
    };
  },
},
```
Here, `intent` is a prop that that the `StatsCard` component accepts and passes it to the style system as follows.
```ts
const styles = useStyleConfig('StatsCard', { intent });
```
With the current implementation of the `StyleSystemFunction` type, it's possible to get this done without any error too because the `StyleFunctionProps` adds `[key: string]: any;` for any additional props the user may pass.

While this works, it doesn't provide type completion for the props of the component. The DX could be much better if we could somehow pass additional types about the props that the variant function receives. 

- Improve DX of creating a variant for a custom component
- Let the user provide type information for the props they can pass to the style system via the `useStyleConfig` hook. 

### Proposed Solution or API

We can change the signature of the `SystemStyleFunction` to let the user pass on additional type information and have TS intellisense do a better job. 
```ts
export type SystemStyleFunction<T extends StyleFunctionProps> = (
  props: T,
) => SystemStyleObject

export type SystemStyleInterpolation = SystemStyleObject | SystemStyleFunction<StyleFunctionProps>
```

This let the user do something like this in the variant function.
```ts
variants: {
  highlighted(props: StyleFunctionProps & StatsCardProps) {
    const { intent } = props;
    let backgroundColor;
    if (intent === 'positive') {
      backgroundColor = 'green.50';
    } else if (intent === 'negative') {
      backgroundColor = 'red.50';
    }
    return {
      backgroundColor,
    };
  },
}
```

This helps with autocompletion. 
image

### Alternatives

Haven't considered other alternatives yet. 

### Additional Information

_No response_

praveenpuglia:
Closes #6796 

## 📝 Description

The type update allows the consumer of the variant functions to add additional type information for the props received from inside the component

The current `SystemStyleFunction` signature doesn't let the consumer pass on additional type information about the props it can receive. 

## 🚀 New behavior

The PR allows the user to pass on any additional props type apart from `StyleFunctionProps`. 

No

### DIFF
--- a/packages/components/styled-system/src/define-styles.ts
+++ b/packages/components/styled-system/src/define-styles.ts
@@ -12,11 +12,13 @@ export type StyleFunctionProps = {
   [key: string]: any
 }
 
-export type SystemStyleFunction = (
-  props: StyleFunctionProps,
+export type SystemStyleFunction<T extends StyleFunctionProps> = (
+  props: T,
 ) => SystemStyleObject
 
-export type SystemStyleInterpolation = SystemStyleObject | SystemStyleFunction
+export type SystemStyleInterpolation =
+  | SystemStyleObject
+  | SystemStyleFunction<StyleFunctionProps>
 
 // ------------------------------------------------------------------ //
 
--- a/.changeset/neat-pets-work.md
+++ b/.changeset/neat-pets-work.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/styled-system": patch
+---
+
+Enhance SystemStyleFunction type to be extensible and allow consumer to pass
+custom types for props


packages/components/styled-system/src/define-styles.ts
```diff
@@ -12,11 +12,13 @@ export type StyleFunctionProps = {
   [key: string]: any
 }
 
-export type SystemStyleFunction = (
-  props: StyleFunctionProps,
+export type SystemStyleFunction<T extends StyleFunctionProps> = (
```
ajitjha393, line 15: Since we know that SystemStyleFunction can only accept T is superset of `StyleFunctionProps` otherwise it will default to `StyleFunctionProps`
Hence we can go for  - 

export type SystemStyleFunction<T extends StyleFunctionProps = StyleFunctionProps>...

packages/components/styled-system/src/define-styles.ts
```diff
@@ -12,11 +12,13 @@ export type StyleFunctionProps = {
...
+  props: T,
 ) => SystemStyleObject
 
-export type SystemStyleInterpolation = SystemStyleObject | SystemStyleFunction
+export type SystemStyleInterpolation =
+  | SystemStyleObject
+  | SystemStyleFunction<StyleFunctionProps>
```
ajitjha393, line 21: Then we don't need to pass StyleFunctionProps as a Generic here

praveenpuglia: correct.

packages/components/styled-system/src/define-styles.ts
praveenpuglia: Makes sense. Will do this change.

segunadebayo, line 15:
```suggestion
export type SystemStyleFunction<T extends StyleFunctionProps = StyleFunctionProps> = (
```

praveenpuglia: So I and @ajitjha393 discussed this change. For me, on webstorm, I didn't get any TS errors but vscode rightly pointed it out. Just doing this will break existing API.

praveenpuglia: image

praveenpuglia: If I understand it correctly, there's no good way to pass the type information received from the dynamically named variant function to  `SystemStyleInterpolation`. Is that correct @ajitjha393 ?

ajitjha393: Yeah as of now, I can't think of an easy way we can capture the type Paramter and make `SystemStyleInterpolation`  generic in `[variant: string] ?: SystemStyleInterpolation`

ajitjha393: https://github.com/chakra-ui/chakra-ui/issues/6796#issuecomment-1281831615

praveenpuglia:
Closing this since there's no real solution at the moment.

PR #6863, opened 2022-10-20
fix(focus): fixed accidental focus thefts from overly eager transitionend handler

Andarist:
## 📝 Description

This is an attempt to fix focus theft when the focus is intentionally moved away from the open element and even when it is moved outside of the current window.

VS Code open with two panels side-by-side. Source code on the left, XState extension on the right

Notice that I have a focus on the popover on the right. We manage the open state of that popover ourselves and we don't use focus trapping in it.

I've noticed that sometimes when I move the focus to the source code on the left then I quickly lose my cursor in the code. I've investigated the issue and it turned out that the popover element was being programmatically focused by Chakra after I manually moved it outside of that window.

After further investigation, I've managed to narrow it down to this `focus` call:
https://github.com/Andarist/chakra-ui/blob/5af9b7dddd0e1c7788ee9459246909758da3adc1/packages/legacy/hooks/src/use-focus-on-show.ts#L53
that was called by a `transitionend` handler here:
https://github.com/Andarist/chakra-ui/blob/5af9b7dddd0e1c7788ee9459246909758da3adc1/packages/legacy/hooks/src/use-focus-on-show.ts#L62

## 🚀 New behavior

I've introduced a ref dance to ensure that the `autoFocus` logic only gets applied after we change from `visible: false` to `visible: true`. If the component stays visible I don't think that `focus` should ever be called from here - the hook is called `useFocusOnShow`, so it should only ever be called on show.

No

### DIFF
--- a/.changeset/small-colts-fry.md
+++ b/.changeset/small-colts-fry.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/react-use-focus-effect": patch
+"@chakra-ui/hooks": patch
+---
+
+Fixed an issue with focus sometimes being moved back to the open element when it
+was intentionally moved away from it.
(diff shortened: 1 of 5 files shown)


segunadebayo (approved):
This is great, @Andarist. It ultimately smells like a problem a "state machine"  would solve.

Thanks for catching this. Reviewed the code, and it looks good to me.

PR #6886, opened 2022-10-23
Feat/vertical button group

### LINKED ISSUE #6885: Attached Vertical Button Group
### Description

I would like to propose adding support for the `isVertical` prop on the `ButtonGroup` component.

Currently, this is not supported and when changing the `flexDir="column"` there are issues where the first and last buttons of a group have border radius' intended for horizontally flexed buttons. Please see the codesandbox here for an example: https://codesandbox.io/s/vertical-attached-button-group-hwzk95?file=/src/index.tsx

### Proposed Solution or API

I would propose adding an `isVertical` prop to group to switch the button group's flex direction:
```tsx
<ButtonGroup isAttached isVertical>
    <Button>Button 1</Button>
    <Button>Button 2</Button>
    <Button>Button 3</Button>
    <Button>Button 4</Button>
</ButtonGroup>
```

### Alternatives

I have tried just changing the `flexDir` on the button group, but the border radius' of the first and last button is still out-of-whack.

### Additional Information

Screen Shot 2022-10-23 at 7 54 34 AM

Here is an example of vertically laid out buttons with the `isVertical` prop.

SpencerDuball:
Closes #6885

## 📝 Description

Adds the `isVertical` prop to `ButtonGroup` component that allows for vertically orientated button groups.

In order to vertically orientate button groups a user would need to specify `flexDir="column"` currently, however even doing so will still present challenges of incorrect border radius' for the first and last buttons. See the codesandbox below for an example: 
https://codesandbox.io/s/vertical-attached-button-group-hwzk95?file=/src/index.tsx:232-233

## 🚀 New behavior

With this feature PR, the `isVertical` prop has been added to `ButtonGroup`. This will update the border radius' correctly for vertical button groups and change the flex direction. This also updates the button theme as there was a prior request to collapse border's when buttons are flush together with `isAttached` prop.

No

### DIFF
--- a/packages/components/button/tests/button-group.test.tsx
+++ b/packages/components/button/tests/button-group.test.tsx
@@ -69,3 +69,55 @@ test("Should flush outline button", () => {
     marginInlineEnd: "",
   })
 })
+
+test("Should flush vertical button", () => {
+  const { getByText } = render(
+    <ButtonGroup isAttached isVertical>
+      <Button>Button 1</Button>
+      <Button>Button 2</Button>
+      <Button>Button 3</Button>
+      <Button>Button 4</Button>
+    </ButtonGroup>,
+  )
+  expect(getByText(/Button 1/i)).toHaveStyle({
+    borderBottomLeftRadius: "0",
+    borderBottomRightRadius: "0",
+  })
+  expect(getByText(/Button 2/i)).toHaveStyle({
+    borderRadius: "0px",
+  })
+  expect(getByText(/Button 3/i)).toHaveStyle({
+    borderRadius: "0px",
+  })
+  expect(getByText(/Button 4/i)).toHaveStyle({
+    borderTopLeftRadius: "0",
+    borderTopRightRadius: "0",
+  })
+})
+
+test("Should flush vertical outline button", () => {
+  const { getByText } = render(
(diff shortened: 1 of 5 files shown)


packages/components/button/tests/button-group.test.tsx
```diff
@@ -69,3 +69,55 @@ test("Should flush outline button", () => {
...
+    // marginBottom: "-1px",
+  })
+  expect(getByText(/Button 2/i)).toHaveStyle({
+    marginInlineEnd: "",
+    // marginBottom: "-1px",
+  })
+  expect(getByText(/Button 3/i)).toHaveStyle({
+    marginInlineEnd: "",
+    // marginBottom: "-1px",
+  })
```
SpencerDuball, line 118: I have tried a crazy amount of combinations to get `marginBottom: "-1px"` to pass and think I might benefit from a contributor's guidance on this. I can confirm in the story (also a committed file) that this is the correct CSS value present, but for the life of me cannot get this attribute to pass. I could not find a way to print out the styles clearly either, could someone give me a pointer on this property/tip to get this test working?

stianlaa:
Thanks for tackling this issue :tada:

PR #6891, opened 2022-10-23
chore: refactor theme - Select

anubra266:
chore: refactor theme - Select


packages/components/theme/src/components/select.ts
```diff
@@ -1,6 +1,7 @@
 import { selectAnatomy as parts } from "@chakra-ui/anatomy"
 import {
   createMultiStyleConfigHelpers,
+  cssVar,
   defineStyle,
 } from "@chakra-ui/styled-system"
 import { mode } from "@chakra-ui/theme-tools"
```
segunadebayo, line 7:
```suggestion
```

Issue #6829, opened 2022-10-14
Style props do not allow to override theme style


bertho-zero:
### Description

When I use the `px` prop on a button with variant link it has no effect.

### Link to Reproduction

https://codesandbox.io/s/zealous-browser-rgsbbq

### Steps to reproduce

By inspecting the 4th button we see that the padding is not applied.

The css rule `padding: 0` from the theme is listed after `padding-inline-start` and `padding-inline-end` from the `px` prop:

![image](https://user-images.githubusercontent.com/8525267/195885264-f14e25e2-5533-4674-bf42-96fdc513694e.png)

Shouldn't the rules coming from the props be after the rules from the theme to be able to override them?

### Chakra UI Version

2.3.5

### Browser

Brave 1.41.96

### Operating System

- [ ] macOS
- [ ] Windows
- [X] Linux

mortezasabihi:
Hey, I have checked this and I found out the issue is not related to padding, it's about height property, look at this:

Ghost styled button:
![image](https://user-images.githubusercontent.com/35567646/196001316-a5bb50fe-3218-44e8-8ead-e06622eb3adf.png)

Link styled button:
![image](https://user-images.githubusercontent.com/35567646/196001341-7db32d85-344b-4458-b088-b20be14e7d2e.png)

bertho-zero:
My problem is that I try to add horizontal padding via px in props and it is ignored, nothing to do with height or visual rendering.

bertho-zero:
It's this `Object.assign` that's the problem:
https://github.com/chakra-ui/chakra-ui/blob/16ba8cdd447be4d8a0a4a4eecc8de81fe04e637c/packages/components/system/src/system.ts#L52-L58

The `px:4` is in both `__css` and `styleProps`, and the `Object.assign` does not put styleProps values after those of __css, `Object.assign` already finding the key in the object it just replaces value without moving it at the end.

---

This test does not pass because padding is after `paddingInlineStart` and `paddingInlineEnd`:

```js
test("should respect priority order", () => {
  const result = toCSSObject({})({
    theme,
    __css: {
      px: 4,
      padding: 0,
    },
    px: 4,
  })

  // Can't test result without stringify because it sorts keys alphabetically
  expect(JSON.stringify(result, null, 2)).toMatchInlineSnapshot(`
    "{
      \\"padding\": \\"0px\\",
      \\"paddingInlineStart\\": \\"var(--chakra-space-4)\\",
[...]

PR #6905, opened 2022-10-26
fix: cascade style respects order

bertho-zero:
Closes https://github.com/chakra-ui/chakra-ui/issues/6829

## 📝 Description

Using `Object.assign` overrides existing properties and does not respect order of precedence.

The order of precedence is not respected.

## 🚀 New behavior

The order of precedence is respected.

No

### DIFF
--- a/packages/legacy/utils/src/object.ts
+++ b/packages/legacy/utils/src/object.ts
@@ -153,3 +153,25 @@ export const fromEntries = <T extends unknown>(entries: [string, any][]) =>
  */
 export const getCSSVar = (theme: Dict, scale: string, value: any) =>
   theme.__cssMap?.[`${scale}.${value}`]?.varRef ?? value
+
+/**
+ * Object.assign which moves properties to end of object rather than replacing them
+ * */
+export const assignAfter = (target: object, ...sources: any[]) => {
+  if (target == null) {
+    // TypeError if undefined or null
+    throw new TypeError("Cannot convert undefined or null to object")
+  }
+
+  const result = Object(target)
+  for (const nextSource of sources) {
+    if (nextSource == null) continue // Skip over if undefined or null
+    for (const nextKey in nextSource) {
+      if (!Object.prototype.hasOwnProperty.call(nextSource, nextKey)) continue
+      if (nextKey in result) delete result[nextKey]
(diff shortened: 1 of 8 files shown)


packages/legacy/utils/src/object.ts
```diff
@@ -153,3 +153,25 @@ export const fromEntries = <T extends unknown>(entries: [string, any][]) =>
  */
 export const getCSSVar = (theme: Dict, scale: string, value: any) =>
   theme.__cssMap?.[`${scale}.${value}`]?.varRef ?? value
+
+/**
+ * Object.assign which moves properties to end of object rather than replacing them
+ * */
+export const assignAfter = (target: object, ...sources: any[]) => {
```
TylerAPfledderer, line 160: I think you meant to stick this in `packages/utilities/object-utils` and not `packages/legacy/utils`? 😅

TylerAPfledderer (commented):
This is awesome, @bertho-zero! I can only provide one tiny comment. 😸

bertho-zero:
@TylerAPfledderer It's done, I moved the function to `object-utils`

.changeset/rich-singers-build.md
```diff
@@ -2,4 +2,5 @@
 "@chakra-ui/radio": patch
 ---
 
-Update type of `value` and `defaultValue` to `string` instead of `string | number`. This reflects the internal implementation
+Update type of `value` and `defaultValue` to `string` instead of
+`string | number`. This reflects the internal implementation
```
bertho-zero, line 6: Why this change after a pull ? :thinking:

k-nearest-neighbor:
After an hour or so tracking down this bug, I see there is a PR that fixes it!
But it is waiting to be merged for 30+ days, and not clear when it will be deployed.

segunadebayo (approved):
Thanks for working on this @bertho-zero.

It aligns with the initial design I had in mind.

I'm a bit weary it might cause some breaking changes for some users, but the solution here is fixes a bug in the system.

Thanks again.

bertho-zero:
Thanks for the merge @segunadebayo, there is another system/styled-system related issue you might be interested in https://github.com/chakra-ui/chakra-ui/issues/7136

PR #6906, opened 2022-10-26
fix: Omit keys by default if keys are not set

TrashF:
This fixes a case where split utility function is called without keys, resulting in "keys is undefined" error message.

Closes # 

## 📝 Description

Stumbled upon a problem while just simply trying to use a `<Menu />` component.

Crashes with a long stack ending up to this object-utils class with an error message stating that keys is undefined.

## 🚀 New behavior

Using a `<Menu />` in my case works.

No.

## 📝 Additional Information

I have no desire to find out why split is getting called with undefined "keys" attribute. It somehow is and this change prepares for the case in a manner I assume is safe for all possible use cases.

### DIFF
--- a/packages/utilities/object-utils/src/index.ts
+++ b/packages/utilities/object-utils/src/index.ts
@@ -38,7 +38,7 @@ export function split<T extends Record<string, any>, K extends keyof T>(
   const omitted: Record<string, any> = {}
 
   for (const [key, value] of Object.entries(object)) {
-    if (keys.includes(key as T[K])) picked[key] = value
+    if (keys && keys.includes(key as T[K])) picked[key] = value
     else omitted[key] = value
   }
 


segunadebayo:
Hey @TrashF,

Can you provide a reproduction to that problem you faced where keys was undefined?

TrashF:
Hi @segunadebayo,

I'm afraid no. I tried reproducing it with an empty project and everything just works. In my actual project case that minor change is the only thing I changed to make it work.

```<Menu>
	<MenuButton
		as={IconButton}
		aria-label='Options'
		variant='outline'
		icon={<Avatar
			size='md'
			name={user.firstname + " " + user.lastname} />}
	/>
	<MenuList>
		<MenuItem icon={<SettingsIcon />}>
			{t("userMenu.settings")}
		</MenuItem>
		<MenuItem icon={<ExternalLinkIcon />} onClick={logout}>
			{t("userMenu.logOut")}
		</MenuItem>
	</MenuList>
</Menu>``` is the code in my troublesome component. It builds without errors, but clicking the button results in "keys is undefined" in browser. Stack trace shows some popper-ish functions so I assume it has something to do with menu positioning. Menu opens and positions itself correctly after my change.

TrashF:
Closed? Just look at the change and you should notice that it is just a common check that was lacking anyway.

TrashF:
Oh well, ended up fixing this with a @rollup/plugin-replace.

`replace({
	delimiters: ['', ''],
	preventAssignment: true,
	values: {
		"(keys.includes(": "(keys && keys.includes(",
	},
}),`

In times like these I remember why some people hate open source.

segunadebayo:
Hi @TrashF,

As I said, reproduction is required for us to act on a PR or issue. Just adding the code works, but we may need to get to the underlying issue.

Open source needs to be stewarded with caution and kindness. Using the word "hate" doesn't seem right, as open-source code is free to modify, fork and update. If you're unwilling to create a reproduction, we can't help much.

I'll find some time to dig into the root cause later though.

TrashF:
@segunadebayo You never said reproduction is a requirement, you asked if I can provide one which I declined. For me, this is a work project, which means extracting a part of it or using client paid time to provide reproduction is not really an option. Despite that, I came here with a PR and an intention to help.

Closing a pull request without commenting is not a form of kindness I know of, but sadly a very common practice nowadays.

Currently, I don't care anymore if the PR gets merged or not - I may or may not be able to remove the kludge from a build pipeline, but the problem I needed to get solved is solved. In terms of code quality, getting the fix merged would likely benefit both your product and my work project. I understand your point of caution, but not taking time to comment or investigate before closing your pull requests is not very friendly. I'm glad to hear you are going to take time to dig into the root cause later.

PR #6910, opened 2022-10-26
fix(pin-input): fix focus management

### LINKED ISSUE #6909: PinInput focus unexpected behavior when you clear it's value onComplete
### Description

I wanted to make request and clear PinInput and to focus on first input

### Link to Reproduction

https://codesandbox.io/s/spring-bash-mfucg1?file=/src/App.tsx

### Steps to reproduce

1. Fill PinInput
2. Trigger onComplete which fetches something and clears its value

### Chakra UI Version

2.3.5

### Browser

Google Chrome 106

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I was working on project that uses ChakraUI
My task wast to send request onComplete and clear `PinInput`'s value 
After fetching something with `async/await` i wanted to get focus on first input

After playing around i didn't manage to make this work

Motherships:
Closes #6909 

## 📝 Description

- Fixes focus when onComplete triggers onChange
- Adds focus management on copy and paste

- If you call onComplete with async/await fetch and clears itself it will lose focus and restrict you from entering new values without changing focus
- If you copy and paste two characters it wont fill itself
- If you copy and paste it wont move your focus

## 🚀 New behavior

- Changes `setValue` to check if index is smaller or the same as nextValue size, move focus to next input
- Checks if two or more characters were copy and pasted
- Moves focus to next input if it was not filled

I don't think so 

## 📝 Additional Information
-

### DIFF
--- a/packages/components/pin-input/src/use-pin-input.ts
+++ b/packages/components/pin-input/src/use-pin-input.ts
@@ -180,7 +180,10 @@ export function usePinInput(props: UsePinInputProps = {}) {
   const setValue = useCallback(
     (value: string, index: number, handleFocus: boolean = true) => {
       const nextValues = [...values]
-      nextValues[index] = value
+      // Index has to fit into our NextValues
+      const currentIndex =
+        nextValues.length >= index ? index : nextValues.length
+      nextValues[currentIndex] = value
       setValues(nextValues)
 
       const isComplete =
@@ -193,7 +196,7 @@ export function usePinInput(props: UsePinInputProps = {}) {
       if (isComplete) {
         onComplete?.(nextValues.join(""))
       } else {
-        if (handleFocus) focusNext(index)
+        if (handleFocus) focusNext(currentIndex)
       }
     },
     [values, setValues, focusNext, onComplete, descendants],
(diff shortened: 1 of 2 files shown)


packages/components/pin-input/src/use-pin-input.ts
```diff
@@ -237,7 +240,8 @@ export function usePinInput(props: UsePinInputProps = {}) {
         }
 
         // in the case of an autocomplete or copy and paste
-        if (eventValue.length > 2) {
+        // We can copy and pase two and more chars
```
anubra266, line 243:
```suggestion
        // We can copy and paste two or more characters
```

anubra266:
https://user-images.githubusercontent.com/30869823/202169226-da01a9a2-c4bc-4148-9512-6139b4ed947f.mp4

@Motherships this fix is not an ideal solution for the issue, because a user expects that when they enter a character into a particular input, it should be recorded into that input. As shown in the video above, this fix enters the character meant for the last input in the first input.

anubra266:
@Motherships to handle the focusing. Create a ref for the first field. Then whenever you clear, just focus it programmatically.

```ts
  const [value, setValue] = React.useState("");
  const firstField = React.useRef<HTMLInputElement>(null)
  const handleClear = async () => {
    console.log("triggers");
    setValue("");
    firstField.current?.focus()
  };
```

```html
<PinInput
 value={value}
 onChange={(v) => setValue(v)}
 onComplete={handleClear}
>
 <PinInputField ref={firstField} />
 <PinInputField />
 <PinInputField />
 <PinInputField />
</PinInput>
```

Motherships:
Actually, thats what I'm doing right now

I just thought that if we have automatic focus management - chakra has to do it for you

Thanks for helping out

PR #6935, opened 2022-11-03
feat(accordion): use zagjs based accordion

cschroeter:


segunadebayo:
Closing this based on our discussed strategy

natBizitza:
Thanks so much for such a great job! 
What will be the next step? When could we expect some changes for the accordion component, especially related to using an index as the primary state of component? Thank you!

PR #6945, opened 2022-11-07
fix(react): add types field to esm exports map

anddoutoi:
#### Edit: This might be bigger than this. It might be that this needs to be done to all modules.

Using `@chakra-ui/react` in a TypeScript project with `"type": "module"` in package.json and `"moduleResolution": "Node16"` in tsconfig.json causes a:
```
TS7016: Could not find a declaration file for module '@chakra-ui/react'. '/…/node_modules/@chakra-ui/react/dist/index.esm.js' implicitly has an 'any' type.
   Try `npm i --save-dev @types/chakra-ui__react` if it exists or add a new declaration (.d.ts) file containing `declare module '@chakra-ui/react';`
```
See screenshot under Current behavior.

See screenshot under New behavior for result if this PR is merged.

More information can be found in the following issue:
- https://github.com/microsoft/TypeScript/issues/50223

And specially in the following comments on above issue:
- https://github.com/microsoft/TypeScript/issues/50223#issuecomment-1222290089
- https://github.com/microsoft/TypeScript/issues/50223#issuecomment-1278044987

Screenshot 2022-11-07 at 12 13 55

## 🚀 New behavior
Screenshot 2022-11-07 at 12 16 50

No.


segunadebayo:
Good catch @anddoutoi. I'll update the other packages

anddoutoi:
Hey @segunadebayo, thanks for the merge ❤️.

This is a heads up comment and a FYI.

I found a disturbing thing when I continued my research on esm/modules/types and we might have been a bit too fast on the merge finger :( sorry.

But reading https://www.typescriptlang.org/docs/handbook/esm-node.html and scrolling to the end you'll find this screenshot:
Screenshot 2022-11-08 at 15 12 31

I have no clue what they mean with this. As it's not an array the order **should not matter** and would indicate poor judgment by the implementer of the package.json parser that TypeScript uses.

It might be nothing and it might work with this PR merged but the official docs states another thing.

anddoutoi:
Nevermind above comment. I tested installing `v0.0.0-dev-20221108124603` and it works.

PR #6968, opened 2022-11-11
feat(props-docs): use ts-morph instead

cschroeter:


segunadebayo:
After trying this several times, I think it's not worth the effort.

I'd recommend we improve the TS Doc comments or try another library to achieve this.

PR #6981, opened 2022-11-14
fix(icon): expose icon to theme

### LINKED ISSUE #6939: Icons have no way of touching defaultProps
### Description

When using the Icon component it does not seem to have any part in the theme object. This means we cannot alter the icon default props. Namely the height set to 1em. This should be auto.

Not every icon is perfectly square. And 1em on the height can chop it a bit. 

The width can be 1em but height should be auto.

### Link to Reproduction

no link

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Chakra UI Version

2.3.7

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

anubra266:
Closes #6939
Expose Icon component to theme


packages/components/icon/src/icon.tsx
```diff
@@ -27,7 +28,7 @@ const fallbackIcon = {
 
 export interface IconProps
   extends Omit<React.SVGAttributes<SVGElement>, keyof ChakraProps>,
-    ChakraProps {}
+    Omit<ChakraProps, "orientation"> {}
```
segunadebayo, line 31: Why the need for this? Just curious

anubra266: @segunadebayo it clashes with the typings for `styleConfig` decided to omit it since Icon doesn't need it.

.changeset/short-deers-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+"@chakra-ui/icon": patch
+---
+
+Hook Icon component to theme
```
segunadebayo, line 5:
```suggestion
Add support for styling the Icon component from the theme object.
```

PR #6982, opened 2022-11-14
feat: allow forced mode render element support

anubra266:
Add Support LightMode/DarkMode to render as element


segunadebayo:
Closing this for now. Don't think it was a good idea.

Issue #7006, opened 2022-11-17
Drawer headers do not provide a valid accessible name


lexanth:
### Description

When I render an open `Drawer`, I expect the a11y dialog to be labelled by the provided header. However, the header does not provide an accessible name.

The `ModalHeader` currently renders a `header` DOM element. As this is in a `div` in a portal, its context is the `body`, so it creates a `banner` element (see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#accessibility)). Banners are not named by their contents (see [WAI-ARIA](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#x5-4-naming-techniques)), so the accessible name of the `header` element is empty.

`Modal` and `AlertDialog` are slightly different here, as they currently render a `<section role="dialog">` or `<section role="alertdialog">`. Based on that, a `header` _should_ be labelled by its contents, but aXe happens to have (what I believe to be) a bug in identifying the `header` in a `section` with a `role` - https://github.com/dequelabs/axe-core/issues/3752

The current a11y unit tests for the modal package are currently giving false negatives, because they're not actually testing the portaled modal itself - it needs to scan `baseElement` rather than `container`.

### Link to Reproduction

https://codesandbox.io/s/flamboyant-moore-nyzywj?file=/src/index.tsx

### Steps to reproduce

1. Open the drawer
2. Run e.g. aXe devtools over it - the `role="dialog"` element does not have an accessible name

### Chakra UI Version

2.4.1

### Browser

Chrome 106

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Most other accessible dialogs I've seen do not use a `header` DOM element inside dialogs - they just use an h2
* https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/alertdialog.html
* https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/dialog
* https://mui.com/material-ui/react-dialog/

segunadebayo:
Closing based on https://github.com/chakra-ui/chakra-ui/pull/7007#issuecomment-1329548733

PR #7007, opened 2022-11-17
fix(Modal): make header an h2 so it is named by its content

lexanth:
Closes https://github.com/chakra-ui/chakra-ui/issues/7006

## 📝 Description

`header` renders a `banner` when scoped to the body, so it isn't named by its contents, so the `Drawer` `role="dialog"` element has no accessible name

`ModalHeader`, `DrawerHeader` and `AlertDialogHeader` render `header` DOM elements

## 🚀 New behavior

`ModalHeader`, `DrawerHeader` and `AlertDialogHeader` render `h2` DOM elements

No API change, no change in visual appearance. If users have snapshotted the DOM they might see it.

### DIFF
--- a/packages/components/modal/tests/alert-dialog.test.tsx
+++ b/packages/components/modal/tests/alert-dialog.test.tsx
@@ -67,9 +67,11 @@ it("renders an element with role='alertdialog' when opened", () => {
 })
 
 it("passes a11y test closed", async () => {
-  await testA11y(<BasicUsage />)
+  const { baseElement } = render(<BasicUsage />)
+  await testA11y(baseElement)
 })
 
 it("passes a11y test opened", async () => {
-  await testA11y(<BasicUsage isOpen />)
+  const { baseElement } = render(<BasicUsage isOpen />)
+  await testA11y(baseElement)
 })
--- a/.changeset/fast-owls-bake.md
+++ b/.changeset/fast-owls-bake.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/modal": patch
+---
+
+Change `ModalHeader` to render an `h2` so that it is named by its content
--- a/packages/components/modal/src/modal-header.tsx
+++ b/packages/components/modal/src/modal-header.tsx
@@ -9,7 +9,7 @@ import { useEffect } from "react"
 
 import { useModalContext, useModalStyles } from "./modal"
 
-export interface ModalHeaderProps extends HTMLChakraProps<"header"> {}
+export interface ModalHeaderProps extends HTMLChakraProps<"h2"> {}
 
 /**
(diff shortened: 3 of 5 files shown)


packages/components/modal/tests/alert-dialog.test.tsx
```diff
@@ -67,9 +67,11 @@ it("renders an element with role='alertdialog' when opened", () => {
...
 it("passes a11y test closed", async () => {
-  await testA11y(<BasicUsage />)
+  const { baseElement } = render(<BasicUsage />)
+  await testA11y(baseElement)
 })
 
 it("passes a11y test opened", async () => {
-  await testA11y(<BasicUsage isOpen />)
+  const { baseElement } = render(<BasicUsage isOpen />)
+  await testA11y(baseElement)
```
lexanth, line 76: These tests currently aren't actually catching the element we're trying to test, because they're portaled

Brennvo:
I'm curious; is there a benefit to keeping a landmark in a Modal as the header (hence the `<header>` implementation inside the modal), especially when the content behind the modal is intended to be inaccessible?

lexanth:
>I'm curious; is there a benefit to keeping a landmark in a Modal as the header (hence the <header> implementation inside the modal), especially when the content behind the modal is intended to be inaccessible?

Strictly a header can contain flow content (e.g. divs etc), whereas a heading (hX) should only contain phrasing content. I suspect this is related to why using a header is less good to act as the accessible name for the dialog/modal - it isn't "just text", it's a grouping of content.

segunadebayo:
@lexanth Right. Its purpose is to act as a group for the heading, close button, and maybe an icon.

I'm reluctant to merge this because it changes the underlying DOM element and might pose a breaking change to many other consumers.

If the Modal header only contains phrasing content (text), feel free to use the `as` prop to change the underlying element.

```jsx
<ModalHeader as="h2">Hello</ModalHeader>

// or

<DrawerHeader>
     <h2>Create your account</h2></DrawerHeader>
 <DrawerBody>
```

Thanks for understanding.

lexanth:
> Its purpose is to act as a group for the heading, close button, and maybe an icon.
None of the examples in the docs have the close button inside the `ModalHeader`/`DrawerHeader` - they have it alongside it

```jsx
<DrawerHeader>
     <h2>Create your account</h2>
 <DrawerHeader>
```
This pattern also doesn't work for providing a name for the dialog - the ID is on the header not the h2, so the h2's content doesn't label the dialog.

I'll work around this within our codebase using the `as="h2"` pattern, but it still stands that the a11y tests for all of the modal components are not actually testing anything. I've raised https://github.com/chakra-ui/chakra-ui/pull/7054 to address the unit test question

PR #7019, opened 2022-11-21
chore: replace tinycolor with color2k

anubra266:
eplace tinycolor with color2k for smaller bundle size.


morellodev:
This PR broke custom themes that uses the `darken` and `lighten` functions, since `tinycolor` [expects](https://github.com/scttcper/tinycolor/blob/master/src/index.ts#L415) the amount value to be in the range [0, 100] and `color2k` [expects](https://github.com/ricokahler/color2k/blob/main/src/darken.ts#L10) it to be in range [0, 1]

Issue #7027, opened 2022-11-24
The ellipsis does not work on Badge component


manakuro:
### Description

The ellipsis does not show up on the Badge component using `noOfLines` and `maxW`. It used to work on `"@chakra-ui/react": "^1.7.3"`.

It's probably because the inner `span` has `display: flex` style. It should work with `display: inline-box`.

### Link to Reproduction

https://codesandbox.io/s/pensive-spence-3pd41y?file=/src/index.js

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Chakra UI Version

2.4.1

### Browser

Google Chrome 107.0.5304.110

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

manakuro:
Can I try to fix it?

anubra266:
Yeah you can try @manakuro

segunadebayo:
Hi @manakuro,

We recently added the `isTruncated` style property back to v2. Once released, you can use this to achieve the ellipsis.

Thanks

PR #7041, opened 2022-11-27
fix(badge): ellipsis show up when using noOfLines and maxW

manakuro:
Closes # https://github.com/chakra-ui/chakra-ui/issues/7027

## 📝 Description
`Badge` is able to truncate the inner text after a specific number of lines, passing the `noOfLines` and `maxW` props.

The ellipsis does not appear on the Badge component using `noOfLines` and `maxW`. It used to work on `@chakra-ui/react": "^1.7.3`.

## 🚀 New behavior
The ellipsis can show up on the Badge component using `noOfLines` and `maxW`.

https://user-images.githubusercontent.com/11571318/204123706-c7d69dbe-5eaf-4f6c-b3f3-e6617464c6fe.mov

No API change, no change in visual appearance.

### DIFF
--- a/packages/components/layout/src/badge.tsx
+++ b/packages/components/layout/src/badge.tsx
@@ -5,8 +5,10 @@ import {
   ThemingProps,
   useStyleConfig,
   HTMLChakraProps,
+  ChakraProps,
 } from "@chakra-ui/system"
 import { cx } from "@chakra-ui/shared-utils"
+import { useMemo } from "react"
 
 export interface BadgeProps
   extends HTMLChakraProps<"span">,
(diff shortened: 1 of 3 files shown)


packages/components/layout/src/badge.tsx
```diff
@@ -21,18 +23,34 @@ export interface BadgeProps
...
   const { className, ...rest } = omitThemingProps(props)
+  const badgeStyle = useMemo((): ChakraProps => {
+    if (props.noOfLines) {
+      return {
+        __css: {
+          verticalAlign: "middle",
+          ...styles,
+        },
+        sx: {
+          display: props.display || "-webkit-inline-box",
```
manakuro, line 34: The `-webkit-line-clamp` CSS property only works in combination with the display property set to `-webkit-box` or `-webkit-inline-box`. In this case, it should use `-webkit-inline-box` instead of `-webkit-box`.

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

segunadebayo:
Hi @manakuro,

I just brought back the `isTruncated` prop, which will help solve this problem effectively. We also want to avoid runtime logic for UI only components like this.

After the next release, if the issue persists, I'll re-open this.

Thank you.

manakuro:
Hi @segunadebayo 
I got it. I appreciate your help.

PR #7074, opened 2022-12-05
fix(skeleton): take children width

### LINKED ISSUE #7065: Skeleton width not taking width of children as described in docs
### Description

The docs state - "Or to wrap another component to take the same height and **_width_**." for the following example - 

```
<Skeleton>
  <div>contents wrapped</div>
  <div>won't be visible</div>
</Skeleton>
```

Whereby the skeleton goes full width, as expected for having `div` children, however if these were `span` elements, or `div` with `display: inline-block`, whereby the children only take a width depending on their text content, the skeleton still goes full width.

### Link to Reproduction

see steps to reproduce

### Steps to reproduce

1. Go to https://chakra-ui.com/docs/components/skeleton/usage
2. Update the example from - 

```
<Skeleton>
  <div>contents wrapped</div>
  <div>won't be visible</div>
</Skeleton>
```

to

```
<Skeleton>
  <span>contents wrapped</span>
  <span>won't be visible</span>
</Skeleton>
```

or

```
<Skeleton>
  <div style={{ display: 'inline-block'}}>contents wrapped</div>
  <div style={{ display: 'inline-block'}}>won't be visible</div>
</Skeleton>
```
3. Observe the skeleton still takes full width
4. Toggle the `<Skeleton isLoaded>` prop and inspect element to see the actual width taking by child elements is **not** full width, and only the width of the text content

### Chakra UI Version

2.4.1

### Browser

Chrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

anubra266:
Closes #7065 
Fix width of skeleton; match width of children.


packages/components/skeleton/stories/skeleton.stories.tsx
```diff
@@ -31,6 +31,12 @@ export const AsContainer = () => (
   </Skeleton>
 )
 
+export const FitContent = () => (
```
segunadebayo, line 34:
```suggestion
export const WithFitContent = () => (
```

PR #7096, opened 2022-12-11
feat(gatsby-plugin): add option to use `ChakraBaseProvider

TylerAPfledderer:
## 📝 Description

This adds an option to the Gatsby plugin for the user to be able to use `ChakraBaseProvider` instead of `ChakraProvider`, which only applies theme tokens initially and not default component themes.

Only the `ChakraProvider` is rendered.

## 🚀 New behavior

Adds the option `isBaseProvider` to render the `ChakraBaseProvider` instead. 

> Default value is `false`

No because the new option is set to `false` so the original behavior is default.

## 📝 Additional Information

This is to keep the plugin up-to-date with the addition of `ChakraBaseProvider` in `v2.4.2`.

> Side note: as of this PR, the docsite should be updated to include `ChakraBaseProvider` for all frameworks.

### DIFF
--- a/tooling/gatsby-plugin/src/theme.js
+++ b/tooling/gatsby-plugin/src/theme.js
@@ -1,3 +1,3 @@
-import { theme } from "@chakra-ui/react"
+import { theme, baseTheme } from "@chakra-ui/react"
 
-export default theme
+export { theme, baseTheme }
--- a/.changeset/wet-pumas-prove.md
+++ b/.changeset/wet-pumas-prove.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/gatsby-plugin": minor
+---
+
+Add option to use `ChakraBaseProvider` instead of `ChakraProvider`. This will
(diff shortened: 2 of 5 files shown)


tooling/gatsby-plugin/src/theme.js
```diff
@@ -1,3 +1,3 @@
-import { theme } from "@chakra-ui/react"
+import { theme, baseTheme } from "@chakra-ui/react"
 
-export default theme
+export { theme, baseTheme }
```
natac13, line 3: Is this not a breaking change??

My gatsby dev would not work until I changed the theme.js overrider from a default export to a named export??

TylerAPfledderer: Hey @natac13! Sorry for the trouble and thank you for bringing up this error.

We have updated it in #7149 to get the original default `theme` object to be a default export within the plugin. We have not noticed any issues from here after some testing. 

Let us know from there if you receive further trouble!

PR #7103, opened 2022-12-13
feat(stack): remove divider around empty space when component is null

amritk:
## 📝 Description

Sometimes you want to hide the logic for hiding a component inside of the component. When you return null, it adds an extra divider and creates an empty space.

Currently, stack with divider will render a divider between a component and an empty component.
image

## 🚀 New behavior

With this PR, the divider between an empty component is removed.
image

no


segunadebayo:
Hi @amritk,

I appreciate your effort to fix this. I'd recommend using flexbox if you ever need this feature. We're slowly moving away from runtime logic for layout primitives.

We'll refactor this code at some point (most likely next major) to use flexbox.

```jsx
<Flex gap="4">
<Item/>
<NullishItem/>
</Flex>
```

Thanks for understanding.

PR #7108, opened 2022-12-14
build: update setup and fix ESM

segunadebayo:
Closes # 

## 📝 Description

- Fix the issue with ESM and the exports
- Switch to Vite build setup


segunadebayo:
I'm going to try again with tsup. No need to switch to Vite.

PR #7117, opened 2022-12-16
feat: add `layout` prop to table

anubra266:
add `layout` prop to table

### DIFF
--- a/packages/components/table/src/table.tsx
+++ b/packages/components/table/src/table.tsx
@@ -19,7 +19,16 @@ const [TableStylesProvider, useTableStyles] = createContext<
 
 export { useTableStyles }
 
-export interface TableOptions {}
+type Globals =
+  | "-moz-initial"
+  | "inherit"
+  | "initial"
+  | "revert"
+  | "revert-layer"
+  | "unset"
+export interface TableOptions {
+  layout?: Globals | "auto" | "fixed"
+}
 
 export interface TableProps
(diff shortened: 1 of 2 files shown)


packages/components/table/src/table.tsx
```diff
@@ -19,7 +19,16 @@ const [TableStylesProvider, useTableStyles] = createContext<
...
+type Globals =
+  | "-moz-initial"
+  | "inherit"
+  | "initial"
+  | "revert"
+  | "revert-layer"
+  | "unset"
+export interface TableOptions {
+  layout?: Globals | "auto" | "fixed"
+}
```
segunadebayo, line 31: Let's simplify this

```suggestion
export interface TableOptions {
  layout?: SystemStyleObject["tableLayout"]
}
```

packages/components/table/src/table.tsx
```diff
@@ -34,13 +36,13 @@ export interface TableProps
  */
 export const Table = forwardRef<TableProps, "table">((props, ref) => {
   const styles = useMultiStyleConfig("Table", props)
-  const { className, ...tableProps } = omitThemingProps(props)
+  const { className, layout = "auto", ...tableProps } = omitThemingProps(props)
```
segunadebayo, line 39: We don't need a default value

```suggestion
  const { className, layout, ...tableProps } = omitThemingProps(props)
```

PR #7124, opened 2022-12-18
fix(use-media-query): return correct value on initial mount if srr is…

sibite:
… enabled

Closes # 6930

## 📝 Description

Fixes problem when responsive components were flashing on mount because base value was used initially before rerendering with wanted value

If component using useBreakpointValue() hook renders at the first time, it uses base value and rerenders to use correct value

## 🚀 New behavior

If component using useBreakpointValue() hook renders at the first time, it uses correct value and does not rerender unnecessarily

No

### DIFF
--- a/.changeset/clean-zoos-teach.md
+++ b/.changeset/clean-zoos-teach.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/media-query": patch
+---
+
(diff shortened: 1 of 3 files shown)


segunadebayo:
Thanks for working on this.

This has been incorporated into the upcoming v3 version

Issue #7125, opened 2022-12-19
Divider allow responsive orientation


JohnFedorchak:
### Description

Allow the Divider to have an orientation set by an object, such as
```jsx
<Divider orientation={{ base: "horizontal", md: "vertical" }} />
```

### Problem Statement/Justification

This should be added to be consistent with the other reactive components and their attributes, 

and to make it easy to display a horizontal line when things are stacked vertically in a mobile layout, but to display a vertical line when things are stacked horizontally on a desktop layout.

### Proposed Solution or API

I think that it needs to be made a `ResponsiveValue<"horizontal" | "vertical">`?

### Additional Information

Currently I'm using something like
```jsx
<Box
  w={{ base: "80%", md: "2px" }}
  h={{ base: "2px", md: "80%" }}
/>
```
to accomplish this.

anubra266:
@JohnFedorchak You can make use of the `useBreakpointValue` hook for this, but responsive orientation is not a design decision we want to take right now, to avoid new complexities in the codebase.

PR #7126, opened 2022-12-19
feat: allow responsive values in orientation divider component

Tisks:
Closes [#7125](https://github.com/chakra-ui/chakra-ui/issues/7125)

## 📝 Description

Allow the Divider to have an orientation set by an object, such as:

`<Divider orientation={{ base: "horizontal", md: "vertical" }} />`

The `Divider` component doesn't allow responsive value in the `orientation` prop.

## 🚀 New behavior

Now the component accepts a responsive value in the `axis` prop (name changed due to conflicts with the default `orientation` type stated in the `ThemingProps`).

The default orientation is still `horizontal` if the `axis` prop is not passed.

Yes.
It changes the `orientation` prop name for `axis` so it will impact in the chakra-ui docs if it gets merged in that the `Divider` docs needs to be updated with that new prop name.

### DIFF
--- a/.changeset/mean-foxes-cheer.md
+++ b/.changeset/mean-foxes-cheer.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/layout": minor
+---
+
+Change the `orientation` prop name for `axis` and apply the `ResponsiveValue`
+type to it in order to be responsive depending on breakpoints in the `Divider`
+component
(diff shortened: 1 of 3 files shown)


anubra266:
@Tisks You can make use of the `useBreakpointValue` hook for this, but responsive orientation is not a design decision we want to take right now, to avoid new complexities in the codebase.

Issue #6901, opened 2022-10-25
CLI support for generating module augmentation theme


lexanth:
### Description

https://github.com/chakra-ui/chakra-ui/pull/5579 added the ability to override the theme types via module augmentation, rather than by writing to node_modules. I think this is a cleaner approach for various reasons - finding the package location doesn't quite work when hoisting in monorepos, it doesn't work with approaches like yarn PnP. Mutating a dependency generally seems unusual.

The theme typings generated by the CLI are ignored if using the module augmentation approach.

### Problem Statement/Justification

We want the best of both worlds - automatic generation of the theme types based on the theme _and_ the clean approach of using module augmentation. Eventually this could become the default and writing to node_modules could be dropped.

### Proposed Solution or API

Add a CLI argument to choose the output mode

Could be `--module-augmentation` as a boolean or a `--template=augmentation` to allow future changes, if we can imagine them

This would just change the `template` in https://github.com/chakra-ui/chakra-ui/blob/main/tooling/cli/src/command/tokens/create-theme-typings-interface.ts#L88 to be something like this
```
`// regenerate by running
// npx @chakra-ui/cli tokens path/to/your/theme.(js|ts)
import { BaseThemeTypings } from "@chakra-ui/styled-system";
declare module "@chakra-ui/styled-system" {
  export interface CustomThemeTypings extends BaseThemeTypings {
    ${sameContentAsOtherTemplate}
  }
}
`
```

### Alternatives

Currently you can work around this with an elaborate file structure
* `shared.types.d.ts`
  * Must have exactly this name, because the CLI output references this
  * Content: `export type { BaseThemeTypings } from '@chakra-ui/styled-system';`
* `generated-types.d.ts`
  * Content: CLI output
* `theme-types.d.ts`
  * Content:
```
import { ThemeTypings } from './generated-types';

declare module '@chakra-ui/styled-system' {
  export interface CustomThemeTypings extends ThemeTypings {}
}
```
But that's a bit of a pain.

### Additional Information

Happy to look at implementing this if it would be accepted (although getting https://github.com/chakra-ui/chakra-ui/pull/5701 merged would be a higher priority)

PR #7127, opened 2022-12-19
feat(cli): allow generating theme typings to use via module augmentation

lexanth:
Closes https://github.com/chakra-ui/chakra-ui/issues/6901

## 📝 Description
https://github.com/chakra-ui/chakra-ui/pull/5579 added the ability to override the theme typings by module augmentation rather than by writing to node_modules (which would better support e.g. Yarn PnP, for example).

This allows generating theme typings in a structure which can be used in this way.

The CLI generates theme typings in a structure which is designed for use in overwriting node_modules.

## 🚀 New behavior
When the CLI is run with `--template augmentation`, theme typings are generated which are appropriate for use with module augmentation.

No - default behaviour is still to generate theme typings in the old format.

### DIFF
--- a/.changeset/serious-numbers-know.md
+++ b/.changeset/serious-numbers-know.md
@@ -5,5 +5,5 @@
 Allow generating theme typings to use via module augmentation
 
 ```bash
-chakra-cli tokens --template augmentation
+chakra-cli tokens --template augmentation --out ./types/chakra-ui__styled-system.d.ts
 ```
--- a/.changeset/serious-numbers-know.md
+++ b/.changeset/serious-numbers-know.md
@@ -1,0 +1,9 @@
+---
+"@chakra-ui/cli": minor
+---
+
+Allow generating theme typings to use via module augmentation
+
+```bash
+chakra-cli tokens --template augmentation
+```
--- a/tooling/cli/src/command/tokens/create-theme-typings-interface.ts
+++ b/tooling/cli/src/command/tokens/create-theme-typings-interface.ts
@@ -54,7 +54,7 @@ function applyThemeTypingTemplate(
   switch (template) {
     case "augmentation":
       return `// regenerate by running
-// npx @chakra-ui/cli tokens path/to/your/theme.(js|ts)
+// npx @chakra-ui/cli tokens path/to/your/theme.(js|ts) --template augmentation --out path/to/this/file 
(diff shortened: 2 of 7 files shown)


.changeset/serious-numbers-know.md
```diff
@@ -0,0 +1,9 @@
+---
+"@chakra-ui/cli": minor
+---
+
+Allow generating theme typings to use via module augmentation
+
+```bash
+chakra-cli tokens --template augmentation
```
TimKolberger, line 8: **suggestion**: I think we should include the `--out` parameter in this changelog entry as well, otherwise the augmented typings will be generated in the node_modules, which might be unexpected.

```suggestion
chakra-cli tokens --template augmentation --out ./types/chakra-ui__styled-system.d.ts
```

tooling/cli/src/command/tokens/create-theme-typings-interface.ts
```diff
@@ -37,11 +37,41 @@ export interface ThemeKeyOptions {
...
+}
+
+function applyThemeTypingTemplate(
+  typingContent: string,
+  template: TypingsTemplate,
+) {
+  switch (template) {
+    case "augmentation":
+      return `// regenerate by running
+// npx @chakra-ui/cli tokens path/to/your/theme.(js|ts)
```
TimKolberger, line 57: **suggestion**: Since we have two different template we could provide the correct command to generate it as well.

```suggestion
// npx @chakra-ui/cli tokens --template augmentation --out path/to/this/file path/to/your/theme.(js|ts)
```

tooling/cli/src/index.ts
```diff
@@ -49,6 +65,8 @@ export async function run() {
             out,
             strictComponentTypes,
             format,
+            strictTokenTypes,
```
TimKolberger, line 68: **praise (non-blocking)**: good catch!

TimKolberger (approved):
Thank you for working on that @lexanth ✨ 
I like the clean integration of this feature into the cli package - good job!

I left a few non blocking comments.

segunadebayo:
Hi @lexanth,

Thanks for working on this. Could you fix the test failure? I'd love to get this in.

lexanth:
@segunadebayo Sorted now!

PR #7173, opened 2022-12-27
feat: transform multiple token vals in shorthand space props

TylerAPfledderer:
## 📝 Description

This PR provides the ability to pass in multiple token values to the shorthand space props `margin`, `m`, `padding`, and `p`, per standard CSS convention.

Only one value can be supplied to the above-mentioned props. So `margin="4"` converts to `margin: "16rem"` for all sides of an element.

## 🚀 New behavior

i.e.
`margin="1 3"` (Note that it's a string!)

converts to
`margin: "0.25rem 0.75rem"` (or the equivalent CSS vars)

Or...

`p="1 10px 4"` (Mixed with unit vals and 3 for top, left/right, bottom)

converts to
`padding: "0.25rem 10px 16rem"`

This should have no effect on existing projects, as no altering of existing implementation is needed for the new transformation to work.

## 📝 Additional Information
This would not be applied to shorthand directional spacing props like `paddingX`, `my`, and similar, as for their purposes should only accept one value to address `top-bottom` or `left-right`, else different spacing values should be addressed in some other way.

### DIFF
--- a/.changeset/big-feet-tan.md
+++ b/.changeset/big-feet-tan.md
@@ -1,0 +1,26 @@
+---
+"@chakra-ui/styled-system": minor
+---
+
+Allow multiple token values to be converted in shortand spacing props
+
+For the props `margin`, `m`, `padding`, and `p` you can supply multiple tokens
+as you would for the standard CSS conventions with `top right bottom left`, and
+the tokens will be converted accordingly.
+
+For example: `margin="3 4"`
+
+converts to `margin: "var(--chakra-space-3) var(--chakra-space-4)"`
+
+Or (mixing with unit values)...
+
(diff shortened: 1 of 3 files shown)


segunadebayo:
This approach is interesting.

The same logic will have to be extended to scroll (margin | padding), margin (start | end), borders, border radius, outline, animation, transition, etc.

In the early days, we also had this idea but settled on keeping it as is due to the internal complexity we needed to manage.

As with CSS-in-JS and runtime CSS, we want to reduce the runtime JS required for style calculations, so I’m weary of merging this PR.

We can only add to this if “absolutely” is required. In this case, I think most users can do without this feature.

PR #7204, opened 2023-01-08
refactor(checkbox): inject default theme from component

yukukotani:
## 📝 Description

> Add a brief description

Inject default theme from component so that we don't need to declare it in `extendBaseTheme` explicitly anymore.

If this approach seems good, I'll expand it to other components.

We need to explicitly declare the default component themes to use in `extendBaseTheme`. The style will be broken if we forget the declaration.

https://codesandbox.io/s/create-react-app-ts-forked-2bwzer?file=/src/index.tsx

## 🚀 New behavior

We don't need to declare the component themes.

https://codesandbox.io/s/create-react-app-ts-forked-5g50lg?file=/src/index.tsx

No

### DIFF
--- a/packages/components/checkbox/package.json
+++ b/packages/components/checkbox/package.json
@@ -43,6 +43,7 @@
     "@chakra-ui/react-use-merge-refs": "workspace:*",
     "@chakra-ui/react-use-update-effect": "workspace:*",
     "@chakra-ui/react-types": "workspace:*",
+    "@chakra-ui/theme": "workspace:*",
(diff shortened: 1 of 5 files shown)


segunadebayo:
Thanks for working on this @yukukotani.

I don't think this is the right approach because it moves the bundle size from the theme into the checkbox. If a user wants to use `extendBaseTheme,` they'll need to be aware of adding the theme themselves.

I wish there was some way to do this automatically, but I can't think of any right now.

PR #7216, opened 2023-01-11
refactor: component api docs generation

segunadebayo:
## 📝 Description

- Refactored the code that generates the props documentation
- Use explicit exports for all component packages (explicit is always better)
- Fix export of `ReactIcon` from the icons package (it was omitted in the previous release)


BenWildeman:
this has caused a breaking change with types. it's now no longer possible to import specific types for TS. For instance, I make use of things like `GridOptions` when building components from scratch using `chakra.div`

Issue #7202, opened 2023-01-07
`useBreakpointValue` doesn't respond to window resizing


adamzerner:
### Description

The [docs](https://chakra-ui.com/docs/hooks/use-breakpoint-value) for `useBreakpointValue` say:

> This hook also responds to the window resizing and returning the appropriate value for the new window size.

However, this is not happening for me in my NextJS app:

```
import { useBreakpointValue } from '@chakra-ui/react';
import Head from 'next/head'

export default function Home() {
  const isMobile = useBreakpointValue({ base: true, md: false }) as boolean;
  
  return (
    <>
      <Head>
        <title>Create Next App</title>
        <meta name="description" content="Generated by create next app" />
        <meta name="viewport" content="width=device-width, initial-scale=1" />
        <link rel="icon" href="/favicon.ico" />
      </Head>
      <main>
        Hello World
        <p>isMobile: {isMobile.toString()}</p>
      </main>
    </>
  )
}
```

### Link to Reproduction

https://github.com/adamzerner/demo-use-breakpoint-value-fail-next

### Steps to reproduce

1. Clone https://github.com/adamzerner/demo-use-breakpoint-value-fail-next
2. `yarn install`
3. `yarn dev`
4. Open http://localhost:3000/
5. Make the window size narrow
6. Observe "isMobile: false" instead of `isMobile: true" in the UI

### Chakra UI Version

2.4.6

### Browser

Latest Firefox

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

bragle:
I have the same issue. You can even see it in the docs. The example button [here](https://chakra-ui.com/docs/hooks/use-breakpoint-value#usage) should change on resize, but it only changes on page reload.

cschroeter:
Can confirm the issue. Fix is on the way

cschroeter:
@adamzerner @vicero  please use `@chakra-ui/react@2.4.7`

BenjaOliva:
Any news on this issue ? Im facing this same problem on `@chakra-ui/react@2.6.1`

PR #7222, opened 2023-01-12
fix(styled-system): fix issue with useBreakpointvalue in Firefox

cschroeter:
Resolves https://github.com/chakra-ui/chakra-ui/issues/7202


cschroeter:
Resolves #7202

PR #7232, opened 2023-01-13
refactor: env detection

### LINKED ISSUE #6374: EnvironmentProvider appends a span to DOM which breaks some unit tests

### LINKED ISSUE #6420: An empty span tag injected by ChakraProvider becomes visible when a modal is rendered inside a stack
### Description

An empty `span` tag injected by `ChakraProvider` becomes visible when a modal is rendered inside a stack. I haven't been able to reproduce the issue when the stack is removed.

### Link to Reproduction

https://codesandbox.io/s/black-cherry-54whdz?file=/src/App.tsx

### Steps to reproduce

1. Go to https://codesandbox.io/s/black-cherry-54whdz?file=/src/App.tsx. This sandbox includes a DOM structure with a button that triggers a modal, both of which are inside a stack.
2. Click on the `Open modal` button
3. You may not see it on the first click, but eventually you'll see a flash of red content when the modal opens. This is caused by an injection of an empty `<span />` tag by `ChakraProvider` somehow becoming visible while the modal opens. I've given this `span` a width, height, and a red background so you can see the visual jank it causes when the modal opens.

### Chakra UI Version

2.2.4

### Browser

Chrome latest

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

* This issue is not present in `v2.2.1`
* https://github.com/chakra-ui/chakra-ui/issues/6374 also complains about the `span` tag causing issues

segunadebayo:
Closes #6374
Closes #6420

## 📝 Description

- Add `disableEnvironment` option to disable the environment-injected span
- Refactored code to avoid state updates within the ref

Partially. Environment is an internal API that isn't documented. Most users shouldn't use this anyways


LeonardDrs:
If I may, the `const showSpan = !disabled || !environmentProp`  statement prevents from using `disableEnvironment` without specifying an `environmentProp` anyway. 
This makes the `disableEnvironment` prop shallow as the `span` was not showing anyway if we had an `environmentProp`.

frontshift:
I tripped over this as well as simply adding `disableEnvironment` to `ChakraProvider` still renders the span if no environment is set. Be good to add this to the documentation.
 ```JSX
const env = useEnvironment()

return (
    <ChakraProvider theme={theme} disableEnvironment environment={env}>
        {children}
    </ChakraProvider>
)
```

PR #7265, opened 2023-01-20
Fix color mode behavior

odiak:
## 📝 Description

Follow system color mode on page load, if `useSystemColorMode` is `true`.

1. set system as "dark mode" (app follows)
2. leave page
3. set system as "light mode"
4. come back to page
5. app is still "dark mode"

## 🚀 New behavior

1. set system as "dark mode" (app follows)
2. leave page
3. set system as "light mode"
4. come back to page
5. app changes to "light mode"


segunadebayo:
Thanks for working on this. I recommend using `next-themes` to manage color mode going forward.

Closing this in favor of the upcoming v3 version.

Issue #7337, opened 2023-02-09
Export useModalManager and/or add index in useModal


bertho-zero:
### Description

I would need the `ModalManager` to create custom components that look like modals.

### Problem Statement/Justification

I am creating a custom component which is similar to a drawer except that above a certain breakpoint it is displayed directly in the dom (without a drawer).
This is to have a drawer on mobile and the content of the drawer visible on desktop without having to duplicate the elements and thus avoid Flash Of Content.

I would like to use `isTopModal` from `ModalManager` to make my component detect as the topmost modal or not.

For that it would be necessary to export `useModalManager` or that `useModal` returns the index of the modal.

### Proposed Solution or API

`useModal` could export `index`, `isTopModal` or `manager`.

`modalManager` and `useModalManager` could be exported.

mikirejf:
Bump. It would be nice to have the manager instance exported so we can have an idiomatic way to check the number of opened modals etc ...

PR #7338, opened 2023-02-09
Feat/export modal manager

bertho-zero:
Closes https://github.com/chakra-ui/chakra-ui/issues/7337

## 📝 Description

Exports `modalManager` and returns the index of the modal from `useModal` to
allow creating custom components similar to modals.

No


packages/components/modal/src/index.ts
```diff
@@ -42,3 +42,4 @@ export { ModalOverlay } from "./modal-overlay"
 export type { ModalOverlayProps } from "./modal-overlay"
 export { useModal } from "./use-modal"
 export type { UseModalProps, UseModalReturn } from "./use-modal"
+export { modalManager, useModalManager } from "./modal-manager"
```
segunadebayo, line 45: I would export only the hook in this case.

```suggestion
export { useModalManager } from "./modal-manager"
```

Issue #7343, opened 2023-02-10
Typescript - Button should not allow disabled prop


lexanth:
### Description

Since https://github.com/chakra-ui/chakra-ui/commit/0f376650efce9eb2d2cba3484ffe4e3a73364504 Button now accepts a `disabled` prop, but does nothing with it - it is always overridden by `isDisabled`, `isLoading` and the button group's `isDisabled`.

It would make it easier to use to either prevent `disabled` being used (in typescript) or include `disabled` in the options which disable a button.

### Link to Reproduction

https://codesandbox.io/s/distracted-matsumoto-iq5sly?file=/src/index.tsx

### Steps to reproduce

1. Click the button
2. Alert is shown

### Chakra UI Version

2.4.9

### Browser

Chrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

The behaviour (ignoring `disabled`) may be valid, but if so it should be excluded from the TS props.

pmpwith2i:
@lexanth I opened a PR for adding a deprecation message when users are using the disabled prop, hope it will help :)

segunadebayo:
Duplicate of #7269

PR #7344, opened 2023-02-10
fix: deprecate button disabeld prop

pmpwith2i:
## 📝 Description

> This PR adds the TS Doc for deprecating the disabled prop in buttons, as requested in the issue #7343 

> Actually users can pass the disabled prop to chakra buttons but as for requirements the button will not be disabled

## 🚀 New behavior

> I'm replacing the prop in order to return a deprecation error if users are using disabled

NO


barrymichaeldoyle:
Looking forward to seeing this merged.

segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version where `disabled` is now native

PR #7370, opened 2023-02-16
feat(styled-system): create `gradients` theme key

TylerAPfledderer:
Add the `gradient` theme key to be able to create tokens for gradient styles.

> Applies to the `bgGradient`, `bgImg`, `bgImage`, and `backgroundImage` style props.

Due to the current setup under the hood of the [Background Gradient API](https://chakra-ui.com/docs/styled-system/gradient#background-gradient-api), I do not believe it is worth the investment to make the logic work in a theme config. There are probably easier and cheaper ways to acquire the color theme tokens;

- CSS Variables explicitly: 
```
lightLinearGradient: 'linear-gradient(to right, var(--chakra-colors-gray-200), var(--chakra-colors-blue-500))'
```
- Importing the colors object and use interpolation: 
```ts
lightLinearGradient: `linear-gradient(to right, ${colors.gray.200}, ${colors.blue.500})`
```


packages/core/styled-system/src/config/background.ts
segunadebayo, line 1: Let's connect the gradients token to `backgroundImage`, `bgImage` and `bgGradient`

TylerAPfledderer:
@segunadebayo made the updates as requested.

.changeset/sixty-otters-pretend.md
```diff
@@ -0,0 +1,27 @@
...
+
+// SomeComponent.tsx
+
+<Box bgGradient='lightBgGradient' />
+```
+
+🚨 NOTE: The
+[Background Gradient API](https://chakra-ui.com/docs/styled-system/gradient#background-gradient-api)
+can not be used in a token as the conversion is done when the api is used
+directly on a prop and not when compiling the theme config
```
TylerAPfledderer, line 27: @segunadebayo did I explain this correctly?

packages/core/styled-system/src/config/background.ts
```diff
@@ -5,7 +5,7 @@ import { t, Token, transforms } from "../utils"
 export const background: Config = {
   background: t.colors("background"),
   backgroundColor: t.colors("backgroundColor"),
-  backgroundImage: t.propT("backgroundImage", transforms.bgImage),
```
segunadebayo, line 8: This one is a breaking change. I was hoping we could add this while preserving the existing transforms for `bgGradient`

TylerAPfledderer: @segunadebayo I ended up moving it's logic over to the `parseGradient` function, hence the tests I added still passing. Will this still mot work?

TylerAPfledderer: @segunadebayo I moved it's logic over to the `parseGradient` function, hence the test I added passed with the `url()`. Is there a check I am missing?

TylerAPfledderer: I'm just now recognizing that I should be adding tests to `parse-gradient.test.ts` instead. But it appears there are no tests failing there at the moment.

segunadebayo: Got it. I'm going to double check that.

PR #7406, opened 2023-03-01
Update avatar-name.tsx

anLpk:
Closes # 

## 📝 Description
Currently, the avatar component has a fallback to generate initials with random and accessible background colours.
But If the users name more than 2 words, initials do not get displayed as expected.

## ⛳️ Current behaviour (updates)

If the user name is `Anil Pak`, it's displayed as `AP`, but If the user name is `Anil Pak Chakra` it will be also displayed as `AP` instead of `AC`

## 🚀 New behavior

With these changes instead of splitting the name as first_name and last_name, we are splitting the string name to the array and taking the first letter from the first and the last word to create initials.

Not sure :)


packages/components/avatar/src/avatar-name.tsx
```diff
@@ -3,10 +3,10 @@ import { useAvatarStyles } from "./avatar-context"
 import { AvatarOptions } from "./avatar-types"
 
 export function initials(name: string) {
-  const [firstName, lastName] = name.split(" ")
```
lukemartin, line 6: Could change this to:

```ts
const [firstName, ...otherNames] = name.split(" ");
const [lastName] = otherNames.reverse();
```

and keep the rest of the function as-is

anLpk: Thank Luke! Looks even better!

packages/components/avatar/src/avatar-name.tsx
```diff
@@ -3,7 +3,8 @@ import { useAvatarStyles } from "./avatar-context"
...
+  const [firstName, ...otherNames] = name.split(" ");
+  const [lastName] = otherNames.reverse();
```
aashutoshrathi, line 7: if you just need last element, you should use:
```ts
 const lastName = otherNames.at(-1);
```

aashutoshrathi (commented):
left a comment

aashutoshrathi (approved):

PR #7412, opened 2023-03-01
feat: add divider label support

Pagebakers:
Closes # 

## 📝 Description

Add support for labels in the Divider component. Support horizontal/vertical orientation.

No label support

## 🚀 New behavior

```tsx
<Divider><DividerLabel>or</DividerLabel></Divider>
```

No


Pagebakers:
@segunadebayo This is finished, but didn't get any feedback since I opened it. It is something that can be merged otherwise I will close it.

segunadebayo:
Hi @Pagebakers,

Thanks for working on this. I'm deeply sorry for the late response here.

Looking through the PR, we don't want to make this feature implicit.

Today, users can achieve the same by combining `AbsoluteCenter` and `Divider`

```jsx
import { Box, Divider, AbsoluteCenter } from '@chakra-ui/react'

export const App = () => {
  return (
    <Box position="relative" padding="10">
      <Divider />
      <AbsoluteCenter bg="white" px="4">Welcome</AbsoluteCenter>
    </Box>
  )
}
```

This should be a recipe in the docs we can add, not an update we need to add to the core.

I'll add this use case to the docs shortly.

segunadebayo:
I just updated the Divider docs to include an example.
https://chakra-ui.com/docs/components/divider/usage#adding-content-within-a-divider

Thanks a lot for attempting this.

PR #7417, opened 2023-03-05
fix: add focusvisible styles to MenuButton component without passing as prop

poteboy:
## 📝 Description

Add keyboard focus indicator to MenuButton component

This pull request adds a default focus style to the MenuButton component so that it is visible when it receives keyboard focus. This change was made to improve the accessibility of the component and to comply with the WCAG Level AA guidelines.

Currently, the MenuButton component does not have a default focus style when the as prop is not passed with the value "Button". This means that users who rely on the keyboard to navigate the page may have difficulty determining which component they are interacting with.

## 🚀 New behavior

With this pull request, the MenuButton component now has a default focus style that is visible when it receives keyboard focus. This ensures that all users, including those who rely on the keyboard, can easily determine which component they are interacting with.

No.

## 📝 Additional Information

N/A.

### DIFF
--- a/packages/components/menu/src/menu-button.tsx
+++ b/packages/components/menu/src/menu-button.tsx
@@ -1,4 +1,9 @@
-import { forwardRef, HTMLChakraProps, chakra } from "@chakra-ui/system"
+import {
+  forwardRef,
+  HTMLChakraProps,
+  chakra,
+  useStyleConfig,
+} from "@chakra-ui/system"
 import { cx } from "@chakra-ui/shared-utils"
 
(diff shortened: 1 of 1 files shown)


segunadebayo:
Hi @poteboy,

We don't want to pull the button styles into the menu button. In the next major, we're looking to switch to a `cloneElement` based model instead to make it much easier to compose menu triggers.

```jsx
<MenuTrigger>
   <Button>Click me</Button>
</MenuTrigger>
```

For now, let's allow the users to handle this in their apps instead.

PR #7427, opened 2023-03-07
feat: add iconPlacement to menu-item-option

### LINKED ISSUE #7422: MenuItemOption radio mode, show icon on right side
### Description

Allow showing icon right side of the item. Eg:
image

Allowing users to flip icon's positions will allow them to make more customisable MenuLists

### Proposed Solution or API

Default `MenuItemOption` in radio mode show icon on the left side, if selected.
I plan to add a flag `rightCheckedIcon` to defined if you want icon to be on right side, and internally, since it's flex, will do `flexDir="row-reverse"`.
OR 
a prop `iconPlacement` -> can be `right|left` defaults to `left`

### Alternatives

use `flexDir="row-reverse"`

### Additional Information

_No response_

aashutoshrathi:
Closes #7422 

## 📝 Description

Adds `iconPlacement` to `MenuItemOption`

> MenuItemOption selected  icon can be placed on the right side now too

## 🚀 New behavior

> MenuItemOption selected icon could only be on the left side of the item

No


ankursehdev:
Need this feature, are we merging it soon?

ankursehdev (changes requested):
Instead of using it in the plain string. Do you think using an enum const makes more sense?

1. Use enum

```
const Placement = {
  LEFT: 'left',
  RIGHT: 'right',
} as const;
```

2. type we can do: `iconPlacement?: typeof Placement[keyof typeof Placement];` 
3. While comparing we can use `iconPlacement === Placement.RIGHT`

segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version, where you can explicitly render the icon where you prefer, rather than use a prop

Issue #7408, opened 2023-03-01
Popover z-index


vitalijalbu:
### Description

The popover doesn't have a specifi z-index.

### Link to Reproduction

https://turo.vercel.app/search

### Steps to reproduce

1. Go to search page
2. Click on some filters buttons

### Chakra UI Version

latest

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

robcrock:
@vitalijalbu if you need a quick workaround, it looks like this should work too.

The sibling could be set to `position: relative` and `z-index: -1`
<img width="1421" alt="image" src="https://user-images.githubusercontent.com/35023759/232233068-7bc28d10-012f-4eb8-8321-fb4c979b47d9.png">

chasegarner:
I opened a new PR for this one that includes a changeset for a faster merge.

segunadebayo:
I just pushed a fix for this. We'll release an update shortly.

If the issue persists after upgrading, I'll re-open it.

PR #7432, opened 2023-03-08
Use correct popover zIndex

krvajal:
📝 Description
Z-Index added in Popover which is taken from the indices object

⛳️ Current behavior (updates)
Functions in accordance to passed index

🚀 New behavior
Functions wrt given popover index

💣 Is this a breaking change (Yes/No):
No

In reference to https://github.com/chakra-ui/chakra-ui/issues/7408


vitalijalbu:
---
**latest chakra ui version**
---

Screenshot 2023-03-31 alle 11 21 41

segunadebayo:
Thanks for working on this.

This has been incorporated into the upcoming v3 version

PR #7464, opened 2023-03-19
feat: add css reset scope

### LINKED ISSUE #6946: Scoping global CSS reset
### Description

We're introducing Chakra into a legacy application, and when we add a `ChakraProvider` to a part of our application, the entire site is changed.

This is due to the global CSS reset that Chakra ships with. I can remove the css reset with the (undocumented) `resetCSS` property, but then everything looks weird.

This makes is really painful to introduce Chakra to legacy codebases.

This would make it possible to introduce Chakra into legacy applications.

### Proposed Solution or API

We could make the ChakraProvider add an auto-generated class name to the CSS reset, and scope it that way. However, adding a class in front of all css resets, makes it override the component specific styles (the specificity is higher). So that won't work.

We could add the same class in front of all emotion generated code as well (somehow). Not sure how that would work though

### Alternatives

_No response_

### Additional Information

_No response_

segunadebayo:
Closes #6946

## 📝 Description

Add support for scoping the css reset to a specific selector.

To use this feature, pass the `resetScope` prop to the `ChakraProvider` or
`ChakraBaseProvider` component.

```jsx live=false
import { ChakraProvider } from "@chakra-ui/react"
function App() {
  return (
    <ChakraProvider resetScope=".ck-reset">
      <App />
    </ChakraProvider>
  )
}
```

No

### DIFF
--- a/.changeset/breezy-deers-judge.md
+++ b/.changeset/breezy-deers-judge.md
@@ -1,0 +1,22 @@
+---
+"@chakra-ui/css-reset": minor
+"@chakra-ui/provider": minor
+"@chakra-ui/theme": minor
+---
+
+Add support for scoping the css reset to specific selector.
+
+To use this feature, pass the `resetScope` prop to the `ChakraProvider` or
+`ChakraBaseProvider` component.
+
+```jsx live=false
+import { ChakraProvider } from "@chakra-ui/react"
+
+function App() {
+  return (
+    <ChakraProvider resetScope=".ck-reset">
+      <App />
+    </ChakraProvider>
+  )
+}
+```
--- a/packages/components/css-reset/src/css-reset.tsx
+++ b/packages/components/css-reset/src/css-reset.tsx
@@ -1,24 +1,30 @@
 import { Global } from "@emotion/react"
 
-const vhPolyfill = `
-  :root, :host {
+const css = String.raw
+
+const vhPolyfill = css`
+  :root,
+  :host {
     --chakra-vh: 100vh;
   }
 
   @supports (height: -webkit-fill-available) {
-    :root, :host {
+    :root,
+    :host {
       --chakra-vh: -webkit-fill-available;
     }
   }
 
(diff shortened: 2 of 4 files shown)


selbekk:
👏 Thanks a ton for this, this will mean so much for adoption across large orgs with existing code bases.

posabsolute:
Has this been released yet?

TylerAPfledderer:
@posabsolute released per this [merged Version Package](https://github.com/chakra-ui/chakra-ui/pull/7461)

[Version 2.5.3 Changelog](https://chakra-ui.com/changelog/2.5.3)

posabsolute:
@TylerAPfledderer 

I think there are still some issues with the scoped CSS, most is scoped but some of the reset still leaks globally.

I would expect HTML would not receive any styling, at least things like font and line height should not be affected.

Screenshot 2023-04-04 at 7 19 23 PM

posabsolute:
The other thing that would help for this feature is where the CSS is added, with styled-components you can define where the styles are added. 

In micro front-ends its very possible we load multiple apps, so it's less error prone to have the CSS leave when the app is unloaded. This is the target option in StyleSheetManager from styled-component

simowaer:
@segunadebayo this is breaking in chrome browsers below version 88 because it changes to using the `:where()` selector (https://developer.mozilla.org/en-US/docs/Web/CSS/:where#browser_compatibility)

posabsolute:
Hey @TylerAPfledderer,

I gave it a go, it's almost there but if you add a ck-reset parent, then it overrides styles from the main components, example below:

Screenshot 2023-07-13 at 9 45 05 AM

Andrei-Ist:
Do you have any updates how to fix this?
Have you tried the disableGlobalStyle feature release of ChakraProvider,  in Chakra version 2.7.1. 
I'm trying to use this new feature, as an alternative to resetScope. So far it seems to have no effect, or I'm using it wrong. I'm trying it in chrome v 115.

PR #7499, opened 2023-03-27
fix(toast): remove invalid role

### LINKED ISSUE #7324: Toast Accessibility
### Description

When I run a web application using the `Toast` component through the [W3C Validator](https://validator.w3.org), I expect it to pass with no errors but instead it produces an error because of an invalid tag / role combination.

Inside the `<chakra-portal>` element, there is a series of `ul` tags with `role="region"` set. This is not a valid role for a `ul` tag as stated by the [UL Tag Technical Summary](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#technical_summary), and thus is invalid HTML.

### Link to Reproduction

https://codesandbox.io/s/chakra-toast-accessibility-bpffqm

### Steps to reproduce

1. Go to the `codesandbox` link.
2. Open dev tools.
3. Copy the `html` element from the example.
4. Run the `html` through the [W3C Validator](https://validator.w3.org).
5. See error (example below).

### Chakra UI Version

2.4.9

### Browser

Chromium 109

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

The output relating to the toast component from https://validator.w3.org

Validator Output

kkudose:
Closes #7324 

## 📝 Description

- `region` is not a valid `role` for `ul`
- the default role is appropriate for this use (`list`)

role is `region`

## 🚀 New behavior

role is `list`

no

### DIFF
--- a/.changeset/nervous-beans-exist.md
+++ b/.changeset/nervous-beans-exist.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/toast": patch
+---
+
+Remove invalid role in toast
--- a/packages/components/toast/src/toast.provider.tsx
+++ b/packages/components/toast/src/toast.provider.tsx
@@ -123,7 +123,6 @@ export const ToastProvider = (props: ToastProviderProps) => {
 
     return (
       <ul
-        role="region"
         aria-live="polite"
         key={position}
         id={`chakra-toast-manager-${position}`}


segunadebayo:
Hi @kkudose,

The role for the toast element is actually valid, it's the element we're rendering that needs to be updated. The role is required to get it working in NVDA.

Issue #7193, opened 2023-01-05
Checkbox animates initial state


vadimdemedes:
### Description

When `Checkbox` is rendered with `isChecked={true}` set right away, it still performs an animation for drawing a check mark inside of it, as if user interacted with that checkbox. I think if checkbox is immediately rendered with `isChecked={true}`, there should be no animation.

P.S. Thanks for a great project!

### Link to Reproduction

https://codesandbox.io/s/cocky-https-4hii19?file=/src/App.js

### Steps to reproduce

1. Click "Toggle checkbox visibility" button
2. Checkbox isn't rendered
3. Click the same button again
4. Checkbox is now rendered, but there's an animation for drawing a check mark

### Chakra UI Version

2.4.6

### Browser

Google Chrome 108

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

In Framer Motion this is solved by setting `initial={false}` to avoid animations on initial render. See https://www.framer.com/docs/animate-presence/##suppressing-initial-animations.

szpytfire:
+1. Having issues in Snapshot tests because of this! This isn't limited to MacOS // Google Chrome; happens on all browsers.

Any chance of this behaviour being reviewed?

selrond:
This is especially problematic when using some kind of virtualization like https://github.com/bvaughn/react-window, which utilizes render props pattern, recreating checkbox components on each render:

https://user-images.githubusercontent.com/6603389/230128154-c1d1fe4c-8daa-462d-a1d6-02bd42f74e5a.mp4

segunadebayo:
I just pushed a fix for this. We'll release an update shortly.

If the issue persists after upgrading, I'll re-open it.

selrond:
For the reference: https://github.com/chakra-ui/chakra-ui/commit/24f952da27ef43febbe20714a08eb94ce87133eb

PR #7510, opened 2023-03-29
fix(checkbox): animate icon only when transitioning to checked state

vadimdemedes:
Closes https://github.com/chakra-ui/chakra-ui/issues/7193

## 📝 Description

> `Checkbox` animates the check icon on first render when `isChecked` or `defaultChecked` is `true`. This creates a wrong impression as if checkbox has just transitioned from unchecked to checked state without user's interaction and causes confusion.

> `Checkbox` always assigns an animation to the check icon, regardless whether state has changed or not.

## 🚀 New behavior

> This PR updates `Checkbox` to track whether state has changed and assign an animation to the check icon only when it did. This avoids animation of the check icon on the initial render.

No.

## 📝 Additional Information

Before:

https://user-images.githubusercontent.com/697676/228465005-8e5b7d3c-391f-49d9-8f63-d2bd59f4d35e.mp4

After:

https://user-images.githubusercontent.com/697676/228464838-863cba94-16bb-4a78-b609-78188b21999f.mp4

### DIFF
--- a/packages/components/checkbox/src/checkbox.tsx
+++ b/packages/components/checkbox/src/checkbox.tsx
@@ -10,7 +10,7 @@ import {
   useMultiStyleConfig,
 } from "@chakra-ui/system"
 import { callAll, cx } from "@chakra-ui/shared-utils"
-import { cloneElement, useMemo } from "react"
+import { cloneElement, useMemo, useState } from "react"
 import { useCheckboxGroupContext } from "./checkbox-context"
 import { CheckboxIcon } from "./checkbox-icon"
 import { CheckboxOptions, UseCheckboxProps } from "./checkbox-types"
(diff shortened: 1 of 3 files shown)


packages/components/checkbox/src/checkbox.tsx
```diff
@@ -134,16 +134,26 @@ export const Checkbox = forwardRef<CheckboxProps, "input">(function Checkbox(
     onChange,
   })
 
+  const [previousIsChecked, setPreviousIsChecked] = useState(state.isChecked)
+  const [animateIcon, setAnimateIcon] = useState(false)
+
+  if (state.isChecked !== previousIsChecked) {
+    setAnimateIcon(true)
+    setPreviousIsChecked(state.isChecked)
+  }
```
vadimdemedes, line 143: Setting state during render is a bit unusual, but apparently it's the recommended way now → https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes.

segunadebayo:
Thanks for this, PR. I've gone ahead to apply some modifications and pushed the fix.

Issue #6328, opened 2022-07-15
Multiple toast when using multiple react roots


mabey01:
### Description

When a page has multiple react roots, showing a toast causes multiple toasts to be rendered.

### Link to Reproduction

https://stackblitz.com/edit/react-ts-2ja3cn?file=index.tsx

### Steps to reproduce

1. Go to [demo](https://stackblitz.com/edit/react-ts-2ja3cn?file=index.tsx)
2. Click on "Show toast"
3. Click on "Update" to update the count of toast rendered.
4. See that it's more than the visual toasts seen on screen.

When downgrading to chakra v1.8.8 and react v17, this issue does not happen. There is just 1 toast rendered.

### Chakra UI Version

2.2.4

### Browser

Chrome 103

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

jonathanwilke:
I am having the same issue. On a "normal" page I only get one toast. But when I am on a page where I have a shadow root element included, I get a second toast.

itkrt2y:
[Workaround] Use `ChakraProvider` of `@chakra-ui/provider` instead of `@chakra-ui/react` for the second and subsequent ones.

```tsx
// Example

import { ChakraProvider } from '@chakra-ui/react';
import { ChakraProvider as RawProvider } from '@chakra-ui/provider';
import React, { ComponentType } from 'react';
import * as ReactDOM from 'react-dom/client';

let initialized = false;

function initializeRoot(Root: ComponentType, containerId: string) {
  let Provider: typeof ChakraProvider;
  if (initialized) {
    Provider = RawProvider;
  } else {
    Provider = ChakraProvider;
    initialized = true;
  }

  const root = document.getElementById(containerId);
  ReactDOM.createRoot(root).render(
    <Provider>
      <Root />
    </Provider>
  );
}
```

The `ChakraProvider` in `@chakra-ui/react` contains the `<ToastProvider>`, but the `@chakra-ui/provider` does not.
https://github.com/chakra-ui/chakra-ui/blob/88e812c3a6d99554965b54856382916b6fbf6774/packages/react/src/chakra-provider.tsx#L16-L25
https://github.com/chakra-ui/chakra-ui/blob/88e812c3a6d99554965b54856382916b6fbf6774/packages/provider/src/chakra-provider.tsx#L63-L96
You can prevent multiple toasts by not using `<ToastProvider>` multiple times.

PR #7514, opened 2023-03-30
feat: add optional toastStore prop to ToastProvider

niklas-paro:
Closes # 
https://github.com/chakra-ui/chakra-ui/issues/6328 (Closed because a workaround was found, but I don't think that workaround is dependable or extends to all situations)

## 📝 Description

> Add a brief description

There are several cases where you may have multiple Toast Providers in your app. For example, you may be developing a small micro-frontend (e.g. a custom dialog) for other apps to consume. Or you may be developing micro-frontends, where you want to ensure that you only show and control your own toasts.

The toastStore is currently a singleton. So if you have multiple ToastProviders in your app for whatever reason, you also get multiple instances of toasts being rendered.

## 🚀 New behavior

This PR adds `toastStore` optional prop to the `ToastProvider`, and exports the `createToastStore` (formerly `createStore`) factory function.

No. The new prop is optional, and without providing it the behaviour stays the same.

### DIFF
--- a/.changeset/cool-dryers-tap.md
+++ b/.changeset/cool-dryers-tap.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/toast": minor
+---
+
+Add optional toastStore prop to ToastProvider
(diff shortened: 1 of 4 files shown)


segunadebayo:
Thanks for working on this.

PR #7529, opened 2023-04-01
feat: improve polymorphic types BREAKING

TimKolberger:
## 📝 Description

This PR addresses the ambiguity of the `as`-prop typings, while fixing current issues with TypeScript 5.0.x.

> The hard work was done by nasheomirro in https://github.com/nasheomirro/react-polymorphed, 
> where the implementation in https://github.com/chakra-ui/polymorphic is based on.
> Shoutout and many thanks to nasheomirro 💖 

```tsx
// the ref is not typed correctly
const ref = React.useRef<HTMLButtonElement | HTMLAnchorElement>(null)
return <chakra.button as="a" ref={ref} />
```

## 🚀 New behavior

Installed and integrated `@polymorphic-factory/react`,
deprecated unused types,
and updated pnpms behaviour to install the mono repo dependencies to make the build green.

```tsx
const ref = React.useRef<HTMLAnchorElement>(null)
return <chakra.button as="a" ref={ref} />
```

Yes, because of the updated type, there is a high chance the userland code will break somewhere.

## 📝 Additional Information

- [ ] create a changeset
- [ ] gather feedback, how is the autocomplete ergonomics, etc?


.npmrc
```diff
@@ -1,3 +1,3 @@
 strict-peer-dependencies=false
 enable-pre-post-scripts=true
-
+node-linker=hoisted
```
TimKolberger, line 3: This addresses the issue during build:
```
The inferred type of "X" cannot be named without a reference to "@polymorphic-factory/react"
```

Fix taken from https://github.com/microsoft/TypeScript/issues/42873#issuecomment-1481530011

PR #7553, opened 2023-04-08
fix: make use-image respect crossOrigin property

AlphaNecron:
fix: make use-image respect crossOrigin property in order to avoid errors when loading image without CORS, leading to false status.

Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior


AlphaNecron:
Can you add a `changeset` for me? My network connection is so fluctuating that pulling npm packages takes forever :eyes:.

PR #7585, opened 2023-04-18
Feat: storybook theme override

bohdanbirdie:
Closes # 

## 📝 Description
This is an attempt to allow theme override in the hosted Storybook as well as the composed storybook page.
Refs:
- https://github.com/chakra-ui/chakra-ui/discussions/6824
- https://github.com/chakra-ui/chakra-ui/discussions/5987#discussioncomment-4907794

And we have exactly the same issues as in the last link above. It's great to have your hosted Storybook page composed into the local version, but it's not so useful without the ability to override the default theme.

Current idea is to allow storing theme JSON in the Local Storage, so the consumer can write a custom theme there and it will be used in the remote, deployed Storybook.

No way to override the default theme in the hosted Storybook

## 🚀 New behavior

Added ability to override default theme in the hosted Storybook

No

## 📝 Additional Information
I will also update the docs if this idea is approved.

image

Local version:
image

### DIFF
--- a/.changeset/wild-tigers-worry.md
+++ b/.changeset/wild-tigers-worry.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/storybook-addon": patch
+---
+
+Updated storybook setup to allow custom theme override in composed version.
+Allow storing theme override in the local storage.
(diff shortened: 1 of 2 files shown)


bohdanbirdie:
Ok, looks like it does not really work because it's cross-domain :(

PR #7603, opened 2023-04-26
fix(avatar): render a single initial when only one name is passed

### LINKED ISSUE #7602: Single name in avatar renders 2 'initials'
### Description

In version 2.6.0 of Chakra UI, the Avatar initials logic has changed:

Before:
```ts
function initials(name) {
  const [firstName, lastName] = name.split(" ");
  return firstName && lastName ? `${firstName.charAt(0)}${lastName.charAt(0)}` : firstName.charAt(0);
}
```

After:
```ts
export function initials(name: string) {
  const names = name.split(" ")
  const firstName = names.at(0) ?? ""
  const lastName = names.at(-1) ?? ""
  return firstName && lastName
    ? `${firstName.charAt(0)}${lastName.charAt(0)}`
    : firstName.charAt(0)
}
```

The 'before' situation would correctly render a single character when passing a single name:

```tsx
<Avatar name="Name" /> // => "N"
```

In 2.6.0 however, this now renders "NN" because for a single name, the `names` array equals `["Name"]`, and `.at(0)` and `.at(-1)` will return the same string.

### Link to Reproduction

https://codesandbox.io/s/nervous-kirch-txe81n

### Steps to reproduce

_No response_

### Chakra UI Version

2.6.0

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

mauricekleine:
Closes #7602

## 📝 Description

This fixes the issue that the 'initials' are rendered twice when a single name is passed to the Avatar component

```tsx
<Avatar name="Name" /> // => "NN"
```

## 🚀 New behavior

```tsx
<Avatar name="Name" /> // => "N"
```

This is how it worked before Chakra 2.6.0

No

### DIFF
--- a/.changeset/breezy-crabs-swim.md
+++ b/.changeset/breezy-crabs-swim.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/avatar": patch
+---
+
+Fix issue where the initial was incorrect when only one name is passed
--- a/packages/components/avatar/src/avatar-name.tsx
+++ b/packages/components/avatar/src/avatar-name.tsx
@@ -5,7 +5,7 @@ import { AvatarOptions } from "./avatar-types"
 export function initials(name: string) {
   const names = name.split(" ")
   const firstName = names.at(0) ?? ""
-  const lastName = names.at(-1) ?? ""
+  const lastName = names.length > 1 ? names.at(-1) : ""
   return firstName && lastName
     ? `${firstName.charAt(0)}${lastName.charAt(0)}`
     : firstName.charAt(0)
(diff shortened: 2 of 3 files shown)


leoliu:
@mauricekleine This change breaks on iOS 14 due to lack of `.at` function. Looking at MDN doc `.at` appears much too recent. WDYT?

mauricekleine:
@leoliu this change does not introduce `.at` - it was already used before. Please open a new issue for this

leoliu:
@mauricekleine sorry, my bad. I was upgrading from 2.5.x to 2.6.x and somehow mixed up when I switched from chrome to safari to comment. It looks like this is being addressed https://github.com/chakra-ui/chakra-ui/pull/7624

Issue #7630, opened 2023-05-05
Type AccordionIconProps failure with 'as' prop but supports it.


FredPerr:
### Description

In TypeScript, when using the `AccordionIconProps`, the `as` props is not recognizable, but ChakraUI does supports it as any other component extending `Icon` does:

### Link to Reproduction

https://codesandbox.io/s/hardcore-carson-ewzofr?file=/src/index.tsx

### Steps to reproduce

1. Follow instruction on how to install Chakra-UI on a project

### Chakra UI Version

2.6.1

### Browser

Firefox 112

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

PR #7642, opened 2023-05-09
'as' prop failure: accordion-icon 

ankitkdev:
Closes #https://github.com/chakra-ui/chakra-ui/issues/7630

## 📝 Description

> Fixed Type AccordionIconProps failure with 'as' prop but supports it.

> It is possible to use the as prop with AccordionIcon but TypeScript does not recognize it. The type is not built properly.

## 🚀 New behavior

> as prop is now recognized.

No


segunadebayo:
Closing this. Already merged a solution

PR #7671, opened 2023-05-18
fix: avatar excess label

### LINKED ISSUE #7666: AvatarGroup does not change the size of text displaying exceeding avatars
### Description

When i use an AvatarGroup
And i set the size to "xs"
Then i expect the font size of the span providing the exceeding avatars number to scale (e.g. be smaller and fit the avatar)

Current behavior:
The font size of the span does not change if providing a size options
This can cause overflow

Expected behavior:
The font size of the span does change according to the size options
https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/avatar/src/avatar-group.tsx#L99

### Link to Reproduction

see screenshot

### Steps to reproduce

_No response_

### Chakra UI Version

2.3.5

### Browser

Google Chrome

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

image

kashikarparth:
Closes #7666

## 📝 Description

> Added a `fontSize` adjustment in the theme for `Avatar` excess label. Applicable for sizes `xs` and `2xs` only, as we don't want the label to grow beyond its default value for larger sizes, but need it to shrink for these small ones.

> The excess label can tend to overflow from the background, as mentioned in the original issue.

## 🚀 New behavior

> Excess label no longer overflows, and shrinks for `xs` and `2xs` sizes, due to introduced changes in `Avatar` theme.

Excess labels xs
Excess labels 2xs

## 📝 Additional Information: N/A

### DIFF
--- a/packages/components/theme/src/components/avatar.ts
+++ b/packages/components/theme/src/components/avatar.ts
@@ -63,7 +63,14 @@ const baseStyle = definePartsStyle((props) => ({
   container: runIfFn(baseStyleContainer, props),
 }))
 
-function getSize(size: keyof typeof themeSizes | "100%") {
+type GetSizeOpts = {
+  adjustExcessLabelFontSize?: boolean
+}
+
+function getSize(
+  size: keyof typeof themeSizes | "100%",
+  { adjustExcessLabelFontSize }: GetSizeOpts = {},
+) {
   const themeSize = size !== "100%" ? themeSizes[size] : undefined
   return definePartsStyle({
     container: {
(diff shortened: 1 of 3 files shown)


sergeyt (approved):

packages/components/theme/src/components/avatar.ts
```diff
@@ -43,41 +47,45 @@ const baseStyleContainer = defineStyle((props) => {
...
       [$border.variable]: "colors.gray.800",
     },
-    borderColor: $border.reference,
-    verticalAlign: "top",
   }
 })
 
+const baseStyleLabel = defineStyle({
+  fontSize: $fs.reference,
+  lineHeight: "1",
```
barrymichaeldoyle, line 69: This `lineHeight` addition breaks text alignment 😢

Issue #7738, opened 2023-06-06
Button: color prop is getting ignored when using layerStyles


srowe0091:
### Description

When I style the buttons with layerStyles, I expect all the styling properties to be applied to the component, but the texts `color` property is not getting applied

### Link to Reproduction

https://codesandbox.io/s/dreamy-tamas-f0d879?file=/src/index.tsx

### Steps to reproduce

codesandbox shows how to reproduce

### Chakra UI Version

2.7.0

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

This worked in 2.6.1 and broke in version 2.7.0

segunadebayo:
This is a design flaw in v2. We've now fixed this in the upcoming v3 version

PR #7746, opened 2023-06-10
fix(button/style-system): stop layerStyles color props from being discarded when injected to component

snovosel:
Closes # [7738](https://github.com/chakra-ui/chakra-ui/issues/7738)

## 📝 Description
This PR fixes an issue when using layerStyles in which the component would not inherit any color related props that are defined within the layerStyles object which exist already in the component's existing styles props.

Currently, if `color` or `bg` styles are added to the layerStyles object, those values would be discarded via the function `getWithPriority` 

## 🚀 New behavior
This PR allows certain style keys to be overwritten if included in the layerStyles object.

No

### DIFF
--- a/.changeset/large-radios-invite.md
+++ b/.changeset/large-radios-invite.md
@@ -1,0 +1,9 @@
+---
+"@chakra-ui/styled-system": patch
+"@chakra-ui/button": patch
+---
+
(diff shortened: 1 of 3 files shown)


segunadebayo:
Thanks for working on this.

Issue #7740, opened 2023-06-07
`__css` has higher priority than `color` for `Icon`


julioolvr:
### Description

When I pass both the `__css` and the `color` prop to the `Icon` component, I expect `color` to have higher priority, but `__css` wins.

From the docs here (emphasis mine) https://chakra-ui.com/docs/styled-system/component-style#return-value:

> Please note that we are passing the styles to the prop __css. It has the same API as [the sx prop](https://chakra-ui.com/docs/styled-system/the-sx-prop), **but has a lower style priority**. This means **you can override the style properties with chakra style props**.

But the following renders a red icon:

```tsx
<PhoneIcon color="blue" __css={{ color: 'red' }}/>
```

### Link to Reproduction

https://codesandbox.io/s/crazy-banzai-9c3yl9?file=/src/index.tsx

### Steps to reproduce

Just checking the rendered result is enough

### Chakra UI Version

2.0.18

### Browser

Google Chrome 114.0.5735.90

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

The code does seem to do this intentionally, but I wonder if it was intended (I found this originally on Chakra 1, and saw the same behavior on Chakra 2)

https://github.com/chakra-ui/chakra-ui/blob/2456a7090439ebd1a1d1cdee36daa472d12ce6f7/packages/components/icon/src/icon.tsx#L63-L64

snovosel:
I feel as though this should be intended. It looks like throughout other components there is a similar pattern of spreading the `_css` prop at the end.

Also, I'm trying to think of a use-case where you would want to pass both the `color` style prop and also `_css` including a different color. I would think the `_css` prop would be used as a way to override the default allowed props as well as add any additional styles. 

Open to any comments though

julioolvr:
Looking at the code for the `Button` component, it seems to rely on `chakra.button` that explicitly sets `__css` as the lowest priority https://github.com/chakra-ui/chakra-ui/blob/2456a7090439ebd1a1d1cdee36daa472d12ce6f7/packages/core/system/src/system.ts#L53-L59

I tried the following (and added it to the codesandbox example for comparison) and the `color` prop effectively wins and text shows as blue:

```tsx
<Button __css={{ backgroundColor: "pink", color: "red" }} color="blue">
  Button
</Button>
```

[...]

PR #7747, opened 2023-06-11
fix(icon-colors): fix priority of __css prop color with relation to icon color prop

snovosel:
Closes # [7740](https://github.com/chakra-ui/chakra-ui/issues/7740)

## 📝 Description
This PR fixes and issue where the `color` prop was merged with the final `__css` prop when rendering an `Icon` component. In the event that the `__css` prop includes a `color` property, the `__css` prop color would take priority as it is spread following the color prop in the `Icon` style object:

[Icon Styles](https://github.com/chakra-ui/chakra-ui/blob/2456a7090439ebd1a1d1cdee36daa472d12ce6f7/packages/components/icon/src/icon.tsx#LL57C3-L66C4)

```
const styles: SystemStyleObject = {
    w: "1em",
    h: "1em",
    display: "inline-block",
    lineHeight: "1em",
    flexShrink: 0,
    color,
    ...__css,
    ...customStyles,
  }
```

When an `<Icon />`  component is given both a `__css` prop which includes a color property, as well as a `color` prop, the `__css` prop color value will take priority.

## 🚀 New behavior
If both props are provided, the `color` prop will take priority

No


snovosel:
Can anyone take a look at this?

rever96:
adding theme to icon with a default color still breaks all the color props

segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version, where you now pass an array of styles. This way, you won't need to rely on __css anymore.

PR #7761, opened 2023-06-13
Use clipboard imperative usage

### LINKED ISSUE #7754: [useClipboard] Allow imperative usage
### Description

I need to copy a current URL from a browser using a SSR app.

The problem is that I cannot use 

```tsx
const clipboard = useClipboard(window.location.toString());
```

since the `window` is undefined on the server side. 

Because of that, I need to use an extra `useEffect` to notify that I'm on the client and then update the state. Notice that `typeof window === "undefined" ? "" : window.location.toString()` is no-go since you might render the value on page so it is not sustainable.

### Proposed Solution or API

The solution would be to allow imperative usage

```tsx
const clipboard = useClipboard();

return <button onClick={() => clipboard.onCopy(window.location.toString())}>Copy URL</button>
```

### Alternatives

_No response_

### Additional Information

In my head, this is not a breaking change, but I might be wrong. Also, I would like to work on PR for this if you folks think it is okay to add it to Chakra UI.

kristian240:
Closes #7754

## 📝 Description

Added possibility to call `copy` from `useClipboard` imperatively. 

To pass in a value that is obtained imperatively, e.g. `prompt` or value from a global object, you would need to either rely on `useEffect` and violate [chains of computations](https://react.dev/learn/you-might-not-need-an-effect#chains-of-computations) or in SSR do a two-pass render to get access to the global object.

## 🚀 New behavior

By calling the `useClipboard` without `value` arguments, i.e. no arguments on only `optionsOrTimeout` argument you are able to pass in value to `onCopy` callback.

## 📝 Additional Information

I will link docs PR here as well once I create it.

-----

Hey @segunadebayo! Here is the PR. Please checkout other `feat` commits as I had multiple ideas on how to solve the DX. I find this the best, but I would like to hear your opinion on this one.

### DIFF
--- a/packages/legacy/hooks/src/use-clipboard.ts
+++ b/packages/legacy/hooks/src/use-clipboard.ts
@@ -23,8 +23,8 @@ export interface UseClipboardOptions {
  *
  * @see Docs https://chakra-ui.com/docs/hooks/use-clipboard
  */
-export function useClipboard(
-  value: string,
+export function useClipboard<TValue extends string | undefined = undefined>(
+  value?: TValue,
   optionsOrTimeout: number | UseClipboardOptions = {},
 ) {
   const [hasCopied, setHasCopied] = useState(false)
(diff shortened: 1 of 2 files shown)


vito:
It looks like the docs for this were merged, but not the feature itself :sweat_smile:

segunadebayo:
Thanks for working on this. We no longer use this hook in v3 since we now have a `Clipboard` component we ship in Chakra.

We'll cover this in the new docs

PR #7770, opened 2023-06-15
fix: add zIndex to closeButton for targetability when overlapped by other content

### LINKED ISSUE #7717: DrawerCloseButton sometimes not full width without DrawerHeader
### Description

When I create a Drawer without having a DrawerHeader as a child of DrawerContent, the DrawerCloseButton appears to be normal visually, but can only be clicked on its right side in some cases. The hover behavior of the DrawerCloseButton also only applies when the cursor is hovering over the right side of the button in these cases.

This happens when an Input Element is used as a child of the DrawerBody component, but I have not checked extensively to see if it happens when using other elements as well.

If DrawerHeader is a required child or DrawerContent, I failed to see it in the documentation. Should there be some kind of default behavior if no DrawerHeader is used?

### Link to Reproduction

https://codesandbox.io/s/zealous-davinci-t6945z

### Steps to reproduce

1. Go to [this example](https://codesandbox.io/s/zealous-davinci-t6945z)
2. When the example finishes loading, click on Open Drawer
3. Hover/Click the left side of the DrawerCloseButton
4. Check intended behaviors
5. Hover/Click the right side of the DrawerCloseButton
6. Check intended behaviors

### Chakra UI Version

2.5.1

### Browser

Google Chrome 113.0.5672.127
Firefox 102.11.0esr

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

_No response_

HJordan35:
Closes #7717

## 📝 Description

This PR fixes the issue of the CloseButton not having a clickable area when DrawerContent elements overlap with the CloseButton's absolute position. 

The lack of a baseline zIndex on the CloseButton allows content inside of the drawer to extend into the clickable area of the CloseButton's absolute position. See example here : https://codesandbox.io/s/zealous-davinci-t6945z

## 🚀 New behavior

This PR adds a simple `zIndex: 1` styles to the CloseButton's base styles to ensure it will appear above drawer content added without a zIndex of its own.

No


segunadebayo:
I would recommend doing this in your app code.

Issue #7775, opened 2023-06-17
Accordion doesn't open if you click it multipe times fast enough, and it loses track of state.


fidalgodev:
### Description

If you click the accordion button 3 times fast, the accordion acts up and loses state. It stays closed, even tho the Arrow icon stays on the open state. Then you need to click it twice for it to open again (first for it to close, even tho the panel is not rendering, and then again to open normally):

https://github.com/chakra-ui/chakra-ui/assets/41859298/027d6c14-e85d-474d-86d7-99ef1aab6912

JSX:

```
<Accordion allowToggle defaultIndex={[0]}>
  <AccordionItem>
    <AccordionButton>
      <Box as="span" flex="1">
        Accounts
      </Box>
      <AccordionIcon />
    </AccordionButton>
    <AccordionPanel flexDirection="column" display="flex">
      <NavLink as={NextLink} href={' '}>
        Monzo
      </NavLink>
      <NavLink as={NextLink} href={' '}>
        Starling
      </NavLink>
      <NavLink as={NextLink} href={' '}>
        Chase
      </NavLink>
    </AccordionPanel>
  </AccordionItem>
</Accordion>
```

### Link to Reproduction

N/A

### Steps to reproduce

1. Create an accordion
2. Click 3 times fast on the Button that triggers the opening
3. See the accordion panel not opening, even tho the arrow state is on the open state.

```
"@chakra-ui/next-js": "^2.1.4",
 "@chakra-ui/react": "^2.7.0",
```

### Chakra UI Version

2.1.4 next-js, 2.7.0 react js

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

henriquesml:
Apparently, the issue occurs in the Collapse component due to the missing display. I've opened a fix proposition PR at #7819

chuganzy:
There seems to be a race condition in `framer-motion` that is causing this to happen. Looking into it later.

chuganzy:
@nelsieborja That is because of the `framer/motion`'s version that website relies on. It uses older one, which does not have this issue.

stefanalexlupu:
Hi, I am also encountering this problem on nested accordions. I reproduced this issue here: https://codesandbox.io/s/nested-accordions-in-chakra-ui-59ncv4?file=/src/App.js

When an accordion is open, if you double click, it closes but the button remains in the open state

framer-motion v10 (mentioning this because comments above indicate it might be a framer motion issue)

[...]

PR #7819, opened 2023-07-01
Opening Collapse correctly through many clicks

henriquesml:
## 📝 Description

As @fidalgodev reported in #7775, the Collapse transition does not work when you rapidly click multiple times.
@segunadebayo, why do we set the `display` property to `block` or `none` depending on the `startingHeight`?.
Another solution for this is to set unmountOnExit to true, but it doesn't seem right to me.

The exit animation doesn't happen.

## 🚀 New behavior

The exit animation runs smoothly.

Not sure.


forrestwilkins:
@henriquesml Was this already resolved elsewhere? It looks like this was intended to solve https://github.com/chakra-ui/chakra-ui/issues/7775.

Issue #7251, opened 2023-01-17
Would be great to have a "Toggle Button" component!


DarrenBaldwin07:
### Description

Currently, Chakra has the Switch (https://chakra-ui.com/docs/components/switch) however this has no indicator (label). I would propose adding a new "Toggle Button" component that simply toggles between 2 different states.

### Problem Statement/Justification

Refer to description.

### Proposed Solution or API

N/A

### Alternatives

N/A

### Additional Information

Here is a example (via: https://turbo.build): 
<img width="115" alt="Screen Shot 2023-01-17 at 7 26 33 AM" src="https://user-images.githubusercontent.com/68653294/212898815-fd6a5280-69ad-4a13-afdc-1a13b618ae11.png">

CaitlinSweeney:
We had to create a `SelectButtonGroup` for our app, basically a Button Toggle, and the option to have multiple buttons. Utilizes the chakra button components. It looks something like this and some style tweaking.

```
import {
  ButtonGroup as CBG,
  ButtonGroupProps,
  forwardRef,
} from '@charka-ui/react';
import React from 'react';

export const ButtonGroup = forwardRef<ButtonGroupProps, 'div'>((props, ref) => {
  const { isAttached, spacing = 2 } = props;
  const styles = {
    flexDirection: { base: 'column', lg: 'row' },
    '& > *:not(style) ~ *:not(style)': {
      marginStart: { base: 0, lg: spacing },
      marginTop: { base: 3, lg: 0 },
    },
  };
  if (!isAttached) {
    return <CBG ref={ref} sx={{ ...styles }} {...props} />;
  }
  return <CBG ref={ref} {...props} />;
});

const inactiveStyles = {
  color: 'gray.700',
  borderColor: 'gray.100',
  bg: 'gray.100',
  _hover: { bg: 'gray.300' },
  _active: { bg: 'gray.200' },
};

/**
 * @example (
 * <SelectButtonGroup onChange={handleChangeValue} value={value}>
 *  <Button value={value1}>label1</Button>
 *  <Button value={value2}>label2</Button>
 * </SelectButtonGroup>
 * )
 * @note
 * must have multiple child elements
 */
export const SelectButtonGroup: React.FC<{
  onChange: (value: string) => void;
  value: string;
  groupProps?: ButtonGroupProps;
}> = ({ onChange, value, children, groupProps }) => {
  if (!children) throw new Error('Children required');
  // iterate over array of child nodes to apply extended props
[...]

Issue #171, opened 2019-10-10
Is a toggle button group being considered?


coodoo:
[Something like this](https://material-ui.com/components/toggle-button/) from `material-ui` would come in real handy, mind giving it a thought and provide it out of the box?

<img width="944" alt="Screen Shot 2019-10-10 at 9 49 20 AM" src="https://user-images.githubusercontent.com/325936/66532786-42f1f480-eb43-11e9-9eef-6b5a1c9841c7.png">

segunadebayo:
Hi @coodoo,

Thanks for your request. I'll add it to the backlog. I'd like to move Chakra to TS before adding any new component. I'll definitely consider this.

ericclemmons:
I wouldn't consider this stale, but "blocked" or "on-hold".

coodoo:
bump? 😬

ualtinok:
bumpy

emileond:
bump 👀

iamminci:
bump plz

cameronelliott:
Thank you for making Chakra, this would be a helpful feature.

rtman:
@segunadebayo any chance this is in the works?

thegreatsantini:
Bump

ksanmugam:
Still looking forward to this request

emileond:
Seems like this is coming in the next major release, but if you need it now. It's very straightforward to build your own using ChakraUI's native components, here's one I built - hope it helps:

https://codesandbox.io/s/chakra-ui-toggle-btn-3n3gkn

maximilianotaverna:
bump

PR #7866, opened 2023-07-18
feat: add toggle button component

spiltbeans:
Closes # 

## 📝 Description

This PR adds a `ToggleButtonGroup` component to Chakra UI.

Development of a Chakra UI `ToggleButton` has been referenced in issues #7251 #171 and discussion #7724.

There is no `ToggleButtonGroup` that keeps track of the selected state.

Currently, a workaround using the existing Chakra UI API would be to use a `ButtonGroup` with `Button` children and use some state to change on an `onClick` event which determines which `Button` child is active.

## 🚀 New behavior

`ToggleButtonGroup` acts as a wrapper for Chakra UI `Button` or `button` components. `ToggleButtonGroup` is opinionated about its children and will coerce either the `Button` or `button` component to the Chakra UI `Button` component. Props passed to `ToggleButtonGroup` children will be passed to the coerced `Button` component.

`ToggleButtonGroup` extends props from `ButtonGroupProps` type so you can pass any props available in the `ButtonGroup` component.

```js
<ToggleButtonGroup>
    <Button>Option 1</Button>
    <Button>Option 2</Button>
    <Button>Option 3</Button>
</ToggleButtonGroup>
```

`ToggleButtonGroup` has the following additional props:
- `initialIndex?: number` which manually sets the initial active button
- `allowNone?: boolean` which allows or denies no buttons being selected
	- By default set to `true`
	- Note: if `false`, the initial index will default to the first child
- `exclusive?: boolean` which allows or denies multi-button selection.
	- By default set to `true`

`ToggleButtonGroup` overrides `onChange` from `ButtonGroupProps`
- `onChange?: (v: (string | number | JSX.Element)[] | null) => any` where `v` is the selected value(s) or `null` if none selected
	- a child's `value` is determined in the following process:
		- `value` prop if it exists
		- `innerHTML` if it is a string
		- defaults to `JSX.Element`

No

## 📝 Additional Information
toggle

The implementation of `onChange` is useful for my use cases but I'd be interested to see if the community expects alternative behavior for `onChange`.

While the implementation expects a `Button` or `button` type, there is no type enforcement for children to be `Button` or `button` type. I'd like to know if others believe type enforcement is necessary since `ButtonGroup` doesn't type enforce.

Another design choice was to have the selected state held internally to the component. Where other UI libraries like Material UI use the pattern of passing a `value` prop from the wrapper, my implementation handles the current value internally and emits the selected values through `onChange`.

```js
<ToggleButtonGroup onChange={v => console.log(v)}>
    <Button>Option 1</Button>
    <Button>Option 2</Button>
    <Button>Option 3</Button>
</ToggleButtonGroup>
```

rather than the pattern found [here](https://mui.com/material-ui/react-toggle-button/#enforce-value-set)

A limitation is that you can't programmatically change the selected values, but only through the `onClick` event. Again, I'd like to hear from the community about which implementation is preferred.


segunadebayo:
Thanks for working on this @spiltbeans.

We'd prefer to integrate the ToggleGroup from our [Zag.js](https://zagjs.com/components/react/toggle-group) library in the coming major release.

Philzen:
I was under the impression from following the discussion in #7724 that @spiltbeans was actually aiming at what Zag.js calls a [Segmented Control](https://zagjs.com/components/react/segmented-control) with this PR. Isn't it @spiltbeans?

Issue #7627, opened 2023-05-04
Using responsive sizes with `IconButton` uses incorrect padding


joekrill:
### Description

When using a responsive `size` value (either array or object syntax) for an `IconButton`, the button padding is different compared to using the same `size` value directly, as seen here:

<img width="529" alt="Screenshot 2023-05-04 at 4 48 20 PM" src="https://user-images.githubusercontent.com/393220/236325734-35aa9377-dc69-400b-8961-7b7bfb44170b.png">

It appears to have to do with `padding-inline` and `padding` values conflicting. In the case of using a non-responsive size, we end up with these rules:

```
padding-inline-start: var(--chakra-space-4);
padding-inline-end var(--chakra-space-4);
padding: 0;
```

so we end up with no padding (`padding:0` overrides the `padding-inline-*` values). But when using responsive sizes, an additional media query rule gets applied which _only_ includes the `padding-inline` values, so those end up taking precedent, overriding the `padding: 0` value. So we en up with a left/right padding value of `var(--chakra-space-4);`

But with a response 

### Link to Reproduction

https://codesandbox.io/s/compassionate-ritchie-52q4eg

### Steps to reproduce

Set a responsive `size` property for an `IconButton`.

### Chakra UI Version

2.6.1

### Browser

Chrome 112.0.5615.137 , Firefox 112.0.2

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

Dhruviika:
hey. i want to work on this issue.

Jhappy77:
Would be great to get this PR merged @segunadebayo , looks like it's been waiting around for a while :)

segunadebayo:
This is now fixed.

PR #7876, opened 2023-07-21
fix(icon-button): changed the definition of padding

lukke-dev:
Closes [#7627](https://github.com/chakra-ui/chakra-ui/issues/7627)

## 📝 Description

Changed the definition of padding to ensure the responsiveness
of the size worked

When using a responsive size value (either array or object syntax) for an IconButton, the button padding is different compared to using the same size value directly

## 🚀 New behavior

Priority was given to padding that had iconbutton passing it with style

No


segunadebayo:
Thanks for working on this.

PR #7907, opened 2023-08-01
fix: Backward comaptability of Button disabled attribute

### LINKED ISSUE #7269: disabled prop does not work for Button and Radio despite the type allowing it
### Description

The type for Button and Radio allows both the disabled and isDisabled props, but only isDisabled works. disabled will be silently ignored with no warning whatsoever. 

Please fix this by either changing the type so disabled is no longer a prop of  Button, Radio or making it so that disabled works.

### Link to Reproduction

https://codesandbox.io/s/intelligent-jang-wu06rc

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Chakra UI Version

2.4.9

### Browser

_No response_

### Operating System

- [x] macOS
- [x] Windows
- [x] Linux

### Additional Information

_No response_

### LINKED ISSUE #7816: Disabled html attribute is not set on Button if passed in as prop to MenuItem
### Description

If you pass in a `Button` as an `as` prop to `MenuItem` and disable it, it will not render with the `disabled` html attribute.

This issue is not present in `@chakra-ui/button: "2.0.15"`

## Examples

```
<!-- button won't have disabled attribute -->
<MenuItem as={Button} isDisabled>My Disabled Item</MenuItem>

<!-- button has disabled attribute -->
<Button isDisabled>My Disabled Button</Button>
```

### Link to Reproduction

https://codesandbox.io/s/modest-hoover-yfrdt6?file=/src/index.js

### Steps to reproduce

_No response_

### Chakra UI Version

2.7.1

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

### LINKED ISSUE #7905: disabled attribute doesn't work on <Button>
### Description

When I update `@chakra-ui/button` from 2.0.13 to 2.1.0 disable attribute doesn't work because of https://github.com/chakra-ui/chakra-ui/commit/0f376650efce9eb2d2cba3484ffe4e3a73364504. That's violation of SemVer because:
> MINOR version when you add functionality in a backward compatible manner

### Link to Reproduction

https://example.com

### Steps to reproduce

_No response_

### Chakra UI Version

2.8.0

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

BANOnotIT:
Closes #7905, closes #7816, closes #7269

## 📝 Description

`disabled` attribute can't be omitted because that requires BREAKING change

All version @chakra-ui/button@>2.0.16 are incompatible with @chakra-ui/button@2.0.15

## 🚀 New behavior

It respects disabled attribute

This is a revert of BREAKING change

## 📝 Additional Information

isDisabled can be a MINOR change but removing disabled prop behavior is a BRAKING change

### DIFF
--- a/.changeset/modern-ties-tickle.md
+++ b/.changeset/modern-ties-tickle.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/button": patch
+---
+
+fix: Backward comaptability of Button disabled attribute
--- a/packages/components/button/src/button.tsx
+++ b/packages/components/button/src/button.tsx
@@ -46,6 +46,7 @@ export const Button = forwardRef<ButtonProps, "button">((props, ref) => {
     spinnerPlacement = "start",
     className,
     as,
+    disabled,
     ...rest
   } = omitThemingProps(props)
 
@@ -87,7 +88,7 @@ export const Button = forwardRef<ButtonProps, "button">((props, ref) => {
       __css={buttonStyles}
       className={cx("chakra-button", className)}
       {...rest}
-      disabled={isDisabled || isLoading}
+      disabled={isDisabled || isLoading || disabled}
     >
       {isLoading && spinnerPlacement === "start" && (
         <ButtonSpinner


Philzen:
@BANOnotIT why not simply `git revert 0f376650efce9eb2d2cba3484ffe4e3a73364504`? It comes down to the same thing w/o changing the Component's signature.

BANOnotIT:
> It comes down to the same thing w/o changing the Component's signature.

It doesn't since in mentioned commit disabled prop is placed after `{...rest}`. If you revert it then `isDisabled={true} disabled={false}` will result in override and element will not be disabled at all. My PR merges current behavior with the fact that `disabled={true}` should disable things anyway.

Philzen:
> It comes down to the same thing w/o changing the Component's signature.

@BANOnotIT indeed i was wrong here. But with your proposed PR for me the problem still exists that no workaround is possible for #7965, as `isLoading` will still switch it to disabled, even if `disabled={false}` (which also feels contradictory), which currently seriously hurts accessibility for visually impaired people. 

Also, adding `disabled` to the Components' signature opposes what i believe @segunadebayo had in mind in the first place with 0f376650efce9eb2d2cba3484ffe4e3a73364504 – the idea here was to bring people to use `isDisabled` instead of `disabled` as far as i understand it.

BANOnotIT:
> which currently seriously hurts accessibility for visually impaired people.

Yeah... That's definitely needs some consideration.

> adding disabled to the Components' signature

`disabled` is already in components' signature and was since 2.0.0. This PR just brings back functionality that was declared in 2.0.0 to comply with Semantic Versioning. It's critical in my project to bring back functionality that was declared and was neither properly deprecated nor given warnings. Mentioned commit is "BREAKING CHANGE" and can only occure in v3 and never in 2.*.

Imo it's more crucial to bring back functionality and then think on how we can design future interface better. But this attribute is used already and used in many projects in many places. One can't simply remove functionality in fix release.

Philzen:
> One can't simply remove functionality in fix release.

We're totally on the same page here, a breaking change in a minor version is unacceptable in a semver world.

> This PR just brings back functionality that was declared in 2.0.0 

No it doesn't (completely). Because in any version before 0f376650efce9eb2d2cba3484ffe4e3a73364504, i could set `disabled={false} isLoading={true}` to work around the accessability problem – but now it's broken and there's no way to keep the focus on the input field. So even if this PR get's merged, for me Chakra will still have a breaking change.

BANOnotIT:
> for me Chakra will still have a breaking change.

Got it. Then yeah, I think then only revert is a right thing to do.

cc @segunadebayo

ShawnFumo:
I understand the thinking for this (always good to help with accessibility!), but is a bit frustrating that seemingly a decision was made to revert back in August and it never happened. Is there anything I can do to help this along? It is a moot point for us now (deploying a fix to prod today), but would like to avoid any other companies getting the unfortunate surprise that we did. Maybe we could get the original patch comment into the full changelog as well, since it seems like it never got added there.

My person opinion is to do a quick revert, then if the functionality is important enough, do a bigger version change to indicate it is a breaking change and use TypeScript's `Omit` feature to remove `disabled` from `Button` entirely. That way it won't even compile and will be very clear what is going on. While there could be a console warning of some sort, I feel like this is a perfect situation where TypeScript types would be doing their job. 

I know this probably feels like it was a small change, but you never know how end users may be using it. We unfortunately happened to be using `disabled` in a basic wrapped component for forms, and happened to be using it instead of `isLoading`, probably due to a stylistic choice back when it was implemented. Somehow we didn't catch it during testing, and ended up with a few weeks of no double-click protection on form submissions in prod. That includes duplicate transactions sent to banks needing reports, reversals, etc. Hoping we can avoid anyone else having to deal with this kind of headache if possible.

Philzen:
> My personal opinion is to do a quick revert, then if the functionality is important enough,

@ShawnFumo Yes, reverting that change would be most appreciated. Otherwise our current codebase is stuck with an older ChakraUI version, b/c we need to be able to set `isLoading={true}` without the button losing focus (which it does when being disabled).

augustoabreu:
@Philzen any updates on this? this is a breaking change that was wrongly introduced and it's been hanging for a while...

Philzen:
> @Philzen any updates on this? this is a breaking change that was wrongly introduced and it's been hanging for a while...

@augustoabreu i've now created #8462 which simply reverts this breaking change. Just checked on the docs playground and it seems to work as expected again, try for yourself: https://chakra-ui-website-git-fork-philzen-7269-7816-7-fafd11-chakra-ui.vercel.app/docs/components/button

segunadebayo:
Merged a recent for this already

PR #7944, opened 2023-08-14
fix: avatar image fallback alt when no name is provided

### LINKED ISSUE #7833: Avatar alt text
### Description

How to include alternative text to `img` without a `name` being provided.

In our specific use case, we are not wanting to display the user initials and just have the fallback be a user icon. The issue with this is that we aren't passing in a `name` so that the fallback initials will be ignored. With doing this, the `alt` attribute isn't getting applied and is causing accessibility errors. 

_Note:_ we are always passing in an `aria-label`.

Things I've tried: 
- Passing in `alt` and `aria-label` but those attributes are getting applied to the `span`. 
- Passing in `ignoreFallback` but this prevents the fallback icon from being rendered.

Screen Shot 2023-07-07 at 10 33 24 AM

### Proposed Solution or API

Possible solutions: 

- Allow for an `alt` prop to be passed in and applied to the `svg`. 
- Create a new `fallbackOrder` prop which will allow consumers to decide the fallback + order (EX: `['src', 'icon']`. With this, we can now pass in the `name` prop.

### Alternatives

_No response_

### Additional Information

_No response_

jeferson-sb:
Closes #7833 

## 📝 Description

Small fix to Avatar image when a `src` is provided but no name is, allow `iconLabel` to be used as fallback similar to the way it does for the fallback svg.

Avatar image alt attribute value only supports `name`

## 🚀 New behavior

Small fix to Avatar image when a `src` is provided but no name is, allow `iconLabel` to be used as fallback similar to the way it does for the fallback svg.

No


segunadebayo:
Thanks for working on this.

PR #7949, opened 2023-08-15
fix: use step increment precision when precisionProp is not specified

markdowney:
## 📝 Description

When using the `NumberInput`, if your value is an Integer (eg: 0 or 1) and you haven't specified a precision,
holding down Ctrl or ⌘, and pressing ⬆ or ⬇ should update the value by 0.1 * step, but it doesn't.

```
<NumberInput value={0} step={1} />
```
Holding down Ctrl or ⌘, and pressing ⬆ or ⬇ doesn't update the value.

## 🚀 New behavior

```
<NumberInput value={0} step={1} />
```
Holding down Ctrl or ⌘, and pressing ⬆ or ⬇ increment or decrements the value by 0.1 * step.

No.

### DIFF
--- a/.changeset/clever-tips-hope.md
+++ b/.changeset/clever-tips-hope.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/number-input": patch
+"@chakra-ui/counter": patch
+---
+
+Fixed increment/decrement of 0.1 \* step when precision is not specified.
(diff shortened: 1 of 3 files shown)


segunadebayo:
Thanks for this PR. We'll address this in v3.

PR #7959, opened 2023-08-23
fix: css inconsistence in input field appearence

### LINKED ISSUE #7939: Input type="search" switches between round and square (borderRadius) on phone
### Description

The search field (input component with type="search") switches between having a high value of borderRadius to having the normal value. This is only when using it on a phone - not when inspecting it on a desktop in phone mode. It is round to begin with and gets square on focus. 

### Link to Reproduction

https://codesandbox.io/s/jolly-parm-tz9t4y?file=/src/index.tsx

### Steps to reproduce

1. Go to [https://codesandbox.io/s/jolly-parm-tz9t4y?file=/src/index.tsx](https://codesandbox.io/s/jolly-parm-tz9t4y?file=/src/index.tsx) on a phone (Iphone in my case).
4. See round search field.
5. Try clicking on it.

### Chakra UI Version

2.5.1

### Browser

Chrome

DavideFrancescon:
Closes #7939

## 📝 Description

Search (and number) input didn't have they appearance reset on IOS causing them to not update their border radius until clicked.

Appearance not reset, border start IOS-like and on click re-render with the right style

## 🚀 New behavior

Appearance now get reset like every other input field, and the border is right at first render

No

### DIFF
--- a/packages/components/css-reset/src/css-reset.tsx
+++ b/packages/components/css-reset/src/css-reset.tsx
@@ -178,12 +178,14 @@ export const CSSReset = ({ scope = "" }: CSSResetProps) => (
         -webkit-appearance: none !important;
       }
 
+      ${scope} input[type="search"],   
       ${scope} input[type="number"] {
-        -moz-appearance: textfield;
+        -moz-appearance: none !important;
       }
 
-      ${scope} input[type="search"] {
-        -webkit-appearance: textfield;
+      ${scope} input[type="search"],   
+      ${scope} input[type="number"] {
+        -webkit-appearance: none !important;
         outline-offset: -2px;
       }
 


adrianflatner:
Hope to get this in🎉 Is there a way to get this prioritized?

packages/components/css-reset/src/css-reset.tsx
```diff
@@ -178,12 +178,14 @@ export const CSSReset = ({ scope = "" }: CSSResetProps) => (
         -webkit-appearance: none !important;
       }
 
+      ${scope} input[type="search"],   
       ${scope} input[type="number"] {
-        -moz-appearance: textfield;
+        -moz-appearance: none !important;
```
segunadebayo, line 183: Do we need to add `!important`?

segunadebayo:
I would recommend applying this to your app code for now

Closing this in favor of the upcoming v3 version which incorporates this

Issue #7964, opened 2023-08-24
Inconsistent tooltip behavior on mobile


nbouvrette:
### Description

It looks like the tooltips have been designed primarily for `hover` events. On mobile this event does not work and we need `onClick` displayable tooltips.

### Link to Reproduction

https://chakra-ui.com/docs/components/tooltip#more-examples

### Steps to reproduce

1. Go to https://chakra-ui.com/docs/components/tooltip#more-examples
2. Put your browser in mobile mode
3. Click on the first button in the "More example section". The button is labeled `Close on Click - true(default)`
4. On the first click, the tooltip shows
5. On the second click the tooltip hides
6. On subsequent clicks, nothing happen

I would expect that on subsequent clicks, the tooltip is toggled.

To fix this I tired using a `state` and `onClick` events, but I for some reason the `Tooltip` component is not letting the `onClick` event through.

Here is my temporary workaround which is behaving as I would expect on mobile:

```tsx
  const [isTooltipOpen, setIsTooltipOpen] = useState(false)

  return (
    <Flex
      onClick={() => {
        setIsTooltipOpen(!isTooltipOpen)
      }}
    >
      <Tooltip
        isOpen={isTooltipOpen}
        onOpen={() => {
          setIsTooltipOpen(true)
        }}
        onClose={() => {
          setIsTooltipOpen(false)
        }}
        label={'Hello world'}
      >
        Help!
      </Tooltip>
    </Flex>
  )
```

### Chakra UI Version

Latest

### Browser

Latest CHrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

jishat:
Can I fix it?

jishat:
I fixed it

segunadebayo:
Tooltips don't work on mobile devices. I recommend using a popover or equivalent instead on mobile.

PR #7967, opened 2023-08-26
fix: #7964 Inconsistent tooltip behavior

jishat:
## 📝 Description:
Did not work hover on mobile

## ⛳️ Current behavior (updates):
We need onClick displayable tooltips.

## 🚀 New behavior:
Now set touch event condition if use from mobile and display tooltips onClick

💣 Is this a breaking change (Yes/No):
No


segunadebayo:
Tooltips should not be used on mobile devices. Prefer to use a popover instead.

PR #7978, opened 2023-08-31
Refactor/Optimize Media Query Module

rustworthy:
## 📝 Description

- making use of 'useMemo' in media-query module and passing exhaustive deps to corresponding useEffect hook (where previously linter-ignore statement was needed to pass linter check);
- removing redudant slicing in MediaQueryListEvent handler (line 57 of packages/components/media-query/src/use-media-query.ts)
- using verbose mediaQueryLists (plural) and  mediaQueryList (singular) instead of 'mql' to avoid cases like `mql.forEach(mql => ...)`

### DIFF
--- a/packages/components/media-query/src/media-query.utils.ts
+++ b/packages/components/media-query/src/media-query.utils.ts
@@ -30,3 +30,5 @@ export function getClosestValue<T = any>(
 
   return undefined
 }
+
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks for working on this @rustworthy.

I've incorporated changes from this into the v3 version we're working on.

Cheers.

PR #8050, opened 2023-10-05
fix: theme takes precedence over component properties

IagoLast:
Fixes the bug reported at
https://github.com/chakra-ui/chakra-ui/discussions/8046

Closes # https://github.com/chakra-ui/chakra-ui/discussions/8046

## 📝 Description

This PR modifies the checkbox label behaviour taking inline props into account 

Theme properties taking precedence over inline ones

## 🚀 New behavior

Inline ones take precedence

It is a bugfix but might break some external apps. 

## 📝 Additional Information

I've added a test using storybook. I got ugly imports. I don't mind removing this file. 
I've also just passed the props to the `getLabelProps` function, not sure of the consecuences of that.
This is my first PR here 😅

Thx for this great project 🙏

### DIFF
--- a/packages/components/checkbox/src/checkbox.tsx
+++ b/packages/components/checkbox/src/checkbox.tsx
@@ -177,7 +177,7 @@ export const Checkbox = forwardRef<CheckboxProps, "input">(function Checkbox(
       {children && (
         <chakra.span
           className="chakra-checkbox__label"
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version which allows for explicit rendering of checkbox parts

PR #8067, opened 2023-10-13
bugfix: removed line-height from avatar fallback label theme

barrymichaeldoyle:
## 📝 Description

> The line height previously set to "1" breaks the alignment of the text inside the Avatar circle. Removing that property fixes the alignment.

> Text was misaligned

Screenshot 2023-10-13 at 10 18 29

## 🚀 New behavior

> Text is no longer misaligned

Screenshot 2023-10-13 at 10 18 37

> No


barrymichaeldoyle:
I don't believe I need to set up a Changeset for this since it is such a minor change.

segunadebayo:
I recommend applying these changes in your app code.

Thanks for understanding.

PR #8073, opened 2023-10-18
CLI support for semantic tokens color scheme

isBatak:
## 📝 Description

This is an opt-in addition to CLI to support inclusion of semantic tokens as `colorScheme` types.

The problem:
Sometimes semantic-tokes could be a complex objects and Designer can use them for defining some some component states:
```ts
 {
        semanticTokens: {
          colors: {
            primary: {
              default: { default: "#64FFDA" },
              strong: { default: "#42A28F" },
              medium: { default: "#24514D" },
              weak: { default: "#152A2D" },
            },
            secondary: {
              default: { default: "#00B0FF" },
              strong: { default: "#0777AA" },
              medium: { default: "#0D3E55" },
              weak: { default: "#11303F" },
            },
          },
        },
      },
```

And if you want to use those colors in for example Button theme, you can define variant like this
```ts
import { defineStyle, defineStyleConfig } from '@chakra-ui/react';
const commonStyles = defineStyle({
	fontSize: '0.6875rem',
	borderRadius: 'base',
	bg: 'neutral.default',
	color: 'background',
});

const baseStyle = defineStyle(({ colorScheme }) => {
	return {
		bg: `${colorScheme}.weak`,
		color: 'white',
		_hover: {
		   bg: `${colorScheme}.medium`,
		},
		_active: {
		  bg: `${colorScheme}.strong`,
		},
	};
});



export const Badge = defineStyleConfig({
	baseStyle
});
```

And then use Button component like this
```tsx
<Button colorScheme="primary" />
```

It's not possible.

## 🚀 New behavior

Explained above

No, because it's an opt-in flag.

### DIFF
--- a/.changeset/grumpy-badgers-sniff.md
+++ b/.changeset/grumpy-badgers-sniff.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/cli": patch
+---
+
+Add CLI flag `--allow-semantic-token-colors-in-color-schemes` to include
+semantic token colors in color schemes type
--- a/tooling/cli/README.md
+++ b/tooling/cli/README.md
@@ -26,6 +26,7 @@ Options:
   --no-format               Disable auto formatting
   --watch [path]            Watch directory for changes and rebuild
   --template <template>     Choose the template to use for the generation (choices: "default", "augmentation", default: "default"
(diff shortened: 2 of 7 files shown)


segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version, which supports both tokens and semantic tokens.

PR #8118, opened 2023-11-13
fix(Heading): using object syntax for fontsize not working in extendTheme (#7807)

### LINKED ISSUE #7807: Heading sizes & device breakpoints don't work well when used in extendTheme
### Description

When I create a heading, using a size attribute, and set the fontSize with multiple values depending on the device breakpoint, this is not working and ignoring it when using the extendTheme, if the fontSize is defined in the heading directly works fine. 

The Text component works fine

### Link to Reproduction

https://codesandbox.io/s/sxgs9c?file=%2FApp.tsx&utm_medium=sandpack

### Steps to reproduce

1. Go to 'the code sandbox'
2. Resize the window
3. You can see how the first heading and the text fontSize change, but the heading in the middle don't chage 

If you inspect the second Headingsize, you will se that only the mint-width query of 48px is generated, and nothing else. 

### Chakra UI Version

React 2.6.1

### Browser

Firefox

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I'm not sure it's an error, seems related to this: https://github.com/chakra-ui/chakra-ui/issues/3501 But Maybe I'm doing something wrong.

Today I investigated more, if the info is set with a list for the heading works, but with the object doesn't work, it generates the default.

surajpheudin:
Closes #7807

## 📝 Description

> Replace array syntax with object syntax for fontSize in heading theme because mergeWith function was used inside useStyleConfigImpl was not merging the array from default Theme and new object from extended Theme.

> fontSize: ["3xl", null, "4xl"] and fontSize: {base: "2xl",  xl: "6xl"} was not merging properly resulting fontSize: ['3xl', null, '6xl', base: '4xl', xl: '6xl']

## 🚀 New behavior

> fontSize: {base: "2xl", md: "4xl"} and fontSize: {base: "2xl",  xl: "6xl"} was merge properly resulting fontSize: {base: "2xl", "md": "4xl", xl: "6xl"}

### DIFF
--- a/packages/components/layout/stories/heading.stories.tsx
+++ b/packages/components/layout/stories/heading.stories.tsx
@@ -1,0 +1,33 @@
+import * as React from "react"
+import { ChakraProvider, extendTheme } from "../../react/src"
+import { Heading } from "../src"
+
+export default {
+  title: "Components / Typography / Heading",
+}
+
+const theme = extendTheme({
+  components: {
+    Heading: {
+      variants: {
+        customCaps: {
+          textTransform: "uppercase",
+          fontSize: {
+            base: "4xl",
+            lg: "6xl",
+          },
+        },
+      },
+    },
+  },
+})
+
+// see https://github.com/chakra-ui/chakra-ui/issues/2464
+export const withVariant = () => (
+  <ChakraProvider theme={theme}>
+    <Heading variant="customCaps">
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version which merges styles correctly.

PR #8122, opened 2023-11-14
fix: Allow 0 as valid child label of form input components - radio, checkbox, switch

jsonhero:
Closes # 

## 📝 Description

- The conditional render check for `children` is too broad of a falsy check which doesn't allow for the number 0 to be rendered with a label. This disrupts the formatting compared to any other value (or number) passed as the child. 
- This change affects `checkbox`, `radio` and `switch` labels.
- Added new helper function for checking 'nullish' children which are not renderable.

- Sandbox demoing current behavior: 
https://codesandbox.io/s/serene-shamir-z8sx54?file=/src/index.js

## 🚀 New behavior

- Renders label for children with the value of 0 without disrupting none label render for nullish children

No.

### DIFF
--- a/.changeset/angry-lies-study.md
+++ b/.changeset/angry-lies-study.md
@@ -1,0 +1,9 @@
+---
+"@chakra-ui/react-children-utils": patch
+"@chakra-ui/checkbox": patch
+"@chakra-ui/switch": patch
+"@chakra-ui/radio": patch
+---
+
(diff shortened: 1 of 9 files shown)


segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version, which allows for explicit rendering of radio, checkbox and switch components

PR #8133, opened 2023-11-23
fix: useCallbackRef now actually works

benwiley4000:
Closes # 

## 📝 Description

fixes useCallbackRef

useCallbackRef doesn't work

## 🚀 New behavior

it works now

hopefully not

### DIFF
--- a/.changeset/itchy-fireants-jump.md
+++ b/.changeset/itchy-fireants-jump.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react-use-callback-ref": patch
+---
+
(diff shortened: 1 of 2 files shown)


benwiley4000:
One alternative option (to honor the dependencies) would be to keep the use effect but pass the dependency array there instead of into the useCallback. But I see little practical reason to only update the callback when the dependencies change.

segunadebayo:
When you say it doesn't work, what do you mean @benwiley4000? could you provide a reproduction?

segunadebayo:
Thanks for your PR. This will be covered in v3

PR #8165, opened 2023-12-12
fix(menu): fix click events in shadow dom

### LINKED ISSUE #7419: Menu "closeOnSelect" does not work within Shadow DOM
### Description

When I set false on `closeOnSelect`, I expect it to keep opened after click.
It works well in normal scenario, but it does not work within shadow DOM. 

### Link to Reproduction

https://codesandbox.io/s/chakra-menu-list-close-on-select-forked-b2ipng

### Steps to reproduce

1. Add a `Menu` component to Shadow DOM;
2. Set `closeOnSelect` to false;
3. Click on an item;
4. See error: the menu is closed.

### Chakra UI Version

2.5.1

### Browser

Google Chrome 110

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux

### Additional Information

Maybe related with [isHtmlElement](https://github.com/chakra-ui/chakra-ui/blob/bea57af3fc039fce0d1b9ac272cdd65c21fd98e1/packages/components/menu/src/use-menu.ts#L817) on [isTargetMenuItem](https://github.com/chakra-ui/chakra-ui/blob/bea57af3fc039fce0d1b9ac272cdd65c21fd98e1/packages/components/menu/src/use-menu.ts#L387)?

mbeckem:
Closes #7419

## 📝 Description

`<Menu />` current will not work correctly when used in a shadow DOM.
Clicks on menu items will cause the menu to close (even if `closeOnSelect={false}` is used), as will repeated clicks on on the menu opener button, see linked issue.

The current logic inside use-menu (and use-outside-click) relies on `element.contains(event.target)` to detect clicks outside the menu.
This causes problems if the menu is used inside a shadow DOM, since the events in question will no longer refer to the original `target` (e.g. a button) but the entire shadow root (or web component) as a result of the event bubbling up. 
Essentially, too many clicks (even clicks on menu items and on the menu opener button) will be treated as "outside" the menu, causing it to close.

## 🚀 New 

`event.composedPath().includes(element)` is used instead to detect whether an event's original target is inside `element`.

This change should be compatible with all modern browsers (see [canisue](https://caniuse.com/mdn-api_event_composedpath)). I could not find any information about chakra support IE11 or not (i hope it's not required).

## 📝 Additional Information

See also
- https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath
- https://lamplightdev.com/blog/2021/04/10/how-to-detect-clicks-outside-of-a-web-component/

### DIFF
--- a/.changeset/hip-spiders-check.md
+++ b/.changeset/hip-spiders-check.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/react-use-outside-click": patch
+"@chakra-ui/hooks": patch
+---
+
+Fix outside click detection in shadow doms by using `event.composedPath()`
+instead of `event.target`
--- a/.changeset/small-seals-complain.md
+++ b/.changeset/small-seals-complain.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/menu": patch
+---
+
+Fix menu button click detection in shadow doms by using `event.composedPath()`
+instead of `event.target`
--- a/packages/components/menu/package.json
+++ b/packages/components/menu/package.json
@@ -56,6 +56,8 @@
     "@chakra-ui/portal": "workspace:*",
     "@chakra-ui/theme": "workspace:*",
     "@chakra-ui/button": "workspace:*",
+    "@emotion/cache": "^11.11.0",
+    "@emotion/react": "^11.11.1",
     "react-icons": "^4.10.1",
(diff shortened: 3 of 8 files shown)


segunadebayo:
Thanks for working on this @mbeckem.

I've gone ahead to cherry-pick the changes in this PR to the v2 branch.

PR #8173, opened 2023-12-18
feature: support anchorRef on PopoverAnchor

xeinebiu:
## 📝 Description

> Adds a possibility on PopoverAnchor component to pass `anchorRef`. This ensures us we can reuse the same popover with different anchors on different use cases.

image

There are third party libraries whose we cannot wrap inside PopoverAnchor to display a Popover on a click event.

## 🚀 New behavior

PopoverAnchor can either accept `children` or `anchorRef` to position the Popover. `anchorRef` can be an element on a 3rd party library which dispatched a click event and we want to display the popover for that element.

- No

### DIFF
--- a/.changeset/witty-files-rescue.md
+++ b/.changeset/witty-files-rescue.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/popover": major
+---
+
+PopoverAnchor can either accept children or anchorRef to position the Popover.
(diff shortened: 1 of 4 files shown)


JAD3N:
Is there any chance this will get merged soon?

segunadebayo:
This will be supported in v3.

PR #8189, opened 2024-01-04
fix(use-outside-click) change mouse to pointer events

Nightbr:
## 📝 Description

- To make it works with some Graph library (Reactflow) that prevent mouseUp/mouseDown events

- When clicking outside, on some cases (on Reactflow Graph for example) when the mouseUp/mouseDown events are prevented, the menu will not close

## 🚀 New behavior

- In every case, the click outside works and the menu it closed properly

No

## 📝 Additional Information

- https://github.com/xyflow/xyflow/issues/2401
- https://github.com/xyflow/xyflow/issues/2341

### DIFF
--- a/.changeset/rude-bugs-try.md
+++ b/.changeset/rude-bugs-try.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react-use-outside-click": patch
+---
+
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks for this PR @Nightbr.

We've accounted for this in the upcoming v3.

PR #8245, opened 2024-01-31
draft: migrate accordion

Pagebakers:
Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior

### DIFF
--- a/packages/components/src/accordion/use-accordion.ts
+++ b/packages/components/src/accordion/use-accordion.ts
@@ -3,29 +3,26 @@ import { mergeRefs } from "@chakra-ui/hooks/use-merge-refs"
 import { callAllHandlers } from "@chakra-ui/utils/call-all"
 import { createContext } from "@chakra-ui/utils/context"
 import { warn } from "@chakra-ui/utils/warn"
-import { useCallback, useEffect, useId, useRef, useState } from "react"
-import {
-  useAccordionDescendant,
-  useAccordionDescendants,
-} from "./accordion-context"
+import { nextById, prevById, queryAll } from "@zag-js/dom-utils"
+import { useCallback, useId, useRef, useState } from "react"
+
+type AccordionValue<T extends boolean | undefined> = T extends undefined
+  ? string
+  : T extends true
+  ? string[]
+  : string
 
 /* -------------------------------------------------------------------------------------------------
  * useAccordion - The root react hook that manages all accordion items
  * -----------------------------------------------------------------------------------------------*/
 
-/**
- * @deprecated - This will be removed in future versions.
- * Please use `number | number[]` instead.
- */
-export type ExpandedIndex = number | number[]
-
-export interface UseAccordionProps {
+export interface UseAccordionProps<Multiple extends boolean = boolean> {
   /**
    * If `true`, multiple accordion items can be expanded at once.
    *
    * @default false
    */
-  allowMultiple?: boolean
+  allowMultiple?: Multiple
   /**
    * If `true`, any expanded accordion item can be collapsed again.
    *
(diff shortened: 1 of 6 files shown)


packages/components/src/accordion/use-accordion.ts
```diff
@@ -3,47 +3,48 @@ import { mergeRefs } from "@chakra-ui/hooks/use-merge-refs"
...
-
-export interface UseAccordionProps {
+export interface UseAccordionProps<Multiple extends boolean> {
   /**
    * If `true`, multiple accordion items can be expanded at once.
    *
    * @default false
    */
-  allowMultiple?: boolean
+  allowMultiple?: Multiple
```
segunadebayo, line 25: We should keep this as a boolean because we want the value always to return an array (regardless of whether it's  multiple or not)

packages/components/src/accordion/use-accordion.ts
```diff
@@ -52,100 +53,93 @@ export interface UseAccordionProps {
...
+  const reactId = useId()
+  const id = `accordion-${idProp ?? reactId}`
 
   /**
    * This state is used to track the index focused accordion
    * button when click on the button, tab on the button, or
    * use the down/up arrow to navigate.
    */
-  const [focusedIndex, setFocusedIndex] = useState(-1)
+  const [focusedId, setFocusedId] = useState<string | null>(null)
```
segunadebayo, line 81: Do we want to call this `focusedValue`?

packages/components/src/tabs/use-tabs.ts
```diff
@@ -433,33 +404,33 @@ export function useTabIndicator(): React.CSSProperties {
 
   // Update the selected tab rect when the selectedIndex changes
   useSafeLayoutEffect(() => {
-    if (selectedIndex == null) return
+    if (selectedValue == null) return
 
-    const tab = descendants.item(selectedIndex)
+    const tab = document.getElementById(`${makeTabId(id, selectedValue)}`)
```
segunadebayo, line 409: We want to make sure this is scoped to the correct owner document.

Can we do something like `ref?.current.ownerDocument`

Pagebakers: I was using querySelector initially on the root element, but that somehow would throw invalid selector every now and then. Your suggestion would work nicely.

packages/components/src/accordion/use-accordion.ts
Pagebakers: I see Ark uses an array as well, will update it.

packages/components/src/accordion/use-accordion.ts
Pagebakers: Internally the value will actually map to a unique id:

```
  /**
   * Generate unique ids for all accordion item components (button and panel)
   */
  const reactId = useId()
  const uid = value || reactId

  const buttonId = makeId("button", id, uid)
  const panelId = makeId("panel", id, uid)
```

packages/components/src/accordion/use-accordion.ts
```diff
@@ -185,9 +166,21 @@ export interface UseAccordionItemProps {
...
+   */
+  value?: string
+}
+
+function getAllItems(root: HTMLElement | null) {
+  return queryAll(root, "[aria-controls]:not([disabled])")
+}
+
+function makeId(type: string, id: string, value: string) {
+  return `accordion-${type}-${id}-${value}`
```
Pagebakers, line 183: @segunadebayo the id might break when value includes spaces or unsupported characters, is there a normalisation function available?

segunadebayo: How about we let the users handle this? By default react ids are escaped properly. We can also use a string escape library

Pagebakers: Escaping whitespace chars now, all other chars should be pretty much valid.

PR #8315, opened 2024-02-25
docs: component namespacing

Omikorin:


segunadebayo:
Thanks for giving this a shot.

We'll most likely create a whole new doc. Adapting the current one is a bit too much work.

Issue #7600, opened 2023-04-25
Unsafe inline style violate Content Security Policy


yinlam22:
### Description

Unsafe inline style is used in some of the chakra UI components that will violate the Content Security Policy
Example: 
- https://github.com/chakra-ui/chakra-ui/blob/d3fc1fd61dcee50a8916dfee9bd403e645bf6cbb/packages/components/toast/src/toast.provider.tsx#L130
- https://github.com/chakra-ui/chakra-ui/blob/d3fc1fd61dcee50a8916dfee9bd403e645bf6cbb/packages/components/toast/src/toast.utils.ts#:~:text=%3A%20undefined-,return%20%7B,%7D,-%7D
```
  return {
    position: "fixed",
    zIndex: "var(--toast-z-index, 5500)",
    pointerEvents: "none",
    display: "flex",
    flexDirection: "column",
    margin,
    top,
    bottom,
    right,
    left,
  }
```

**Expected behavior**
- A way to add nonce globally to all chakra UI components
- Avoid the use of unsafe inline style

### Link to Reproduction

Not really possible to show this in CodeSandbox

### Steps to reproduce

Add the Content-Security-Policy header on a website with this value:

```
"base-uri; child-src; connect-src 'self' *; default-src; font-src; form-action; frame-ancestors 'self' *; frame-src 'self'; img-src 'self' * data:; manifest-src 'self'; media-src 'self'; object-src 'none'; script-src 'self'; style-src 'self' ;"
```

### Chakra UI Version

2.4.1

### Operating System

- [x] macOS
- [x] Windows
- [ ] Linux

segunadebayo:
Hi @yinlam22,

The styles used in that component are not unsafe, except I misunderstand what you're saying. Adding styles in the `style` attribute is not an issue with React.

As far as I know, this will only be an issue when style elements are injected into the document.

yinlam22:
Hi @segunadebayo 
Sorry, I should have provided more clarification and shared my finding on this. 

What I meant by unsafe inline style on that chakra UI component is it is blocked by the Content Security Policy when the "**unsafe-inline**" is not added to the Content-Security-Policy: style-src 

if the unsafe-inline source is not specified in the CSP style-src, then it will block the inline style attributes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_inline_styles

I reproduced the scenario using codesandbox
[...]

PR #8338, opened 2024-03-04
Allow to set a style-src CSP without unsafe-inline

### LINKED ISSUE #8043: CSP error thrown by color-mode.util
### Description

When I use Chakra in a project with a content security policy, I see the following error:

```
Refused to apply inline style because it violates the following Content Security Policy directive: "style-src http://my-domain.test/ 'nonce-LBVC3WAK89jF6Nh0lWF7IQ=='". Either the 'unsafe-inline' keyword, a hash ('sha256-GNF74DLkXb0fH3ILHgILFjk1ozCF3SNXQ5mQb7WLu/Y='), or a nonce ('nonce-...') is required to enable inline execution.

preventTransition @ color-mode.utils.ts:55
setDataset @ color-mode.utils.ts:17
(anonymous) @ color-mode-provider.tsx:68
(anonymous) @ color-mode-provider.tsx:86
commitHookEffectListMount @ react-dom.development.js:23150
commitPassiveMountOnFiber @ react-dom.development.js:24926
commitPassiveMountEffects_complete @ react-dom.development.js:24891
commitPassiveMountEffects_begin @ react-dom.development.js:24878
commitPassiveMountEffects @ react-dom.development.js:24866
flushPassiveEffectsImpl @ react-dom.development.js:27039
flushPassiveEffects @ react-dom.development.js:26984
performSyncWorkOnRoot @ react-dom.development.js:26076
flushSyncCallbacks @ react-dom.development.js:12042
commitRootImpl @ react-dom.development.js:26959
commitRoot @ react-dom.development.js:26682
finishConcurrentRender @ react-dom.development.js:25981
performConcurrentWorkOnRoot @ react-dom.development.js:25809
workLoop @ scheduler.development.js:266
flushWork @ scheduler.development.js:239
performWorkUntilDeadline @ scheduler.development.js:533
```

This is not resolved by setting a nonce in my emotion cache, or by adding a `ColorModeScript` with a nonce.

### Link to Reproduction

https://github.com/chakra-ui/chakra-ui/compare/main...apexskier:chakra-ui:csp-style-issue

### Steps to reproduce

Run the vite-js example project from my fork, run behind https and add a content security policy matching the one in the vite config example. (I used something similar to `ngrok` and this chrome plugin https://chrome.google.com/webstore/detail/modheader-modify-http-hea/idgpnmonknjnojddfkpgkljpfnnfcklj.

### Chakra UI Version

default branch

### Browser

Google Chrome 116.0.5845.187

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

The error comes from the `<style>` node creation here: https://github.com/chakra-ui/chakra-ui/blob/6d12ea81f4791f4df2953533b570178f59551b74/packages/components/color-mode/src/color-mode.utils.ts#L49-L55

To fix this the same way CSP is supported for the main script injection, a nonce also needs to be passed into this style injection. It would make sense to me to reuse the same nonce, which would be most seamless by using the same from emotion cache provider. (or it could be passed through in some other way)

An example/verification of this fix can be found here https://github.com/apexskier/chakra-ui/compare/csp-style-issue...apexskier:chakra-ui:csp-style-issue-fix (the first commit, or combo of first and second commits both work).

ArthurWD:
Closes #8043, #7600

## 📝 Description

This PR allows to set a style-src CSP without unsafe-inline

Chakra currently uses both the style attribute on HTML elements and an injected style element. Both require unsafe-inline style-src.

## 🚀 New behavior

*  Uses direct style props like `flex={1}` instead of `style={{ flex: 1 }}`.
*  Uses `__css` instead of `style` where an style object is used.
*  Adds the option to pass a `nonce` to the ColorModeProvider where the style element is injected.

## 💣 Is this a breaking change: 
No

## 📝 Additional Information
I was unable to test the behaviour locally, because I had issues linking to the local directory from my package.json. I haven't tested all changes myself, so please take a good look.

Also, there are some places where a style attribute is set on `motion.div`. I was not sure how to handle this, so I haven't fixed those issues.

### DIFF
--- a/packages/components/src/checkbox/checkbox-icon.tsx
+++ b/packages/components/src/checkbox/checkbox-icon.tsx
@@ -5,12 +5,10 @@ function CheckIcon(props: PropsOf<typeof chakra.svg>) {
     <chakra.svg
       width="1.2em"
       viewBox="0 0 12 10"
-      style={{
-        fill: "none",
-        strokeWidth: 2,
-        stroke: "currentColor",
-        strokeDasharray: 16,
-      }}
+      fill="none"
+      stroke="currentColor"
+      strokeDasharray={16}
+      strokeWidth={2}
       {...props}
     >
       <polyline points="1.5 6 4.5 9 10.5 1" />
@@ -23,7 +21,8 @@ function IndeterminateIcon(props: PropsOf<typeof chakra.svg>) {
     <chakra.svg
       width="1.2em"
       viewBox="0 0 24 24"
-      style={{ stroke: "currentColor", strokeWidth: 4 }}
+      stroke="currentColor"
+      strokeWidth={4}
       {...props}
     >
       <line x1="21" x2="3" y1="12" y2="12" />
@@ -54,12 +53,10 @@ export function CheckboxIcon(props: CheckboxIconProps) {
 
   return isChecked || isIndeterminate ? (
     <chakra.div
-      style={{
-        display: "flex",
-        alignItems: "center",
-        justifyContent: "center",
-        height: "100%",
-      }}
+      alignItems="center"
+      display="flex"
+      height="100%"
+      justifyContent="center"
     >
       <BaseIcon {...rest} />
     </chakra.div>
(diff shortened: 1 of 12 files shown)


benmcginnis:
it would be really nice to get this merged, even though I know y'all are working on the v3 migration in earnest. this would be really helpful for those of us who're going to be on v2 for a while

ra-lfaro:
+1 Would also like this support in

segunadebayo:
Closing this in favor of #8523

We'll get that merged and release an update shortly

PR #8347, opened 2024-03-10
refactor: remove unused framer features

mezotv:
## 📝 Description

> Changes the motion import to the more lightweigth m component. 

> Currently uses the motion import which loads all of the features even the ones you don't need.

## 🚀 New behavior

> Uses the lightweight m component with only the animation features loaded into. Since thats the only thing the components use.

> Not a breaking change!

## 📝 Additional Information

Docs for this change: https://www.framer.com/motion/guide-reduce-bundle-size/

### DIFF
--- a/.changeset/slow-dolphins-cross.md
+++ b/.changeset/slow-dolphins-cross.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react": patch
+---
+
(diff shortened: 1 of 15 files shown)


segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version which no longer requires framer-motion

PR #8364, opened 2024-03-15
fix(theme): allow for simplified input styling

tom-engineering:
## 📝 Description
This change allows users to style `Input` components via `InputGroup` without those styles being overridden by the `Input` itself.

The `Input` component can be styled by setting some custom CSS variables:
```
--input-height
--input-font-size
--input-padding
--input-border-radius
```
These CSS variables are also used by `InputGroup` and the `InputLeftAddon/InputRightAddon` and `InputLeftElement/InputRightElement` components to allow them to share size information with the `Input` to create a seamless end result.

Unfortunately, if you want to adjust any of these values in a one-off scenario, you have to apply the new value to the `InputGroup` **and** to the `Input`, because the `Input` has its own copy of the CSS variables which take precedence.

```typescript
<InputGroup sx={{ '--input-height': '50px' }}>
  <Input placeholder="Search for something..." sx={{ '--input-height': '50px' }} />
  <InputRightElement>
    <Search />
  </InputRightElement>
 </InputGroup>
```

## 🚀 New behavior

This PR changes the keys of the CSS variables set on the `Input` itself to this set:

```
--input-field-height
--input-field-font-size
--input-field-padding
--input-field-border-radius
```
It then adjusts the corresponding properties of an `Input` to use the original set with a fallback to the new set:

```css
height: var(--input-height, var(--input-field-height));
font-size: var(--input-font-size, var(--input-field-font-size));
padding-inline-start: var(--input-padding, var(--input-field-padding));
padding-inline-end: var(--input-padding, var(--input-field-padding));
border-radius: var(--input-border-radius, var(--input-field-border-radius));
```
This lets a user set the original variable once on the `InputGroup` and have it inherit throughout the other group components.

No - if a user is already dealing with this issue by setting the original CSS vars in two places, that will continue to work due to the CSS variable fallback arrangement.

## 📝 Additional Information

It's my understanding that the reason this has to be solved with two CSS variables is because Chakra doesn't allow theme tokens in CSS variable fallback values. Having the fallback be another variable which _can_ be set with a theme token means we can avoid hard-coding any CSS values.

### DIFF
--- a/.changeset/afraid-moons-teach.md
+++ b/.changeset/afraid-moons-teach.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/theme": patch
+---
+
+Prevented Input from overwriting InputGroup styling variables
--- a/packages/theme/src/components/input.ts
+++ b/packages/theme/src/components/input.ts
@@ -4,15 +4,23 @@ import {
   cssVar,
   defineStyle,
 } from "@chakra-ui/styled-system"
-import { getColor, mode } from "@chakra-ui/theme-tools"
+import { type CSSVar, getColor, mode } from "@chakra-ui/theme-tools"
 
 const { definePartsStyle, defineMultiStyleConfig } =
   createMultiStyleConfigHelpers(parts.keys)
 
-const $height = cssVar("input-height")
-const $fontSize = cssVar("input-font-size")
-const $padding = cssVar("input-padding")
-const $borderRadius = cssVar("input-border-radius")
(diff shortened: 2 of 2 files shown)


tom-engineering:
Would this work better against the `v2` branch?

segunadebayo:
Thanks for working on this.

Closing this in favor of the upcoming v3 version, which no longer requires `InputGroup`

Issue #7965, opened 2023-08-24
isLoading & accessibility problem b/c of losing focus


Philzen:
### Description

When i interact with an element (i.e. a button) i expect the focus to stay on, even when it switches to `isLoading`, but instead it is lost – which makes user's relying on accessibility features miss potentially important information about the status of the button.

Background:
We have a device on/off-button in our application implemented as an `IconButton`. Now, when the cam is started, we're showing the loading state while it is being activated, which sometimes can take a few seconds and update the title & aria-label, b/c so user's know the next click will stop the cam.

An accessibility review was conducted and flagged this problem: as soon as the button switches to loading state, it will make screen reader describe the page instead b/c the focus is immediately lost.

### Link to Reproduction

https://codesandbox.io/s/intelligent-jang-wu06rc

### Chakra UI Version

1.8.8

### Browser

Firefox 116.0.2 & Chomium 115.0.5790.170

### Operating System

- [ ] macOS
- [ ] Windows
- [X] Linux

Philzen:
The culprit here of course is the fact that `isLoading` seems also *disable* the element. The only possible workaround i could come up with so far is to explicitly set `disabled={false}`.

Tragically, the above mentioned workaround seems not to be possible in the latest release due to #7269 … and somehow feels wrong to me anyway – imho i'd prefer an element not to be disabled unless i explicitly set `isDisabled`.

As Chakra UI otherwise plays so nice with A11Y out-of-the-box (and also b/c of my astonishment learning about this problem when i tried using screen reader myself to debug this behaviour), i wanted to quickly raise this here and am looking forward to any perspectives on this.

Philzen:
@segunadebayo @tioluwani94 would it be possible get a 2.8.3 release anytime soon? With the last 2.x release now almost 9 months old, there are a couple of fixes that are desperately needed in the upstream for a project that i'm working on, specifically the fix that merged via #8468, which re-enabled the only known workaround for this issue.

Kindly advise if i can do anything to support.

segunadebayo:
Working on a release atm

segunadebayo:
v2.9.1 is now released

Philzen:
> v2.9.1 is now released

[...]

Issue #8104, opened 2023-11-03
Exposing the useAlertContext - custom component theming issues


CaitlinSweeney:
### Description

Hello,
I'm upgrading from `Chakra-UI@v1 `to `Chakra-UI@v2` (finally). I have my custom components in a npm package so I can re-use the theme and custom components in 2 applications. I am looking to customize some components a bit easier.

### Problem Statement/Justification

I'm having a lot of issues getting the custom theme files to access the theme provider. It works as expected here: https://codesandbox.io/s/chakra-ui-v2-with-custom-themes-76r2xy

 I understand you can override the `AlertIcon` by passing a child component. Would it be possible to export the `useAlertContext` so the `status` is exposed? Makes it easier to create a reusable component for overriding the AlertIcon. Right now the `alertStyles` is exported.

### Proposed Solution or API

Option for some customization when implementing:  
```
const STATUSES = {
  default: { icon: ErrorIcon, colorScheme: 'gray' },
  info: { icon: ErrorIcon, colorScheme: 'blue' },
  warning: { icon: ErrorIcon, colorScheme: 'orange' },
  success: { icon: CheckIcon, colorScheme: 'green' },
  error: { icon: ErrorIcon, colorScheme: 'red' },
}
export type AlertStatus = keyof typeof STATUSES

export function getStatusIcon(status: AlertStatus) {
  return STATUSES[status].icon
}
 export const AlertIcon = forwardRef<AlertIconProps, 'span'>((props, ref) => {
  const { status } = useAlertContext()
  const CustomIcon= getStatusIcon(status)
  
  return (
    <ChakraAlertIcon
      {...props}
      {...ref}
    >
      {props.children ||  <CustomIcon h="100%"w="100%" /> )}
    </ChakraAlertIcon>
  )
})
``` 

### Alternatives

Alternatively, I can just use this.
```
export const AlertGroup = forwardRef<AlertGroupProps, 'div'>((props, ref) => {
  const {
    children,
    description,
    descriptionProps,
    iconProps,
    status = 'error',
    title,
    titleProps,
    orientation,
    variant,
    ...sx
  } = props

  const container = {
    status,
    orientation,
    ref,
    variant,
    ...sx,
  }

  const getIcon = (() => {
    switch (status) {
    case 'error': return <ErrorIcon color="red.500" />
    case 'info': return <ErrorIcon color="blue.500"  />
    case 'success': return <CheckIcon color="green.500" />
[...]

PR #8462, opened 2024-04-24
fix: Undo Breaking Change by re-allowing disabled prop on button

### LINKED ISSUE #7269: disabled prop does not work for Button and Radio despite the type allowing it

### LINKED ISSUE #7816: Disabled html attribute is not set on Button if passed in as prop to MenuItem

### LINKED ISSUE #7905: disabled attribute doesn't work on <Button>

Philzen:
## 📝 Description

Closes #7269
Closes #7816
Closes #7905

Also re-enables the only currently known workaround for the accessibility problem described in  #7965. 

Removing the ability to use the disabled-prop posed a breaking change, which is hereby fixed.

Despite the `disabled` prop being exposed in the type system, any value given will be discarded and overwritten by `isDisabled || isLoading`.

## 🚀 New behavior

Values given to `disabled`-prop are now respected again. 

One use case (which was possible in 1.x up until 2.0.15) is the combination of `isLoading disabled={false}` to ensure that a button does not lose focus while in loading-state. 

This reverts a breaking change.

## 📝 Additional Information

This closes #7907, effectively replacing it.

### DIFF
--- a/.changeset/nine-bags-rest.md
+++ b/.changeset/nine-bags-rest.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react": patch
+---
+
+Revert breaking change by re-allowing disabled prop on button
--- a/packages/components/src/button/button.tsx
+++ b/packages/components/src/button/button.tsx
@@ -76,6 +76,7 @@ export const Button = forwardRef<ButtonProps, "button">((props, ref) => {
 
   return (
     <chakra.button
+      disabled={isDisabled || isLoading}
       ref={useMergeRefs(ref, _ref)}
       as={as}
       type={type ?? defaultType}
@@ -84,7 +85,6 @@ export const Button = forwardRef<ButtonProps, "button">((props, ref) => {
       __css={buttonStyles}
       className={cx("chakra-button", className)}
       {...rest}
-      disabled={isDisabled || isLoading}
     >
       {isLoading && spinnerPlacement === "start" && (
         <ButtonSpinner


Philzen:
@segunadebayo Kindly have a look. This reverts a breaking change in v2 that affected quite a lot of users, see linked issues for reference.

Speaking for our team: we have an application that we'd love to update to v2 to bring it in line with React 18, however we cannot upgrade as it would break accessibility of toggle buttons losing their focus when `isLoading={true}` and using the `disabled` prop is the only current workaround we know of. 

On a side note, I haven't reviewed the state machine config for buttons in zagjs, so i'm just referencing #7724 and #7965 for your attention, which may get solved in v3. Really thrilled about zag+ark btw :rocket: 

Thanks in advance.

Philzen:
@segunadebayo Many thanks for merging! Can you say when 2.8.3 is going to be published so this (along with other improvements such as #8104) will become available?

PR #8494, opened 2024-05-21
feature: export CreateIconOptions interface

matzar:
## 📝 Description

When using `createIcon`, I found myself needing the `CreateIconOptions` interface.

In this scenario, I am defining and passing my own objects to the `createIcon` method, which should conform to the `CreateIconOptions` type. However, I had to copy and paste it from `@chakra-ui/icons`. Given that this is a real-world scenario, I believe the `CreateIconOptions` interface should be exported from `@chakra-ui/icons`.

`CreateIconOptions` interface is not being exported from `@chakra-ui/icons`.

## 🚀 New behavior

exports `CreateIconOptions` interface from `@chakra-ui/icons`.

No

### DIFF
--- a/packages/react/src/components/icon/create-icon.tsx
+++ b/packages/react/src/components/icon/create-icon.tsx
@@ -3,7 +3,7 @@
 import { Children, forwardRef } from "react"
 import { Icon, type IconProps } from "./icon"
 
-interface CreateIconOptions {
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this PR @matzar. I've added it to the v2 branch.

Will do a release shortly

PR #8575, opened 2024-06-14
[v3] chore(theme): align theme recipe name with panda

isBatak:
Align theme recipe name with panda


packages/react/src/components/skip-nav/skip-nav-link.tsx
```diff
@@ -40,7 +40,7 @@ const baseStyle = defineStyle({
  */
 export const SkipNavLink = forwardRef<HTMLAnchorElement, SkipNavLinkProps>(
   function SkipNavLink(props, ref) {
-    const recipe = useRecipe("SkipLink", props.recipe)
+    const recipe = useRecipe("skipLink", props.recipe)
```
segunadebayo, line 43: This should be skipNavLink. It might have been an error on my side earlier.

isBatak: Here is the change https://github.com/chakra-ui/chakra-ui/pull/8575/commits/4282fe968e20548d4f2b9f7c290023d3ce011af3
I hope I didn't change to many things :D

PR #8590, opened 2024-06-18
fix: optimize image loading in Image component

stepanve:
## 📝 Description
- fix to load crossOrigin before src.

- In Safari browser, crossOrigin may be evaluated before src, causing a CORS error for the image tag.

## 🚀 New behavior
- The CORS error in Safari browser is resolved.

- No

## 📝 Additional Information
- References: https://github.com/vuejs/core/issues/4680


segunadebayo:
Thanks for this. We've addressed it in v3

PR #8675, opened 2024-07-25
fix: popover focus when using framer-motion >=11.2

### LINKED ISSUE #8555: Framer-motion 11.2.0 breaks Popover `initialFocusRef` functionality
### Description

When updating `framer-motion` from 11.1.9 to 11.2.0, ChakraUI's Popover component doesn't automatically focus on the element referenced by the `initialFocusRef` prop (the expected behaviour is documented [here](https://v2.chakra-ui.com/docs/components/popover/usage#focus-an-element-when-popover-opens))

To be clear, this works with `framer-motion` version 11.1.9, but **not** with version 11.2.0. It's possibly this is a bug, with `framer-motion`, but I'm raising the issue here as it's possibly a compatibility issue with Chakra's use of the library.

Thanks 🙏  

### Link to Reproduction

https://codesandbox.io/p/sandbox/sweet-lucy-sz3k82?file=/package.json:11,29

### Steps to reproduce

1. Setup the Popover component with an Input within the PopoverBody.
2. Ensure that the input ref is passed into the `initialFocusRef` prop for the Popover
3. Click on the trigger to open the Popover content
4. See that the Input is not automatically focused.

### Chakra UI Version

2.8.2

### Browser

Chrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

zanuarmirza:
Closes #8555

## 📝 Description
Fix popover focus when using framer-motion >=11.2.
Set visibility value to `visible` immediately

framer-motion >=11.2.0 has an update about [Binary visibility interpolation](https://github.com/framer/motion/commit/a8b752f6f422a39b66654800715d1df96cfca138). It broke the popover focus action. This happens because the visibility value at the first frame still holds `hidden` value and that makes the focus action not work properly.

## 🚀 New behavior

by adding visibility transition duration to zero, It makes the visibility value change before focus action is triggered.

No


zanuarmirza:
Framer motion 11.3.25 has fixed this issue.

PR #8676, opened 2024-07-26
[v3] docs(migration): css prop nuances

isBatak:
Add `css` prop nuances to the migration guide.


segunadebayo:
Thanks for this PR @isBatak.

I've extracted this into the main docs itself.

PR #8703, opened 2024-08-03
feat(footer): add hover effect on social icons

Jagroop2001:
Closes # 

## 📝 Description
Added a `hoverColor` property to each link in the links array and updated the `FooterLink` component to change the icon color on hover using Chakra UI’s _hover pseudo-prop.

Before this implementation, the icons in the footer did not have any hover effect, resulting in a static appearance without visual feedback when a user interacted with them.

## 🚀 New behavior
Currently, the icons in the footer now change color when hovered over, providing a visual indication of interactivity.

### DIFF
--- a/website/src/components/footer.tsx
+++ b/website/src/components/footer.tsx
@@ -18,11 +18,17 @@ type FooterLinkProps = {
   icon?: React.ElementType
   href?: string
   label?: string
+  hoverColor?: string
 }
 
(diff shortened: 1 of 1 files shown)


Jagroop2001:
https://github.com/chakra-ui/chakra-ui/issues/8704

segunadebayo:
Thanks for this PR. We’re no longer accepting PR to the legacy docs.

PR #8745, opened 2024-08-18
fix(docs): update field usage docs and migration notes

7alip:
## 📝 Description

Updates field component usage docs.

Components are not visible and "ReferenceError: Label is not defined" is displayed on the field page

## 🚀 New behavior

Removes error message

No

## 📝 Additional Information

The NumberInput example still causes the page to crash.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -530,15 +530,15 @@ Form control has now been renamed to `Field` to better reflect its purpose as an
 element that represents a form field.
 
 ```tsx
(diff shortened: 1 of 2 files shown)


segunadebayo:
@7alip Thanks for your PRs.

The actual website for the next chakra is here:
https://github.com/chakra-ui/chakra-ui/tree/main/apps/www/content

I will replace the next.chakra-ui.com once we've made good progress.

Issue #8779, opened 2024-08-29
(v2) Mark as `this: void` for callback values which is used by user. (@typescript-eslint/unbound-method)


LumaKernel:
### Description

Some parameters user can use are callbacks. But these callbacks are not marked as `this: void`. So we can't destruct these values from object safely in type-level confidence even if it's safe in runtime.

### Problem Statement/Justification

Like following code, some parameters user can use are callbacks. But these callbacks are not marked as `this: void`.
If so, this bound is not guaranteed to be unrelated by type-level.

```ts
import { useToast } from '@chakra-ui/react';

useToast({
  render: ({ onClose }) => {
    // onClose();
  }
});
```

We cannot do this. Instead, we should do as following to keep this binding:

```ts
import { useToast } from '@chakra-ui/react';

useToast({
  render: (renderProps) => {
    // renderProps.onClose();
  }
});
```

This is because, `onClose` is defined as follows: ( `toast.types-24f022fd.d.ts` )

```ts
interface RenderProps extends UseToastOptions {
    /**
     * Function to close the toast
     */
    onClose(): void;
}
```

This might be fixed as following:

```ts
interface RenderProps extends UseToastOptions {
    /**
     * Function to close the toast
     */
    onClose(this: void): void;
    // or,
    // onClose: () => void;
}
```

Actually, this is restricted by the typescript-eslint rule [@typescript-eslint/unbound-method](https://typescript-eslint.io/rules/unbound-method/).

You can see quick example: https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=JYOwLgpgTgZghgYwgAgILIN4ChnJge3wAoBKALmQGcwpQBzAbh2QHoW9CiwALYSigG75gAE3JUa9Bq3YB5ANLM2HfBVLIAvAD4JtEIxnIFWAL5YREBABs4UFAnwhqyOBVRMHTsJhXITmlyYCYhIGIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJgIwHsOATWgFtE%2BABbc%2B6KImjRu0SODABfEMqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Proposed Solution or API

- Add `this: void` for these callbacks.
[...]

PR #8780, opened 2024-08-29
fix(toast.types): ToastOptions callbacks are actually this-bound free, so mark as this:void

LumaKernel:
Closes # (partial resolving)
Partially resolving https://github.com/chakra-ui/chakra-ui/issues/8779
This PR is not fully resolving all problems, but it's good to describe the idea what should we do to solve the issue.

## 📝 Description

As we can see `packages/components/src/toast/toast.tsx`, the render parameter object is not class, and these values are destructed right before to pass. That means, the props' callback are this-bound free.

```ts
  const renderToast: React.FC<RenderProps> = (props) => {
    if (typeof render === "function") {
      return render({ ...props, ...options }) as JSX.Element
    }
    return <ToastComponent {...props} {...options} />
  }
```

## 🚀 New behavior

No runtime behaviour change.
Now, ToastOptions can be destructed this-bound freely.

No

### DIFF
--- a/packages/components/src/toast/toast.types.ts
+++ b/packages/components/src/toast/toast.types.ts
@@ -35,7 +35,7 @@ export interface ToastOptions {
   /**
    * Function that removes the toast from manager's state.
    */
-  onRequestRemove(): void
+  onRequestRemove(this: void): void
 
   /**
    * The position of the toast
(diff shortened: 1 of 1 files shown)


segunadebayo:
I've updated the code to use the arrow-function version instead.

Will release an update shortly

PR #9005, opened 2024-10-27
feat: expose utils

Pagebakers:
Since the @chakra-ui/utils package is deprecated it's no longer possible to use utilities like `cx` in v3.

This exposes the utils at `@chakra-ui/react/utils` so design system builders can use the utils to build on top of chakra.


segunadebayo:
Like we talked about in Slack. We don't want to expose the utils for now.

Please copy over the code you need.

PR #9047, opened 2024-10-31
Fix/avatar root as prop

### LINKED ISSUE #9046: Avatar.Root does not respect `as` prop.
### Description

When I set `<Avatar.Root as="span">` I expected the DOM element to be `span`, but instead it's `div`.

### Link to Reproduction

https://stackblitz.com/edit/vitejs-vite-ntrqra?file=src%2Fcomponents%2Fui%2Favatar.tsx

### Steps to reproduce

1. Go to `avatar.tsx`
2. See that there's `as="span"`
3. Inspect the avatar element in the preview
4. See that it's still `div` and not `span`

### Chakra UI Version

3.0.1

### Browser

Google Chrome 130

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

codewithsupra:
Closes #9046

## 📝 Description

This pull request addresses Issue #9046 by adding support for the `as` prop in the `Avatar.Root` component, enabling users to specify the HTML element for the root (e.g., `span`, `section`, etc.). Previously, the `Avatar.Root` component defaulted to rendering as a `<div>` even when the `as` prop was set.

Currently, setting `<Avatar.Root as="span">` or any other element type does not affect the rendered HTML. The `Avatar` component always defaults to a `<div>` for the root element, regardless of the `as` prop value.

## 🚀 New behavior

With this change:
- The `Avatar.Root` component now respects the `as` prop, rendering the specified HTML element (e.g., `<span>`).
- This enhancement provides more flexibility in markup and alignment for various use cases, particularly when specific elements are required for styling or accessibility.

No, this change is backward-compatible.

## 📝 Additional Information

- Verified the update using Storybook to confirm that the `as` prop renders the specified element.
- Link to reproduction provided in the original issue for verification: [Reproduction on Stackblitz](https://stackblitz.com/edit/vitejs-vite-ntrqra?file=src%2Fcomponents%2Fui%2Favatar.tsx)

---

By supporting the `as` prop, Chakra UI's `Avatar` component can now better adapt to various semantic and layout needs.

### DIFF
--- a/.changeset/fast-snails-travel.md
+++ b/.changeset/fast-snails-travel.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react": major
+---
+
+Adds support for the 'as' prop in the Avatar component to allow root element
+customization.
--- a/apps/compositions/src/ui/avatar.tsx
+++ b/apps/compositions/src/ui/avatar.tsx
@@ -7,6 +7,7 @@ import { forwardRef } from "react"
 type ImageProps = React.ImgHTMLAttributes<HTMLImageElement>
 
 export interface AvatarProps extends ChakraAvatar.RootProps {
+  as?: React.ElementType // Add `as` prop type here
   name?: string
   src?: string
   srcSet?: string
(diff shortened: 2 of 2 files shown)


segunadebayo:
Appreciate the PR. The proposed fix doesn't address the underlying issue.

I should have a fix for this shortly.

PR #9114, opened 2024-11-06
omit valid element for button props

kristian240:
> No issue for this one

## 📝 Description

In the latest Next.js `15.0.2` when a react element prop is passed to `useStyleConfig` the `mergeWith` function on https://github.com/chakra-ui/chakra-ui/blob/v2/packages/components/src/system/use-style-config.ts#L33 enters an infinite loop and VDOM is never finished therefore the response is never returned. As a result, your request will never finish.

This was not the case on Next.js `v14`.

With the latest Next.js `15.0.2` the Button from v2 **does not work** when a react element is passed as a prop (e.g. for `leftIcon`).

## 🚀 New behavior

With the latest Next.js `15.0.2` the Button from v2 **does work** when a react element is passed as a prop (e.g. for `leftIcon`).

No

## 📝 Additional Information

Let me know if this should be applied to any other component.

cc: @isbatak

### DIFF
--- a/packages/components/src/button/button.tsx
+++ b/packages/components/src/button/button.tsx
@@ -4,8 +4,8 @@ import {
   SystemStyleObject,
   ThemingProps,
 } from "@chakra-ui/styled-system"
-import { cx, dataAttr } from "@chakra-ui/utils"
-import { useMemo } from "react"
+import { cx, dataAttr, objectFilter } from "@chakra-ui/utils"
+import { isValidElement, useMemo } from "react"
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this PR @kristian240 

This was fixed 3 weeks ago here: https://github.com/chakra-ui/chakra-ui/commit/171b124c68c7562bbd7fb7ff427951c4322fe500

I assume you're still on 2.8.x that's why it's happening. I will release a new patch for v2 that should resolve it all.

PR #9161, opened 2024-11-13
Update popover snippet import

aboveyunhai:
`React.RefObject<HTMLElement>` is assuming we have `React` UMD global like if we are in `Next.js` environment.
Just simply add `import React` so the code snippet can work everywhere regardless frameworks.

## 📝 Description

> Add a brief description

## 🚀 New behavior


segunadebayo:
Thanks for this PR, Yunhai.

Pushed an update for this across all components.

PR #9207, opened 2024-11-19
docs: fix disabled button state in Toast Pause and Play example

joekrill:
I noticed that the [Pause and Play](https://www.chakra-ui.com/docs/components/toast#pause-and-play) example for the Toast component didn't seem to work. 

It turns out that the `disabled` properties were inverted causing the buttons to be disabled when they should have been enabled, and vice versa.

Also: 
- Updates the example to disable the _Show Toast_ button while the toast is visible, and re-enable it once the toast is dismissed.
- Adds clarification to the documentation on what pause/resume actually does, because I found that a little unclear when I was reading this.


segunadebayo:
Thank you @joekrill

PR #9333, opened 2024-12-15
[fix] Tooltip: Avoid React 19 deprecation warning when accessing `element.ref`

### LINKED ISSUE #9329: [v2] Ensure components don't access `ref` for React 19 compatibility
### Description

In react 19 any components ~~could be conditionally exported without being wrapped in `forwardRef` when React v19 is used.~~ should refrain from `element.ref` access.

In React 19, any components ~~using `forwardRef`~~ [accessing `element.ref` will trigger a warning](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-element-ref):

> Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.

[This is because `ref` has become a regular prop in React 19](https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop). 

Side note: This is actually nice b/c it removes the need to wrap any components in `forwardRef` (something noteworthy to add to the docs).

<details><summary>forwardRef migration for libs that want it</summary>
<p>

```tsx
import { version } from 'react'

function _Tooltip (props) {
  // …
  {
    // …
    ref,
    ...rest
  } = ownProps
  // …
}

export const Tooltip = version.startsWith('19.') 
  ? _Tooltip 
  : forwardRef<TooltipProps, "div">((props, ref) => _Tooltip({ ...props, ref }))
```

</p>
</details> 

### Proposed Solution or API

A commonly used component that triggers the warning is the popular chakra `<Tooltip />`:

https://github.com/chakra-ui/chakra-ui/blob/ebf39c93d1a88982d14230d69dd9404f18b4f8b9/packages/components/src/tooltip/tooltip.tsx#L125-L128

The offending bit is `child.ref` which used to be a normal field, but in React 19 was changed to a getter that will still return the value, but trigger the warning:

```
Object.getOwnPropertyDescriptor(temp0, "ref").get // function elementRefGetterWithDeprecationWarning()
```

Not relying on this getter being there forever, nor it's name, the most stable future-proof solution i could come up with is this one-line change

```diff
      tooltip.getTriggerProps(
        child.props,
-       child.ref,
+       Object.getOwnPropertyDescriptor(child, 'ref').value
      )
```

`Object.getOwnPropertyDescriptor(child, 'ref').value` will return the `child.ref` in react versions before 19, and `undefined` starting with react 19 – which is what we want b/c there it is now available as `child.props.ref`.

### Alternatives

Above proposed solution basically works like a charm in initial tests. However i'm not aware if there may be any drawbacks here ~~(i.e. will treeshaking still work the same?)~~.

Alternatives i could think of that all work without evaluating `react.version` or the name of the getter function in the code itself:
1. Provide a separate package  
If there is some kind of npm mechanism to resolve to different package versions depending on what react peerDependency version is detected, that would be awesome to automatically provide the correct one. Otherwise, not a big deal to have developers manually change package.json to use the react19-compatible chakra package when they migrating.  
Of course there is the drawback of maintaining this.
2. Provide react19-compatible component versions as separate imports  
~~So alongside of the existing Tooltip export we could introduce a second one not wrapped in `forwardRef` called `Tooltip19` for instance. Would definitely work very clean with treeshaking.~~
3. ~~Provide a wrapper component that has a prop for this  
So if that prop (i.e. `noForwardRef`) on the wrapper is set, the child will be rendered accordingly. Again, not sure how well that works with treeshaking.~~  
Evaluate the react version in code  
    ```tsx  
    import { version } from 'react'
  
    //…
    
          tooltip.getTriggerProps(
            child.props,
            version.startsWith('19.') ? undefined : child.ref
          )
    ```

Philzen:
Closes #9329

Closes # 

## 📝 Description

Accessing `element.ref` [will trigger a deprecation warning in React 19](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-element-ref) which this PR will avoid.

Using the Tooltip component with React 19 will give:

> Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.

… decorated with a long an ugly stack trace in the console.

## 🚀 New behavior

This change ensures `element.ref` is still picked from the child when using React <19, but picked from the child props when using React >=19.

No

### DIFF
--- a/packages/components/src/tooltip/tooltip.tsx
+++ b/packages/components/src/tooltip/tooltip.tsx
@@ -124,7 +124,10 @@ export const Tooltip = forwardRef<TooltipProps, "div">((props, ref) => {
     }
     trigger = cloneElement(
       child,
-      tooltip.getTriggerProps(child.props, child.ref),
+      tooltip.getTriggerProps(
+        child.props, 
+        Object.getOwnPropertyDescriptor(child, 'ref').value
+      ),
     )
   }
 


n1xx1:
This definitely feels like the wrong fix, and just bypassing the deprecation warning. Shouldn't it be something like this instead?

```
tooltip.getTriggerProps(child.props, child.props.ref ?? child.ref)
```

jcpendesign (approved):

yusuke-fukuhara:
when will this PR be released?

segunadebayo:
Thanks for this PR. I've extended it to other components as well.

Will publish an new patch shortly.

Philzen:
Thanks, looking forward to the patch release.

@segunadebayo i'm wondering, what is the advantage in 1efcb51c09353da877884b47d00e2d980e3247f0 over https://github.com/chakra-ui/chakra-ui/pull/9333/commits/f6897203192c44207c64a529865f379200f871c3 ?

Issue #9440, opened 2025-01-07
`typegen` does not work with Yarn PnP


joekrill:
### Description

I can't use the `typegen` command because I'm using Yarn PnP. Yarn PnP does not use the traditional `node_modules` folder, so when the Chakra CLI tool attempt to generate types, I get the following error:

```
Error: EROFS: read-only filesystem, open '/node_modules/@chakra-ui/react/dist/types/styled-system/generated/conditions.gen.d.ts'
```

(actually 4 of these errors for each file that is generated).

The typegen tool really shouldn't be writing to the `node_modules` folder anyway, or assuming it's writable. Would it be possible to specify a custom output folder for these types? Would that even work with how the library reads these types? 

This is causing other type errors in my codebase. In particular if I try to use `useSlotRecipe` for a custom component it returns a `never` type because - as far as it is concerned - the slot doesn't even exist.

### Link to Reproduction

n/a

### Chakra UI Version

3.2.5

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

segunadebayo:
Good point. We should consider using DTS module augmentation instead. That should work everywhere.

segunadebayo:
After looking into this over the weekend, I don't see an easy way around this and would prefer not to support Yarn PnP instead of overcomplicating the codebase.

We've improved the types for `useSlotRecipe` and `useRecipe` to avoid the returned `never` type.

Sorry for the inconvenience. Please consider setting up nodeLinker for Yarn PnP.
https://yarnpkg.com/configuration/yarnrc#nodeLinker

joekrill:
> After looking into this over the weekend, I don't see an easy way around this and would prefer not to support Yarn PnP instead of overcomplicating the codebase.

That's unfortunate. It's not about specifically supporting Yarn PnP. It's about not making assumptions about the specific package manager - and really the development environment - being used in general. Even NPM, pnpm, Bun, and Yarn (even with the node linker) all vary slightly in how they handle `node_modules`.  

I think it's a really weird and unexpected behavior to be modifying files inside `node_modules`. I can't imagine most people expect this sort of thing to be happening.

joekrill:
@segunadebayo Would you still consider the ability to specify a custom output directory for generated types? That would allow a decent workaround. One could then use Yarn's [Package patching](https://yarnpkg.com/cli/patch) feature to have correct type information, with a few extra steps:

1. `yarn patch @chakra-ui/react`
[...]

PR #9502, opened 2025-01-20
feat(cli): allow specifying outdir when running typegen command

joekrill:
Adds a workaround for #9440

## 📝 Description

Adds the ability to specify an optional `outdir` when running the `typegen` cli command.

Currently the output directory is hard-coded. 

## 🚀 New behavior

The default value is still the same, but an optional `--outdir` value can be specified to override it and generate the  type files in a different location.

No, the default remains the same when no `outdir` is specified.

## 📝 Additional Information

As mentioned in [#9440](https://github.com/chakra-ui/chakra-ui/issues/9440#issuecomment-2590764193), this will provide a usable workaround for anyone using Yarn PnP to be able to run the `typegen` command and take advantage of the generated types. May also be useful for debugging or other purposes.

@segunadebayo I mentioned putting up a PR for this as a workaround in #9440, but since that issue is closed I wasn't sure if it would get noticed. Since it's a pretty simple change I just decided to throw it up and see how it goes over. An alternative to the CLI parameter could be an environment variable, but this way seemed to be more consistent with the other commands.

### DIFF
--- a/packages/cli/src/commands/typegen.ts
+++ b/packages/cli/src/commands/typegen.ts
@@ -3,7 +3,8 @@ import * as p from "@clack/prompts"
 import { Command } from "commander"
 import createDebug from "debug"
 import { writeFileSync } from "fs"
-import { resolve } from "path"
+import { createRequire } from "node:module"
+import { join, resolve } from "node:path"
 import { generateCondition } from "../utils/generate-conditions"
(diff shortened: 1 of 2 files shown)


Oladunjoye:
thanks for putting this up. hugely impacting many folks!

PR #9511, opened 2025-01-22
fix(cli): properly check for existing files when not run with --force

tryone144:
## 📝 Description

Fix `chakra snippet add` overwriting existing files even without `--force` specified.

`chakra snippet add` overwrites existing files even without specifying `--force`.

## 🚀 New behavior

`chakra snippet add` checks if the file for the snippet already exists and only overwrites with `--force`.

No, but users might have relied on the automatic overwriting.

## 📝 Additional Information

The `chakra snippet add` command erroneously relied on the snippet `id` to detect whether it was already existing. This does not work, because the `id` comes without the file extensions and therefore `existsSync(id)` will always fail.

This is fixed by extracting the filename from the registry instead and falling back on the id, if no entry was found. The latter might be better handled by proper error messages and early returns but no such facility exists at the moment.


tryone144:
I am happy to receive suggestions on wording for the changeset entry.

PR #9593, opened 2025-02-02
fix(v2): styles removed when Nextjs (app) rehydrates

### LINKED ISSUE #9579: Styles removed when Nextjs (app) rehydrates
### Description

When Nextjs have multiple nested Suspense boundaries it results with server side rendered html losing Chakra style for a brief moment (when rehydrating?). It's expected that resolving suspended subtrees doesn't affect style and prerendered html remains styled.

This issue starts happening in the 2.9.1 version. After diving deeper into the changes it seems that [commit](https://github.com/chakra-ui/chakra-ui/commit/44d1469a82e7c0e9607a4bc6db6d05ad16e7dc2a) has introduced the issue by wrapping `ColorProvider` with `withEmotionCache` ([code](https://github.com/chakra-ui/chakra-ui/commit/44d1469a82e7c0e9607a4bc6db6d05ad16e7dc2a#diff-523ef608bc6420238d0ab43d908e2b63f9a6cf1032a5d1d643ef1ba1fb28dd39R40)). Locally, I removed the emotion cache wrapper in `node_modules` (in ColorProvider) and everything seems to be working fine.

Chakra used to provide a solution for a similar problem with `CacheProvider` in Nextjs apps, but this solution doesn't work anymore. I've tried using ` useEmotionCache` from `chakra/next-js` and wrapping the application with Emotion cache provider, this does solve the issue but it also duplicates the injected styles and results with significantly larger document.

### Link to Reproduction

https://stackblitz.com/edit/nextjs-rjpcesea?file=package.json

### Steps to reproduce

1. Go to `/test/route` page
2. You can notice that splash screen is "white" and text unstyled for a brief moment
3. Refresh if needed

### Chakra UI Version

2.9.1

Emotion version `11+`
Nextjs version `14+`

### Browser

_No response_

### Operating System

- [x] macOS
- [x] Windows
- [x] Linux

### Additional Information

Note that `withEmotionCache` in Emotion has a different rendering logic depending on env, see [withEmotionCache](https://github.com/emotion-js/emotion/blob/main/packages/react/src/context.tsx#L49).

Some questions:
-  Why is `ColorProvider` wrapped with `withEmotionCache`? It seems like it's a very specific component that adds a general cache to the application. It might be better to do the composition somewhere else and additional escape hatch for the consumer to do a different integration/composition.
- Any ideas why everything works when `withEmotionCache` is removed from `ColorProvider`?

cc @isBatak @oyrad

isBatak:
Closes #9579

## 📝 Description
 
This PR updates the implementation to use `__unsafe_useEmotionCache` instead of `withEmotionCache`, as Chakra UI injects CSS into the document head for certain transitions, requiring a nonce.  

- Chakra UI handles CSS injection in the client for specific transitions, requiring a nonce. [Reference](https://github.com/chakra-ui/chakra-ui/blob/1899431ca9ed583cd6cdc99ababca1ada43a9870/packages/components/src/color-mode/color-mode.utils.ts#L58)
- Since this behavior occurs only on the client, we can safely use the `__unsafe_useEmotionCache` hook. [Reference](https://github.com/emotion-js/emotion/blob/cce67ec6b2fc94261028b4f4778aae8c3d6c5fd6/packages/react/src/context.tsx#L26)
- `__unsafe_useEmotionCache` is labeled "unsafe" because it may break Emotion’s automatic SSR support. However, in this case, Chakra UI's logic runs exclusively on the client, making this usage safe. On the server, the hook simply returns `null`. [Changelog Reference](https://github.com/emotion-js/emotion/blob/cce67ec6b2fc94261028b4f4778aae8c3d6c5fd6/packages/react/CHANGELOG.md?plain=1#L267)

No

## 📝 Additional Information

Users should update Emotion to v11.14.0


packages/components/src/checkbox/checkbox.test.tsx
```diff
@@ -535,11 +535,3 @@ test("On resetting form, all checkboxes in the form should reset to its default
   expect(checkbox1).toBeChecked()
   expect(checkbox2).not.toBeChecked()
 })
-
-test("Checkbox with an ID should have a for on the label", () => {
```
isBatak, line 539: @segunadebayo It seems this `for` attribute was removed here https://github.com/chakra-ui/chakra-ui/commit/34802a2e1ae17daf35441fd4f2da9cbe82094a41

segunadebayo: Good catch! I've added it back

segunadebayo:
@isBatak can you rebase this? and I'll have it merged

PR #9707, opened 2025-02-20
fix(slider-v2): Correctly offset custom size slider thumb

### LINKED ISSUE #9151: [Regression] Slider thumb offset issue #2
### Description

We have recently attempted to update `@chakra-ui/react` from `v2.8.2` to `v2.10.4` but noticed a bug with the `Slider` component. 

Although the following is no longer the issue for the default slider thumb, it still appears to be a problem for custom size slider thumbs.

It looks like the `SliderThumb` component gets rendered with a slight offset compared to the actual value. The following appears to be the case for both keyboard and mouse interactions. Interestingly, due to the following issue, the `SliderThumb` component could be moved beyond the `SliderTrack`.

<video src="https://github.com/user-attachments/assets/f33fb3fa-5b11-4ffe-97d7-6da54223f8a1
"></video>

### Link to Reproduction
You can find an example of such a slider under the `Customizing the Slider` section on the V2 Chakra UI docs:
https://v2.chakra-ui.com/docs/components/slider/usage#customizing-the-slider

### Steps to reproduce

1. Click in the centre of the slider thumb
2. Notice how the slider thumb jumps slightly to the right
3. Move the slider to the end of the available track
4. Notice how the slider thumb moves beyond the available track

### Chakra UI Version

2.10.4

### Browser

Google Chrome 129.0.6668.101

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

This is a follow-up to https://github.com/chakra-ui/chakra-ui/issues/8945#issuecomment-2464216634

andrey-mitko:
Closes #9151

## 📝 Description
The following pull request addresses an issue in V2 of Chakra UI, where the custom size `SliderThumb` component has an incorrect offset relative to the `SliderTrack`. 

At the moment, the `SliderThumb` component gets rendered with a slight offset compared to the actual value. This appears to be the case for both keyboard and mouse interactions and appears to be caused by reliance on the hardcoded `--slider-thumb-size` variable, meaning that a custom-size slider thumb was incorrectly offset. As a side effect, this also meant that the `SliderThumb` component could be moved beyond the `SliderTrack`.

<video src="https://github.com/user-attachments/assets/f33fb3fa-5b11-4ffe-97d7-6da54223f8a1"></video>

You can find more details about this issue, along with steps to reproduce it, in the [following GitHub issue](https://github.com/chakra-ui/chakra-ui/issues/9151).

## 🚀 New behavior
Since the slider thumb offset is now dynamic and no longer reliant on the `--slider-thumb-size` variable, the `SliderThumb` component gets rendered as expected. 

<video src="https://github.com/user-attachments/assets/3f6ddccd-3150-422a-9184-b9955b2101cc"></video>

Please note that `SliderThumb` components using the standard size will continue to work as expected.

No

## 📝 Additional Information
Please do let me know if you think there is any way to improve the following PR!

### DIFF
--- a/packages/components/src/slider/slider-utils.ts
+++ b/packages/components/src/slider/slider-utils.ts
@@ -39,7 +39,7 @@ export function getStyles(options: {
       ...orient({
         orientation,
         vertical: {
-          bottom: `calc(${thumbPercents[i]}%`,
+          bottom: `${thumbPercents[i]}%`,
           transform: "translate(-50%, 50%)",
         },
         horizontal: {
--- a/packages/components/src/slider/slider-utils.ts
+++ b/packages/components/src/slider/slider-utils.ts
@@ -43,7 +43,7 @@ export function getStyles(options: {
           transform: "translate(-50%, 50%)",
         },
         horizontal: {
-          left: `calc(${thumbPercents[i]}%`,
+          left: `${thumbPercents[i]}%`,
           transform: "translate(-50%, -50%)",
         },
       }),
--- a/packages/components/src/slider/slider-utils.ts
+++ b/packages/components/src/slider/slider-utils.ts
@@ -39,10 +39,12 @@ export function getStyles(options: {
       ...orient({
         orientation,
         vertical: {
-          bottom: `calc(${thumbPercents[i]}% - var(--slider-thumb-size) / 2)`,
+          bottom: `calc(${thumbPercents[i]}%`,
+          transform: "translate(-50%, 50%)",
         },
         horizontal: {
-          left: `calc(${thumbPercents[i]}% - var(--slider-thumb-size) / 2)`,
+          left: `calc(${thumbPercents[i]}%`,
+          transform: "translate(-50%, -50%)",
         },
       }),
     }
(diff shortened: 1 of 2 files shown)


packages/components/src/slider/slider-utils.ts
```diff
@@ -39,10 +39,12 @@ export function getStyles(options: {
       ...orient({
         orientation,
         vertical: {
-          bottom: `calc(${thumbPercents[i]}% - var(--slider-thumb-size) / 2)`,
+          bottom: `calc(${thumbPercents[i]}%`,
```
segunadebayo, line 42:
```suggestion
          bottom: `${thumbPercents[i]}%`,
```

packages/components/src/slider/slider-utils.ts
```diff
@@ -39,10 +39,12 @@ export function getStyles(options: {
...
+          transform: "translate(-50%, 50%)",
         },
         horizontal: {
-          left: `calc(${thumbPercents[i]}% - var(--slider-thumb-size) / 2)`,
+          left: `calc(${thumbPercents[i]}%`,
```
segunadebayo, line 46:
```suggestion
          left: `${thumbPercents[i]}%`,
```

PR #9763, opened 2025-03-04
feat: Unplug Chakra UI

### LINKED ISSUE #9757: Let Package Managers (Yarn PnP specifically) know to unplug the package for the type generation
### Description

We found issues with running chakra-cli typegen command in yarn pnp mode.
Referring to this issue: https://github.com/chakra-ui/chakra-ui/issues/9440, we found that we could use the patch command to patch the Types into the chakra UI. But this resulted in us needing to patch the chakra-cli package each time the theme was changing.

We've found that just unplugging the package would help us with the typegen, and resolve all of these issues.

Chakra-UI could benefit from letting yarn know to unplug the package by default until Chakra UI supports Module Augmentation. 

### Proposed Solution or API

We just need to add a [preferUnplugged](https://yarnpkg.com/configuration/manifest#preferUnplugged) field in the package.json

### Alternatives

_No response_

### Additional Information

This is the suggested approach by the Yarn team as well: https://yarnpkg.com/advanced/rulebook#packages-should-never-write-inside-their-own-folder-outside-of-postinstall

agarwalvaibhav0211:
Closes #9757 

## 📝 Description

> Let Package Managers (Yarn PnP specifically) know to unplug the package for the type generation

> Chakra UI Package is stored as a .zip file in yarn pnp mode making it non-writable

## 🚀 New behavior

> Yarn PnP will by default unplug the `@chakra-ui/react` package till the time the package implements module augmentation

This may be a breaking change for users using zero-install feature of Yarn PnP. [Check the note on unplugged folder](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored). The user will need to run yarn install on every system after every upgrade to unplug and make sure that the unplugged chakra-ui is upgraded. There's a [pending issue](https://github.com/yarnpkg/berry/issues/5379) on yarn regarding this

TLDR; this will cause zero install to break in yarn pnp repo's.

### DIFF
--- a/.changeset/few-geese-matter.md
+++ b/.changeset/few-geese-matter.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react": minor
+---
+
+Unplug @chakra-ui/react to facilitate type generation in Yarn PnP mode
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -68,6 +68,7 @@
     "typedocs": "tsx scripts/typedocs.ts",
     "typecheck": "tsc --noEmit"
   },
+  "preferUnplugged": true,
   "exports": {
     ".": {
       "source": "./src/index.ts",


segunadebayo:
I'm a bit weary of merging this given the breaking change potential.

joekrill:
For what it's worth, I'd personally rather _not_ have it be forced to be unplugged. I prefer using the patch method. I may be misunderstanding how this settings works, but wouldn't this force it to be unplugged, no matter what? As it currently stands, one at least has the option to manually unplug it. But I don't think it works the other way around if unplugged is the default?

PR #9772, opened 2025-03-07
fix: checkbox / radio perf

### LINKED ISSUE #9771: trackFocusVisible causing performance issues
### Description

We recently upgraded from Chakra v2.0.2 to 2.10.5. Everything went pretty well but on some pages we noticed big performance issues when typing into the search bar. 

### The Performance Problem:

We noticed a 40x slowdown when strict mode is on:
On v2.0.2 focusing the search field and typing a character took 40ms
On v2.10.5 **it takes ~1600ms** 😬 

### The Root Cause

We narrowed the issue down to something that changed between v2.1.2 and 2.2.0. In the chrome devtools performance flame charts we noticed a lot of Checkboxes re-rendering. And found [this suspicious line added in 2.2.0](https://github.com/chakra-ui/chakra-ui/pull/6153/files#diff-445514188dc73f74a48ca922881ed7816acdcf8acca6e0bf2fb39c2e4ab340bfR167-R169).

```tsx
  useEffect(() => {
    return trackFocusVisible(setIsFocusVisible)
  }, [])
```

### Is This a Me Problem? Yes and No.

We _do_ have a lot of checkboxes because this is a List page with a lot of filters in a sidebar to help filter the list data. We have identified things we need to improve on this page that can help mitigate the introduction of this slow code **BUT** even if we have too many checkboxes a 40x slowdown is _just too much_ and there is clearly an opportunity for perf gains here.

### Here's a video of the modality changes triggering re-renders

In this video it shows how when the modality changes it causes so many other unrelated things on the screen to rerender. We are interacting with the search input but when I focus with my mouse (and modality goes to "pointer") you see the checkboxes flash with purple (using react-scan). When I then type the first letter into the search (and modality goes to "keyboard") you see the checkboxes flash again.

https://github.com/user-attachments/assets/e35df07e-bf27-4423-b6ba-16eb9f4bf816

This is essentially an unchecked global state that has the power to cause too many things to re-render too often, especially while using forms—lots of mouse to keyboard to mouse when using forms.

### Here's a video of the perf recording:

The 1658ms is the time it takes for the "a" that I typed to show up. It's a very laggy experience.

https://github.com/user-attachments/assets/3248ddc5-618d-40b5-8478-c9a02fd2167a

And a screenshot of the flame chart

Image

### Solution 🧠 💡 

Within the useEffect where trackFocusVisible is called, I think we should find a way to only call the `setFocusVisible` to `true` if the component's `isFocused` state is also true. It's pointless to have isFocusVisible thrashing around form true to false when it ultimately doesn't matter unless `isFocused` is true.

Something like this:
```tsx
  // Maybe store isFocused into a ref so we don't have to setup the effect more than once?
  const isFocusedRef = useRef(isFocused)
  isFocusedRef.current = isFocused
  useEffect(() => {
    return trackFocusVisible((focusVisible: boolean) => {
      setIsFocusVisible(isFocusedRef.current && focusVisible)
    })
  }, [])
```

### Link to Reproduction

Here's my attempt to reproduce but it must be something more complicated than this because it's not happening here: https://stackblitz.com/edit/chakra-ui-v3-corzlaz4?file=package.json,src/main.tsx,index.html

### Steps to reproduce

1. Have a page with a decent amount of checkboxes on it (or other components with the trackFocusVisible thing), like 100-200 of them
2. Have a text input as well
3. Have Strict Mode ON
4. Interact with the text input in a way that causes modality to change from mouse to keyboard and back. Click the input, type a little, click away. Each time the modality changes observe that the checkboxes re-render.

### Chakra UI Version

2.10.5

### Browser

_No response_

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

timkindberg:
Avoid re-rendering when focus visible tracker emits a value. Greatly improves performance on pages with lots of checkboxes and/or radios.

Fixes #9771

### DIFF
--- a/.changeset/rare-bugs-clap.md
+++ b/.changeset/rare-bugs-clap.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react": patch
+---
+
+Fix performance issue where all checkboxes and radios would rerender whenever
+trackFocusVisible modality changed, e.g. click into an input then type
--- a/packages/components/src/checkbox/use-checkbox.ts
+++ b/packages/components/src/checkbox/use-checkbox.ts
@@ -62,13 +62,15 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
   const onBlurProp = useCallbackRef(onBlur)
   const onFocusProp = useCallbackRef(onFocus)
 
-  const [isFocusVisible, setIsFocusVisible] = useState(false)
   const [isFocused, setFocused] = useState(false)
   const [isHovered, setHovered] = useState(false)
   const [isActive, setActive] = useState(false)
 
+  const isFocusVisibleRef = useRef(false)
   useEffect(() => {
-    return trackFocusVisible(setIsFocusVisible)
+    return trackFocusVisible((state) => {
+      isFocusVisibleRef.current = state
+    })
   }, [])
 
   const inputRef = useRef<HTMLInputElement | null>(null)
@@ -188,7 +190,7 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
         "data-hover": dataAttr(isHovered),
         "data-checked": dataAttr(isChecked),
         "data-focus": dataAttr(isFocused),
-        "data-focus-visible": dataAttr(isFocused && isFocusVisible),
+        "data-focus-visible": dataAttr(isFocused && isFocusVisibleRef.current),
         "data-indeterminate": dataAttr(isIndeterminate),
         "data-disabled": dataAttr(isDisabled),
         "data-invalid": dataAttr(isInvalid),
@@ -209,7 +211,6 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
       isChecked,
       isDisabled,
       isFocused,
-      isFocusVisible,
       isHovered,
       isIndeterminate,
       isInvalid,
@@ -225,7 +226,7 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
       "data-hover": dataAttr(isHovered),
       "data-checked": dataAttr(isChecked),
       "data-focus": dataAttr(isFocused),
-      "data-focus-visible": dataAttr(isFocused && isFocusVisible),
+      "data-focus-visible": dataAttr(isFocused && isFocusVisibleRef.current),
       "data-indeterminate": dataAttr(isIndeterminate),
       "data-disabled": dataAttr(isDisabled),
       "data-invalid": dataAttr(isInvalid),
@@ -236,7 +237,6 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
       isChecked,
       isDisabled,
       isFocused,
-      isFocusVisible,
       isHovered,
       isIndeterminate,
       isInvalid,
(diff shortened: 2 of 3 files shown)


timkindberg:
I am having trouble running the tests locally.  Getting this output:

```bash
> chakra-ui@2.0.0 test /Users/tim.kindberg/projects/chakra-ui
> vitest

/Users/tim.kindberg/projects/chakra-ui/node_modules/.pnpm/rollup@4.6.1/node_modules/rollup/dist/native.js:64
                throw new Error(
                      ^

Error: Cannot find module @rollup/rollup-darwin-arm64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at requireWithFriendlyError (/Users/tim.kindberg/projects/chakra-ui/node_modules/.pnpm/rollup@4.6.1/node_modules/rollup/dist/native.js:64:9)
    at Object.<anonymous> (/Users/tim.kindberg/projects/chakra-ui/node_modules/.pnpm/rollup@4.6.1/node_modules/rollup/dist/native.js:73:48)
    ... 3 lines matching cause stack trace ...
    at Module._load (node:internal/modules/cjs/loader:1022:12)
    at cjsLoader (node:internal/modules/esm/translators:366:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:315:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24) {
  [cause]: Error: Cannot find module '@rollup/rollup-darwin-arm64'
  Require stack:
  - /Users/tim.kindberg/projects/chakra-ui/node_modules/.pnpm/rollup@4.6.1/node_modules/rollup/dist/native.js
      at Module._resolveFilename (node:internal/modules/cjs/loader:1143:15)
      at Module._load (node:internal/modules/cjs/loader:984:27)
      at Module.require (node:internal/modules/cjs/loader:1231:19)
      at require (node:internal/modules/helpers:179:18)
      at requireWithFriendlyError (/Users/tim.kindberg/projects/chakra-ui/node_modules/.pnpm/rollup@4.6.1/node_modules/rollup/dist/native.js:62:10)
      at Object.<anonymous> (/Users/tim.kindberg/projects/chakra-ui/node_modules/.pnpm/rollup@4.6.1/node_modules/rollup/dist/native.js:73:48)
      at Module._compile (node:internal/modules/cjs/loader:1369:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1427:10)
      at Module.load (node:internal/modules/cjs/loader:1206:32)
      at Module._load (node:internal/modules/cjs/loader:1022:12) {
    code: 'MODULE_NOT_FOUND',
    requireStack: [
      '/Users/tim.kindberg/projects/chakra-ui/node_modules/.pnpm/rollup@4.6.1/node_modules/rollup/dist/native.js'
    ]
  }
}

Node.js v20.12.2
 ELIFECYCLE  Test failed. See above for more details.

```

PR #9903, opened 2025-04-07
fix: dialog drawer backdrop

MohamedSayed008:
Closes # 

## 📝 Description

 
 

The zIndex: `-1` introduced an issue where the backdrop appears in front of the positioner when the dialog is opened, even though it fixed the drawer issue. This happens because `--z-index` is set to `1400` for the drawer, but `1402` for the dialog.

By default, the positioner uses the modal z-index `(var(--chakra-z-index-modal))`, which equals `1400`. Previously, it was using a zIndex of `1402`, which caused it to appear in front of the positioner. The earlier fix ([commit e3d7db7#r154974953](https://github.com/chakra-ui/chakra-ui/pull/9903#)) adjusted it from `1402` to `1401`.

Now, since it's added to the layer styles, the value is more explicit, ensuring the overlay consistently appears above the positioner.

This fix ensures it's always -1 from the positioner zIndex, which solves the problem. 
 
 
## 🚀 New behavior

No

## 📝 Additional Information

Environment used is nextJs dev environment
Browser: Chrome latest version

### DIFF
--- a/packages/react/src/theme/recipes/dialog.ts
+++ b/packages/react/src/theme/recipes/dialog.ts
@@ -12,7 +12,7 @@ export const dialogSlotRecipe = defineSlotRecipe({
       top: 0,
       w: "100vw",
       h: "100dvh",
-      zIndex: "calc(var(--z-index) - 2) ",
+      zIndex: "calc(var(--z-index) - 2)",
       _open: {
         animationName: "fade-in",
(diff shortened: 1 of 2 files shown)


segunadebayo:
Really good findings, Maradona.

I've improved on the PR changes in the main branch.

I appreciate you, buddy.

PR #9962, opened 2025-04-22
fix(hooks): fix critical React hooks bugs and memory leaks

Codimow:
## 📝 Description
This PR fixes several critical bugs in React hooks implementation that could lead to memory leaks, stale closures, and unexpected behavior in components using these hooks.

Currently, there are several issues in our hooks implementation:
1.use-overlay.tsx uses array indices as React keys which can cause issues when overlays are added/removed
2.use-callback-ref.ts has incorrect dependency handling, potentially causing stale closures
3.use-element-rect.ts doesn't properly clean up event listeners or calculate initial values
4.use-media-query.ts doesn't properly handle query changes in the dependency array

These issues can lead to memory leaks, UI inconsistencies, and degraded performance in applications.

## 🚀 New behavior

1.use-overlay.tsx now uses stable overlay IDs as keys and properly manages overlay lifecycle
2.use-callback-ref.ts correctly handles dependencies to prevent stale closures
3.use-element-rect.ts calculates initial values properly and ensures full cleanup
4.use-media-query.ts properly responds to query changes and handles cleanup correctly

These improvements make the hooks more reliable and prevent memory leaks and performance issues.

No

## 📝 Additional Information
These fixes should be backward compatible with existing code while providing more stable behavior and better performance. The changes follow React best practices for hooks implementation.

### DIFF
--- a/packages/react/src/hooks/use-callback-ref.ts
+++ b/packages/react/src/hooks/use-callback-ref.ts
@@ -14,10 +14,12 @@ export function useCallbackRef<Args extends unknown[], Return>(
     throw new Error("Cannot call an event handler while rendering.")
   })
 
+  // Use useInsertionEffect to update the ref before React processes effects
   useInsertionEffect(() => {
     callbackRef.current = callback
-  })
+  }, [callback])
 
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useCallback((...args: Args) => callbackRef.current?.(...args), deps)
+  // We're removing the dependency array because we want this callback to be stable
+  // across renders, and only the internal ref.current should change
+  return useCallback((...args: Args) => callbackRef.current?.(...args), [])
 }
--- a/packages/react/src/hooks/use-element-rect.ts
+++ b/packages/react/src/hooks/use-element-rect.ts
@@ -1,29 +1,44 @@
 "use client"
 
 import { useEnvironmentContext } from "@ark-ui/react"
-import { useEffect, useRef, useState } from "react"
+import { useCallback, useEffect, useRef, useState } from "react"
 
 export function useElementRect() {
   const ref = useRef<HTMLElement | null>(null)
   const [rect, setRect] = useState<DOMRect | null>(null)
   const env = useEnvironmentContext()
 
+  // Extract the calculation logic to a callback
+  const updateRect = useCallback(() => {
+    if (ref.current) {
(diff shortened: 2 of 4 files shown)


packages/react/src/hooks/use-callback-ref.ts
```diff
@@ -14,10 +14,12 @@ export function useCallbackRef<Args extends unknown[], Return>(
...
   useInsertionEffect(() => {
     callbackRef.current = callback
-  })
+  }, [callback])
 
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useCallback((...args: Args) => callbackRef.current?.(...args), deps)
+  // We're removing the dependency array because we want this callback to be stable
+  // across renders, and only the internal ref.current should change
+  return useCallback((...args: Args) => callbackRef.current?.(...args), [])
```
isBatak, line 24:
```suggestion
  return useCallback((...args: Args) => callbackRef.current?.(...args), deps)
```
It will be stable because of the default `[]` value on the line 11 `deps: React.DependencyList = [],`.
`deps` allows us to surgically update reference form the user-land which is useful in some edge-cases, but it doesn't make this callback unstable.

isBatak (changes requested):

packages/react/src/hooks/use-element-rect.ts
```diff
@@ -1,29 +1,44 @@
...
-import { useEffect, useRef, useState } from "react"
+import { useCallback, useEffect, useRef, useState } from "react"
 
 export function useElementRect() {
   const ref = useRef<HTMLElement | null>(null)
   const [rect, setRect] = useState<DOMRect | null>(null)
   const env = useEnvironmentContext()
 
+  // Extract the calculation logic to a callback
+  const updateRect = useCallback(() => {
```
isBatak, line 12: This seems wrong. We added extra work and didn't solve anything. I would revert this change back to `handleResize`. 
Real solution would be [`useEffectEvent`](https://react.dev/reference/react/experimental_useEffectEvent) but it's not released yet.
`useCallback` is meant to only be used for optimising callback props to prevent component re-rendering (in combination with `memo`), not to be used inside `useEffects`. This could easily get unoptimised if someone accidentally add dependency to the array in the future update.

segunadebayo:
Closing due to inactivity. Kindly address the comments above to re-open

PR #10063, opened 2025-05-30
fix(utils): prevent infinite loop in objectToArrayNotation with trailing nulls

stechermichal:
There is no corresponding issue to this PR

## 📝 Description

v2 - Fixed infinite loop in `objectToArrayNotation` function that occurs when the last value in the result array is null.

The function uses `while (lastItem === null)` but `lastItem` is assigned once before the loop and never updated, causing an infinite loop when the last array element is null.

## 🚀 New behavior

The function now checks the array length and re-evaluates the last item on each iteration: `while (result.length > 0 && result[result.length - 1] === null)`

No

## 📝 Additional Information
This bug was introduced during a refactor and affects responsive props handling when trailing breakpoint values are null.

### DIFF
--- a/packages/utils/src/responsive.ts
+++ b/packages/utils/src/responsive.ts
@@ -33,8 +33,9 @@ export function objectToArrayNotation(
   bps = breakpoints,
 ) {
   const result = bps.map((br) => obj[br] ?? null)
-  const lastItem = result[result.length - 1]
-  while (lastItem === null) result.pop()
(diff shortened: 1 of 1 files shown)


stechermichal:
@segunadebayo Hi, hope it's okay to tag you here so that someone could have a look, thanks.

kalisaNkevin (commented):
Add a changeset to this since this is a published package fix.

PR #10071, opened 2025-06-02
fix(chart-tooltip): apply formatter prop to display formatted values

gaurav-132:
Sure! Here's a filled-out version of your PR template based on your recent change (`formatter` prop in `ChartTooltip`):

---

Closes # 

## 📝 Description

> This PR adds support for a `formatter` prop in the `ChartTooltip` component to allow formatting values using custom logic (e.g., number formatting, currency, percentage).

> Currently, `ChartTooltip` displays raw values without formatting, making it difficult to localize or display them with consistent precision or styles.

## 🚀 New behavior

> With this update, users can pass a `formatter` function as a prop to `ChartTooltip`, which will be used to format the values shown in the tooltip.

No

## 📝 Additional Information

This is a non-breaking enhancement and does not introduce new dependencies. Useful for displaying values in a user-friendly or localized format inside charts.

---

### DIFF
--- a/packages/charts/src/chart/chart.tsx
+++ b/packages/charts/src/chart/chart.tsx
@@ -197,6 +197,7 @@ export interface ChartTooltipProps extends TooltipProps<string, string> {
   fitContent?: boolean
   nameKey?: string
   indicator?: "line" | "dot" | "dashed"
+  formatter?: (value: any) => React.ReactNode
   render?: (item: Payload<string, string>) => React.ReactNode
 }
 
@@ -211,6 +212,7 @@ export function ChartTooltip(props: ChartTooltipProps) {
     showTotal,
     fitContent,
     nameKey,
+    formatter,
     render,
   } = props
 
@@ -273,7 +275,9 @@ export function ChartTooltip(props: ChartTooltipProps) {
                     fontWeight="medium"
                     fontVariantNumeric="tabular-nums"
                   >
-                    {item.value.toLocaleString()}
+                    {formatter
+                      ? formatter(item.value)
+                      : item.value.toLocaleString()}
                   </Text>
                 )}
               </HStack>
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks @gaurav-132. Can you include an example for this?

gaurav-132:
example like what i have done or why?

gaurav-132:
This is a utility function that returns a formatter.

const formatter = chart.formatNumber({ style: "currency", currency: "USD" });


chart.formatNumber(...) takes formatting options.

It returns a function that you can use to format numbers.

Example:- formatter(1000); // "$1,000.00"

2. Passing the formatter as a prop
Let’s say you are using a component like ChartTooltip that accepts a formatter prop:

<ChartTooltip value={1000} formatter={formatter} />

Here:

value={1000} is the number you want to display.

formatter={formatter} is the function used to format value.

Final Output in UI
If you pass value={1000} and formatter={chart.formatNumber(...)}, the output will be:  $1,000.00

segunadebayo:
Let's match the API for the tooltip formatter in recharts. Notice it supports multiple parameters (value and name)

In this case, let's add `name` as well

CleanShot 2025-06-02 at 19 25 32@2x

gaurav-132:
I noticed that we are already displaying the name, but the formatted value is not currently being shown. From what I understand, our main goal here is to format and display the value correctly.

Could you please clarify if the name prop is still required in this case? If so, I’d really appreciate your guidance on how it would be used in this context. As a beginner in open source contributions, I’m eager to learn and ensure this PR is completed successfully with your support.

Thank you!

gaurav-132:
Implemented tooltip value formatting aligned with Recharts' formatter behavior. If any issues arise, please advise on necessary adjustments.

gaurav-132:
Hi @segunadebayo, why i am not showing in contributors list?

PR #10121, opened 2025-06-28
fix: add special handling for top-level selectors

adonig:
## Description

This PR adjusts how selector strings are transformed during serialization by improving support for top-level selectors like `:host` and `:host-context`. This avoids incorrect nesting or formatting when dealing with Web Components and Shadow DOM contexts.

## Current behavior

Currently, selectors that begin with `:host` or `:host-context` are treated like any other nested selector, which can result in incorrect CSS output. For example, `:host(.foo)` might become `&:host(.foo)` or similar, which breaks the intended semantics.

## New behavior

Selectors that begin with `:host` or `:host-context` are now recognized as top-level and are serialized accordingly with `:host &` formatting, preserving proper structure and specificity when used inside Shadow DOM or custom elements.

## Is this a breaking change?

No

## Additional Information

The logic was added in the `createSerializeFn` helper. A new helper function `isTopLevelSelector()` checks for `:host` and `:host-context` prefixes and adjusts selector composition accordingly during serialization.

### DIFF
--- a/packages/react/src/styled-system/serialize.ts
+++ b/packages/react/src/styled-system/serialize.ts
@@ -13,7 +13,7 @@ export function createSerializeFn(
 
         if (!conditions.has(prop) && !isValidProperty(prop)) {
           return parseSelectors(prop)
-            .map((s) => "&" + s)
+            .map((s) => (isTopLevelSelector(s) ? `${s} &` : `&${s}`))
             .join(", ")
         }
 
@@ -21,6 +21,10 @@ export function createSerializeFn(
       },
     })
   }
+}
+
+function isTopLevelSelector(s: string): boolean {
+  return s.startsWith(":host") || s.startsWith(":host-context")
 }
 
 function parseSelectors(selector: string): string[] {


adonig:
I now believe it might even be better to move this into `conditions.has`:

```TypeScript
  function has(key: string) {
    return keys().includes(key)
     || /^@|&|&$/.test(key)
     || key.startsWith("_")
     || key.startsWith(":host")
     || key.startsWith(":host-context")
  }
```

adonig:
I’m still running into an issue during development where this style gets injected automatically:

```html
<style data-emotion="root" data-s="">
  .root-pcu6ol:host(.dark) .root-pcu6ol {
    background: var(--chakra-colors-black);
    --bg-currentcolor: var(--chakra-colors-black);
  }
</style>
```

I’ve tried placing `console.log` statements in various parts of the Chakra and Emotion internals to trace the origin, but so far none of them seem to trigger, which makes me think the injection might be happening in a place I’m overlooking.

@segunadebayo Could you point me in the right direction where this style is being generated or injected during dev mode? I’d like to understand what’s triggering this so I can adjust or disable it if needed.

Thanks in advance!

segunadebayo:
Thanks for helping with this @adonig 

This must be happening due to the global css and the `_dark` condition
https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/theme/global-css.ts#L45
https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/preset-base.ts#L187

segunadebayo:
I just pushed a fix for this, taking ideas from your PR. We'll release an update shortly.

PR #10141, opened 2025-07-08
docs: update breakpoint unit faq

AmAzing129:
## 📝 Description

breakpoints' unit is "em", and the article below is recommending "em", too. But the FAQ is "rem". I'm missing something or it should be "em"?

image


segunadebayo:
The breakpoint unit is actually rem. Will remove the link to that article

PR #10181, opened 2025-07-22
feat: allow nested conditions to be used in tokens

tryone144:
## 📝 Description

Allow using 'nested' conditions when defining (semantic) tokens. One such example is the existing `_hover` condition that resolves to
```js
[ "@media (hover: hover)",
  "&:is(:hover, [data-hover]):not(:disabled, [data-disabled])" ]
```
for nesting the actual selector inside the `@media`-query.

Using a condition that is not a simple `string` fails with `TypeError: selector.startsWith is not a function` [here](https://github.com/chakra-ui/chakra-ui/blob/11423e71a5f8b90f9b89d9aa26db44c209976ad9/packages/react/src/styled-system/system.ts#L138). This is true for the default `_hover` condition [here](https://github.com/chakra-ui/chakra-ui/blob/11423e71a5f8b90f9b89d9aa26db44c209976ad9/packages/react/src/preset-base.ts#L74-L77).

## 🚀 New behavior

This PR updates the check for conditions that consist of *only* media-queries to support conditions that are of type `string[]` (like the default `_hover`). This then allows users to add and use their own conditions.

Our project would like to overwrite the default `_dark` selector to only apply on `@media not print` which requires this feature.

No


Adebesin-Cell:
Thanks for putting this together 🙏

This is on our radar and will land as part of a broader direction for the token system down the line. Going to close for now; the idea's a good one, and we'll pick it back up when we get there. Really appreciate the contribution 🙏

PR #10185, opened 2025-07-25
feat(button): add Compact variant examples to Storybook

Varnika060306:
## 📝 Description

This PR adds a new example to the Chakra UI Storybook demonstrating the `isCompact` prop for the `Button` component.  
The example showcases four compact button variants:
- Compact Default  
- Compact Blue  
- Compact Outline  
- Compact Small  

These additions improve the discoverability and documentation of the `isCompact` prop for developers using Chakra UI.

---

## 🧰 Current behavior (updates)

The current Storybook does not include any visual reference or usage examples of the `isCompact` prop in the Button component.

---

## 🚀 New behavior

This PR introduces a new story titled **"Compact"** under the `Button` section in Storybook.  
It renders multiple compact-styled buttons using the `isCompact` prop, helping users understand its effect.

---

## 🧨 Is this a breaking change (Yes/No):

**No**

---

## ℹ️ Additional Information

- Tests were run with `pnpm test` using `vitest` — all tests passed successfully ✅  
- Visual output was confirmed under `localhost:6006` in Storybook  
- Story file updated: `button.stories.tsx`

### DIFF
--- a/apps/compositions/src/examples/button-with-compact.tsx
+++ b/apps/compositions/src/examples/button-with-compact.tsx
@@ -3,14 +3,14 @@ import { Button, Stack } from "@chakra-ui/react"
 export const ButtonWithCompact = () => {
   return (
     <Stack direction="row" spacing="4">
-      <Button isCompact>Compact Default</Button>
-      <Button isCompact colorScheme="blue">
+      <Button compact>Compact Default</Button>
+      <Button compact colorScheme="blue">
         Compact Blue
       </Button>
-      <Button isCompact variant="outline">
+      <Button compact variant="outline">
         Compact Outline
       </Button>
-      <Button isCompact size="sm" colorScheme="green">
+      <Button compact size="sm" colorScheme="green">
         Compact Small
       </Button>
     </Stack>
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
   "dependencies": {
     "@babel/core": "7.28.0",
     "@babel/eslint-parser": "7.28.0",
-    "@chakra-ui/cli": "workspace:*",
     "@changesets/changelog-github": "0.5.1",
     "@changesets/cli": "2.29.5",
     "@changesets/get-release-plan": "4.0.13",
@@ -114,5 +113,8 @@
   "packageManager": "pnpm@10.13.1",
   "engines": {
     "node": ">=20.x"
+  },
+  "devDependencies": {
+    "@chakra-ui/cli": "3.0.0"
   }
 }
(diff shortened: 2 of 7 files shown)


Varnika060306:
Thank you for reviewing! Let me know if any changes are needed.

isBatak:
The recommended way would be to add boolean variant in the button recipe.
You can extend your recipe like this:
```js
export const buttonRecipe = defineRecipe({
  variants: {
     compact: {
       true: {
          paddingInline: "0.5rem",
          paddingBlock: "0.25rem",
          fontSize: "0.875rem",
          height: "auto",
        },
      },
    },
    //...
 });
 ```
 
Use the Chakra UI CLI typegen to generate new types.
 
 Then use it like this:
 
 `<Button compact size="sm" colorScheme="green">Button</Button>`
 
Here is the Table recipe example https://github.com/chakra-ui/chakra-ui/blob/2cafb0ee474fe50e6fb3432ac1dbf02301ffe6ac/packages/react/src/theme/recipes/table.ts#L40
 
Chakra UI v3 doesn't prefix boolean props with `is` anymore.

I don't think this `compact` prop should be a part of Chakra core, it should stay in the user-land.
What you can contribute is to add docs for the boolean variant.

Varnika060306:
Hey! I’ve implemented the compact: true boolean variant in the button recipe as per your suggestion, and also added a usage example to Storybook.

Changes made:
	•	Added compact boolean variant inside defineRecipe (under variants.true)
	•	Did not use isCompact, as Chakra UI v3 uses raw booleans
	•	Added <Button compact size="sm" /> in the usage story
	•	Kept everything in user-land (not Chakra core)

Still stuck on:
	•	Running Chakra CLI typegen to generate types for the compact prop – I wasn’t sure how to run it correctly inside the monorepo (especially under packages/react).

Would love your help/guidance on how to run that step properly! Let me know if anything else needs fixing. Thanks again

segunadebayo:
Closing based on https://github.com/chakra-ui/chakra-ui/pull/10185#issuecomment-3117792205

This prop is not something we want to support in Chakra UI. You should add it to your own project.

isBatak:
We had a misunderstanding earlier.
When I mentioned “user-land,” I was referring to your project, not this repository.

The compact button variant seems to be a specific requirement for your project. Since this type of button recipe extension can easily be implemented on the user side, I don’t think it belongs in the Chakra repo. We aim to keep the default recipes simple to reduce maintenance overhead.

With that in mind, instead of this PR, you could open a new one that updates the documentation. Specifically, it could explain how to extend a recipe with a boolean variant somewhere on this page https://www.chakra-ui.com/docs/theming/recipes. That kind of contribution would add significantly more value.

PR #10279, opened 2025-08-28
docs: add radiocard examples

frittate:
Closes # 

## 📝 Description

This PR updates the docs for the Radio Card component by adding a "controlled radio card" section.
The section closely aligns with the "controlled" section of the Radio Group component.
- added a new section in the docs markdown, describing how to use the RadioCard in a controlled fashion
- added a new example code that uses the RadioCard in a controlled fashion
- updated the props to display the "value" and "onValueChange" props.

- updated one example in the RadioCard to use the correct prop `size` for an icon, instead of `fontSize`.

The RadioCard allows for a controlled use, but the docs omit this.

## 🚀 New behavior

The docs describe how to use the RadioCard in a controlled way.

No.

### DIFF
--- a/apps/compositions/src/examples/radio-card-controlled.tsx
+++ b/apps/compositions/src/examples/radio-card-controlled.tsx
@@ -1,12 +1,13 @@
 "use client"
 
 import { HStack, RadioCard } from "@chakra-ui/react"
+import { useState } from "react"
 
 export const RadioCardControlled = () => {
-  const [value, setValue] = useState<string | null>(null)
-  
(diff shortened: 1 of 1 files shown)


apps/compositions/src/examples/radio-card-controlled.tsx
```diff
@@ -0,0 +1,28 @@
+import { HStack, RadioCard } from "@chakra-ui/react"
```
segunadebayo, line 1:
```suggestion
"use client"

import { HStack, RadioCard } from "@chakra-ui/react"
```

PR #10331, opened 2025-09-19
bugfix: Update shadows.ts

agashane:
## 📝 Description

Change values for `black` color token to `{colors.black}`

After installing the preset from npm, it could not work in pandacss since `black` color token could not be found. I had to manually correct it and use `{colors.black}` instead.

## 🚀 New behavior

Update in `shadows.ts` semantic token and changed values `black` to `{colors.black}`

No


segunadebayo:
This is an underlying bug in the Chakra theme itself. I'm working on fixing it right now.

PR #10339, opened 2025-09-22
fix: make CodeBlock line numbers consistently 1-based

Mr-KRAMA:
- Fix documentation error in Line focus section
- Update example to use correct 1-based indexing


segunadebayo:
Thanks for this PR. I've gone ahead to implement the fix in the main branch due to the complexity in yours.

Mr-KRAMA:
@segunadebayo Hey, thanks for applying the fix! Since the solution originated from this PR, would it be possible to add me as a co-author in the commit history or contributors list?”

Issue #10327, opened 2025-09-18
[CodeBlock] Add some padding/margin to a code row


zhutmost:
### Description

When I scroll to the right end of a code row:
<img width="787" height="610" alt="Image" src="https://github.com/user-attachments/assets/5306b8e7-4afe-4266-8ea0-03ee289c58de" />
There is no blank between the right border and the word `environment`. I think there should be some margin or padding to make it more beautiful.

### Problem Statement/Justification

n/a

### Proposed Solution or API

n/a

rajappa-adabala:
I would like to work on this issue.

rajappa-adabala:
Hi @zhutmost 👋

I’d like to work on this issue.

I have added right padding to the Code component so that very long lines of code no longer touch the right border. This ensures a small margin is always present, making it visually cleaner.

I’ve tested it locally using Storybook, and the change applies correctly when scrolling long code lines.

I’ll open a PR with this fix shortly.

Coderxrohan:
Made chnages and seen in deployment, now issue has beed resolved. @segunadebayo 
Please verify and see pull request.

segunadebayo:
I just pushed a fix for this. We'll release an update shortly.

If the issue persists after upgrading, I'll re-open it.

PR #10370, opened 2025-10-07
fix(code): add right padding to prevent text from touching border

Coderxrohan:
Closes #<i#10327>  https://github.com/chakra-ui/chakra-ui/issues/10327

## 📝 Description

This PR fixes a small UI issue in the `<Code>` component where long lines of code touch the right border without any spacing.  
I added a small right padding to the base style so the text always has a little breathing room.

Right now, if a line of code is too long and scrolls horizontally, the last character touches the right border, which looks cramped.

## 🚀 New behavior

With this change, there’s always a small margin on the right side.  
Now even long lines scroll cleanly without sticking to the edge.

No – this is just a minor styling fix, no breaking changes.

## 📝 Additional Information

Verified in Storybook with long code lines.  
This improves readability and keeps spacing consistent with the left side.

### DIFF
--- a/packages/react/src/theme/recipes/code.ts
+++ b/packages/react/src/theme/recipes/code.ts
@@ -4,13 +4,11 @@ import { badgeRecipe } from "./badge"
 const { variants, defaultVariants } = badgeRecipe
 
 export const codeRecipe = defineRecipe({
-  className: "chakra-code",
   base: {
     fontFamily: "mono",
-    alignItems: "center",
-    display: "inline-flex",
-    borderRadius: "l2",
+    fontSize: "sm",
+    px: "0.2em",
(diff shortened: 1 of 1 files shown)


Coderxrohan:
A small fix to add right padding in Code so long lines don’t stick to the border. Keeps things cleaner and more readable.
@Coderxrohan & @segunadebayo

segunadebayo:
This isn't the proper fix. Will look into it myself

Coderxrohan:
I’ve updated the fix to scope the padding only for block variants so inline code stays unaffected. This should resolve the long-line border issue cleanly.
https://github.com/chakra-ui/chakra-ui/pull/10371

PR #10393, opened 2025-10-16
feat(menu): add isActive to MenuItem mapping to aria-current="page"

Samii2383:
This PR adds an isActive prop to MenuItem that, when true, maps to aria-current="page". This communicates the active (current) item to assistive technologies without changing visuals or behavior.
What’s included:
MenuItemProps extended with optional isActive?: boolean.
Prop mapping injects aria-current="page" when isActive is true.
Unit test asserting aria-current="page" is present for active items.
Why:
Improves accessibility by conveying the current selection in a menu using the correct ARIA semantics.
No styling or layout changes; purely semantic enhancement aligned with Chakra’s accessibility goals.
Logic explained:
The MenuItem wrapper maps isActive to an ARIA attribute:
If isActive is true → add aria-current="page".
Otherwise → render unchanged.
This pattern follows ARIA guidance for indicating the current item in a navigation/menu context, helping screen readers announce the active entry reliably.
Implementation uses prop mapping on the Chakra component wrapper so consumers retain the same API and composition patterns, with an opt-in a11y upgrade.
Test plan:
pnpm install
Run tests at the workspace root (or package-level) and verify:
New menu test passes, asserting aria-current="page" appears when isActive is set.
No regressions across existing tests.
Notes:
No visual or breaking API changes; isActive is optional and additive.
Works with existing Menu composition patterns.

### DIFF
--- a/.changeset/menu-aria-current.md
+++ b/.changeset/menu-aria-current.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react": minor
+---
+
+feat(menu): add isActive to MenuItem mapping to aria-current="page"
+


Samii2383:
Hi! Thanks for the quick review. I proposed aria-current="page" on MenuItem to communicate an active entry, but I understand this may be semantically incorrect for command menus. Could you clarify the preferred pattern for indicating an “active/current” entry in Chakra’s Menu?
If it’s a navigation use case, would you prefer a docs example using Link with aria-current instead (outside Menu)?
If it’s selection, should we steer users to MenuCheckboxItem/MenuRadioItem (via aria-checked) or expose a prop that maps to the correct ARIA semantics?
I’m happy to rework this as a docs-only update or the appropriate component change based on your guidance. Thanks!
New issue (open in upstream to align on direction)
Title: Clarify recommended a11y pattern for “active/current” state in Menu
Body:
Context: I attempted to expose an isActive prop on MenuItem that mapped to aria-current="page", which was closed as semantically incorrect for command menus. I’d love guidance on the recommended pattern.
Questions:
For navigation-like menus, should we recommend Link + aria-current outside of Menu (docs example)?
For selection inside Menu, should users rely on MenuCheckboxItem/MenuRadioItem (announced via aria-checked)?
Would a docs example for “active styling only” (no ARIA) be appropriate to avoid misleading semantics?
I can open a revised PR (docs-only or correct component) once we agree on the approach.
Links
Upstream Chakra UI: https://github.com/chakra-ui/chakra-ui
Your fork: https://github.com/Samii2383/chakra-ui

PR #10409, opened 2025-10-21
Fix collapse width animation to use width instead of height

danielfdsilva:
## 📝 Description

Collapse width animation is using the height property, but it should be height instead.


segunadebayo:
The panda preset is auto-generated. I'll update it soon.

PR #10410, opened 2025-10-23
feat: add tour component

Adebesin-Cell:
## 📝 Description

Introduces the new `Tour` component to Chakra UI, a guided tour feature that helps users understand interfaces through interactive steps with contextual highlights and messaging.

N/A—this is a new component. Chakra UI did not previously have a Tour component.

## 🚀 New behavior

**New Tour component features:**
- `Tour.Root` / `Tour.RootProvider` - container with `useTour` hook support
- `Tour.Backdrop` 
- `Tour.Spotlight`
- `Tour.Positioner`
- `Tour.Content`, `Tour.Title`, `Tour.Description` - step messaging
- `Tour.Arrow`, `Tour.ArrowTip` - pointer styling
- `Tour.Control`, `Tour.ProgressText`, `Tour.ActionTriggers` - navigation UI
- `Tour.CloseTrigger` - dismiss button

**Tour features:**
- Multiple step types: `dialog`, `tooltip`, `floating`, `wait`
- Configurable placement, backdrop, arrow visibility per step
- Programmatic control: `start()`, `next()`, `prev()`, `setStep()`, progress tracking
- Step effects and lifecycle hooks (`onStepChange`, `onStatusChange`)
- Wait steps for conditional progression

**No**

This is a new component with no impact on existing functionality.

## 📝 Additional Information

Built on top of Ark UI's Tour primitive with Chakra styling and composition patterns. Follows Chakra's slot recipe structure for customization.


spiftire:
Looks great. When will this be ready?

Adebesin-Cell:
> Looks great. When will this be ready?

Thanks! @spiftire 

There's no specific time it will be released yet. When it's released, there will be an announcement for it.

gsanta:
Can I help sth with this component? We would like to use it

kalisaNkevin (commented):
Nice work — this is a big surface and most of it is right.

The part I'd single out is the **`positioner` recipe**. Branching on `&[data-type=dialog], &[data-type=floating]` instead of inventing recipe variants is exactly the right call: the step type already lives in the machine, so the styling follows the data rather than asking the user to keep a prop in sync with it. And the `pointerEvents: "none"` on the full-viewport positioner paired with `pointerEvents: "auto"` on `content` is the detail that's easy to miss and painful to debug later — a tour that silently eats clicks on the page it's pointing at is a nasty bug, and it's already handled here. Same for the `--tour-z-index` + `--layer-index` arithmetic keeping the backdrop one layer under the content.

Re-exporting `tourAnatomy` straight from Ark rather than `extendWith`-ing custom parts is also the right instinct — it keeps the recipe in lockstep with upstream and means there's nothing to re-sync on the next Ark bump. And thanks for remembering `packages/panda-preset/src/slot-recipes/tour.ts`; that one gets forgotten a lot.

Most of what follows is one root-wiring issue plus smaller things. Only the first is blocking from my side.

packages/react/src/components/tour/tour.tsx
```diff
@@ -0,0 +1,227 @@
...
+////////////////////////////////////////////////////////////////////////////////////
+
+export interface TourRootBaseProps
+  extends Assign<ArkTour.RootBaseProps, SlotRecipeProps<"tour">>,
+    UnstyledProp {}
+
+export interface TourRootProps
+  extends HTMLChakraProps<"div", TourRootBaseProps> {}
+
+export const TourRoot = withProvider<HTMLDivElement, TourRootProps>(
```
kalisaNkevin: `ArkTour.Root` renders no DOM element. Its implementation destructures only `{ children, tour }` (plus presence/render-strategy props) and returns a provider stack — `TourProvider` → `RenderStrategyPropsProvider` → `PresenceProvider`. Everything else is dropped on the floor.

So wiring it through `withProvider(..., "root", { forwardAsChild: true })` with `TourRootProps extends HTMLChakraProps<"div", ...>` advertises an API that doesn't exist:

```tsx
<Tour.Root tour={tour} bg="red" p="4" asChild ref={ref} />
// typechecks, renders nothing, ref never attaches
```

The `"root"` slot doesn't exist in the anatomy either — zag's tour parts are `content, actionTrigger, closeTrigger, progressText, title, description, positioner, arrow, arrowTip, backdrop, spotlight`, and Ark adds `control`. No `root`.

`withRootProvider` is the pattern for this shape; `Dialog` does the same thing for the same reason:

```ts
export const TourRoot = withRootProvider<TourRootProps>(ArkTour.Root)
```

Worth fixing before merge rather than after — it's the only finding that changes the public API surface.

packages/react/src/components/tour/tour.tsx
```diff
@@ -0,0 +1,227 @@
...
+////////////////////////////////////////////////////////////////////////////////////
+
+export interface TourRootProviderBaseProps
+  extends Assign<ArkTour.RootBaseProps, SlotRecipeProps<"tour">>,
+    UnstyledProp {}
+
+export interface TourRootProviderProps
+  extends HTMLChakraProps<"div", TourRootProviderBaseProps> {}
+
+export const TourRootProvider = withProvider<
```
kalisaNkevin: This is identical to `TourRoot` below — same component, same props type, same slot, same options. Ark Tour has no `RootProvider` export at all; the `tour` prop from `useTour()` *is* the provider mechanism, which is what every example already does.

So this gives users two names for one thing, and the name implies a `value={...}` API that isn't there. I'd drop it and keep `Tour.Root`.

packages/react/src/theme/slot-recipes.ts
```diff
@@ -105,6 +106,7 @@ export const slotRecipes = {
   tagsInput: tagsInputSlotRecipe,
   toast: toastSlotRecipe,
   tooltip: tooltipSlotRecipe,
+  tour: tourSlotRecipe,
```
kalisaNkevin, line 109: `tour` is registered here, but `packages/react/src/styled-system/generated/recipes.gen.ts` has zero Tour entries — no `TourSlot`, no `TourVariant`, nothing added to `ConfigSlotRecipes`.

That means `SlotRecipeProps<"tour">` resolves to nothing, so consumers get no variant typing for the component. Looks like the file just wasn't regenerated after the recipe landed.

packages/react/src/theme/recipes/tour.ts
```diff
@@ -0,0 +1,105 @@
...
+      _open: {
+        animationName: "fade-in",
+        animationDuration: "slow",
+      },
+      _closed: {
+        animationName: "fade-out",
+        animationDuration: "moderate",
+      },
+    },
+    content: {
```
kalisaNkevin, line 41: `content` is a fixed `width: "sm"` with `p: "6"` and no overflow handling, so a step with a long description grows the popover instead of scrolling inside it — and the arrow is positioned against that box, so it drifts as the box grows.

Worth comparing to how [moduix handles this](https://moduix.dev/docs/tour): they add a `Tour.Body` part specifically to keep long step content scrollable "without clipping the arrow or action row". Either a `body` part or a `maxHeight` + `overflowY: auto` on the text region would cover it.

apps/www/content/docs/components/tour.mdx
```diff
@@ -0,0 +1,217 @@
...
+## Usage
+
+```jsx
+import { Tour, useTour } from "@chakra-ui/react"
+```
+
+```jsx
+const tour = useTour({ steps })
+
+<Tour.Root tour={tour}>
```
kalisaNkevin, line 22: None of the 12 examples or this anatomy use `Portal`. The recipe leans on `position: fixed` for the positioner, which breaks inside any ancestor with a `transform`, `filter`, or `overflow: hidden` — those create a containing block and the "fixed" overlay gets clipped to it.

That's a real risk for a component whose whole job is anchoring to arbitrary elements anywhere on the page. moduix portals the overlay subtree by default for this reason. Worth at least showing `Portal` in the anatomy so the common case is the safe one.

packages/react/src/components/tour/tour.tsx
```diff
@@ -0,0 +1,227 @@
...
+  extends Omit<TourActionTriggerProps, "action"> {}
+
+export const TourActionTriggers = (props: TourActionTriggersProps) => {
+  const api = useTourContext()
+  const actions = api.step?.actions ?? []
+
+  return (
+    <For each={actions}>
+      {(action) => (
+        <TourActionTrigger key={action.label} action={action} {...props}>
```
kalisaNkevin: `key={action.label}` collides if a step has two actions with the same label (two "Skip"s, or a localized string that repeats). Index is safer here since the list is positional and never reordered.

Separately: Ark ships `TourActions` as a render-prop (`props.children(step?.actions ?? [])`) which isn't re-exported. The docs teach the verbose `Tour.Context` pattern as the escape hatch, but `Tour.Actions` is the natural middle ground between that and this component. Might be worth surfacing both.

packages/react/src/theme/recipes/tour.ts
```diff
@@ -0,0 +1,105 @@
+import { tourAnatomy } from "../../anatomy"
+import { defineSlotRecipe } from "../../styled-system"
+
+export const tourSlotRecipe = defineSlotRecipe({
+  slots: tourAnatomy.keys(),
+  className: "tour",
```
kalisaNkevin: Nit, take it or leave it: 67 of 76 recipes use the `chakra-` prefix. There's precedent for going bare (`date-picker`, `splitter`, `fieldset`, `tags-input`), so this isn't wrong — just flagging since the majority convention is `chakra-tour`.

kalisaNkevin (approved):
GTM

PR #10480, opened 2025-11-18
Stub for feature to Auto generate ai components

Jchang4:
it would be great to auto generate ui components for users that fits their theme/style preferences.
Let's build that!


Jchang4:
sry i keep putting htis in the wrong place

PR #10492, opened 2025-11-22
fix(dialog): fix backdrop issue when opening concurrently

manuuuskp:
Closes # 

## 📝 Description

When opening the dialog multiple times, the backdrop incorrectly appears above the dialog content on the second and subsequent opens, making the dialog content go behind the backdrop.

Root Cause:

The issue occurs due to different unmount behaviors between dialog components:

Backdrop - Fully unmounts when the dialog closes (removes from DOM)
Positioner & Content - Stay mounted but hidden when the dialog closes (remain in DOM)

DOM ordering issue explained:

on Dialog First open:

<Portal>
  <Backdrop />      
  <Positioner>      
    <Content />
  </Positioner>
</Portal>

After close:

<Portal>
  
  <Positioner>      
    <Content />
  </Positioner>
</Portal>

on Dialog Second open:

<Portal>
  <Positioner>      
    <Content />
  </Positioner>
  <Backdrop />      
</Portal>

Since all three elements had the same z-index, the browser uses DOM order for stacking. The backdrop, now appearing last in the DOM, renders on top.

## 🚀 New behavior

Ensuring proper z-index layering regardless of DOM order:

Backdrop: z-index: calc(var(--dialog-z-index) + var(--layer-index, 0) - 1)
Positioner: z-index: calc(var(--dialog-z-index) + var(--layer-index, 0))
Content: Remove explicit z-index (uses natural stacking within positioner)

This guarantees the backdrop is always behind the dialog content, independent of DOM order or mount/unmount behavior.

## 📝 Additional Information

https://github.com/user-attachments/assets/cd6b4294-1b9a-4b93-8c70-d63b17e48a5d


segunadebayo:
Thanks for looking into this @manuuuskp. That was a clever fix.

I managed to fix the root cause, so the content unmounts as intended.

Thanks again.

PR #10531, opened 2025-12-11
feat(slider): add markerLabel to anatomy

### LINKED ISSUE #10514: Slider marker label is not themeable
### Description

The label for a slider marker is just a span with a classname: https://github.com/chakra-ui/chakra-ui/blob/5db1102559e34c2748c16b5e97f9f19df9436d9e/packages/react/src/components/slider/slider.tsx#L198

IMO, the marker label should be part of the component's anatomy, so we can set eg. `textOverflow` on it.

### Link to Reproduction

-

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Chakra UI Version

3.x.x

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

JatinMehta007:
Closes #10514

## Description

Added `markerLabel` to the `sliderAnatomy` and updated the Slider component to use `<chakra.span>` for the marker label. This allows the label to be styled via the theme system.

##  Current behavior (updates)

Currently, the slider marker label is a plain HTML `<span>` with a hardcoded class name (`chakra-slider__marker-label`). Because it is not part of the component anatomy or a chakra element, it ignores theme styles and cannot be customized via the theme config.

##  New behavior

* Added `markerLabel` to the `sliderAnatomy` definition.
* Updated the `Slider` component to render the label as a `<chakra.span>` instead of a plain `<span>`.
* Connected the label to the theme system using `__css={styles.markerLabel}`.

This allows users to style the marker label (e.g., adding `textOverflow`, changing colors, or adjusting font size) directly through the theme.

##  Is this a breaking change (Yes/No):

No

## Additional Information

Verified the changes in Storybook using the "With Marks" story. The existing class name `chakra-slider__marker-label` was preserved for backward compatibility.

### DIFF
--- a/.changeset/polite-lemons-move.md
+++ b/.changeset/polite-lemons-move.md
@@ -1,0 +1,5 @@
+---
+"@chakra-ui/react": minor
+---
+
+feat: add markerLabel to slider anatomy
--- a/packages/react/src/anatomy.ts
+++ b/packages/react/src/anatomy.ts
@@ -153,7 +153,10 @@ export const comboboxAnatomy = arkComboboxAnatomy.extendWith(
   "empty",
 )
 
-export const sliderAnatomy = arkSliderAnatomy.extendWith("markerIndicator")
+export const sliderAnatomy = arkSliderAnatomy.extendWith(
+  "markerIndicator",
+  "markerLabel",
+)
 
 export const statAnatomy = createAnatomy("stat").parts(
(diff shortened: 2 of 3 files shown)


segunadebayo:
Thanks for this PR. Worked on it already

PR #10565, opened 2026-01-05
fix(docs): show a correct JSX for Next.js example of the image-component

yannis-schmiedel:
Closes # 

## 📝 Description
fix a type in the docs of the image components:
- removed the ";" before the chakra-image-component, which does not belog into the xml

there is an additional semicolon

## 🚀 New behavior

i removed the semicolon

No


yannis-schmiedel:
Instead of only removing the missplaced semicolon, i deicided to include a proper demo component like everywhere else in the docs

PR #10638, opened 2026-02-12
feat: improve createContext types

isBatak:
Closes # 

## 📝 Description

Fix the return type of `createContext` when `strict` is set to `false` so that the generated hook correctly includes `undefined` in its return type.

Currently, when using `createContext({ strict: false })`, the return type of `useContext()` does not include `undefined`, even though the `context` can be `undefined` at runtime.

Even though `context` can be `undefined`, its type does not include `undefined`. As a result, when using it conditionally:
```
disabled={context?.disabled}
```
ESLint auto-fix removes the optional chaining (?.) because the type system indicates that `context` is always defined.

## 🚀 New behavior

When strict is false, the return type of the generated hook becomes:
```
T | undefined
```
This correctly reflects runtime behavior and allows safe conditional usage like:
```
disabled={context?.disabled}
```
TypeScript and ESLint will no longer incorrectly treat the value as always defined.

No.

This change only improves the TypeScript return type to better reflect the existing runtime behavior when strict is false.

## 📝 Additional Information


Adebesin-Cell (approved):
Looks good, thanks!

PR #10660, opened 2026-02-24
fix(theme): Add cursor style for segment group component

gregthebusker:
Closes # 

## 📝 Description

Added cursor style for segment group component

https://github.com/user-attachments/assets/c1832123-9ab9-4a70-b3d5-21c3d7e3eda8

## 🚀 New behavior

https://github.com/user-attachments/assets/595e8280-d825-4e7f-94d2-8d77592891f8

No


segunadebayo:
Kindly change this in your userland instead. We don't want to add this right now.

Thanks for understanding

PR #10727, opened 2026-03-23
feat(hooks): useMediaQuery migrate to syncExternalStore

isBatak:
## 📝 Description

Migrate `useMediaQuery` from `useState` + `useEffect` to `useSyncExternalStore`.

Media query values are set via `useEffect`, causing a hydration flicker (SSR fallback → real value) and potential tearing in concurrent mode.

## 🚀 New behavior
No hydration flicker — `useSyncExternalStore` provides the correct value synchronously on the client
No tearing in concurrent mode
`ssr` option kept for backward compatibility but is now a no-op
Dropped legacy `addListener`/`removeListener` fallback (Safari <14)

No

## 📝 Additional Information

N/A

TODO:
- [ ] Test on a large real-world project to validate hydration behavior end-to-end

### DIFF
--- a/.changeset/migrate-use-media-query-sync-external-store.md
+++ b/.changeset/migrate-use-media-query-sync-external-store.md
@@ -1,0 +1,12 @@
+---
+"@chakra-ui/react": minor
+---
+
+Migrate `useMediaQuery` hook from `useState + useEffect` to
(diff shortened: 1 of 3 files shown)


segunadebayo:
Feel free to reopen this when you have a working solution

PR #10741, opened 2026-03-29
fix(docs): set the aria-label for the extra options menu of the copy page widget

joaopalmeiro:
Related to https://github.com/chakra-ui/ark/issues/3835

## 📝 Description

This PR sets the `aria-label` for the extra options menu next to the "Copy Page" (as Markdown) button. This way, screen reader users can learn more about the menu without having to open and navigate through it. 

The PR was tested using VoiceOver + Safari and VoiceOver + Chrome.

The extra options menu next to the "Copy Page" button on the documentation pages is announced only as a generic menu.

## 🚀 New behavior

The menu is now introduced as "More Markdown options":

https://github.com/user-attachments/assets/b3841bf7-a59d-4717-a5ad-1c975e61c5c8

No

### DIFF
--- a/apps/www/components/llms-copy-widget.tsx
+++ b/apps/www/components/llms-copy-widget.tsx
@@ -30,7 +30,11 @@ export const LLMSCopyWidget = (props: { data: Docs }) => {
         mx="-1px"
         zIndex="1"
       />
-      <ActionMenu data={data} borderStartRadius="0">
+      <ActionMenu
+        data={data}
+        borderStartRadius="0"
+        aria-label="More Markdown options"
+      >
         <LuChevronDown />
       </ActionMenu>
     </ButtonGroup>


joaopalmeiro:
> Clean fix - an icon-only button with no label is invisible to screen readers. One thought: since this menu's options include both Markdown and plain text formats, the label "More Markdown options" might be slightly misleading. Something like "More copy format options" or "Additional export options" could be more accurate. Minor nitpick on an otherwise solid one-liner.

Thanks for the feedback, @HardikDewra!

I opted for this option to stay consistent with what was merged in the Ark UI repo: https://github.com/chakra-ui/ark/pull/3837

I understand your concerns, but I feel that mentioning "copy" (nothing is copied to the clipboard) or "export" (nothing is downloaded) might be more misleading, given that this menu is mostly for opening certain links, some of them with URL parameters (and all of them based on the Markdown version of the page). 

Either way, I don't have a strong opinion, let me know if I should change the label, please. And thanks for sharing your thoughts. I think these labels are really important and far from being nitpicks! 😄

PR #10779, opened 2026-04-17
fix(compositions): add aria-label to icon-only carousel and pagination buttons

akinshaywai:
## Summary

All `Carousel.PrevTrigger` / `Carousel.NextTrigger` examples across the compositions library were rendering icon-only `IconButton` components **without an accessible name**. Screen readers would announce these as generic "button" elements, giving keyboard/AT users no indication of their purpose.

The same issue existed in `pagination-with-count-text.tsx`.

### Changes

- **17 carousel example files**: Added `aria-label="Previous slide"` and `aria-label="Next slide"` to every icon-only prev/next `IconButton` (including the vertical carousel which uses up/down chevrons)
- **`pagination-with-count-text.tsx`**: Added `aria-label="Previous page"` and `aria-label="Next page"` to the pagination icon buttons

### Why not fix this in the component defaults?

`Carousel.PrevTrigger` and `Carousel.NextTrigger` intentionally accept `asChild` for custom rendering. Adding a default `aria-label` in `defaultProps` would be overridden by users who pass their own, and the correct label text is context-dependent (e.g., could be "Previous image" or "Previous item"). Fixing the examples is the right layer — it sets a clear pattern for consumers.

### WCAG reference

WCAG 2.1 SC 4.1.2 — Name, Role, Value: interactive controls must have a programmatically determinable name.

## Test plan

- [ ] Review the diff — every modified `IconButton` that wraps a chevron/arrow icon now has `aria-label`
- [ ] Open any carousel example in a browser with VoiceOver or NVDA — prev/next buttons are now announced with their purpose

### DIFF
--- a/apps/compositions/src/examples/carousel-basic.tsx
+++ b/apps/compositions/src/examples/carousel-basic.tsx
@@ -19,7 +19,7 @@ export const CarouselBasic = () => {
 
       <Carousel.Control justifyContent="center" gap="4">
         <Carousel.PrevTrigger asChild>
-          <IconButton size="xs" variant="ghost">
+          <IconButton size="xs" variant="ghost" aria-label="Previous slide">
             <LuChevronLeft />
           </IconButton>
         </Carousel.PrevTrigger>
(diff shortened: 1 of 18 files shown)


segunadebayo:
If you check the DOM closely, you'll see that the aria-label already exists

CleanShot 2026-04-17 at 12 10 26

PR #10785, opened 2026-04-20
fix: reject invalid css unit decimals

Bojun-Vvibe:
**Repo:** chakra-ui/chakra-ui (⭐ 38000)
**Type:** bugfix
**Files changed:** 2
**Lines:** +19/-1

## What
This change fixes the `isCssUnit` helper in `packages/react/src/utils/unit.ts` so it only accepts a literal decimal point when parsing CSS length values. It also adds a focused Vitest regression test covering valid values like `1.5rem` and `100%`, plus invalid inputs such as `1a5rem`, `1-5rem`, and `12 px`.

## Why
The previous regex used `.?` instead of `\\.?`, which allowed any single character to act like a decimal separator. That meant malformed values such as `1a5rem` and `1-5rem` could be treated as valid CSS units and flow into spacing-related styling logic. Tightening the regex closes that gap with a minimal, low-risk fix and a test that prevents regressions.

## Testing
Attempted to run `pnpm vitest run packages/react/__tests__/unit.test.ts`, but `pnpm` could not execute in this sandbox because it tried to access blocked global/cache paths and then the network. As a lightweight local check, I verified the corrected regex behavior with `node`, confirming valid units pass and malformed separators are rejected.

## Risk
Low / The change narrows a utility regex to reject invalid input and adds direct regression coverage for the intended behavior.

### DIFF
--- a/packages/react/__tests__/unit.test.ts
+++ b/packages/react/__tests__/unit.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, test } from "vitest"
+import { isCssUnit } from "../src/utils/unit"
+
+describe("isCssUnit", () => {
+  test("accepts valid CSS length values", () => {
+    expect(isCssUnit("12px")).toBe(true)
+    expect(isCssUnit("1.5rem")).toBe(true)
+    expect(isCssUnit("-.5em")).toBe(true)
+    expect(isCssUnit("1e3px")).toBe(true)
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks for the fix! I've landed this directly on main with a changeset in commit 16f8329ac. Appreciate the regression test — kept it as-is. Closing in favor of that commit.

PR #10803, opened 2026-05-03
fix(www): account for banner height in docs sidebars

alstmd9902:
## 📝 Description
Fixes the docs sidebar height calculation when the top promo banner is visible.

The docs sidebars use `--content-height`, which subtracts only the sticky header height from
the viewport height.

When the promo banner is visible, it also takes up vertical space. On shorter desktop
viewports, this can make the sidebar scroll container extend past the visible viewport,
causing the bottom of the sidebar content to appear clipped.

## 🚀 New behavior
Adds a sidebar-specific height variable that accounts for the promo banner height, and applies
it only to the docs sidebars.

This keeps the existing docs content height unchanged while preventing the sidebar scroll area
from extending below the viewport when the banner is visible.

### Before
image

The sidebar scroll area extends past the visible viewport when the promo banner is visible.

### After
image

The sidebar scroll area accounts for the promo banner height.

No

## 📝 Additional Information
Tested with:
- `pnpm --dir apps/www exec eslint app/theme.ts app/docs/sidebar.tsx`
- Manually checked `/docs/get-started/installation`

### DIFF
--- a/apps/www/app/docs/sidebar.tsx
+++ b/apps/www/app/docs/sidebar.tsx
@@ -42,7 +42,7 @@ export const SidebarStart = (props: BoxProps) => {
       py="8"
       flexShrink="0"
       ref={containerRef}
-      height="var(--content-height)"
+      height="var(--sidebar-height)"
       overflowY="auto"
       overscrollBehavior="contain"
       width="16rem"
@@ -77,7 +77,7 @@ export const SidebarEnd = (props: BoxProps) => {
       pb="8"
       px="2"
       flexShrink="0"
-      height="var(--content-height)"
+      height="var(--sidebar-height)"
       overflowY="auto"
       overscrollBehavior="contain"
       width="16rem"
--- a/apps/www/app/theme.ts
+++ b/apps/www/app/theme.ts
@@ -11,8 +11,10 @@ export const system = createSystem(defaultConfig, {
   },
   globalCss: {
     ":root": {
+      "--banner-height": "44px",
       "--header-height": { base: "64px", md: "104px" },
       "--content-height": "calc(100dvh - var(--header-height))",
+      "--sidebar-height": "calc(var(--content-height) - var(--banner-height))",
     },
   },
 })


kalisaNkevin (commented):
> Nice fix and great screenshots. The intent is clear. A couple of things to discuss before merging:
>
> 1. `--banner-height: 44px` is a hardcoded magic number. Is the banner always exactly 44px? What happens on narrow viewports where it might wrap? Consider a comment linking it to the banner component, or setting it dynamically.
> 2. Is the banner always visible? If it can be dismissed or conditionally hidden, subtracting its height unconditionally would make the sidebar too short when the banner isn't shown.

Adebesin-Cell (changes requested):
Nice catch 👏 

That said, I'm not sold on hardcoding `--banner-height: 44px`. A couple of things bug me:

- The banner height isn't actually fixed in `pro-banner.tsx`. It lands around ~44px on desktop, but on mobile the text has `maxW="32ch"` and `textAlign="start"`, so it wraps to two lines, and you're closer to ~70px. So `44px` quietly under-subtracts on mobile (which I think is what @kalisaNkevin was getting at). Any copy change or swapping in `BlackFridaySaleBanner` silently breaks the calc again.

- Even on desktop, we're subtracting `44px` unconditionally. Once you scroll the banner out, the sidebar still loses those `44px,` and you end up with a permanent gap at the bottom of the viewport. The banner only really costs layout while it's visible.

I think the cleaner move is to measure it instead:

* small `ResizeObserver` on the banner
* write the actual height to `--banner-height` on `:root.`

That handles wrapping, breakpoints, copy changes, etc., for free.

An alternative would be making the banner sticky/fixed so it behaves like permanent chrome alongside the header, and then the existing `--content-height` math works without introducing a second variable.

What do you think?

segunadebayo:
Going to skip this PR for now. Thanks for the effort

alstmd9902:
Thanks a lot for the thoughtful feedback and review.

You’re absolutely right that hardcoding `--banner-height` is brittle once wrapping, conditional rendering, and scroll
visibility come into play.

While working on this, I explored a few different approaches, including CSS-only layout changes, passing layout values down, and measuring the banner with `ResizeObserver`. Each approach addressed part of the problem, but I wasn’t able to land on a solution that felt clean and robust enough for the docs layout as a whole.

Your feedback helped clarify the tradeoffs here, especially around sticky positioning and viewport-based sizing. I learned a lot from the discussion and really appreciate the time you took to review this.

Thanks again.

PR #10841, opened 2026-06-05
fix(build): resolve windows path and import issues across scripts and tests

### LINKED ISSUE #10840: Build and tests fail on Windows due to script path resolution errors
### Description

When I try to build the repository locally on Windows using  `pnpm build` , I expected the build to complete successfully, but it crashed immediately with  `ERR\_UNSUPPORTED\_ESM\_URL\_SCHEME`  and module resolution errors instead.

### Link to Reproduction

N/A (This is a local build environment bug on Windows, not a runtime component bug, so it cannot be reproduced in Stackblitz).

### Steps to reproduce

1. Clone the repository on a Windows OS.
2. Run  `pnpm install`
3. Run  `pnpm build`  or  `pnpm test`
4. See build script failures.

### Chakra UI Version

3.x (main branch)

### Browser

N/A

### Operating System

- [ ] macOS
- [x] Windows
- [ ] Linux

### Additional Information

There are three specific Windows-environment bugs happening in the repository's scripts:

1. `scripts/build/main.ts`  and  `scripts/build/config.ts`  are doing dynamic `import(resolve("package.json"))` , which fails on Windows because the ESM loader expects a `file://`  protocol, not  `C:\` .
2. `packages/panda-preset/scripts/sync.ts`  uses Node's  `relative()`  path method, which outputs backslashes ( `\` ) on Windows. This generates broken typescript files (e.g.  `import { x } from ".. \def"` ).
3. The test suite in  `packages/cli/**tests**/resolve-tsconfig.test.ts`  uses manual path concatenations with backslashes, causing 5 tests to fail against standardized forward-slash outputs.

bijanmurmu:
Closes #10840 

## 📝 Description

This PR fixes multiple Windows-specific path resolution issues that cause  `pnpm build` ,  `pnpm theme:eject` , and `pnpm test`  to crash immediately for Windows developers.

1. `scripts/build/main.ts`  and  `scripts/build/config.ts`  fail with an  `ERR\_UNSUPPORTED\_ESM\_URL\_SCHEME`  because dynamic  `import(resolve("package.json"))`  does not handle Windows absolute drive paths (e.g.,  `C:...` ) correctly.
2. `packages/panda-preset/scripts/sync.ts`  generates invalid TypeScript imports (like  `import { defineSlotRecipe } from "..\def"` ) because  `node:path` 's  `relative()`  returns backslashes on Windows.
3. `packages/cli/**tests**/resolve-tsconfig.test.ts`  fails 5 assertions on Windows because the expected paths use backslashes ( `\` ) while  tsconfck  returns forward slashes ( `/` ).

## 🚀 New behavior

1. Changed dynamic  `import`  to  `JSON.parse(readFileSync(...))`  in build scripts, which is cross-platform safe and guarantees  `dependencies`  are extracted properly for rollup.
2. Added  `jsx: "automatic"`  to  `esbuild`  config so it correctly transpiles JSX without falling back to classic mode.
3. Normalizes all backslashes to forward slashes ( `.replaceAll("\\", "/")` ) during code generation in  `sync.ts` and inside the CLI test assertions.

No.

## 📝 Additional Information
With these fixes, all 972 automated tests in the repository now pass perfectly on Windows.

### DIFF
--- a/packages/cli/__tests__/resolve-tsconfig.test.ts
+++ b/packages/cli/__tests__/resolve-tsconfig.test.ts
@@ -25,7 +25,9 @@ describe("resolveTsconfig", () => {
       join(testDir, "src/index.ts"),
       tsconfigPath,
     )
-    expect(result).toBe(tsconfigPath)
+    expect(result?.replaceAll("\\", "/")).toBe(
+      tsconfigPath?.replaceAll("\\", "/"),
+    )
   })
 
   it("finds tsconfig.json for a source file", async () => {
@@ -42,7 +44,9 @@ describe("resolveTsconfig", () => {
     writeFileSync(join(testDir, "src/index.ts"), "export default {}")
 
     const result = await resolveTsconfig(join(testDir, "src/index.ts"))
-    expect(result).toBe(tsconfigPath)
+    expect(result?.replaceAll("\\", "/")).toBe(
+      tsconfigPath?.replaceAll("\\", "/"),
+    )
   })
 
   it("resolves solution-style tsconfig with references", async () => {
@@ -82,7 +86,9 @@ describe("resolveTsconfig", () => {
     writeFileSync(join(testDir, "src/index.ts"), "export default {}")
 
     const result = await resolveTsconfig(join(testDir, "src/index.ts"))
-    expect(result).toBe(appTsconfigPath)
+    expect(result?.replaceAll("\\", "/")).toBe(
+      appTsconfigPath?.replaceAll("\\", "/"),
+    )
   })
 
   it("falls back to first reference when no paths found", async () => {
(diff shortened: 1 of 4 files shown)


segunadebayo:
Thanks for the PR and for the investigation in #10840. We landed a fix on `main` (39ef8b542) that takes a slightly different approach, so I'm closing this one:

- Reading `package.json` with `readFileSync` via a shared helper instead of dynamic `import()`
- Normalizing paths inside `resolveTsconfig` so the tests pass unchanged
- The `jsx: "automatic"` flag turned out to be unnecessary: `tsconfig.build.json` already sets `jsx: react-jsx` and the esbuild plugin picks it up per file

Appreciate the detailed report, it made the root cause easy to confirm.

PR #10865, opened 2026-06-21
feat(field): horizontal field layout improvements

isBatak:
## 📝 Description

Implement horizontal field layout with proper label-to-input alignment using CSS Grid.

Storybook:
image

Docs:
image

Horizontal field layouts had helper and error text appearing on the same line as the label and input, making it difficult to manage multiple rows of information.

## 🚀 New behavior

- Horizontal layout now uses CSS Grid with two columns: fixed label width + flexible input area
- New part `<Field.InputElement>` is added, it's optional and used just for styling purposes
- Label centers vertically to the input element
- Input/select/textarea/switch elements align to the end of their column
- Helper text and error text stack below the input in the second column

No

## 📝 Additional Information

**What changed:**
- Tried the `<Field.Group>` approach that we discussed internally, but it didn't work because there was no simple way to align the label with the input.
- Ditched the group approach and implemented the `InputElement` slot that wraps the input and sets the root to a grid layout. This solved all alignment issues.
- Label uses `alignSelf: center` to vertically center to input height, input elements use `justifySelf: end` for right alignment (useful for switches), and helper/error text flows naturally in the second column based on HTML order.


packages/react/src/components/field/field.tsx
```diff
@@ -61,6 +61,16 @@ export const FieldLabel = withContext<HTMLLabelElement, FieldLabelProps>(
 
 ////////////////////////////////////////////////////////////////////////////////////
 
+export interface FieldInputElementProps
+  extends HTMLChakraProps<"input">, UnstyledProp {}
+
+export const FieldInputElement = withContext<
+  HTMLDivElement,
+  FieldInputElementProps
+>("input", "inputElement")
```
Adebesin-Cell: This is the type mismatch I mentioned, the ref is typed `HTMLDivElement`, but the props and base tag are `"input"`, so they don't line up. And in practice this is always an `asChild` grid wrapper (it wraps `Switch.Root` too, which is a div), so it isn't really an input. A neutral `div` keeps the ref/tag/props consistent and matches how it's used:

```suggestion
export interface FieldInputElementProps
  extends HTMLChakraProps<"div">, UnstyledProp {}

export const FieldInputElement = withContext<
  HTMLDivElement,
  FieldInputElementProps
>("div", "inputElement")
```

Adebesin-Cell (commented):
Hey Ivica, here's what I meant.

packages/react/src/theme/recipes/field.ts
```diff
@@ -48,15 +48,30 @@ export const fieldSlotRecipe = defineSlotRecipe({
           flexDirection: "column",
           alignItems: "flex-start",
         },
+        group: {
+          flexDirection: "row",
+        },
```
Adebesin-Cell, line 53: The `group` here doesn't have a matching slot in the field anatomy, so this block never renders.

```suggestion
```

packages/react/src/styled-system/generated/system.gen.ts
```diff
@@ -227,6 +227,7 @@ export interface SystemProperties {
   containerType?: ConditionalValue<CssProperties["containerType"] | undefined | AnyString> | undefined
   content?: ConditionalValue<CssProperties["content"] | undefined | AnyString> | undefined
   contentVisibility?: ConditionalValue<CssProperties["contentVisibility"] | undefined | AnyString> | undefined
+  // cornerShape?: ConditionalValue<CssProperties["cornerShape"] | undefined | AnyString> | undefined
```
Adebesin-Cell, line 230: Could we also remove this?

isBatak: I explained this problem here https://chakraui.slack.com/archives/C049UEYUK7S/p1782042834416699?thread_ts=1781767849.968099&cid=C049UEYUK7S

If I remove this completely, it will appear again on the next code gen. If I uncomment it, we will have TS issue. 
I think we need to contribute to the `csstype` and add `cornerShape` type definition.

Adebesin-Cell: How did I miss that? 🤦

Adebesin-Cell (approved):

packages/react/src/styled-system/generated/system.gen.ts
isBatak: I've figured it out and added local CornerShape augmentation for now
https://github.com/chakra-ui/chakra-ui/pull/10865/changes/dabe8ae813d301b800a602446642570074c6e152

Maybe we should consider migrating to https://github.com/dev-five-git/csstype-extra or do the same auto-gen ourselves

>Automatically generated, up-to-date TypeScript definitions for CSS based on Mozilla's MDN data. Inspired by [csstype](https://github.com/frenic/csstype)

PR #10877, opened 2026-07-01
feat: add DateInput component

kalisaNkevin:
Closes # 

## 📝 Description

Added Date Input component

> Add a brief description

## 🚀 New behavior


BraianS:
Hello. Any update?

packages/react/src/theme/recipes/date-input.ts
```diff
@@ -0,0 +1,179 @@
...
+      textStyle: "label",
+    },
+
+    control: {
+      display: "flex",
+      alignItems: "center",
+    },
+
+    segmentGroup: {
+      width: "100%",
```
Adebesin-Cell, line 28: Every example sets `maxWidth="14rem"` because the recipe forces the control full-width here (plus `root: width: full` above), then the docs rein it back in. The range example even adds `maxW="fit-content"` to each `Segments` to undo it — that's the recipe fighting itself.

A date input is segmented, so its natural width is its content (`12 / 31 / 2026`), not the container. The precedent is `PinInput`, the other segmented control:

```ts
control: { display: "inline-flex", gap: "2", isolation: "isolate" } // no width
```

It hugs its content, so its examples set no width at all. Text-field inputs (`NumberInput`, `DatePicker`) stretch on purpose — DateInput isn't in that family.

Suggest:

```ts
root: {
  // width: "full",  ← drop
},
segmentGroup: {
  width: "fit-content", // was "100%"
},
```

Then every `maxWidth="14rem"` comes out of the examples, the range hacks go away, and the DatePicker composition still stretches because it opts in with `flex="1"` on the control (flex beats width). Anyone who wants full-width adds `width="full"`, same as PinInput. We don't pick the layout for them, we just don't force one.

packages/react/src/theme/recipes/date-input.ts
```diff
@@ -0,0 +1,179 @@
...
+          },
+        },
+      },
+
+      subtle: {
+        segmentGroup: {
+          borderWidth: "1px",
+          borderColor: "transparent",
+          bg: "bg.muted",
+          focusVisibleRing: "inside",
```
Adebesin-Cell, line 138: The focus-ring fix only landed on `outline`. The changeset explains why the group needs plain `focusRing`: focus lands on the child segments and reflects via `data-focus`, so `:focus-visible` never matches the group. From the utilities:

```ts
focusRing:        "&:is(:focus, [data-focus])"                   // matches ✅
focusVisibleRing: "&:is(:focus-visible, [data-focus-visible])"   // matches neither ❌
```

So `subtle` here (`focusVisibleRing: "inside"`) and `flushed` below (`_focusVisible`) still won't show a ring, same bug the changeset describes, just on the other two variants. They need `focusRing`/`_focus` too.

packages/react/src/styled-system/generated/recipes.gen.ts
```diff
@@ -567,6 +567,25 @@ export type DataListVariantMap = {
   [K in keyof DataListVariant]: Array<DataListVariant[K]>
 }
 
+// DateInput
+
+export type DateInputSlot = "root" | "label" | "control" | "segmentGroup" | "segment" | "hiddenInput"
```
Adebesin-Cell, line 572: This file has changes beyond DateInput. It drops `label | control | input | trigger | clearTrigger | positioner` from `DatePickerSlot` and adds a `FloatingPanel` recipe, neither is on `main`. Looks like it was regenerated from a branch that had drifted, and as-is it'll break DatePicker's types.

Could you rebase on `main` and re-run the recipe generation so the diff here is only the DateInput additions?

.changeset/date-input-control-focus-ring.md
```diff
@@ -0,0 +1,9 @@
+---
+"@chakra-ui/react": patch
+---
+
+Fix `DateInput` control focus ring: the `control` slot isn't natively focusable
```
Adebesin-Cell, line 5: DateInput is new in this PR, so a separate patch entry for "fix the focus ring" reads oddly in the changelog, there's nothing shipped to fix yet. Fold this into `add-date-input-component.md` and drop this file.

Adebesin-Cell (commented):
Looks good! 

A few things to tighten before merge. The width one is the main one: it removes the repeated `maxWidth` from the examples by letting the control hug its content instead of us forcing a layout.

packages/react/src/theme/recipes/date-input.ts
kalisaNkevin: Done. Dropped `width: full` from `root` and switched `segmentGroup` to `width: fit-content`.

One addition to your sketch: `control` needs `fit-content` too. `root` is a flex column, so its default `align-items: stretch` stretches `control` back to full width even once the segment group hugs. That bit the clear-trigger example, where `marginStart="auto"` threw the X button to the far right of the page. Removed that prop and gave the control an explicit `gap="2"` instead.

All the `maxWidth="14rem"` props are out, and so are the range `maxW="fit-content"` hacks. Two I left in place:

- `with-hook-form`: the `maxW` is on the `Stack` and constrains the plain `Input` sitting next to it, nothing to do with DateInput.
- `with-date-picker`: it's on `DatePicker.Root`, which sets `width: full` in its own recipe. I also had to push the opt-in one level down to `<DateInput.Segments flex="1" />`, since `DatePicker.IndicatorGroup` is absolutely positioned against `DatePicker.Control` and the calendar icon detaches from the field if the bordered box doesn't stretch.

packages/react/src/theme/recipes/date-input.ts
kalisaNkevin: Good catch, fixed both. `subtle` uses `focusRing: "inside"` and `flushed` uses `_focus` now.

packages/react/src/styled-system/generated/recipes.gen.ts
kalisaNkevin: I don't think either half of this holds up, so I've left the file as-is.

The `DatePickerSlot` entries aren't removals, they're duplicates. `label`, `control`, `input`, `trigger`, `clearTrigger` and `positioner` all still appear earlier in the same union, `main` just lists them twice. `@zag-js/date-picker@1.41.2` declares all six in its base `parts()`, and our `arkDatePickerAnatomy.extendWith("indicatorGroup")` is unchanged from `main`. Duplicate union members are a no-op, so DatePicker's types aren't affected either way.

FloatingPanel is `main`'s generated file being stale rather than drift on this branch. `main` already has `theme/recipes/floating-panel.ts` registered in `slot-recipes.ts`, but its `recipes.gen.ts` contains no `FloatingPanel` at all, so regenerating on a clean `main` produces the same additions. There's also no `package.json` diff against `main`, so no Ark bump is in play here.

.changeset/date-input-control-focus-ring.md
kalisaNkevin: Dropped the file. I didn't fold anything into `add-date-input-component.md`, for the reason you gave: DateInput is new in this PR, so there's no released focus-ring behaviour to describe. The existing minor entry covers it.

packages/react/src/styled-system/generated/recipes.gen.ts
```diff
@@ -1559,6 +1572,31 @@ export type MarqueeVariantMap = {
   [K in keyof MarqueeVariant]: Array<MarqueeVariant[K]>
 }
 
+// FloatingPanel
+
+export type FloatingPanelSlot =
```
Adebesin-Cell, line 1577: These were the changes I meant, but it's alright 👍, I think it'd still come out in some other generation.

Adebesin-Cell (approved):

PR #10883, opened 2026-07-07
feat(cmdk): implementation of command palette

kalisaNkevin:
Closes # 

## 📝 Description

Implementation of Command Palette component

> Add a brief description

## 🚀 New behavior

## 📝 Additional Information
Screenshot 2026-07-09 at 12 28 05
Screenshot 2026-07-09 at 12 27 36
Screenshot 2026-07-09 at 12 29 03
Screenshot 2026-07-09 at 12 28 56
Screenshot 2026-07-09 at 12 28 47
Screenshot 2026-07-09 at 12 30 20
Screenshot 2026-07-09 at 12 30 13
Screenshot 2026-07-09 at 12 29 56
Screenshot 2026-07-09 at 12 29 48
Screenshot 2026-07-09 at 12 28 40
Screenshot 2026-07-09 at 12 31 14
Screenshot 2026-07-09 at 12 31 01
Screenshot 2026-07-09 at 12 30 52
Screenshot 2026-07-09 at 12 30 46
Screenshot 2026-07-09 at 12 30 39
Screenshot 2026-07-09 at 12 30 30
Screenshot 2026-07-09 at 12 30 25
Screenshot 2026-07-09 at 12 29 41
Screenshot 2026-07-09 at 12 29 33
Screenshot 2026-07-09 at 12 29 24
Screenshot 2026-07-09 at 12 29 13
Screenshot 2026-07-09 at 12 28 33
Screenshot 2026-07-09 at 12 28 27
Screenshot 2026-07-09 at 12 28 20
Screenshot 2026-07-09 at 12 28 11


jayalfredprufrock:
Is this likely to get merged at some point? 

I know the Listbox docs have a dialog/command palette example so I'm curious what this new component brings to the table besides OOTB styling vs composing the Listbox manually. Regardless having all the different examples in the docs is awesome so I'd love to see this get added. Thanks!

packages/react/src/components/command-palette/command-palette.tsx
```diff
@@ -0,0 +1,385 @@
...
+  // zag's listbox wires the label to the list but not to the input
+  // (unlike combobox), so the input would have no accessible name.
+  const listbox = useListboxContext()
+  const labelId = listbox.getLabelProps().id
+
+  // Clear the input on Escape before any dismissable layer (dialog, drawer,
+  // popover) handles the key. Those layers listen on the document in the
+  // capture phase, so a window capture listener is the only spot that runs
+  // earlier — a bubble handler on the input never gets the chance.
+  useEffect(() => {
```
Adebesin-Cell: This is patching gaps in Listbox's input (value control, Escape, accessible name) from userland. I opened chakra-ui/zag#3256 to fix it upstream, once it lands, this effect and the `setInputValue` hack can go. A `// TODO(zag#3256)` here for now.

apps/www/components/site/hero.section.tsx
```diff
@@ -147,9 +147,9 @@ export const HeroSection = () => (
     <Container>
       <Stack gap={{ base: "5", md: "10" }} mb="20">
         <Announcement alignSelf="flex-start" asChild>
-          <Link href="/shipped">
+          <Link href="/docs/components/command-palette">
```
Adebesin-Cell: This repoints the homepage announcement to CommandPalette. That pill is a release-level decision, so let's keep it out of a component PR, mind reverting this file?

apps/compositions/src/examples/command-palette-within-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [open, setOpen] = useState(false)
+
+  const { contains } = useFilter({ sensitivity: "base" })
+
+  const { collection, filter } = useListCollection({
+    initialItems: commands,
+    filter: contains,
+  })
+
+  useEffect(() => {
```
Adebesin-Cell: Every overlay example re-hand-rolls this ⌘K listener. Worth shipping a small hotkey helper (or one reusable snippet) so it's not re-taught each time. Not blocking.

Adebesin-Cell (commented):
Nice! Great work.

packages/react/src/styled-system/generated/recipes.gen.ts
```diff
@@ -1184,6 +1178,37 @@ export type ComboboxVariantMap = {
   [K in keyof ComboboxVariant]: Array<ComboboxVariant[K]>
 }
 
+// CommandPalette
+
+export type CommandPaletteSlot =
```
Adebesin-Cell, line 1202: While regenerating: this file also drops DatePicker's slots + adds an unrelated FloatingPanel block from being behind `main`. A rebase + regenerate should clear it.

apps/www/content/docs/get-started/migration.mdx
```diff
@@ -324,7 +324,7 @@ Before:
 ```tsx
 import { Icon } from "@chakra-ui/react"
 
-<Icon viewBox="0 0 24 24" color="red.500" boxSize={6}>
+;<Icon viewBox="0 0 24 24" color="red.500" boxSize={6}>
```
Adebesin-Cell: This (and the stack-separator changeset) is unrelated Prettier churn from a global format run — worth reverting to keep the diff command-palette-only.

apps/www/components/site/hero.section.tsx
kalisaNkevin: Reverted, the file is back to `main`'s `/shipped` pill. Agreed on the principle.

packages/react/src/components/command-palette/command-palette.tsx
kalisaNkevin: Added a `TODO(zag#3256)` above the block, covering the accessible name, the Escape handling and the `setInputValue` helper together, since all three go away with the same fix. Thanks for opening the upstream issue.

apps/www/content/docs/get-started/migration.mdx
kalisaNkevin: Reverted both, and it turned out to be worth digging into: this wasn't a global format run.

`main`'s copy of `migration.mdx` isn't Prettier-clean, and `.lintstagedrc` runs `prettier --write` over every staged `.mdx`. So the moment that file is staged in any commit, the hook rewrites it and the `;<Icon` prefixes reappear. I actually hit this while fixing it here, my first revert was silently undone inside the commit and I had to redo it with `--no-verify`.

Both files are byte-identical to `main` now, and dropping the stack-separator changeset also cleared the merge conflict, since it was already deleted on `main` after release.

Might be worth formatting `migration.mdx` on `main` in its own PR so this stops ambushing whoever touches it next.

packages/react/src/styled-system/generated/recipes.gen.ts
kalisaNkevin: I don't think this one holds, so I've left the file as-is.

The `DatePickerSlot` entries aren't drops, they're duplicates. `label`, `control`, `input`, `trigger`, `clearTrigger` and `positioner` all still appear earlier in the same union; `main` just lists each of them twice. Duplicate union members are a no-op, so DatePicker's types are unaffected.

FloatingPanel is `main`'s generated file being stale rather than this branch being behind. `main` already registers `floatingPanel` in `theme/slot-recipes.ts`, but its `recipes.gen.ts` contains no `FloatingPanel` at all, so regenerating on a clean `main` produces the same block. This branch is merged up to `main` as of now and the output is unchanged.

Same thing came up on #10877, so this is the identical situation in both PRs.

apps/compositions/src/examples/command-palette-within-dialog.tsx
kalisaNkevin: Left this one for a follow-up if that's alright.

Checking the actual spread: of the 25 command-palette examples, only `command-palette-within-dialog` wires up a ⌘K listener. `within-drawer` and `within-popover` don't have one. So it's taught once rather than re-taught per overlay, and a shared helper would be a slightly bigger API question (scoping, sequences, SSR) than I'd want to fold into this PR.

Happy to open a separate issue for a `useHotkey` style helper if you think it's worth having.

kalisaNkevin:
Also reworked this to sit on **Combobox + Dialog** instead of Listbox. The goal isn't to turn Listbox into another Combobox, and looking at the machines that's exactly what the workarounds here were doing.

zag's listbox machine has no input-value concept at all — only `selectionMode` and `typeahead`. Combobox has `inputValue`, `onInputValueChange`, `allowCustomValue`, `inputBehavior` and `selectionBehavior`. That's why this PR was reaching for a native `HTMLInputElement.prototype.value` setter: there was no state to drive. Combobox also has `disableLayer` ("whether to disable registering this a dismissable layer"), which is the built-in answer to the Escape-versus-dialog fight I was solving with a window capture listener, and it ships `clearTrigger`, `control` and `list` parts plus the label→input wiring.

So all three workarounds are gone, and the `TODO(zag#3256)` with them.

The palette now owns its overlay rather than being composed into one:

```tsx
<CommandPalette.Root collection={collection} onInputValueChange={(e) => filter(e.inputValue)}>
  <CommandPalette.Trigger />
  <Portal>
    <CommandPalette.Backdrop />
    <CommandPalette.Positioner>
      <CommandPalette.Panel>
        <CommandPalette.Control>
          <CommandPalette.Indicator />
          <CommandPalette.Input />
        </CommandPalette.Control>
        <CommandPalette.List>…</CommandPalette.List>
      </CommandPalette.Panel>
    </CommandPalette.Positioner>
  </Portal>
</CommandPalette.Root>
```

`Root` binds `mod+k` by default (`hotkey={null}` opts out), which also closes out your hotkey comment — that's now one implementation rather than something each example re-teaches.

Worth flagging, since these are real behavior changes:

- **Escape closes the palette** instead of clearing the input first. The dialog owns dismissal now; `ClearTrigger` covers clearing.
- **`onSelect` gives `{ value: string[], itemValue: string }`**, where listbox gave a bare string. Call sites use `itemValue`.
- **Unselected items have no `aria-selected`** — combobox omits it where listbox emitted `"false"`.
- **`within-drawer` and `within-popover` examples are dropped**, since the palette is no longer composed into an arbitrary overlay. `within-dialog` is now just the default.
- **`Root` takes no style props** — it renders no DOM. Layout goes on `Panel`.

`selectionMode="none" | "single" | "multiple"` is unchanged, mapped onto combobox's `multiple` + `selectionBehavior`.

Adebesin-Cell (approved):
Cool!

PR #10891, opened 2026-07-12
fix: quote argument-hint YAML values so Copilot CLI ≥1.0.65 loads all skills (closes #10890)

### LINKED ISSUE #10890: Skills fail to load in GitHub Copilot CLI ≥1.0.65 due to array-shaped `argument-hint`
## Bug

`argument-hint` in the YAML frontmatter of 2 SKILL/command file(s) is declared with **bare brackets** (a YAML flow sequence), so downstream slash-command loaders that validate `argument-hint` as a **string** — notably [GitHub Copilot CLI ≥ 1.0.65](https://github.com/github/copilot-cli) — silently reject the SKILL on load. The command disappears from the CLI menu.

```yaml
# Broken (parses as an array of one string)
argument-hint: [issue-url]

# Fixed (parses as a string)
argument-hint: "[issue-url]"
```

Claude Code currently renders the accidental array by concatenation, which is why nobody caught this until Copilot CLI tightened the type check. Both [Claude Code slash-command docs](https://code.claude.com/docs/en/slash-commands) and [VS Code Agent Skills docs](https://code.visualstudio.com/docs/agent-customization/agent-skills) document `argument-hint` as a string.

## Affected files

- `.claude/commands/github-reviewer.md`
- `.claude/commands/ark.md`

## Fix

Wrap the value after `argument-hint:` in double quotes. One-character change per line, no behaviour change on Claude Code.

## PR

I'm sending a PR with the exact fix — see the linked PR from this issue.

thejesh23:
Closes #10890.

## Summary

Purely mechanical single-character transform to 2 file(s): wrap the value after `argument-hint:` in double quotes so YAML parses it as a **string** instead of a flow-sequence array. Fixes GitHub Copilot CLI ≥ 1.0.65 silently dropping the skill.

```diff
- argument-hint: [foo]
+ argument-hint: "[foo]"
```

Bare `[...]` is YAML flow-sequence syntax → the loader receives `["foo"]` (a list), fails its `str` validation, and rejects the SKILL without an error. Claude Code, VS Code Agent Skills, and every other loader document `argument-hint` as a string; the quoted-string form is what all reference docs show.

## Files (2)

- `.claude/commands/github-reviewer.md`
- `.claude/commands/ark.md`

## Verification

- YAML round-trip via `yaml.safe_load` on every changed frontmatter reports `argument-hint` as `str`
- Regression sweep: no unquoted-bracket `argument-hint:` lines remain in these files
- No vulnerable value contained `"` or `\`, so bare double-quote wrapping is safe

## Sources

- [Claude Code slash-commands / frontmatter reference](https://code.claude.com/docs/en/slash-commands) — `argument-hint` documented as a string
- [VS Code Agent Skills docs](https://code.visualstudio.com/docs/agent-customization/agent-skills) — same

### DIFF
--- a/.claude/commands/ark.md
+++ b/.claude/commands/ark.md
@@ -1,6 +1,6 @@
 ---
 description: Implement a new Ark UI component
-argument-hint: [componentName]
+argument-hint: "[componentName]"
 ---
 
 ## Context
--- a/.claude/commands/github-reviewer.md
+++ b/.claude/commands/github-reviewer.md
@@ -1,6 +1,6 @@
 ---
 description: Draft a response for a GitHub issue
-argument-hint: [issue-url]
+argument-hint: "[issue-url]"
 ---
 
 ## Context


thejesh23:
@segunadebayo Can you please review this

PR #10898, opened 2026-07-13
fix(react): remove obsolete TypeScript suppressions

wanxiankai:
Closes #10892

## 📝 Description

Removes four obsolete or incorrect TypeScript suppression comments from the React package and replaces the two real type mismatches with explicit, narrow typing.

- `AlertRoot` suppresses the mismatch between its responsive `status` prop and the alert status context.
- `useConst` initializes an untyped null ref and suppresses assignment of its generic value.
- The recipe and slot recipe context helpers retain suppressions even though their input props are now typed as `any` and compile without them.

## 🚀 New behavior

- Narrows the alert context value to `StatusProps["status"]`.
- Types the `useConst` ref as `T | null` and narrows function initializers before invoking them.
- Removes the two obsolete recipe context suppressions.
- Keeps runtime behavior and the public API unchanged.

No.

## 📝 Additional Information

Validation:

- `pnpm --filter=@chakra-ui/react typecheck`
- `pnpm exec vitest run packages/react/__tests__/recipe.test.ts packages/react/__tests__/recipe-types.test.ts packages/react/__tests__/display-name.test.ts` (29 tests passed)
- `pnpm --filter=@chakra-ui/react build:fast`
- `pnpm exec prettier --check` for all four changed files
- `pnpm lint` (0 errors; two unrelated existing warnings)
- `git diff --check`

No changeset is included because this is an internal type-safety refactor with no runtime or public API change.

### DIFF
--- a/packages/react/src/components/alert/alert.tsx
+++ b/packages/react/src/components/alert/alert.tsx
@@ -50,11 +50,9 @@ export const AlertRoot = withProvider<HTMLDivElement, AlertRootProps>(
   {
     forwardAsChild: true,
     wrapElement(element, props) {
+      const status = (props.status || "info") as StatusProps["status"]
       return (
-        // @ts-ignore fix later
-        <AlertStatusProvider value={{ status: props.status || "info" }}>
-          {element}
-        </AlertStatusProvider>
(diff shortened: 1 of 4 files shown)


segunadebayo:
Pushed a better fix for this TS issues.

PR #10899, opened 2026-07-13
fix(number-input): prevent control overflow when using NumberInput.Label

### LINKED ISSUE #10889: Use of NumberInput.Label causing control to overflow
### Description

When I tried to provide the label to `NumberInput.Label`, the expectation was to show the label appropriately along with NumberInput.Input and NumberInput.Control. but Outcome or output was the NumberInput.Control was overflowing

Image

### Link to Reproduction

https://stackblitz.com/edit/r14eagyb?file=src%2FApp.tsx&showSidebar=0

### Steps to reproduce

1. Go to stackblitz link form https://chakra-ui.com/docs/components/number-input
2. Add an `NumberInput.Label` next to Input or Control
3. Let the UI reload (HMR)
4. see the spinner or control component. its overflowing.

### Chakra UI Version

3.36.0

### Browser

Chrome 150

### Operating System

- [ ] macOS
- [x] Windows
- [ ] Linux

### Additional Information

_No response_

mdqasim786:
Closes #10889

## Description

Fixes the layout issue where `NumberInput.Control` overflows when `NumberInput.Label` is used.

## Current behavior

When a label is rendered with `NumberInput.Label`, the control is positioned relative to the root container and extends into the label area, causing an overflow.

## New behavior

The control is anchored to the input height so it no longer overlaps the label while preserving the existing layout for other examples.

## Verification

- Reproduced the issue locally
- Verified the fix in Storybook
- `pnpm test` passed
- `pnpm typecheck` passed


Adebesin-Cell:
Thanks for the fix and the clear repro, @mdqasim786 🙏

#10104 already handles this more completely — it moves the control into a `NumberInput.TriggerGroup` and fixes the label positioning there. So I'll close this in favor of that one. Thanks again!

PR #10916, opened 2026-07-27
fix(cli): replace prettier with formatly for generated file formatting

Mr-KRAMA:
Base branch: main
Head branch: fix/cli-replace-prettier-with-formatly

Summary
@chakra-ui/cli previously shipped Prettier as a hard runtime dependency and used it to format every generated type file (conditions.gen.d.ts, token.gen.d.ts, etc.). This forced Prettier into every project's install even when the project already uses Biome, dprint, or deno fmt — and it formatted generated files according to Prettier's rules rather than the project's own formatter config.

This PR removes the hard Prettier dependency and replaces it with [Formatly](https://github.com/JoshuaKGoldberg/formatly), which detects and delegates to whatever formatter the consuming project already has configured.

Root Cause
packages/cli/src/utils/pretty.ts had a static top-level import:

import { type Options, format } from "prettier"

Copy
ts
This made Prettier a mandatory install for every consumer of @chakra-ui/cli, regardless of their formatter choice. There was no fallback, no detection, and no way to opt out.

Fix
Two files changed, one new test file added.

packages/cli/src/utils/pretty.ts
- import { type Options, format } from "prettier"
-
- export function pretty(value: any, options: Options = {}) {
-   return format(value, {
-     parser: "typescript",
-     ...options,
-   })
- }

+ export async function pretty(value: string): Promise<string> {
+   try {
+     const { format } = await import("prettier")
+     return format(value, { parser: "typescript", ... })
+   } catch {
+     // prettier not installed — return raw string,
+     // formatly handles the on-disk pass in io.write
+     return value
+   }
+ }

Copy
diff
Reason: Static import replaced with a dynamic await import() inside try/catch. Prettier is now optional — if it is not installed in the consuming project, pretty() returns the raw string without throwing. The on-disk formatting pass is handled separately by formatly in io.write.

packages/cli/src/utils/io.ts
+ import { formatly } from "formatly"

  export const write = async (path, file, content) => {
+   const filePath = outPath(path, file)
    try {
-     await writeFile(outPath(path, file), await content)
+     await writeFile(filePath, await content)
+     // Delegates to the project's configured formatter.
+     // If none is detected, formatly reports ran=false silently.
+     await formatly([filePath], { cwd: process.cwd() })
    } catch (error) {
-     throw new Error(`Failed to write file ${outPath(path, file)}: ...`)
+     throw new Error(`Failed to write file ${filePath}: ...`)
    }
  }

Copy
diff
Reason: Formatly's API is file-based — it takes file paths and formats them on disk. After writing each generated file, we call formatly with that file's path. Formatly detects the project's formatter (Biome, dprint, deno fmt, Prettier) and runs it. If no formatter is detected, it returns { ran: false } silently — no error, no crash.

packages/cli/package.json
- "prettier": "3.6.2",
+ "formatly": "^0.3.0",

Copy
diff
Reason: Removes the hard Prettier dep. Formatly is the only new runtime dependency. No other packages changed.


segunadebayo:
Closing this. Formatly only formats files on disk, so it can't replace Prettier for in-memory CLI output. Thanks for the PR.

PR #10917, opened 2026-07-29
docs(compositions): add toggle-tip usage examples

Fadhilamadan:
## 📝 Description

Adds four new composition examples for the ToggleTip component: arrow, closeOnEscape, closeOnInteractOutside, and size variants. Also updates the docs page with usage sections and exports them as Storybook stories.

The toggle-tip docs page has basic usage but lacks examples for sizes, arrow, closeOnEscape, and closeOnInteractOutside props. These aren't shown in Storybook either.

## 🚀 New behavior

- New composition examples: ToggleTipArrow, ToggleTipCloseOnEscape, ToggleTipCloseOnInteractOutside, ToggleTipSizes
- Docs page now has sections for Sizes, Arrow, Close on Escape, and Close on Interact Outside
- All four examples exported as Storybook stories

No


apps/compositions/src/examples/toggle-tip-close-on-escape.tsx
```diff
@@ -0,0 +1,13 @@
+import { Flex } from "@chakra-ui/react"
+import { InfoTip } from "compositions/ui/toggle-tip"
+
+export const ToggleTipCloseOnEscape = () => {
+  return (
+    <Flex justify="center">
+      <InfoTip closeOnEscape={false}>
+        Try pressing Escape — this popover stays open because closeOnEscape is
+        set to false.
+      </InfoTip>
```
Adebesin-Cell, line 10: Let's keep the tip content short and let the prose above do the explaining; that's the pattern in every other example (`"This is the tooltip content"` in the tooltip variants, `"This is some additional information."` in toggle-tip basic). The MDX already says "Set to `false` to keep it open on Escape", so the tip is repeating it, and since `InfoTip`'s trigger is just the icon, the explanation belongs in the prose.

```suggestion
      <InfoTip
        closeOnEscape={false}
        content="This is some additional information."
      />
```

(Also uses the `content` prop so it matches `basic`/`info-tip`/`arrow`/`sizes`.)

Adebesin-Cell (commented):
Nice! Thanks for your contribution. Two things before merge:

**1. Regenerate the examples registry.** `apps/www/public/r/examples/toggle.json` and `index.json` still list only the two original examples, the four new ones aren't in them. The docs site renders fine (it imports the files directly), but that committed registry feeds the CLI/Stackblitz and is now out of sync. Could you run `pnpm www generate:examples` and commit the result?

**2. Tip copy**: left inline on the two "close on…" examples.

apps/compositions/src/examples/toggle-tip-close-on-interact-outside.tsx
```diff
@@ -0,0 +1,13 @@
+import { Flex } from "@chakra-ui/react"
+import { InfoTip } from "compositions/ui/toggle-tip"
+
+export const ToggleTipCloseOnInteractOutside = () => {
+  return (
+    <Flex justify="center">
+      <InfoTip closeOnInteractOutside={false}>
+        Click outside this popover — it stays open because
+        closeOnInteractOutside is set to false.
+      </InfoTip>
```
Adebesin-Cell, line 10: Same as the Escape example — short, neutral content, let the prose explain.

```suggestion
      <InfoTip
        closeOnInteractOutside={false}
        content="This is some additional information."
      />
```

apps/compositions/src/examples/toggle-tip-arrow.tsx
```diff
@@ -0,0 +1,13 @@
+import { Flex } from "@chakra-ui/react"
+import { InfoTip } from "compositions/ui/toggle-tip"
+
+export const ToggleTipArrow = () => {
+  return (
+    <Flex justify="center">
+      <InfoTip
+        showArrow
+        content="This toggle tip has an arrow pointing to the trigger."
+      />
```
Adebesin-Cell, line 10: Optional: milder version of the same thing, `tooltip-with-arrow` uses the neutral placeholder rather than describing the arrow in the content. Fine to leave, but for consistency:

```suggestion
      <InfoTip showArrow content="This is some additional information." />
```

PR #10931, opened 2026-08-13
fix(RadioCard): preserve outline border-width when disabled and checked

### LINKED ISSUE #10930: RadioCard outline variant visually lose border-width when disabled
### Description

When setting the `RadioCard` variant outline to disabled, I would expect the border width not to change, only the color. It is the background color on the element that has the `data-state="checked"` that seems to hide the shadow:

with background

Image

without background (not saying removing the background is the solution, just pointing out the cause of the issue)

Image

### Link to Reproduction

https://stackblitz.com/edit/hpkwxjuk?file=src%2FApp.tsx&showSidebar=0

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Chakra UI Version

3.36.1

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

waterWang:
## Summary

Fixes #10930: the RadioCard outline variant visually loses its
border-width when disabled and checked.

### Root cause

In the outline variant, the checked `item` renders a 2px visual border
composed of a real `borderWidth: 1px` plus an inner ring simulated with
`boxShadow: inset 0 0 0 1px var(--shadow-color)`.

When the RadioCard is disabled, the `itemControl` (which fills the
`item` via `flex: 1` and `rounded: inherit`) gets a `bg: bg.muted`
background from the base recipe. That opaque background paints over the
`item`'s inset box-shadow ring, so the checked border appears as thin as
1px instead of 2px — the visual border-width change described in the
issue.

### Fix

Override the disabled `itemControl` background in the outline variant
with `bg: unset` (matching the existing pattern already used by the
solid variant). The disabled state remains clearly visible through the
`item`'s existing `opacity: 0.5`, while the checked 2px outline is
preserved.

## Changes

- `packages/panda-preset/src/slot-recipes/radio-card.ts`
- Add `.changeset/radio-card-outline-disabled.md`

> Pure contribution, no bounty (chakra-ui is not a bounty project).

### DIFF
--- a/.changeset/radio-card-outline-disabled.md
+++ b/.changeset/radio-card-outline-disabled.md
@@ -1,0 +1,10 @@
+---
+"@chakra-ui/react": patch
+---
+
+- **RadioCard**: Fix `outline` variant border-width appearing reduced when
+  disabled and checked. The `itemControl`'s disabled background
+  (`bg.muted`) was covering the `item`'s checked inset box-shadow ring,
+  making the visual border appear thinner. Override the disabled
+  background in the `outline` variant to prevent the shadow from being
+  hidden.
--- a/packages/panda-preset/src/slot-recipes/radio-card.ts
+++ b/packages/panda-preset/src/slot-recipes/radio-card.ts
@@ -217,6 +217,11 @@ export const radioCardSlotRecipe = defineSlotRecipe({
             borderColor: "colorPalette.solid",
           },
         },
+        itemControl: {
+          _disabled: {
+            bg: "unset",
+          },
+        },
         itemIndicator: {
           borderWidth: "1px",
           borderColor: "border.emphasized",


segunadebayo:
Superseded by #10938, which carries the same fix.

The change here only touches `packages/panda-preset/src/slot-recipes/radio-card.ts`, which is generated rather than authored — `panda-preset`'s `prebuild` runs `pnpm clean && pnpm theme`, and `cleanFiles()` deletes `src/**/*.{ts,tsx}` before `chakra eject` regenerates it from the React theme. So this edit would be overwritten on the next build, while the changeset would still announce the fix in `@chakra-ui/react`.

The file that needs the change is `packages/react/src/theme/recipes/radio-card.ts`, which #10938 patches. Your changeset text was good and has been carried over there, with credit.

Thanks for the report and the diagnosis — the root cause you identified was correct.

PR #10932, opened 2026-08-13
feat(Button): add accessibleWhenDisabled prop for accessible disabled state (Closes #10929)

### LINKED ISSUE #10929: Add support for `accessibleWhenDisabled` on Button
### Description

Add an `accessibleWhenDisabled` prop to `Button` that keeps a disabled button focusable and accessible to assistive technologies.

When enabled, Chakra would use `aria-disabled="true"` instead of the native `disabled` attribute.

Currently, when a button becomes disabled, for example, when entering a loading state, it loses focus because the native disabled attribute removes it from the focus order.

After clicking the button, focus should ideally remain on the same control while the action is in progress instead of being lost when disabled becomes true.

I think this behavior is useful beyond loading/pending buttons, so a generic API makes more sense than something tied specifically to an isPending or loading state.

Ariakit uses the name `accessibleWhenDisabled`, which I prefer over something like f`ocusableWhenDisabled`. It describes the intent rather than one particular consequence of the behavior.

### Proposed Solution or API

Add an accessibleWhenDisabled prop:
```
<Button disabled accessibleWhenDisabled>
  Submit
</Button>

<Button loading accessibleWhenDisabled>
  Submit
</Button>
```

Without the prop, the current behavior remains unchanged:
```
<button disabled>
  Submit
</button>
```
With accessibleWhenDisabled, Chakra would render:
```
<button aria-disabled="true">
  Submit
</button>
```
Chakra would also need to prevent the button’s action from being triggered while `aria-disabled="true"`, since unlike the native `disabled` attribute, `aria-disabled` does not prevent interaction by itself.

### Alternatives

If this behavior is adopted, it may also be worth considering implementing it in the Zag.js button machine. The current button implementation already contains interaction logic around disabled and loading states, and moving this behavior into the shared state machine could reduce framework-specific code while making the feature available across all frameworks built on Zag.js.

### Additional Information

Similar APIs / discussions:

* Ariakit uses accessibleWhenDisabled https://ariakit.com/reference/button#accessiblewhendisabled
* Base UI discusses preserving focus for buttons while loading:
    https://base-ui.com/react/components/button#loading-states
* Related discussion from Diego Haz:
    https://x.com/diegohaz/status/2086960304324325482

waterWang:
## Summary

Adds an `accessibleWhenDisabled` prop to the `Button` component. When enabled, the button uses `aria-disabled` instead of the native `disabled` attribute, keeping the button focusable and accessible to assistive technologies.

## Problem

When a button becomes disabled (e.g., during a loading state), it loses focus because the native `disabled` attribute removes it from the focus order. After clicking the button, focus should ideally remain on the same control while the action is in progress.

## Solution

- Added `accessibleWhenDisabled` prop to `ButtonBaseProps`
- When `true` and the button is disabled:
  - Uses `aria-disabled="true"` instead of the native `disabled` attribute
  - Sets `data-disabled` for CSS styling (Panda CSS `_disabled` selector matches `[data-disabled]`)
  - Suppresses click events to prevent action triggers
- IconButton and CloseButton inherit the feature via `ButtonProps`

## Usage

```tsx
<Button disabled accessibleWhenDisabled>Submit</Button>
<Button loading accessibleWhenDisabled>Submit</Button>
```

Closes #10929

### DIFF
--- a/packages/react/src/components/button/button.tsx
+++ b/packages/react/src/components/button/button.tsx
@@ -38,7 +38,19 @@ export interface ButtonLoadingProps {
 }
 
 export interface ButtonBaseProps
-  extends RecipeProps<"button">, UnstyledProp, ButtonLoadingProps {}
+  extends RecipeProps<"button">, UnstyledProp, ButtonLoadingProps {
+  /**
+   * When `true`, the button will use `aria-disabled` instead of the native
+   * `disabled` attribute, keeping the button focusable and accessible to
+   * assistive technologies.
+   *
+   * When enabled, the button's click handler will be suppressed to prevent
+   * action triggers while the button is in a disabled state.
+   *
+   * @default false
+   */
+  accessibleWhenDisabled?: boolean | undefined
+}
 
 export interface ButtonProps extends HTMLChakraProps<
   "button",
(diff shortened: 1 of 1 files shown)


isBatak:
@waterWang could you also update the docs?

PR #10935, opened 2026-08-14
refactor: use zag attribute utilities

isBatak:
Closes #

## 📝 Description

Re-export `dataAttr` and `ariaAttr` from `@zag-js/dom-query` instead of maintaining a local implementation.

Chakra maintains its own `dataAttr` utility and does not expose `ariaAttr`.

## 🚀 New behavior

Both utilities are re-exported from `@zag-js/dom-query`.

No.

## 📝 Additional Information

Ideally, Ark UI would expose these utilities so Chakra would not need to install `@zag-js/dom-query` directly and keep its version aligned with Ark UI.

### DIFF
--- a/.changeset/use-zag-attr-utils.md
+++ b/.changeset/use-zag-attr-utils.md
@@ -1,0 +1,6 @@
+---
+"@chakra-ui/react": patch
+---
+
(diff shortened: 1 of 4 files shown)


Adebesin-Cell:
Closing in favor of https://github.com/chakra-ui/chakra-ui/pull/10949, thanks for the refactor upstream!

PR #10943, opened 2026-08-20
docs: add Emotion style registry for Next.js App Router (Fixes #10942)

### LINKED ISSUE #10942: Documented Next.js App Router setup fails hydration for any content streamed through <Suspense>
### Description

Following Chakra v3's [Next.js App Router guide](https://chakra-ui.com/docs/get-started/frameworks/next-app) exactly, and then putting a `<Suspense>` boundary on a route, produces a hydration failure on every load of that route:

```
Uncaught Error: Hydration failed because the server rendered HTML didn't match the client.

+ <div className="chakra-stack css-15d3pce">    <- client
- <style data-emotion="css 15d3pce">            <- server
```

The guide's provider sets up `ChakraProvider` with no Emotion style registry, so Emotion emits each rule as a `<style>` element inline in the React tree. For a route rendered in one pass that is harmless. For content streamed into a Suspense boundary it is not: the chunk arrives after the client has hydrated, and the client renders no `<style>` at that position.

This matters more under Next.js 16 than it used to. With [Cache Components](https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheComponents), streaming through Suspense is the **default** rendering model rather than an opt-in, so the documented setup now fails on the happy path rather than in an unusual configuration.

### Link to Reproduction

https://github.com/FinnIckler/chakra-ui-v3-emotion-cache

### Steps to reproduce

1. Open the reproduction and run `npm install && npm run dev`.
2. Navigate to `/stream`.
3. Open the browser console.

Image

The only non-default setting is `cacheComponents: true`; `components/ui/provider.tsx` is copied verbatim from the App Router guide. The route's shell renders `Container` + `Heading`, and its Suspense child renders `VStack`/`HStack`/`Badge`/`Text` — components the shell does not use, so Emotion inserts those rules for the first time inside the streamed chunk.

**Expected:** the streamed subtree hydrates cleanly.
**Actual:** the hydration error above.

> **Use a cold server.** Emotion's default cache is a module singleton on the server, so once a rule has been inserted it is not re-emitted on later requests in the same process. A warm dev server masks this entirely — after a recompile-and-reload you may see nothing.

### Chakra UI Version

3.36.1

### Browser

Firefox and Chrome

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux

### What the server actually sends

With no registry, each rule is emitted immediately before the element that uses it, inside React's streaming container:

```html
<div hidden id="S:0">
  <style data-emotion="css 15d3pce">...</style><div class="chakra-stack css-15d3pce">
  <style data-emotion="css 1x0y7ph">...</style><div class="chakra-stack css-1x0y7ph">
  <style data-emotion="css 929j8">...</style><span class="chakra-badge css-929j8">streamed</span>
```

Those `<style>` elements are siblings *within the reconciled tree*, so hydration fails at the first one. On a cold server the reproduction returns 11 `data-emotion` tags on that route, all after `</head>`, 8 of them interleaved directly before the element they style.

Checkable without a browser:

```bash
curl -s http://localhost:3000/stream | grep -c '<style data-emotion'
```

### Chakra v2 shipped a registry for exactly this

`@chakra-ui/next-js@2.4.4`, `dist/esm/use-emotion-cache.mjs`:

```js
function useEmotionCache(options) {
  const cache = useMemo(() => {
    const cache2 = createCache({ key: "css", prepend: true, ...options });
    cache2.compat = true;
    return cache2;
  }, [options]);

  useServerInsertedHTML(
    () => createElement("style", {
      key: cache.key,
      "data-emotion": `${cache.key} ${Object.keys(cache.inserted).join(" ")}`,
      nonce: cache.nonce,
      dangerouslySetInnerHTML: { __html: Object.values(cache.inserted).join(" ") }
    })
  );

  return cache;
}
```

The package's npm dist-tags show the v3 line was started and abandoned:

```
latest:    2.4.2
v2-latest: 2.4.4
next:      0.0.0-v3-20240109123510
dev:       0.0.0-dev-20240315145938
```

So the registry was not moved elsewhere — it was dropped, and the v3 App Router guide has no replacement.

### Suggested fix

Ship a first-party registry for v3, or document one in the App Router guide. `@mui/material-nextjs`'s `AppRouterCacheProvider` is the equivalent precedent.

**One caveat if the v2 implementation is ported directly:** it emits `Object.keys(cache.inserted)` — everything inserted so far — on *every* flush, re-sending rules on each pass. That may well explain #8009 ("All chakra styles are duplicated in initial render") and #7893 ("styles applied twice in next 13"). The approach in the [Next.js CSS-in-JS docs](https://nextjs.org/docs/app/guides/css-in-js) wraps `cache.insert` to track only the delta since the last flush, which avoids it.

### Workaround

Wrapping `ChakraProvider` in a delta-tracking registry resolves it. The interleaved tags collapse into one consolidated tag carrying every class name at once, emitted through Next's inserted-HTML channel rather than into the component tree:

```html
<style data-emotion="css 15d3pce 1x0y7ph 929j8 10wrni 1eaio2p ...">
```

Because it is not a React element at a tree position, it takes no part in hydration and the error goes away. (The reproduction includes this file, unwired, so it can be toggled on to confirm.)

```tsx
"use client";

import { useState } from "react";
import { useServerInsertedHTML } from "next/navigation";
import createCache from "@emotion/cache";
import { CacheProvider } from "@emotion/react";

export default function EmotionRegistry({
  children,
}: {
  children: React.ReactNode;
}) {
  const [{ cache, flush }] = useState(() => {
    const cache = createCache({ key: "css" });
    cache.compat = true;

    const previousInsert = cache.insert;
    let inserted: string[] = [];

    cache.insert = (...args) => {
      const [, serialized] = args;
      if (cache.inserted[serialized.name] === undefined) {
        inserted.push(serialized.name);
      }
      return previousInsert(...args);
    };

    const flush = () => {
      const previouslyInserted = inserted;
      inserted = [];
      return previouslyInserted;
    };

    return { cache, flush };
  });

  useServerInsertedHTML(() => {
    const names = flush();
    if (names.length === 0) return null;

    const styles = names.map((name) => cache.inserted[name]).join("");

    return (
      <style
        data-emotion={`${cache.key} ${names.join(" ")}`}
        dangerouslySetInnerHTML={{ __html: styles }}
      />
    );
  });

  return <CacheProvider value={cache}>{children}</CacheProvider>;
}
```

This omits v2's `prepend: true`, so Chakra's rules land in normal document order rather than at the start of `<head>`. Anyone relying on their own stylesheets overriding Chakra will want to keep `prepend: true`.

### Related

- #8970 — Hydration Error with NextJS 15 (v2, closed)
- #8009 — All chakra styles are duplicated in initial render (v2, closed)
- #7893 — styles applied twice in next 13 (v2, closed)

These are v2-era and stem from the `@chakra-ui/next-js` implementation. This report is about v3 having no registry at all.

waterWang:
## Summary

Adds an Emotion style registry to the Next.js App Router setup to prevent hydration errors when content is streamed through `<Suspense>` boundaries.

### Problem

Emotion emits CSS rules as inline `<style>` elements inside the React tree. When a Suspense chunk is streamed, these `<style>` elements arrive after the client has hydrated, causing a hydration mismatch.

Under Next.js 16 with Cache Components (default), streaming through Suspense is the default rendering model, so the documented setup now fails on the happy path.

### Changes

1. **`sandbox/next-app/app/provider.tsx`** — Added `EmotionRegistry` component using `CacheProvider` + `useServerInsertedHTML` with delta-tracking (tracks only rules inserted since the last flush, avoiding the duplicated-styles bug from v2's `@chakra-ui/next-js`)
2. **`sandbox/next-app/package.json`** — Added `@emotion/cache` dependency
3. **`apps/www/content/docs/get-started/frameworks/next-app.mdx`** — Added "Add an Emotion style registry (SSR)" section with full code example and layout usage, updated install command to include `@emotion/cache`

### Related

Fixes #10942

### DIFF
--- a/.changeset/codemod-dry-run-output.md
+++ b/.changeset/codemod-dry-run-output.md
@@ -1,0 +1,13 @@
+---
+"@chakra-ui/codemod": patch
+---
+
+Fix three bugs in the `upgrade` command:
+
+- `--dry` now prints a summary of which transforms would change files. Before,
+  it silenced jscodeshift and showed nothing.
+- `.ts` files now use the `ts` parser instead of `tsx`, so type assertions and
+  generic arrows no longer throw and get skipped.
+- The unused-package check no longer scans `node_modules`, so it stops flagging
+  `framer-motion` and `@emotion/styled` as used when only a dependency imports
+  them (and runs much faster).
--- a/.changeset/codemod-fast-upgrade-preview.md
+++ b/.changeset/codemod-fast-upgrade-preview.md
@@ -1,0 +1,16 @@
+---
+"@chakra-ui/codemod": minor
+---
+
+Speed up the `upgrade` command and add a readable dry-run preview.
+
+- Run transforms in-process instead of spawning a Node process per transform,
+  taking a full run from ~30s to a few seconds.
+- `upgrade --dry` now lists the files that would change and, per file, the
+  components and props it touches (e.g. `Button, color palette`).
+- Large projects write the full breakdown to `chakra-codemod-report.md`; smaller
+  ones print it inline.
+- Stop the `color-mode` transform from installing a snippet during a dry run,
+  and use the current CLI flag instead of the removed `--yes`.
+- Stop the `steps` transform from adding a stray `Steps` import to files that
+  don't use a stepper.
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -14,6 +14,7 @@
   },
   "ignore": [
     "next-app",
+    "next-app-streaming",
     "next-pages",
     "vite-ts",
     "vite-jsx",
--- a/apps/compositions/src/ui/input-group.tsx
+++ b/apps/compositions/src/ui/input-group.tsx
@@ -1,12 +1,24 @@
-import type { BoxProps, InputElementProps } from "@chakra-ui/react"
-import { Group, InputElement } from "@chakra-ui/react"
+import {
+  type BoxProps,
+  Group,
+  InputAddon,
+  type InputAddonProps,
+  InputElement,
+  type InputElementProps,
+} from "@chakra-ui/react"
 import * as React from "react"
 
 export interface InputGroupProps extends BoxProps {
-  startElementProps?: InputElementProps
-  endElementProps?: InputElementProps
   startElement?: React.ReactNode
+  startElementProps?: InputElementProps
   endElement?: React.ReactNode
+  endElementProps?: InputElementProps
+
+  startAddon?: React.ReactNode
+  startAddonProps?: InputAddonProps
+  endAddon?: React.ReactNode
+  endAddonProps?: InputAddonProps
+
   children: React.ReactElement<InputElementProps>
   startOffset?: InputElementProps["paddingStart"]
   endOffset?: InputElementProps["paddingEnd"]
@@ -20,6 +32,10 @@ export const InputGroup = React.forwardRef<HTMLDivElement, InputGroupProps>(
       endElement,
       endElementProps,
       children,
+      startAddon,
+      startAddonProps,
+      endAddon,
+      endAddonProps,
       startOffset = "6px",
       endOffset = "6px",
       ...rest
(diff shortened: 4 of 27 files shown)


FinnIckler:
So you will not supply your own registry as part of the package like you did in v2?

segunadebayo:
@FinnIckler, just a copy-paste of the emotion registry is a better solution for now.

When we get off the emotion train, we'll no longer need this.

PR #10951, opened 2026-08-27
fix: preserve user onClick on Dialog and Drawer ActionTrigger

dfedoryshchev:
### Description

`Dialog.ActionTrigger` and `Drawer.ActionTrigger` accept `onClick` in their public props type, but silently discard it.

Both render `<chakra.button type="button" {...props} ref={ref} onClick={() => dialog.setOpen(false)} />`, declaring the internal close handler *after* the props spread, so a caller's `onClick` is replaced rather than chained (`dialog.tsx:144`, `drawer.tsx:144`). The dialog closes, the handler never runs, and nothing warns: `<Dialog.ActionTrigger onClick={handleSave}>Save</Dialog.ActionTrigger>` is a silent no-op for `handleSave`.

### Fix

`CodeBlockCollapseTrigger` in this package already does this correctly, calling `restProps.onClick?.(e)` before its own action (`code-block.tsx:355`). The change makes both triggers follow that existing convention rather than introducing a new one.

Passing the handler to a child through `asChild` is unaffected, because `mergeProps` chains `on[A-Z]` handlers. Only a handler passed directly to `ActionTrigger` is lost today.

### Changes

One regression test per component, and a patch changeset. The full package suite is 1001 passing.

### DIFF
--- a/.changeset/action-trigger-onclick.md
+++ b/.changeset/action-trigger-onclick.md
@@ -1,0 +1,7 @@
+---
+"@chakra-ui/react": patch
+---
+
+- Fix `Dialog.ActionTrigger` and `Drawer.ActionTrigger` discarding a caller's
+  `onClick`. The internal close handler was declared after the props spread, so
+  it replaced the user handler instead of running alongside it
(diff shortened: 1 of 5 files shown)


Adebesin-Cell (approved):
Thank you ❤️

dfedoryshchev:
I'm happy my contributions are appreciated!

PR #10957, opened 2026-08-29
feat(v4): Panda v2 styling engine — zero-Emotion, build-time (POC/RFC)

Adebesin-Cell:
This is an experimental version. To join in the conversation, please see the design notes in [`design-notes/`](https://github.com/Adebesin-Cell/chakra-ui/tree/feat/v4-static-factory/design-notes) (engine design, Panda v2 gaps, Emotion opt-in) and the discussion below.

## 📝 Description

A proof-of-concept making **Panda v2 the styling engine** for Chakra v4 — build-time CSS extraction, **zero Emotion in the core**, RSC-friendly by construction. Opening as a draft to share the direction and gather feedback.

Chakra styles at runtime via Emotion: an `@emotion/react` provider is required, styles are serialized and inserted at render, and the cost lands on every app (perf, SSR/RSC friction, mandatory provider).

## 🚀 New behavior

- **`@chakra-ui/styled-system`** — a standalone package Panda codegen writes to (css/recipes/patterns/jsx/types); `@chakra-ui/react` consumes it by package name. The styling layer becomes a swappable package.
- **Static, context-free factory** — `chakra.*` composes class names; prop forwarding via Panda's `isCssProperty`. No runtime style insertion.
- **Layout primitives → Panda patterns** — `Stack`/`HStack`/`VStack`/`Flex`/`Center` extract statically.
- **Recipes → generated contexts** — `Button` (recipe), `Card`/`Alert` (slot recipes) use Panda's generated `createRecipeContext`/`createSlotRecipeContext` with static per-recipe imports, so unused recipes tree-shake out.
- **`@chakra-ui/emotion`** — an opt-in, self-contained runtime styling package for consumers who want Emotion back. Not in the bundle unless installed.

Proven end-to-end in `sandbox/vite-ts` (0 `@emotion`/`insertStyles` in output, tree-shaking confirmed); `packages/react` typechecks clean.

Yes — v4. Styling moves from runtime to build time; apps run a Panda build step (to be wrapped in a bundler plugin). Emotion returns via the opt-in `@chakra-ui/emotion` package.

## 📝 Additional Information

Not for merge — a direction to discuss. Still open: RSC demonstrated in a Next app-router sandbox, rolling the migration across all components, and `@chakra-ui/styled-system` distribution (`designSystem` / a bundled bundler plugin).


sandbox/next-app/postcss.config.cjs
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  plugins: {
+    "@pandacss/dev/postcss": {},
```
isBatak, line 3: Could this be replaced with https://github.com/chakra-ui/panda/tree/v2/packages/webpack?

packages/styled-system/panda.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "@pandacss/dev"
+
+export default defineConfig({
```
isBatak, line 3: Why exactly do we need `@chakra-ui/styled-system` and `panda.config.ts`?
`sandbox/next-app` and `sandbox/vite-ts` have their own `panda.config.ts`.

packages/panda-preset/src/patterns.ts
isBatak, line 1: I propose dropping the `@chakra-ui/panda-preset` package and creating a separate `@chakra-ui/theme` that exports the Panda preset. 
Something like this https://github.com/isBatak/chakra-ui/pull/12/changes#diff-0e42e3a162546b3bd24c4afc55c82bf278f4576f004dd604d30a025b93fc4c0eR58

packages/panda-preset/src/recipes/button.ts
```diff
@@ -1,7 +1,7 @@
 import { defineRecipe } from "../def"
 
 export const buttonRecipe = defineRecipe({
-  className: "button",
+  className: "chakra-button",
```
isBatak, line 4: We are also missing [JSX tracking](https://panda-css.com/docs/concepts/recipes#advanced-jsx-tracking).
I like the idea of keeping the tracking definition close to the component code in `tracking.ts` like [this](https://github.com/isBatak/chakra-ui/pull/12/changes#diff-df93c8d8e2780f4ddaeb1eb2f7af102505ea7fe03b45b7680326e2f9f679481aR1) 
and import it into the theme recipe like [this](https://github.com/isBatak/chakra-ui/pull/12/changes#diff-0e42e3a162546b3bd24c4afc55c82bf278f4576f004dd604d30a025b93fc4c0eR93)

sandbox/next-app/postcss.config.cjs
Adebesin-Cell: Yes 👍

Adebesin-Cell:
Closing this since we established pattern and working to go through, would open smaller PRs so it's easier to work and isolate features 👍

packages/styled-system/panda.config.ts
Adebesin-Cell: The sandbox configs are app configs.

In implementation, we'd need to configure Chakra UI as `designSystem: "@chakra-ui/react"`, though it's a huge chunk of work for a single PR, which is why the think pieces are in one PR 😄.

packages/panda-preset/src/patterns.ts
Adebesin-Cell: Yep, sounds good 👍. We should split the work per PR why I'm closing this huge one.

packages/panda-preset/src/recipes/button.ts
Adebesin-Cell: Cool! I agree
I love your idea of `defineChakraRecipe`. We should probably implement something similar to avoid manually adding `chakra-` to every recipe

PR #10966, opened 2026-09-04
fix(flex, square): apply the array form of the css prop

dfedoryshchev:
## 📝 Description

`Flex` and `Square` merge the incoming `css` prop into their base styles with an object spread. The `css` prop is typed to accept an array of style objects, and object-spreading an array produces index keys, so nothing in the array reaches the style output.

`AspectRatio`, `Bleed` and `Float` already pass `css={[baseStyles, props.css]}`, and `toArray` in `use-resolved-props.ts` exists to support exactly this. `Flex` and `Square` are the two remaining `...props.css` sites in the package, so this closes the class rather than sampling it.

`Circle` renders through `Square`, so it is fixed by the same change and needs no edit of its own.

`<Flex css={[{ position: "fixed" }]}>` applies nothing, while the same array on `<Box>` applies it.

## 🚀 New behavior

Both components pass `css={[baseStyles, props.css]}`. Precedence is unchanged for the object form: base styles first, the caller's `css` second.

No.

## 📝 Additional Information

Added `packages/react/__tests__/css-prop.test.tsx` with eight cases. Three fail on `main` - the array form on `Flex`, `Square` and `Circle`. The other five pass before and after: the array form on `Box` as a control, the object form on `Flex` and `Square`, and two checking that a `css` value still wins over a base style it collides with. That last pair is the one I most wanted, since switching to the array form changes how precedence is resolved and nothing in the repo covered it. Full suite 1005/1005 with the change; changeset included.

### DIFF
--- a/.changeset/fix-flex-square-css-array.md
+++ b/.changeset/fix-flex-square-css-array.md
@@ -1,0 +1,10 @@
+---
+"@chakra-ui/react": patch
+---
+
+- **Flex, Square**: Fix the array form of the `css` prop being silently dropped.
+  Both components merged the incoming `css` into their base styles with an
+  object spread, which turns an array into index keys instead of merging its
+  entries. They now pass `css={[baseStyles, props.css]}`, matching
+  `AspectRatio`, `Bleed` and `Float`. `Circle` renders through `Square`, so it
+  is fixed too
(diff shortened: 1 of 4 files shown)


Adebesin-Cell (approved):
Thank you! ❤️

PR #11002, opened 2026-09-13
docs(action-bar): document root props

Adebesin-Cell:
Closes #

## 📝 Description

Adds the `ActionBar` root props to the generated types so the props table renders on the docs page.

The ActionBar root props table was empty because the type entry was not generated.

## 🚀 New behavior

`generate-types` now emits the root props entry for ActionBar, and the docs page picks it up.

No

## 📝 Additional Information

<h2><a href="https://app.greptile.com/api/retrigger?id=63700869"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the documentation lookup and generated metadata match the ActionBar public API.

<details><summary><h3>Summary</h3></summary>

- Points the documentation table at `ActionBar.Root` instead of `Popover.Root`.
- Removes inherited Popover parts that ActionBar does not export.
- Excludes `positioning`, which the public `ActionBarRootProps` type explicitly omits.
</details>

<sub>Reviews (1) · Last reviewed commit: ["docs(action-bar): document root props"](https://github.com/chakra-ui/chakra-ui/commit/3790f295dc25e5931c08bc56899a6d68246a34a2)</sub>


apps/www/public/r/types/action-bar.json
segunadebayo, line 1: This feels wrong to remove all the other parts. Please check the code to see if other parts are exported, then their props should be exposed as well.

Adebesin-Cell: Yeah, those (`Anchor`, `Arrow`, `ArrowTip`, `Description`, `Indicator`, `Title`, `Trigger`) are Popover parts, not ActionBar parts; they only show up because `arkPropsMap` maps `action-bar → popover` and pulls in the whole set. 

I did try dropping non-exported parts automatically instead of hardcoding, but it breaks ~15 other components; it looks like we lean on documenting Ark parts that the namespace doesn't re-export (Field shows `Input`/`Select`/`Textarea` even though `field/namespace.ts` doesn't). ActionBar's the only one wrapping a subset, so the manual removal feels like the safe call here.

